PR #38, opened 2026-04-03
fix: defer message_end until all content blocks are closed

dimakis:
## Summary
- Defers `message_end` emission until all `content_block_stop` events have fired, fixing tool blocks and text being orphaned from finished messages
- Adds session restore fixes — `getMessages` returns v2 format, guards against stale localStorage caches
- Pairs tool results with tool calls in session restore

## Root cause
The `assistant` SDK event fires before all `content_block_stop` events. Server was mapping `assistant` → `message_end` directly, causing `block_end` events to arrive after the message was already finalized. Frontend's `finishCurrent()` would capture the message without those blocks.

## Test plan
- [ ] Build passes (`npm run build`)
- [ ] Tests pass (`npm test`)
- [ ] Tool pills render with names during streaming
- [ ] Tool pills show input summary after completion
- [ ] Tools persist in finished messages (scroll up and they're still there)
- [ ] Session resume doesn't crash
- [ ] Follow-up messages appear in UI

### DIFF
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -260,6 +260,38 @@ describe('runQueryLoop', () => {
     expect(sessionEnds).toHaveLength(1);
   });
 
+  it('defers message_end until all blocks are closed', async () => {
+    const events: Record<string, unknown>[] = [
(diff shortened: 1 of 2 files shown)


dimakis:
Already merged via other agent

PR #44, opened 2026-04-03
fix(security): helmet, request size limit, login rate limiting

dimakis:
## Summary

Three quick security hardening changes — zero behavior change for normal usage.

- **helmet** — security headers (CSP, HSTS, X-Frame-Options, etc.) with websocket connect-src allowed
- **express.json limit** — 10MB cap on request bodies (was unlimited)
- **Login rate limiting** — 5 attempts per minute per IP via express-rate-limit

## Test plan

- [x] All 265 tests pass (route tests cover auth endpoints)
- [x] Typecheck clean
- [ ] CI green
- Manual: verify login works, chat works, file upload works from phone

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,10 @@ jobs:
           cache: npm
 
       - name: Install root dependencies
+        # TODO: Restore `npm ci` once lockfile is generated on Linux.
+        # jsdom's optional @emnapi/* native deps vary by platform — a macOS-generated
+        # lockfile omits Linux variants, causing npm ci to fail on Ubuntu runners.
+        # Fix: run `npm install` in a Linux container and commit that lockfile.
         run: npm install --no-audit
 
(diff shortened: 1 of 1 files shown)


dimakis (commented):
Blocking issue:

The new Helmet CSP in `server/app.ts` is likely to break existing frontend behavior that relies on `data:` image URLs.

Current policy sets `defaultSrc 'self'` and does not define `img-src`, so browsers fall back to `default-src` for images. Mitzo currently renders image previews and CSS icons from data URLs, so this can regress attachment preview and some UI assets.

Suggested fix:
- Add an explicit `imgSrc` directive that includes at least `'self'` and `data:`.
- Then run a quick manual verification for chat image preview and basic UI rendering on mobile Safari.

Non-blocking follow-up:
- Add route-level tests for login rate limiting (`429`) and a test/assertion for CSP headers so this doesn’t regress silently.

PR #84, opened 2026-04-05
fix(worktree): address multi-repo PR review feedback

dimakis:
## Summary

Addresses all issues raised in review of #83.

- **Security**: Internal token now passed via `MITZO_INTERNAL_TOKEN` env var instead of CLI args (was visible in `ps aux`)
- **Cleanup**: `worktreePaths` stores `{path, wtId}` — eliminates fragile regex extraction of worktree ID during session teardown
- **Config**: Repo paths validated at load time (must exist and be a git repo); invalid entries warn and are excluded
- **Test**: Idempotency test renamed to 'first-write-wins' with explicit path assertion to document intent
- **Multi-repo**: Repo MCP server injected regardless of sandbox mode — sandbox shouldn't be required for multi-repo access
- **Reattach**: Stale `worktree_opened` events filtered during replay if path no longer exists (e.g. after server restart); valid ones repopulate `worktreePaths`

## Test plan

- [x] All 391 tests pass

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -1,4 +1,5 @@
 import 'dotenv/config';
+import { existsSync } from 'fs';
 import { createServer } from 'http';
 import { WebSocketServer, WebSocket } from 'ws';
 import { verifyWsAuth } from './auth.js';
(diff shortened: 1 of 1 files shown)


dimakis:
Nudging CI.

PR #85, opened 2026-04-05
feat(hooks): bridge project hooks to Agent SDK

dimakis:
## Summary

- Adds `server/hook-bridge.ts` — reads hook definitions from a project's `.claude/settings.json` and bridges shell command hooks into SDK `HookCallback` functions
- Passes the constructed `hooks` object to the SDK's `query()` call in `chat.ts`
- Enables `SessionStart` hooks (and any future hook events) to fire during Mitzo sessions — e.g. mgmt's boot context injector now works

## What this fixes

Mitzo uses the Agent SDK but never passed the `hooks` option to `query()`. This meant `SessionStart` hooks defined in project settings (like mgmt's profile/architecture boot context) were silently skipped. The user's working style, communication preferences, and workspace architecture were missing from every Mitzo session.

## Design

General-purpose bridge, not SessionStart-specific:
- Reads any hook event type from settings.json
- Expands `$CLAUDE_PROJECT_DIR` and other env vars in commands
- Spawns shell commands, pipes `HookInput` as JSON to stdin
- Maps CLI hook JSON output to SDK's `SyncHookJSONOutput` format
- Graceful failures — broken hooks never crash sessions

## Test plan

- [x] 25 new tests in `hook-bridge.test.ts` (all pass)
- [x] Full suite: 416 tests across 41 files (all pass)
- [ ] Manual: start Mitzo session against mgmt repo, verify boot context appears in Claude's system prompt

### DIFF
--- a/server/__tests__/hook-bridge.test.ts
+++ b/server/__tests__/hook-bridge.test.ts
@@ -203,6 +203,16 @@ describe('createCommandCallback', () => {
       },
     });
   });
+
+  it('returns empty object when aborted', async () => {
+    const controller = new AbortController();
+    // sleep command that we'll abort immediately
+    const cb = createCommandCallback('sleep 10', TEST_DIR, 10000);
+    const promise = cb({} as never, undefined, { signal: controller.signal });
+    controller.abort();
+    const result = await promise;
+    expect(result).toEqual({});
+  });
 });
 
 // --- loadProjectHooks ---
--- a/server/hook-bridge.ts
+++ b/server/hook-bridge.ts
@@ -109,7 +109,7 @@ export function createCommandCallback(
   toolUseID: string | undefined,
   options: { signal: AbortSignal },
 ) => Promise<HookJSONOutput> {
-  return async (input, _toolUseID, _options) => {
+  return async (input, _toolUseID, options) => {
     const hookEventName =
       input && typeof input === 'object' && 'hook_event_name' in input
         ? String((input as Record<string, unknown>).hook_event_name)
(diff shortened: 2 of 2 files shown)


dimakis (commented):
## Review

Overall: solid work. Clean abstraction, good test coverage (25 tests), graceful failure handling throughout. The integration into `chat.ts` is minimal and correct.

---

### Issues worth addressing

**1. AbortSignal is ignored (medium)**

`createCommandCallback` accepts `options.signal` but never wires it to the child process. If the session is cancelled mid-hook, the subprocess lingers.

```ts
// _options is accepted but signal is never used
return async (input, _toolUseID, _options) => {
```

Fix: kill the child on abort:
```ts
const child = execFile('/bin/sh', ['-c', command], { cwd, timeout: timeoutMs }, (err, stdout) => { ... });
options.signal.addEventListener('abort', () => child.kill());
```

**2. `reason` in `SDK_TOP_LEVEL_KEYS` (minor)**

`reason` isn't a standard `SyncHookJSONOutput` top-level field per the SDK types — it'll pass through as top-level but the SDK may silently ignore it. Probably harmless, but worth confirming against SDK source.

**3. `timeout` on `HookCallbackMatcher` (minor)**

`loadProjectHooks` spreads `matcher.timeout` onto the SDK matcher object. If `HookCallbackMatcher` doesn't have a `timeout` field in the SDK types, this is silently dropped. The timeout is already applied correctly to `execFile` via `timeoutMs`, so functionality is fine — just a type cleanliness issue.

**4. stdin write reliability (low)**

`child.stdin?.write(JSON.stringify(input))` — a single `write()` without a drain check could silently truncate for very large inputs. Unlikely to matter in practice for the payloads the SDK sends, but worth noting.

---

### What's good

- General-purpose bridge design (not `SessionStart`-specific) is the right call
- `expandEnvVars` leaving unknown variables unchanged is correct behaviour
- `mapHookOutput` split between top-level/hook-specific fields is clean
- The `hasHooks` guard preventing an empty object return is good
- Graceful failure on broken hooks (never crashes the session) is exactly right

---

**Verdict:** Approvable after fixing the AbortSignal issue. The rest are minor. Don't forget to check off the manual test item (boot context appearing in system prompt) before merging.

PR #88, opened 2026-04-05
feat(skills): slash-command parsing and prompt rendering

dimakis:
## Summary
- Adds `parseSlashCommand()` for splitting `/name args` input
- Adds `resolveSlashCommand()` — server-authoritative resolution (native > skill > error)
- Adds `renderSkillPrompt()` with `$ARGUMENTS` substitution and skill envelope
- Levenshtein-based suggestions for unknown commands

Step 2 of 6. Stacked on #87.

## Test plan
- [x] 15 new tests in `server/__tests__/skill-rendering.test.ts` — all pass
- [x] Full suite passes (flaky `socket hang up` in routes is pre-existing)

### DIFF
--- a/server/slash-commands.ts
+++ b/server/slash-commands.ts
@@ -1,0 +1,151 @@
+import type { SkillRegistry, SkillScope } from './skills.js';
+
+// --- Parsing ---
+
+export interface ParsedSlash {
+  name: string;
+  arguments: string;
+}
+
+/**
+ * Parse a potential slash command from user input.
+ * Returns null if the input is not a slash command.
+ */
+export function parseSlashCommand(input: string): ParsedSlash | null {
+  const trimmed = input.trim();
+  if (!trimmed.startsWith('/')) return null;
+
+  const spaceIdx = trimmed.indexOf(' ');
+  const name = spaceIdx === -1 ? trimmed.slice(1) : trimmed.slice(1, spaceIdx);
+  const args = spaceIdx === -1 ? '' : trimmed.slice(spaceIdx + 1).trim();
+
+  if (!name) return null;
+
+  return { name, arguments: args };
+}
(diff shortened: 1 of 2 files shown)


server/slash-commands.ts
```diff
@@ -0,0 +1,151 @@
...
+      message: string;
+      suggestions: string[];
+    };
+
+/**
+ * Resolve a user message to a slash command result.
+ * This is the single server-authoritative resolution point.
+ */
+export function resolveSlashCommand(
+  input: string,
```
dimakis: `resolveSlashCommand` takes `nativeNames: Set<string>` as a separate parameter, meaning callers must manually keep it in sync with the actual `NativeCommandRegistry`. Consider making `NativeCommandRegistry` (or a wrapper that exposes its names set) a first-class dependency here — pass the registry itself so the source of truth is always authoritative.

server/slash-commands.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export function renderSkillPrompt(
+  body: string,
+  args: string,
+  meta: { name: string; scope: string },
+): string {
+  const substituted = body.replace(/\$ARGUMENTS/g, args);
+
+  return [`[Skill: /${meta.name} | source: ${meta.scope}]`, '', substituted].join('\n');
+}
```
dimakis: The envelope string interpolates `meta.name` and `meta.scope` directly with no escaping. With current naming rules this is safe, but if those rules are ever relaxed, a newline embedded in a skill name or scope value would inject content into the rendered prompt. Consider validating or sanitising name/scope at registration time, or using a delimiter that cannot appear in valid names.

dimakis (commented):
Review of feat/slash-parsing

PR #92, opened 2026-04-05
feat(skills): bundled skill pack — /simplify, /risk-scan, /pr-review

dimakis:
## Summary
- `/simplify` — find complexity, duplication, and cleanup opportunities (Read/Glob/Grep only)
- `/risk-scan` — surface failure modes, missing tests, unsafe assumptions (Read/Glob/Grep only)
- `/pr-review` — compact mobile-friendly diff/branch review (adds Bash for git diff)
- All skills are analysis-first: no Write/Edit, explicit approval gates in prompts

Step 6 of 6. Stacked on #91. **This completes the skills v1 system.**

## Test plan
- [x] 11 new tests: discovery, uniqueness, descriptions, tool restrictions, approval gates, size limits, scope
- [x] Full suite (489 tests) passes

### DIFF
--- a/server/__tests__/bundled-skills.test.ts
+++ b/server/__tests__/bundled-skills.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from 'vitest';
+import { statSync } from 'fs';
+import { join, dirname } from 'path';
+import { fileURLToPath } from 'url';
+import { SkillRegistry, SKILL_SIZE_LIMIT } from '../skills.js';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const SKILLS_DIR = join(__dirname, '..', '..', 'skills');
+
+describe('bundled skills', () => {
+  const registry = new SkillRegistry({ bundledDir: SKILLS_DIR });
+  const skills = registry.list();
+
+  it('discovers all three bundled skills', () => {
+    const names = skills.map((s) => s.name).sort();
+    expect(names).toEqual(['pr-review', 'risk-scan', 'simplify']);
+  });
+
+  it('has unique names', () => {
+    const names = skills.map((s) => s.name);
+    expect(new Set(names).size).toBe(names.length);
+  });
+
+  it('all skills have descriptions', () => {
+    for (const skill of skills) {
+      expect(skill.description).toBeTruthy();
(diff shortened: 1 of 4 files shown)


server/__tests__/bundled-skills.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  });
+
+  it('each bundled skill body contains an approval gate instruction', () => {
+    for (const skill of skills) {
+      const body = registry.getBody(skill.name);
+      expect(body).toBeDefined();
+      // Check for approval gate language
+      const hasGate =
+        body!.toLowerCase().includes('approval') ||
```
dimakis: nit: The approval gate check is a broad OR over five different strings (`approval`, `wait for`, `present`, `before making changes`, `do not modify`). A skill body that happens to contain any one of these words for an unrelated reason would pass silently. Suggest checking for a single canonical phrase that all bundled skills are required to include (e.g. `'Present'` + `'before making changes'` as a conjunction, or a fixed sentinel string), so the test is a meaningful contract rather than a loose heuristic.

server/__tests__/bundled-skills.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  it('/pr-review includes Bash for git diff access', () => {
+    const prReview = skills.find((s) => s.name === 'pr-review');
+    expect(prReview).toBeDefined();
+    expect(prReview!.allowedTools).toBeDefined();
+    expect(prReview!.allowedTools).toContain('Bash');
+    // But not Write/Edit — analysis first
+    expect(prReview!.allowedTools).not.toContain('Write');
+    expect(prReview!.allowedTools).not.toContain('Edit');
+  });
+
```
dimakis: There is no end-to-end test of `$ARGUMENTS` substitution using real bundled skill content. The skill body tests above check metadata and policy properties, but not that `renderSkillPrompt` correctly substitutes `$ARGUMENTS` when called with actual loaded SKILL.md content. Suggest adding at least one test that calls `renderSkillPrompt(registry.getBody('simplify')!, 'src/lib', { name: 'simplify', scope: 'bundled' })` and asserts the token is replaced.

PR #146, opened 2026-04-08
fix: address review findings for PR #144

dimakis:
## Centaur Auto-Fix

Fixes for review findings on PR #144 (`refactor(server): load repo config dynamically instead of caching at startup`).

### Changes
- Moved getRepoConfig() into its own try/catch block so that if it throws (e.g., config file not found or parse error), the error is logged and startup continues with just the primary repo cleanup rather than skipping all worktree cleanup

---
*Generated by Centaur*


dimakis:
Parent PR #144 was closed; reviving as a fresh PR

PR #147, opened 2026-04-08
fix: address review findings for PR #143

dimakis:
## Centaur Auto-Fix

Fixes for review findings on PR #143 (`fix(context-blocks): XML sanitization, size limits, and naming`).

### Changes
- Added `r.ok` check before calling `r.json()` — throws on non-OK HTTP responses so the catch handler fires
- Added `typeof data.contextBlocks === 'object'` runtime guard before iterating with `Object.entries`
- Wrapped `toFixed(1)` calls with `parseFloat()` to strip trailing zeros (e.g. '1.0' becomes '1')
- Skipped hardcoded URL change — no configured base URL exists in the codebase, and the catch handler already covers fetch failures
- Removed duplicate USER_SEND dispatch and forceScrollToBottom from interruptMessage to prevent duplicate user bubbles — the caller is responsible for dispatching the user message
- Collapsed identical if/else branches in sendMessage: only wsSetRunning differs when not already running, so moved the shared wsSend/dispatch/forceScrollToBottom after a conditional wsSetRunning call
- Replaced direct payload mutation (payload.contextBlocks = contextBlocks) with object spread into a new object to avoid mutating the return value of buildSendPayload, also making the pattern consistent with interruptMessage
- Added `ContextBlockContent` type alias for `string` with a JSDoc comment clarifying it represents serialized context block content, and used it in place of `string[]` for the `contextBlocks` field in `FinishedMessage`
- Replaced fixed TMP_DIR path with `mkdtempSync(join(tmpdir(), 'assemble-prompt-'))` to use OS temp directory with unique suffix, preventing parallel CI conflicts and leftover directory issues
- Added `import { mkdtempSync } from 'fs'` and `import { tmpdir } from 'os'`
- Changed TMP_DIR from const to let since it is now assigned per-test in beforeEach
- Added test for XML-unsafe characters in file content (e.g., `<script>` tags) to verify content is preserved verbatim
- Added test for empty (0-byte) context block files
- Added test for duplicate names in the contextBlocks array, asserting the block is injected only once
- Added type annotation `Array<{ data: string; mediaType: string }>` to the images variable to catch signature changes
- Added `resolve` import from 'path' for path traversal validation
- Added `SAFE_NAME_RE` constant to validate contextBlock keys match `/^[a-zA-Z0-9_-]+$/`
- Extracted `isStringRecord(val)` helper to deduplicate the `obj.X && typeof obj.X === 'object' && !Array.isArray(obj.X)` pattern used for toolTierOverrides, repos, and contextBlocks
- Added name validation in contextBlocks loop — keys not matching the safe pattern are skipped with a warning
- Added path traversal check in contextBlocks — resolved paths that escape the repo root are skipped with a warning
- Changed contextBlocks to use `resolve()` instead of `join()` so path traversal sequences are normalized before the containment check
- Extracted shared ContextBlocksSchema with size limits: max 20 items, each string max 100,000 chars
- Replaced inline contextBlocks definitions in SendMessage and InterruptMessage with the shared schema

---
*Generated by Centaur*

### DIFF
--- a/frontend/src/components/ContextPicker.tsx
+++ b/frontend/src/components/ContextPicker.tsx
@@ -19,8 +19,8 @@ function formatSize(bytes: number): string {
   if (bytes === 0) return '0 B';
   if (bytes < 1024) return `${bytes} B`;
   const kb = bytes / 1024;
-  if (kb < 1024) return `${kb.toFixed(1)} KB`;
-  return `${(kb / 1024).toFixed(1)} MB`;
+  if (kb < 1024) return `${parseFloat(kb.toFixed(1))} KB`;
+  return `${parseFloat((kb / 1024).toFixed(1))} MB`;
 }
 
 export function ContextPicker({ selected, onToggle, onClose }: Props) {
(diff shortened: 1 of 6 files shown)


dimakis:
Parent PR #143 was superseded by PR #118; no remaining value

PR #152, opened 2026-04-08
fix(chat): correlation ID dedup + /review-response skill

dimakis:
## Summary

- Fix duplicate user message bubble caused by mismatched optimistic/server IDs
- **Proper fix**: client generates a `clientMsgId` nonce, sends it with the WS message, server echoes it back — exact ID match deduplication works perfectly
- Replaces the fragile content-based matching from the initial commit
- Add `/review-response` bundled skill for triaging PR review comments

## Commits

1. `fix(chat): deduplicate optimistic user message against server echo` — initial content-based fix
2. `refactor(chat): replace content-based dedup with correlation ID` — proper long-term fix
3. `feat(skills): add /review-response bundled skill` — new skill for review triage

## Test plan

- [x] Correlation ID dedup: USER_SEND + server echo with same clientMsgId → single message (717 tests pass)
- [x] Duplicate text with different clientMsgIds → both preserved correctly
- [x] Reconnect scenario (no optimistic message) → server message added normally
- [x] /review-response skill discovered, has Bash in allowed-tools, passes approval gate check
- [ ] Manual: send follow-up message in live session — appears once
- [ ] Manual: disconnect and reconnect — user messages survive

### DIFF
--- a/frontend/src/hooks/useChatMessages.ts
+++ b/frontend/src/hooks/useChatMessages.ts
@@ -377,11 +377,25 @@ export function chatMessagesReducer(
 
     case 'USER_MESSAGE_RECEIVED': {
       // Server-side user_message event (from reattach replay or live emit).
-      // Deduplicate: USER_SEND already added this message client-side during
-      // the live session, so skip if a message with this ID already exists.
+      // Exact ID match — already have this message (reattach replay).
       if (state.messages.some((m) => m.messageId === action.messageId)) {
         return state;
       }
+      // Content match — optimistic USER_SEND already rendered this message.
+      // Upgrade the ID to the server's canonical one so reconnect/restore
+      // can find it by the persisted ID.
+      const optimisticIdx = state.messages.findIndex(
+        (m) =>
+          m.role === 'user' &&
+          m.messageId.startsWith('user-') &&
+          m.blocks[0]?.content === action.text,
+      );
+      if (optimisticIdx !== -1) {
+        const updated = [...state.messages];
+        updated[optimisticIdx] = { ...updated[optimisticIdx], messageId: action.messageId };
+        return { ...state, messages: updated };
+      }
+      // No match — add as new (reconnect/reattach scenario).
       return {
         ...state,
         messages: [
--- a/frontend/src/hooks/useChatMessages.ts
+++ b/frontend/src/hooks/useChatMessages.ts
@@ -61,7 +61,13 @@ export type ChatMessagesAction =
   // Session / UI lifecycle
   | { type: 'ERROR'; error: string }
   | { type: 'SESSION_INFO'; branch: string; isWorktree: boolean }
-  | { type: 'USER_SEND'; text: string; images?: string[]; contextBlocks?: string[] }
+  | {
+      type: 'USER_SEND';
+      text: string;
+      clientMsgId: string;
+      images?: string[];
+      contextBlocks?: string[];
+    }
   | { type: 'SET_RUNNING'; running: boolean }
   | { type: 'CONNECTION_LOST' }
   | { type: 'PERMISSION_REQUEST'; payload: PermissionRequest }
@@ -376,26 +382,13 @@ export function chatMessagesReducer(
     }
 
     case 'USER_MESSAGE_RECEIVED': {
-      // Server-side user_message event (from reattach replay or live emit).
-      // Exact ID match — already have this message (reattach replay).
+      // Server echo of a user message (live confirmation or reattach replay).
+      // The client's USER_SEND uses the same clientMsgId that the server
+      // echoes back, so exact ID match deduplicates correctly.
       if (state.messages.some((m) => m.messageId === action.messageId)) {
         return state;
       }
-      // Content match — optimistic USER_SEND already rendered this message.
-      // Upgrade the ID to the server's canonical one so reconnect/restore
-      // can find it by the persisted ID.
-      const optimisticIdx = state.messages.findIndex(
-        (m) =>
-          m.role === 'user' &&
-          m.messageId.startsWith('user-') &&
-          m.blocks[0]?.content === action.text,
-      );
-      if (optimisticIdx !== -1) {
-        const updated = [...state.messages];
-        updated[optimisticIdx] = { ...updated[optimisticIdx], messageId: action.messageId };
-        return { ...state, messages: updated };
-      }
-      // No match — add as new (reconnect/reattach scenario).
+      // No match — add as new (reconnect where no optimistic message exists).
       return {
         ...state,
         messages: [
@@ -421,7 +414,7 @@ export function chatMessagesReducer(
         messages: [
           ...state.messages,
           {
-            messageId: `user-${Date.now()}`,
+            messageId: action.clientMsgId,
             role: 'user',
             blocks: [],
             images: action.images,
@@ -431,7 +424,7 @@ export function chatMessagesReducer(
               ? {
                   blocks: [
                     {
-                      blockId: `user-text-${Date.now()}`,
+                      blockId: `user-text-${action.clientMsgId}`,
                       blockType: 'text' as BlockType,
                       content: action.text,
                     },
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -19,6 +19,10 @@ import { useVoice } from '../hooks/useVoice';
 import { useAutoSpeak } from '../hooks/useAutoSpeak';
 import type { FinishedBlock, ImageAttachment } from '../types/chat';
 
+function generateClientMsgId(): string {
+  return `user-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+}
+
 export function ChatView() {
   const { sessionId } = useParams<{ sessionId?: string }>();
   const [searchParams] = useSearchParams();
@@ -134,10 +138,6 @@ export function ChatView() {
   });
 
   const hasStarted = msgState.messages.some((m) => m.role === 'user');
-
-  function generateClientMsgId(): string {
-    return `user-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
-  }
 
   function buildSendPayload(
     text: string,
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -135,10 +135,19 @@ export function ChatView() {
 
   const hasStarted = msgState.messages.some((m) => m.role === 'user');
 
-  function buildSendPayload(text: string, images?: ImageAttachment[]): Record<string, unknown> {
+  function generateClientMsgId(): string {
+    return `user-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+  }
+
+  function buildSendPayload(
+    text: string,
+    clientMsgId: string,
+    images?: ImageAttachment[],
+  ): Record<string, unknown> {
     const payload: Record<string, unknown> = {
       type: 'send',
       prompt: text,
+      clientMsgId,
       model: sessionState.model,
       mode: sessionState.mode,
     };
@@ -167,19 +176,20 @@ export function ChatView() {
     // Stop TTS playback when user sends a new message
     voice.stopSpeaking();
 
-    const payload = buildSendPayload(text, images);
+    const clientMsgId = generateClientMsgId();
+    const payload = buildSendPayload(text, clientMsgId, images);
     if (contextBlocks?.length) payload.contextBlocks = contextBlocks;
     const previews = images?.map((img) => img.preview);
 
     if (msgState.running) {
       // Server queues it natively — no client-side stop+re-send needed.
       wsSend(poolKey, payload);
-      dispatch({ type: 'USER_SEND', text, images: previews, contextBlocks });
+      dispatch({ type: 'USER_SEND', text, clientMsgId, images: previews, contextBlocks });
       forceScrollToBottom();
     } else {
       wsSetRunning(poolKey, true);
       wsSend(poolKey, payload);
-      dispatch({ type: 'USER_SEND', text, images: previews, contextBlocks });
+      dispatch({ type: 'USER_SEND', text, clientMsgId, images: previews, contextBlocks });
       forceScrollToBottom();
     }
 
@@ -192,15 +202,17 @@ export function ChatView() {
     contextBlocks?: string[],
   ): void {
     if (!wsIsOpen(poolKey) || !msgState.running) return;
+    const clientMsgId = generateClientMsgId();
     const imagePayload = images?.map((img) => ({ data: img.data, mediaType: img.mediaType }));
     const previews = images?.map((img) => img.preview);
     wsSend(poolKey, {
       type: 'interrupt',
       prompt: text,
+      clientMsgId,
       images: imagePayload,
       ...(contextBlocks?.length ? { contextBlocks } : {}),
     });
-    dispatch({ type: 'USER_SEND', text, images: previews, contextBlocks });
+    dispatch({ type: 'USER_SEND', text, clientMsgId, images: previews, contextBlocks });
     forceScrollToBottom();
   }
 
--- a/server/__tests__/bundled-skills.test.ts
+++ b/server/__tests__/bundled-skills.test.ts
@@ -101,15 +101,6 @@ describe('bundled skills', () => {
     expect(prReview!.allowedTools).not.toContain('Edit');
   });
 
-  it('/review-response includes Bash for gh CLI access', () => {
-    const reviewResponse = skills.find((s) => s.name === 'review-response');
-    expect(reviewResponse).toBeDefined();
-    expect(reviewResponse!.allowedTools).toBeDefined();
-    expect(reviewResponse!.allowedTools).toContain('Bash');
-    expect(reviewResponse!.allowedTools).not.toContain('Write');
-    expect(reviewResponse!.allowedTools).not.toContain('Edit');
-  });
-
   it('all bundled skills are scoped as bundled', () => {
     for (const skill of skills) {
       expect(skill.scope).toBe('bundled');
--- a/server/__tests__/bundled-skills.test.ts
+++ b/server/__tests__/bundled-skills.test.ts
@@ -12,9 +12,9 @@ describe('bundled skills', () => {
   const registry = new SkillRegistry({ bundledDir: SKILLS_DIR });
   const skills = registry.list();
 
-  it('discovers all three bundled skills', () => {
+  it('discovers all four bundled skills', () => {
     const names = skills.map((s) => s.name).sort();
-    expect(names).toEqual(['pr-review', 'risk-scan', 'simplify']);
+    expect(names).toEqual(['pr-review', 'review-response', 'risk-scan', 'simplify']);
   });
 
   it('has unique names', () => {
@@ -82,6 +82,15 @@ describe('bundled skills', () => {
     expect(riskScan!.allowedTools).not.toContain('Edit');
   });
 
+  it('/review-response includes Bash for gh CLI access', () => {
+    const reviewResponse = skills.find((s) => s.name === 'review-response');
+    expect(reviewResponse).toBeDefined();
+    expect(reviewResponse!.allowedTools).toBeDefined();
+    expect(reviewResponse!.allowedTools).toContain('Bash');
+    expect(reviewResponse!.allowedTools).not.toContain('Write');
+    expect(reviewResponse!.allowedTools).not.toContain('Edit');
+  });
+
   it('/pr-review includes Bash for git diff access', () => {
     const prReview = skills.find((s) => s.name === 'pr-review');
     expect(prReview).toBeDefined();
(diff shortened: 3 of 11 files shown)


dimakis (commented):
## Centaur Review

### `frontend/src/hooks/__tests__/useChatMessages.test.ts`
_Good coverage of the core optimistic-dedup flow; minor gaps around multi-optimistic and whitespace-normalization edge cases._

- 🔵 **missing_tests**: No test covers the case where multiple optimistic `user-*` messages exist and the server echo matches one of them — could the dedup logic accidentally upgrade the wrong optimistic message? `[fixable]`
- 🟡 **unsafe_assumptions** (L410): Dedup matching relies solely on exact text equality. If the server normalizes whitespace or trims the message, the optimistic and echoed texts won't match and the message will be duplicated instead of upgraded. `[fixable]`
- 🔵 **style** (L392): The inline comment '// umsg-100 is a server ID (not optimistic user-*), so same text should still add' is helpful but would be clearer as the test name itself. Consider making the `it` description capture this nuance so the comment isn't needed. `[fixable]`


### `frontend/src/hooks/useChatMessages.ts`
_Content-based dedup is the right idea but the text-equality heuristic is too loose — it will misfire on duplicate sends and multi-block messages; add a nonce/correlation ID instead._

- 🔴 **bugs** (L387): Content match only checks `m.blocks[0]?.content`, so multi-block messages (e.g. text + image attachment) will match any message whose first block shares the same text, producing a false-positive dedup and swallowing the new message. `[fixable]`
- 🟡 **unsafe_assumptions** (L387): Matching on `blocks[0]?.content === action.text` assumes the server echo carries a flat `text` field while the client stores content in a `blocks` array. If the user sends the same text twice in quick succession (before the first server echo arrives), the second echo will match the first optimistic message instead of the second, corrupting message ordering. `[fixable]`
- 🟡 **unsafe_assumptions** (L386): `m.messageId.startsWith('user-')` is a fragile convention for detecting optimistic messages. If the ID prefix changes or a server-assigned ID happens to start with 'user-', the guard breaks. Consider a dedicated `optimistic: boolean` flag on the message object. `[fixable]`
- 🟡 **missing_tests**: No tests cover the new content-match dedup path. Key cases needed: (1) optimistic message upgraded on server echo, (2) duplicate user text sent twice, (3) multi-block message, (4) no optimistic match falls through to append. `[fixable]`
- 🔵 **style** (L384): The `findIndex` + array clone + indexed update is fine but could be extracted into a small helper (e.g. `replaceMessageId`) to keep the reducer cases scannable. `[fixable]`

frontend/src/hooks/useChatMessages.ts
```diff
@@ -417,7 +424,7 @@ export function chatMessagesReducer(
               ? {
```
dimakis: 🔵 **style**: `blockId: \`user-text-${action.clientMsgId}\`` derives a second ID by string concatenation. Consider a consistent scheme (e.g., the caller provides both IDs, or use a helper) to avoid ad-hoc ID construction scattered across the reducer. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -135,10 +135,19 @@ export function ChatView() {
 
   const hasStarted = msgState.messages.some((m) => m.role === 'user');
```
dimakis: 🟡 **unsafe_assumptions**: `Date.now()` + 6 random chars gives ~2B × ~2M ≈ 4×10¹⁵ combinations — practically sufficient, but `Math.random()` is not cryptographically secure and two rapid calls in the same millisecond could theoretically collide. Consider `crypto.randomUUID()` for a truly unique ID with zero collision risk. `[fixable]`

dimakis: 🔵 **style**: `generateClientMsgId` is a pure utility with no dependency on component state — extract it out of the component body to avoid re-creation on every render. `[fixable]`

server/__tests__/bundled-skills.test.ts
```diff
@@ -92,6 +101,15 @@ describe('bundled skills', () => {
     expect(prReview!.allowedTools).not.toContain('Edit');
```
dimakis: 🔴 **bugs**: Duplicate test: '/review-response includes Bash for gh CLI access' is added twice (lines 85-91 and lines 104-110). The second copy is identical and should be removed. `[fixable]`

server/__tests__/send-to-chat.test.ts
```diff
@@ -48,9 +48,35 @@ describe('sendToChat emits user_message via WebSocket', () => {
       type: 'user_message',
       text: 'Hello from user',
     });
+    // Falls back to server-generated umsg-* when no clientMsgId provided
```
dimakis: 🔵 **style**: The added comment '// Falls back to server-generated umsg-* when no clientMsgId provided' is helpful but was added to a pre-existing test rather than as part of the new test — minor noise in the diff. `[fixable]`

server/__tests__/send-to-chat.test.ts
```diff
@@ -48,9 +48,35 @@ describe('sendToChat emits user_message via WebSocket', () => {
...
+    registry.register(CLIENT_ID, {
+      ws,
+      abortController: new AbortController(),
+      mode: 'agent',
+      sessionAllowList: new Set(),
+    });
+
+    const session = registry.get(CLIENT_ID)!;
+    session.sessionId = 'sess-client-id';
+    session.inputQueue = { push: pushSpy, close: vi.fn() };
```
dimakis: 🔵 **unsafe_assumptions**: sendToChat is called with positional undefined args (sendToChat(CLIENT_ID, 'Hello', undefined, undefined, 'user-1234-abc')). If the function signature changes or parameters are reordered, these tests will silently pass the wrong values. Consider using a named-options pattern or at minimum adding a comment naming the skipped params. `[fixable]`

server/__tests__/ws-schemas.test.ts
```diff
@@ -6,13 +6,22 @@ describe('WS message schemas', () => {
     const result = IncomingWsMessage.safeParse({
       type: 'send',
       prompt: 'hello',
+      clientMsgId: 'user-1-abc',
       model: 'claude-sonnet-4-6',
       mode: 'agent',
```
dimakis: 🔵 **style**: The 'rejects send without clientMsgId' test also omits model/mode fields. It's unclear whether the parse fails due to missing clientMsgId or missing model. Include model/mode so the test isolates the clientMsgId requirement. `[fixable]`

server/chat.ts
```diff
@@ -452,11 +453,12 @@ export function sendToChat(
   prompt: string,
   images?: Array<{ data: string; mediaType: string }>,
   contextBlocks?: string[],
+  clientMsgId?: string,
 ): boolean {
   const session = registry.get(clientId);
   if (!session?.inputQueue) return false;
```
dimakis: 🟡 **unsafe_assumptions**: clientMsgId is accepted from the client without validation. A malicious or buggy client could send duplicate, empty, or adversarially crafted IDs (e.g., colliding with the `umsg-` prefix convention). Consider validating format/length or at minimum rejecting empty strings. `[fixable]`

dimakis (commented):
## Centaur Review

Found **36** issue(s) (3 critical) (17 warning).


### `frontend/src/hooks/__tests__/chatMessagesReducer.test.ts`
_Tests correctly update USER_SEND payloads with the new required field, but critically miss testing the deduplication behavior the PR is meant to introduce._

- 🟡 **missing_tests**: No test covers the actual deduplication behavior — i.e., dispatching a server echo message with a matching clientMsgId and asserting the optimistic message is replaced rather than duplicated. The existing tests only add clientMsgId to USER_SEND actions but never exercise the dedup logic the PR title describes. `[fixable]`
- 🔵 **unsafe_assumptions**: clientMsgId values are hardcoded static strings (e.g. 'user-1-abc'). If production code generates UUIDs or uses a specific format, consider using a helper or at least a comment indicating these are arbitrary test fixtures, to avoid false confidence in format validation. `[fixable]`


### `frontend/src/hooks/__tests__/useChatMessages.test.ts`
_Clean, well-structured tests covering the key dedup scenarios; consider adding a server-assigned-ID case to exercise the text-based fallback path._

- 🔵 **missing_tests**: No test for the case where the server echo arrives with a different messageId than the clientMsgId (server assigns its own ID). This is a common real-world scenario that would verify the text-based fallback dedup path. `[fixable]`
- 🔵 **style** (L398): The smart quote in "Yep. Let's go" (curly apostrophe U+2019) is intentional test data but could silently break if the reducer does any normalization or encoding. Consider using a plain ASCII apostrophe for robustness, or adding a comment that the smart quote is deliberate. `[fixable]`


### `frontend/src/hooks/useChatMessages.ts`
_Solid dedup fix; verify caller generates collision-resistant IDs and add reducer tests for the dedup path._

- 🟡 **unsafe_assumptions**: The diff doesn't show where `clientMsgId` is generated by the caller. If it's generated via `crypto.randomUUID()` or similar, it's fine. If it reuses `Date.now()`, the dedup could still collide on rapid sends. Verify the caller generates a truly unique ID.
- 🟡 **missing_tests**: No tests visible for the reducer. Key cases to cover: (1) USER_SEND followed by USER_MESSAGE_RECEIVED with same clientMsgId deduplicates, (2) USER_MESSAGE_RECEIVED with unknown ID adds the message, (3) two rapid USER_SENDs produce distinct IDs. `[fixable]`
- 🔵 **style** (L424): `blockId: \`user-text-${action.clientMsgId}\`` derives a second ID by string concatenation. Consider a consistent scheme (e.g., the caller provides both IDs, or use a helper) to avoid ad-hoc ID construction scattered across the reducer. `[fixable]`


_ID generation approach is reasonable but the actual dedup logic (reducer side) is not visible in this diff — verify it exists and is tested._

- 🟡 **unsafe_assumptions** (L137): `Date.now()` + 6 random chars gives ~2B × ~2M ≈ 4×10¹⁵ combinations — practically sufficient, but `Math.random()` is not cryptographically secure and two rapid calls in the same millisecond could theoretically collide. Consider `crypto.randomUUID()` for a truly unique ID with zero collision risk. `[fixable]`
- 🟡 **regressions**: The diff adds `clientMsgId` to the `USER_SEND` dispatch but doesn't show the reducer/state side. If the reducer doesn't store `clientMsgId` on the optimistic message and the dedup logic isn't implemented, this is dead data and the duplication bug isn't actually fixed by this diff alone.
- 🟡 **missing_tests**: No tests for `generateClientMsgId` uniqueness or format, and no test verifying that the reducer deduplicates an echoed server message against the optimistic message with the same `clientMsgId`.
- 🔵 **style** (L137): `generateClientMsgId` is a pure utility with no dependency on component state — extract it out of the component body to avoid re-creation on every render. `[fixable]`


_Duplicate test block must be removed; otherwise the new skill coverage looks correct._

- 🔴 **bugs** (L101): Duplicate test: '/review-response includes Bash for gh CLI access' is added twice (lines 85-91 and lines 104-110). The second copy is identical and should be removed. `[fixable]`
- 🔵 **style**: PR title mentions 'deduplicate optimistic user message against server echo' but the diff only adds a new bundled skill to tests — title/diff mismatch suggests either wrong file or wrong PR description.


_Tests verify clientMsgId passthrough but miss the actual deduplication behavior promised by the PR title._

- 🟡 **missing_tests**: No test for edge cases around clientMsgId: what happens when clientMsgId is empty string, or contains invalid characters? If the dedup logic relies on this ID, malformed values could cause subtle bugs. `[fixable]`
- 🟡 **missing_tests**: No test verifying that duplicate messages with the same clientMsgId are actually deduplicated. The PR title mentions deduplication, but these tests only verify the ID passthrough — the core dedup behavior is untested in this file. `[fixable]`
- 🔵 **style** (L51): The added comment '// Falls back to server-generated umsg-* when no clientMsgId provided' is helpful but was added to a pre-existing test rather than as part of the new test — minor noise in the diff. `[fixable]`
- 🔵 **unsafe_assumptions** (L68): sendToChat is called with positional undefined args (sendToChat(CLIENT_ID, 'Hello', undefined, undefined, 'user-1234-abc')). If the function signature changes or parameters are reordered, these tests will silently pass the wrong values. Consider using a named-options pattern or at minimum adding a comment naming the skipped params. `[fixable]`


### `server/__tests__/ws-schemas.test.ts`
_Solid coverage for the happy path; add a rejection test for interrupt without clientMsgId and tighten the negative test to isolate the missing field._

- 🟡 **missing_tests**: No test verifies that 'interrupt' type also requires clientMsgId — only 'send' has a rejection test. Add a 'rejects interrupt without clientMsgId' case for symmetry. `[fixable]`
- 🔵 **missing_tests**: No test validates the shape/format of clientMsgId (e.g. empty string, non-string types). If the schema enforces a pattern or min-length, those constraints should be covered. `[fixable]`
- 🔵 **style** (L11): The 'rejects send without clientMsgId' test also omits model/mode fields. It's unclear whether the parse fails due to missing clientMsgId or missing model. Include model/mode so the test isolates the clientMsgId requirement. `[fixable]`


### `server/chat.ts`
_Straightforward ID-passthrough change; main concerns are missing input validation on the client-supplied ID and no test coverage._

- 🟡 **unsafe_assumptions** (L459): clientMsgId is accepted from the client without validation. A malicious or buggy client could send duplicate, empty, or adversarially crafted IDs (e.g., colliding with the `umsg-` prefix convention). Consider validating format/length or at minimum rejecting empty strings. `[fixable]`
- 🔵 **regressions** (L309): clientMsgId is added to the startChat options type but there's no corresponding usage shown in startChat's body. If startChat also generates message IDs internally, the dedup intent may be incomplete — callers of startChat won't benefit from the same ID passthrough. `[fixable]`
- 🟡 **missing_tests**: No tests visible for: (1) clientMsgId passthrough producing the expected messageId, (2) fallback to Date.now()-based ID when clientMsgId is undefined, (3) behavior when clientMsgId is an empty string (falsy but not undefined). `[fixable]`
- 🔵 **unsafe_assumptions** (L459): The nullish coalescing (`??`) means an empty string `""` will be used as the messageId rather than falling back to the generated ID. If the intent is to treat any falsy value as "not provided", use `||` instead of `??`. `[fixable]`


### `server/index.ts`
_Straightforward plumbing of `clientMsgId` through all message paths — looks correct in isolation, but verify the receiving functions accept the new parameter and add tests for the dedup behavior._

- 🟡 **regressions**: The diff adds `clientMsgId` as a new parameter to `sendToChat`, `startChat`, and `interruptChat`, but their function signatures are not shown. If those functions don't accept this parameter yet, this will cause a type error or silently ignore the value. Verify that all three functions have been updated to handle `clientMsgId`. `[fixable]`
- 🟡 **unsafe_assumptions**: `msg.clientMsgId` is passed through without validation. If a malicious or buggy client sends a non-string or excessively long value, it could propagate into downstream message handling. Ensure the incoming WebSocket message schema validates `clientMsgId` as an optional string with a bounded length. `[fixable]`
- 🟡 **missing_tests**: No tests visible for the deduplication behavior. Key cases to cover: (1) server echo with matching `clientMsgId` is suppressed on the client, (2) messages without `clientMsgId` still work (backward compat), (3) `clientMsgId` is correctly forwarded through `sendToChat`, `startChat`, and `interruptChat` paths. `[fixable]`
- 🔵 **style**: The `clientMsgId` parameter is added as a positional argument to `sendToChat` and `interruptChat` (now 5 positional args). Consider using an options object to improve readability and avoid parameter ordering mistakes. `[fixable]`


### `server/ws-schemas.ts`
_Adding required fields to existing message schemas is a breaking change — consider making clientMsgId optional or coordinating a synchronized client+server deploy._

- 🔴 **regressions** (L17): Making clientMsgId required (non-optional) in SendMessage is a breaking change. Any existing client that sends a 'send' message without clientMsgId will now fail Zod validation. Ensure all clients are updated simultaneously, or make this field optional (.optional()) during a migration period. `[fixable]`
- 🔴 **regressions** (L31): Same breaking change for InterruptMessage — adding a required field breaks existing clients that don't send clientMsgId. `[fixable]`
- 🟡 **unsafe_assumptions** (L17): clientMsgId is validated only as a non-empty string. If this is used for deduplication, consider adding .uuid() or .max(128) to prevent abuse (e.g., megabyte-sized strings) and enforce a consistent format. `[fixable]`
- 🟡 **missing_tests**: No tests visible for the updated schemas. Add validation tests confirming: (1) messages with clientMsgId pass, (2) messages without clientMsgId are rejected, (3) messages with empty-string clientMsgId are rejected. `[fixable]`


### `skills/review-response/SKILL.md`
_Well-structured skill definition with clear triage workflow; main concerns are placeholder resolution in gh API calls and a mismatch between PR title and file content._

- 🟡 **regressions** (L12): The gh api call uses `{owner}/{repo}` as literal placeholders but doesn't specify how to resolve them. Should either use `gh api repos/:owner/:repo/...` (gh's built-in placeholder syntax) or instruct the agent to detect owner/repo from `gh repo view --json owner,name`. `[fixable]`
- 🔵 **unsafe_assumptions** (L12): Fetches `/reviews` and `/comments` but doesn't mention `/pulls/{number}/reviews/{id}/comments` for threaded review comments. Depending on the review style, some comments may be missed. Consider also fetching issue comments via `/issues/{number}/comments` for top-level PR conversation comments. `[fixable]`
- 🔵 **style** (L1): PR title says 'fix(chat): deduplicate optimistic user message against server echo' but this file is a review-response skill definition — the PR title and file content are completely unrelated. This suggests either a wrong file in the PR or a misleading PR title.
- 🔵 **style** (L10): `$ARGUMENTS` is referenced but never defined in the frontmatter or description. If this is a Claude Code skill convention, it's fine, but a brief inline comment like `<!-- $ARGUMENTS = PR number or empty -->` would aid readability. `[fixable]`

server/chat.ts
```diff
@@ -306,6 +306,7 @@ export async function startChat(
     worktree?: boolean;
     images?: Array<{ data: string; mediaType: string }>;
     contextBlocks?: string[];
+    clientMsgId?: string;
```
dimakis: 🔵 **regressions**: clientMsgId is added to the startChat options type but there's no corresponding usage shown in startChat's body. If startChat also generates message IDs internally, the dedup intent may be incomplete — callers of startChat won't benefit from the same ID passthrough. `[fixable]`

server/chat.ts
dimakis: 🔵 **unsafe_assumptions**: The nullish coalescing (`??`) means an empty string `""` will be used as the messageId rather than falling back to the generated ID. If the intent is to treat any falsy value as "not provided", use `||` instead of `??`. `[fixable]`

server/ws-schemas.ts
```diff
@@ -14,6 +14,7 @@ export const ReattachMessage = z.object({
 export const SendMessage = z.object({
   type: z.literal('send'),
   prompt: z.string().min(1),
+  clientMsgId: z.string().min(1),
```
dimakis: 🔴 **regressions**: Making clientMsgId required (non-optional) in SendMessage is a breaking change. Any existing client that sends a 'send' message without clientMsgId will now fail Zod validation. Ensure all clients are updated simultaneously, or make this field optional (.optional()) during a migration period. `[fixable]`

server/ws-schemas.ts
```diff
@@ -27,6 +28,7 @@ export const SendMessage = z.object({
 export const InterruptMessage = z.object({
   type: z.literal('interrupt'),
   prompt: z.string().min(1),
+  clientMsgId: z.string().min(1),
```
dimakis: 🔴 **regressions**: Same breaking change for InterruptMessage — adding a required field breaks existing clients that don't send clientMsgId. `[fixable]`

server/ws-schemas.ts
dimakis: 🟡 **unsafe_assumptions**: clientMsgId is validated only as a non-empty string. If this is used for deduplication, consider adding .uuid() or .max(128) to prevent abuse (e.g., megabyte-sized strings) and enforce a consistent format. `[fixable]`

skills/review-response/SKILL.md
```diff
@@ -0,0 +1,35 @@
...
+allowed-tools:
+  - Read
+  - Glob
+  - Grep
+  - Bash
+---
+
+Triage all review comments on $ARGUMENTS.
+
+If `$ARGUMENTS` is a PR number, fetch comments with `gh api repos/{owner}/{repo}/pulls/{number}/reviews` and `gh api repos/{owner}/{repo}/pulls/{number}/comments`. If no number is given, detect the current branch's open PR with `gh pr view --json number,url`.
```
dimakis: 🟡 **regressions**: The gh api call uses `{owner}/{repo}` as literal placeholders but doesn't specify how to resolve them. Should either use `gh api repos/:owner/:repo/...` (gh's built-in placeholder syntax) or instruct the agent to detect owner/repo from `gh repo view --json owner,name`. `[fixable]`

dimakis: 🔵 **unsafe_assumptions**: Fetches `/reviews` and `/comments` but doesn't mention `/pulls/{number}/reviews/{id}/comments` for threaded review comments. Depending on the review style, some comments may be missed. Consider also fetching issue comments via `/issues/{number}/comments` for top-level PR conversation comments. `[fixable]`

skills/review-response/SKILL.md
```diff
@@ -0,0 +1,35 @@
+---
```
dimakis: 🔵 **style**: PR title says 'fix(chat): deduplicate optimistic user message against server echo' but this file is a review-response skill definition — the PR title and file content are completely unrelated. This suggests either a wrong file in the PR or a misleading PR title.

skills/review-response/SKILL.md
```diff
@@ -0,0 +1,35 @@
...
+description: Triage PR review comments, investigate each finding, and fix properly
+allowed-tools:
+  - Read
+  - Glob
+  - Grep
+  - Bash
+---
+
+Triage all review comments on $ARGUMENTS.
```
dimakis: 🔵 **style**: `$ARGUMENTS` is referenced but never defined in the frontmatter or description. If this is a Claude Code skill convention, it's fine, but a brief inline comment like `<!-- $ARGUMENTS = PR number or empty -->` would aid readability. `[fixable]`

PR #154, opened 2026-04-09
refactor(server): load repo config dynamically

dimakis:
## Summary
- Replace startup-cached `repoConfig` with `getRepoConfig()` function that reads `.mitzo.json` fresh on each call
- Config changes now take effect without server restart
- Includes Centaur review fix: isolated try/catch for repo config in worktree cleanup

## Context
Revived from closed PR #144 (rebased onto main, incorporated fix PR #146).

## Test plan
- [x] All 717 tests pass
- [ ] Verify config changes are picked up without restart

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -18,7 +18,7 @@ import {
   AVAILABLE_MODELS,
   BASE_REPO,
   getMcpServerNames,
-  repoConfig,
+  getRepoConfig,
   eventStore,
   registry,
 } from './chat.js';
@@ -209,7 +209,7 @@ app.get('/api/repos', (req, res) => {
     res.status(401).json({ error: 'Internal token required' });
     return;
   }
-  res.json(Object.entries(repoConfig.repos).map(([name, path]) => ({ name, path })));
+  res.json(Object.entries(getRepoConfig().repos).map(([name, path]) => ({ name, path })));
 });
 
 app.post('/api/repos/open', (req, res) => {
@@ -225,9 +225,10 @@ app.post('/api/repos/open', (req, res) => {
     return;
   }
 
-  const repoPath = repoConfig.repos[repoName];
+  const config = getRepoConfig();
+  const repoPath = config.repos[repoName];
   if (!repoPath) {
-    const available = Object.keys(repoConfig.repos).join(', ');
+    const available = Object.keys(config.repos).join(', ');
     res.status(404).json({ error: `Unknown repo "${repoName}". Available: ${available}` });
     return;
   }
@@ -308,12 +309,13 @@ app.get('/api/auth/check', (_req, res) => res.json({ ok: true }));
 app.get('/api/models', (_req, res) => res.json(AVAILABLE_MODELS));
 
 app.get('/api/config', (_req, res) => {
-  const actions = repoConfig.quickActions.map((a) => ({
+  const config = getRepoConfig();
+  const actions = config.quickActions.map((a) => ({
     ...a,
     cwd: a.cwd ? join(BASE_REPO, a.cwd) : undefined,
   }));
   const contextBlocks: Record<string, { path: string; sizeBytes: number }> = {};
-  for (const [name, path] of Object.entries(repoConfig.contextBlocks)) {
+  for (const [name, path] of Object.entries(config.contextBlocks)) {
     let sizeBytes = 0;
     try {
       sizeBytes = statSync(path).size;
@@ -384,7 +386,7 @@ app.put('/api/sessions/:id/rename', async (req, res) => {
 
 app.get('/api/worktrees', (_req, res) => {
   const worktrees = listWorktrees(BASE_REPO).map((wt) => ({ ...wt, repo: 'primary' }));
-  for (const [name, repoPath] of Object.entries(repoConfig.repos)) {
+  for (const [name, repoPath] of Object.entries(getRepoConfig().repos)) {
     try {
       const repoWts = listWorktrees(repoPath).map((wt) => ({ ...wt, repo: name }));
       worktrees.push(...repoWts);
@@ -401,11 +403,12 @@ export function isAllowedPath(filePath: string): boolean {
   const resolved = resolve(filePath);
   if (BASE_REPO && resolved.startsWith(resolve(BASE_REPO))) return true;
   if (BASE_REPO && resolved.startsWith(resolve(`${BASE_REPO}-sessions`))) return true;
-  for (const repoPath of Object.values(repoConfig.repos)) {
+  const config = getRepoConfig();
+  for (const repoPath of Object.values(config.repos)) {
     if (resolved.startsWith(resolve(repoPath))) return true;
     if (resolved.startsWith(resolve(`${repoPath}-sessions`))) return true;
   }
-  for (const extra of repoConfig.allowedPaths) {
+  for (const extra of config.allowedPaths) {
     if (resolved.startsWith(resolve(extra))) return true;
   }
   return false;
@@ -439,7 +442,7 @@ app.get('/api/git/info', (_req, res) => {
     branch: getGitBranch(wt.path),
     repo: 'primary',
   }));
-  for (const [name, repoPath] of Object.entries(repoConfig.repos)) {
+  for (const [name, repoPath] of Object.entries(getRepoConfig().repos)) {
     try {
       const repoWts = listWorktrees(repoPath).map((wt) => ({
         ...wt,
@@ -455,7 +458,7 @@ app.get('/api/git/info', (_req, res) => {
 });
 
 app.get('/api/files/roots', (_req, res) => {
-  res.json(repoConfig.roots);
+  res.json(getRepoConfig().roots);
 });
 
 app.get('/api/files', (req, res) => {
@@ -548,7 +551,7 @@ app.put('/api/files/write', (req, res) => {
 // --- Inbox API ---
 
 app.get('/api/inbox', (_req, res) => {
-  const inboxPath = repoConfig.resolvedInboxPath;
+  const inboxPath = getRepoConfig().resolvedInboxPath;
   if (!inboxPath) {
     res.json([]);
     return;
@@ -557,7 +560,7 @@ app.get('/api/inbox', (_req, res) => {
 });
 
 app.post('/api/inbox', (req, res) => {
-  const inboxPath = repoConfig.resolvedInboxPath;
+  const inboxPath = getRepoConfig().resolvedInboxPath;
   if (!inboxPath) {
     res.status(404).json({ error: 'Inbox not configured' });
     return;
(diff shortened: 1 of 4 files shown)


dimakis:
## Centaur Review

Found **12** issue(s) (1 critical) (7 warning).


### `server/app.ts`
_Straightforward mechanical refactor from static import to dynamic getter; main concerns are inconsistent caching of the getter result and missing test coverage for the dynamic-reload behavior._

- 🟡 **unsafe_assumptions** (L386): getRepoConfig() is called inline inside a loop iterator (`Object.entries(getRepoConfig().repos)`). If getRepoConfig() is expensive (e.g., reads from disk each call), this is fine for correctness but wasteful. More importantly, the pattern is inconsistent: some handlers cache the result in a local `const config = getRepoConfig()` (lines 228, 312, 405) while others call it inline (lines 212, 389, 445, 461, 554, 563, 585, 599, 614). The inline calls in `/api/worktrees` and `/api/git/info` iterate over the result, so if `getRepoConfig()` ever returns a fresh object each call, the behavior is still correct — but the inconsistency hurts readability. `[fixable]`
- 🔵 **style** (L389): For consistency with the pattern used in other handlers (lines 228, 312, 405), extract `const config = getRepoConfig()` into a local variable in the `/api/worktrees`, `/api/git/info`, `/api/repos`, and `/api/files/roots` handlers rather than calling inline. `[fixable]`
- 🟡 **missing_tests**: No tests appear to cover that config is now loaded dynamically per-request. A key benefit of this refactor is that config changes are picked up without restart — this should be validated by a test that mutates the underlying config between requests and asserts the new value is reflected.
- 🔵 **regressions** (L401): `isAllowedPath` is a security-sensitive exported function (path traversal guard). The change from a static `repoConfig` to `getRepoConfig()` means every call re-reads config. Verify that `getRepoConfig()` cannot return a partially-constructed or empty config object (e.g., if the config file is being written to concurrently), as that could momentarily widen or narrow the set of allowed paths.


### `server/chat.ts`
_Dynamic reload is a good goal but the current approach re-reads config multiple times per request (latency + TOCTOU risk) and loses the override-reset path; read once per session start and pass the config object through instead._

- 🔴 **unsafe_assumptions** (L94): `getRepoConfig()` calls `loadRepoConfig()` on every invocation. `sdkEnv()`, `buildMcpAllowedTools()`, `buildRepoMcpServer()`, and `buildRepoSystemPrompt()` are all called during session setup, resulting in 4+ redundant file reads per chat start. If `loadRepoConfig` does synchronous I/O (likely, given `loadRepoConfig` naming), this adds latency and creates a TOCTOU inconsistency: config can change between calls within the same request, causing e.g. `buildMcpAllowedTools` to include repo tools while `buildRepoMcpServer` returns null if the file is edited mid-request. Consider reading once per session start and threading the config object through. `[fixable]`
- 🟡 **regressions** (L316): `applyTierOverrides` is now called on every `startChat` invocation but there is no corresponding reset/clear before applying. If `.mitzo.json` removes an override between sessions, the old override persists in-process because `applyTierOverrides` likely only adds/sets, never unsets. The original code applied overrides once at startup which was consistent; now repeated application without a reset can accumulate stale state. `[fixable]`
- 🟡 **regressions** (L316): The startup log `log.info('applied tool tier overrides from .mitzo.json', ...)` was removed entirely. Operators lose visibility into whether overrides are active. Consider logging inside `startChat` when overrides are applied. `[fixable]`
- 🟡 **missing_tests**: No tests cover the new dynamic-reload behavior: (1) config changes between calls are picked up, (2) `applyTierOverrides` is called per-session, (3) `getRepoConfig` is invoked the expected number of times. Unit tests with a mocked `loadRepoConfig` would prevent regressions. `[fixable]`
- 🔵 **style** (L69): `getRepoConfig()` is exported but is a thin wrapper with no caching. The name suggests a cheap getter but it performs file I/O. Consider naming it `loadRepoConfig()` or adding a brief JSDoc noting the I/O cost, so callers don't treat it as free. `[fixable]`


### `server/index.ts`
_Clean refactor with appropriate graceful degradation; fix the `let` → `const` nit and consider whether a silent warn is sufficient for config load failures._

- 🟡 **regressions** (L323): Previously, a failing `repoConfig` would prevent server startup (fail-fast). Now it silently swallows the error and only cleans up the primary repo's worktrees. If `getRepoConfig()` fails due to a misconfiguration the operator should notice, a warn log may not be enough — consider whether this should be surfaced more prominently or whether skipping secondary repos is acceptable.
- 🔵 **style** (L319): `repoEntries` is declared with `let` but is only reassigned via `.push()`. Use `const` since the binding itself is never reassigned. `[fixable]`
- 🟡 **missing_tests**: No tests cover the new error-handling path. A test verifying that worktree cleanup still runs for BASE_REPO when `getRepoConfig()` throws would prevent future regressions.

dimakis:
Centaur has opened PR #155 with fixes for the flagged issues.

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Tier override cleanup bug (stale overrides persist when removed from config) and significant performance regression from uncached synchronous file reads on every call — needs a TTL cache or single-read-per-request pattern._

- 🔴 **bugs** (L316): Stale tier overrides persist across sessions. When `.mitzo.json` removes overrides, the `if (Object.keys(...).length > 0)` guard skips `applyTierOverrides`, so the previous session's overrides remain in the module-level `activeTiers`. Fix: call `applyTierOverrides(currentConfig.toolTierOverrides)` unconditionally — the function already resets to `DEFAULT_TOOL_TIERS` before merging. `[fixable]`
- 🟡 **unsafe_assumptions** (L69): `getRepoConfig()` performs synchronous file I/O (`readFileSync` + `JSON.parse` + multiple `existsSync` calls) on every invocation. A single `startChat` call triggers it ~5-6 times (directly + via `sdkEnv`, `buildMcpAllowedTools`, `buildRepoMcpServer`, `buildRepoSystemPrompt`, `assemblePrompt`). Every HTTP request in `app.ts` also triggers a fresh read. Consider a short TTL cache (e.g. 5-10 seconds) to get dynamic reloading without redundant disk I/O on every call. `[fixable]`
- 🟡 **missing_tests** (L69): The core behavioral change — dynamic reload of `.mitzo.json` — has no test coverage. There are no tests verifying that config changes are picked up without restart, or that multiple calls return fresh data. The only test change is updating the mock shape in `routes.test.ts`. `[fixable]`


### `server/app.ts`
_Tier override cleanup bug (stale overrides persist when removed from config) and significant performance regression from uncached synchronous file reads on every call — needs a TTL cache or single-read-per-request pattern._

- 🟡 **unsafe_assumptions** (L403): `isAllowedPath` is called on file-serving requests and does synchronous `getRepoConfig()` (disk read) on every invocation. This is a hot path — a file write or read triggers path validation, which now does a full config reload from disk each time. `[fixable]`


### `server/index.ts`
_Tier override cleanup bug (stale overrides persist when removed from config) and significant performance regression from uncached synchronous file reads on every call — needs a TTL cache or single-read-per-request pattern._

- 🔵 **style** (L319): `repoEntries` is declared with `let` but should be `const` since the array reference is never reassigned (`.push()` mutates the array, not the binding). `[fixable]`

server/chat.ts
```diff
@@ -316,6 +311,12 @@ export async function startChat(
   const { cwd, worktreePath } = resolveWorktree(ws, baseCwd, options);
   const fullPrompt = assemblePrompt(prompt, cwd, options.images, options.contextBlocks);
 
+  // Apply tier overrides from current .mitzo.json (re-read each session start)
+  const currentConfig = getRepoConfig();
+  if (Object.keys(currentConfig.toolTierOverrides).length > 0) {
```
dimakis: 🔴 **bugs**: Stale tier overrides persist across sessions. When `.mitzo.json` removes overrides, the `if (Object.keys(...).length > 0)` guard skips `applyTierOverrides`, so the previous session's overrides remain in the module-level `activeTiers`. Fix: call `applyTierOverrides(currentConfig.toolTierOverrides)` unconditionally — the function already resets to `DEFAULT_TOOL_TIERS` before merging. `[fixable]`

server/chat.ts
```diff
@@ -65,18 +65,11 @@ const MODE_TO_SDK: Record<MitzoMode, string> = {
...
-const repoConfig = loadRepoConfig(BASE_REPO);
-export { repoConfig };
-
-if (Object.keys(repoConfig.toolTierOverrides).length > 0) {
-  applyTierOverrides(repoConfig.toolTierOverrides);
-  log.info('applied tool tier overrides from .mitzo.json', {
-    overrides: repoConfig.toolTierOverrides,
-  });
+/** Load repo config fresh on each call so edits to .mitzo.json take effect without restart. */
+export function getRepoConfig() {
```
dimakis, line 73: 🟡 **unsafe_assumptions**: `getRepoConfig()` performs synchronous file I/O (`readFileSync` + `JSON.parse` + multiple `existsSync` calls) on every invocation. A single `startChat` call triggers it ~5-6 times (directly + via `sdkEnv`, `buildMcpAllowedTools`, `buildRepoMcpServer`, `buildRepoSystemPrompt`, `assemblePrompt`). Every HTTP request in `app.ts` also triggers a fresh read. Consider a short TTL cache (e.g. 5-10 seconds) to get dynamic reloading without redundant disk I/O on every call. `[fixable]`

dimakis, line 73: 🟡 **missing_tests**: The core behavioral change — dynamic reload of `.mitzo.json` — has no test coverage. There are no tests verifying that config changes are picked up without restart, or that multiple calls return fresh data. The only test change is updating the mock shape in `routes.test.ts`. `[fixable]`

server/app.ts
```diff
@@ -401,11 +403,12 @@ export function isAllowedPath(filePath: string): boolean {
   const resolved = resolve(filePath);
```
dimakis, line 403: 🟡 **unsafe_assumptions**: `isAllowedPath` is called on file-serving requests and does synchronous `getRepoConfig()` (disk read) on every invocation. This is a hot path — a file write or read triggers path validation, which now does a full config reload from disk each time. `[fixable]`

server/index.ts
```diff
@@ -316,7 +316,16 @@ checkPort(PORT).then((inUse) => {
     const protocol = USE_TLS ? 'https' : 'http';
     log.info(`Chat Agent running on ${protocol}://localhost:${PORT}${USE_TLS ? ' (TLS)' : ''}`);
     // Clean up stale worktrees across all repos
-    for (const [label, repoPath] of [['primary', BASE_REPO], ...Object.entries(repoConfig.repos)]) {
+    let repoEntries: [string, string][] = [['primary', BASE_REPO]];
```
dimakis: 🔵 **style**: `repoEntries` is declared with `let` but should be `const` since the array reference is never reassigned (`.push()` mutates the array, not the binding). `[fixable]`

PR #158, opened 2026-04-09
feat(context): add contexgin dependency

dimakis:
## Summary
- Adds `contexgin` as a local file dependency (`file:../../projects/contexgin`)
- Smoke tests verify the public API is importable (compile, discoverSources, integrity, navigation)
- No behavioral changes — contexgin is wired but not used yet

## Context
Step 1 of the contexgin integration plan. See dimakis/mitzo#157 for baseline snapshots.

## Test plan
- [x] `npm test` — 726 tests passing (723 + 3 new smoke tests)
- [x] No production code changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #159, opened 2026-04-09
fix: address review findings for PR #158

dimakis:
## Centaur Auto-Fix

Fixes for review findings on PR #158 (`feat(context): add contexgin dependency`).

### Changes
- Replaced `file:../../projects/contexgin` with `github:dimakis/contexgin` to remove the machine-specific hardcoded path and make the dependency portable for any contributor
- Reverted vitest from ^4.1.4 back to ^4.1.2 to keep the unrelated dependency bump out of this feature PR
- Moved the dynamic import into a single beforeAll block to avoid redundant module resolution across test cases
- Added a test that actually invokes compile with trivial input to catch runtime errors beyond typeof checks

---
*Generated by Centaur*

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "@anthropic-ai/vertex-sdk": "^0.15.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
     "better-sqlite3": "^12.8.0",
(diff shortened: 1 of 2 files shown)


dimakis:
Fixed in #158 directly

PR #162, opened 2026-04-10
feat(frontend): desktop three-panel layout with persistent context

dimakis:
## Summary
- Extracts foundation hooks and ChatArea component for reuse
- Adds desktop three-panel layout: SessionPanel (left) | Chat (center) | ContextPanel + FileBrowser (right)
- Collapsible sidebars with localStorage persistence, 768px breakpoint
- Fixes stale message bleed when switching sessions on desktop
- Adds `/api/files/list` endpoint and `fileViewerRoots` to `/api/config` for FileBrowserPanel

## Test plan
- [x] Tests pass (793/793 — includes new tests for desktop components)
- [ ] Manual: verify mobile layout unchanged below 768px
- [ ] Manual: verify desktop layout renders correctly above 768px
- [ ] Manual: verify session switching clears messages

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -3,9 +3,11 @@ import { useState, useEffect } from 'react';
 import { Login } from './pages/Login';
 import { SessionList } from './pages/SessionList';
 import { ChatView } from './pages/ChatView';
+import { DesktopChatView } from './pages/DesktopChatView';
 import { FileViewer } from './pages/FileViewer';
 import { InboxView } from './pages/InboxView';
 import { ErrorBoundary } from './components/ErrorBoundary';
+import { useIsDesktop } from './hooks/useMediaQuery';
 
 function ProtectedRoute({ children }: { children: React.ReactNode }) {
   const [auth, setAuth] = useState<'loading' | 'ok' | 'denied'>('loading');
@@ -21,6 +23,16 @@ function ProtectedRoute({ children }: { children: React.ReactNode }) {
   return <>{children}</>;
 }
 
+function HomeRoute() {
+  const isDesktop = useIsDesktop();
+  return isDesktop ? <DesktopChatView /> : <SessionList />;
+}
+
+function ChatRoute() {
+  const isDesktop = useIsDesktop();
+  return isDesktop ? <DesktopChatView /> : <ChatView />;
+}
+
 export function App() {
   return (
     <ErrorBoundary>
@@ -31,7 +43,9 @@ export function App() {
             path="/"
             element={
               <ProtectedRoute>
-                <SessionList />
+                <ErrorBoundary>
+                  <HomeRoute />
+                </ErrorBoundary>
               </ProtectedRoute>
             }
           />
@@ -40,7 +54,7 @@ export function App() {
             element={
               <ProtectedRoute>
                 <ErrorBoundary>
-                  <ChatView />
+                  <ChatRoute />
                 </ErrorBoundary>
               </ProtectedRoute>
             }
@@ -50,7 +64,7 @@ export function App() {
             element={
               <ProtectedRoute>
                 <ErrorBoundary>
-                  <ChatView />
+                  <ChatRoute />
                 </ErrorBoundary>
               </ProtectedRoute>
             }
--- a/frontend/src/components/ContextPanel.tsx
+++ b/frontend/src/components/ContextPanel.tsx
@@ -1,0 +1,66 @@
+import { useState, useEffect } from 'react';
+import type { ContextBlockEntry } from './ContextPicker';
+
+export interface ContextPanelProps {
+  selected: string[];
+  onToggle: (name: string) => void;
+}
+
+function formatSize(bytes: number): string {
+  if (bytes === 0) return '0 B';
+  if (bytes < 1024) return `${bytes} B`;
+  const kb = bytes / 1024;
+  if (kb < 1024) return `${kb.toFixed(1)} KB`;
+  return `${(kb / 1024).toFixed(1)} MB`;
+}
+
+export function ContextPanel({ selected, onToggle }: ContextPanelProps) {
+  const [blocks, setBlocks] = useState<ContextBlockEntry[]>([]);
+  const [loaded, setLoaded] = useState(false);
+
+  useEffect(() => {
+    fetch('/api/config', { credentials: 'include' })
+      .then((r) => r.json())
+      .then((data: { contextBlocks?: Record<string, { path: string; sizeBytes: number }> }) => {
+        const entries: ContextBlockEntry[] = [];
+        if (data.contextBlocks) {
+          for (const [name, info] of Object.entries(data.contextBlocks)) {
+            entries.push({ name, path: info.path, sizeBytes: info.sizeBytes });
+          }
+        }
+        setBlocks(entries);
+        setLoaded(true);
+      })
+      .catch(() => {
+        setLoaded(true);
+      });
+  }, []);
+
+  if (!loaded) return null;
+
+  return (
+    <div className="context-panel">
+      <div className="context-panel-header">Context</div>
+      {blocks.length === 0 ? (
+        <p className="context-panel-empty">No context blocks</p>
+      ) : (
+        <div className="context-panel-list">
+          {blocks.map((block) => {
+            const isSelected = selected.includes(block.name);
+            return (
+              <button
+                key={block.name}
+                className={`context-panel-item${isSelected ? ' context-panel-item--selected' : ''}`}
+                onClick={() => onToggle(block.name)}
+              >
+                <span className="context-panel-check">{isSelected ? '✓' : ''}</span>
+                <span className="context-panel-name">{block.name}</span>
+                <span className="context-panel-size">{formatSize(block.sizeBytes)}</span>
+              </button>
+            );
+          })}
+        </div>
+      )}
+    </div>
+  );
+}
--- a/frontend/src/components/ContextPanel.tsx
+++ b/frontend/src/components/ContextPanel.tsx
@@ -4,6 +4,8 @@ import type { ContextBlockEntry } from './ContextPicker';
 export interface ContextPanelProps {
   selected: string[];
   onToggle: (name: string) => void;
+  blocks?: ContextBlockEntry[];
+  loaded?: boolean;
 }
 
 function formatSize(bytes: number): string {
@@ -14,11 +16,18 @@ function formatSize(bytes: number): string {
   return `${(kb / 1024).toFixed(1)} MB`;
 }
 
-export function ContextPanel({ selected, onToggle }: ContextPanelProps) {
-  const [blocks, setBlocks] = useState<ContextBlockEntry[]>([]);
-  const [loaded, setLoaded] = useState(false);
+export function ContextPanel({
+  selected,
+  onToggle,
+  blocks: externalBlocks,
+  loaded: externalLoaded,
+}: ContextPanelProps) {
+  const [selfBlocks, setSelfBlocks] = useState<ContextBlockEntry[]>([]);
+  const [selfLoaded, setSelfLoaded] = useState(false);
 
+  // Self-fetch config only when not provided via props (standalone usage)
   useEffect(() => {
+    if (externalBlocks !== undefined) return;
     fetch('/api/config', { credentials: 'include' })
       .then((r) => r.json())
       .then((data: { contextBlocks?: Record<string, { path: string; sizeBytes: number }> }) => {
@@ -28,13 +37,14 @@ export function ContextPanel({ selected, onToggle }: ContextPanelProps) {
             entries.push({ name, path: info.path, sizeBytes: info.sizeBytes });
           }
         }
-        setBlocks(entries);
-        setLoaded(true);
+        setSelfBlocks(entries);
+        setSelfLoaded(true);
       })
-      .catch(() => {
-        setLoaded(true);
-      });
-  }, []);
+      .catch(() => setSelfLoaded(true));
+  }, [externalBlocks]);
+
+  const blocks = externalBlocks ?? selfBlocks;
+  const loaded = externalLoaded ?? selfLoaded;
 
   if (!loaded) return null;
 
--- a/frontend/src/components/FileBrowserPanel.tsx
+++ b/frontend/src/components/FileBrowserPanel.tsx
@@ -1,0 +1,139 @@
+import { useState, useEffect, useCallback } from 'react';
+
+interface DirEntry {
+  name: string;
+  isDir: boolean;
+}
+
+interface FileRoot {
+  label: string;
+  path: string;
+}
+
+export function FileBrowserPanel() {
+  const [roots, setRoots] = useState<FileRoot[]>([]);
+  const [activeRoot, setActiveRoot] = useState('');
+  const [currentDir, setCurrentDir] = useState('');
+  const [entries, setEntries] = useState<DirEntry[]>([]);
+  const [preview, setPreview] = useState<{ name: string; content: string } | null>(null);
+  const [loaded, setLoaded] = useState(false);
+
+  // Fetch roots from config
+  useEffect(() => {
+    fetch('/api/config', { credentials: 'include' })
+      .then((r) => r.json())
+      .then((data: { fileViewerRoots?: FileRoot[] }) => {
+        const r = data.fileViewerRoots ?? [];
+        setRoots(r);
+        if (r.length > 0) {
+          setActiveRoot(r[0].path);
+        }
+        setLoaded(true);
+      })
+      .catch(() => setLoaded(true));
+  }, []);
+
+  // Fetch directory listing when root or dir changes
+  useEffect(() => {
+    if (!activeRoot) return;
+    const params = new URLSearchParams();
+    if (currentDir) params.set('dir', currentDir);
+    else params.set('dir', activeRoot);
+
+    fetch(`/api/files/list?${params}`, { credentials: 'include' })
+      .then((r) => r.json())
+      .then((data: { entries: DirEntry[]; currentDir: string }) => {
+        setEntries(data.entries ?? []);
+        setCurrentDir(data.currentDir);
+      })
+      .catch(() => setEntries([]));
+  }, [activeRoot, currentDir]);
+
+  const navigateToDir = useCallback(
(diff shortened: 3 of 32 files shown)


dimakis (commented):
## Centaur Review

### `frontend/src/components/FileBrowserPanel.tsx`
_Solid extraction of ChatArea and useSessionList, but FileBrowserPanel has an infinite-fetch-loop bug from the currentDir dependency cycle, and DesktopChatView has a CLEAR/RESTORE race condition on session switch. Large sendMessage duplication between ChatView and DesktopChatView will cause maintenance burden._

- 🔴 **bugs** (L50): Infinite fetch loop: the useEffect depends on `currentDir`, but the fetch response sets `currentDir` via `setCurrentDir(data.currentDir)`. If the server normalizes the path (e.g., resolves symlinks, adds trailing slash), the returned `currentDir` will differ from what was sent, triggering a re-render which re-triggers the effect in an infinite loop. Even when the path matches, this is fragile. Either remove `currentDir` from the response handler (trust client state) or compare before setting. `[fixable]`
- 🟡 **unsafe_assumptions** (L56): Parent directory navigation via string splitting (`currentDir.split('/').slice(0, -1).join('/')`) can produce an empty string for a single-segment path, falling back to `activeRoot`. But if `currentDir` is the root itself (which `isSubdir` should prevent), or if paths have double slashes, this could navigate to unexpected locations. More importantly, absolute server paths are exposed to the client and used for directory traversal — while the server validates via `isAllowedPath`, the client constructs paths by concatenation (`${currentDir}/${dirName}`) without sanitization. `[fixable]`


_Solid extraction of ChatArea and useSessionList, but FileBrowserPanel has an infinite-fetch-loop bug from the currentDir dependency cycle, and DesktopChatView has a CLEAR/RESTORE race condition on session switch. Large sendMessage duplication between ChatView and DesktopChatView will cause maintenance burden._

- 🟡 **bugs** (L82): Race condition between CLEAR and RESTORE: when `sessionId` changes, one effect dispatches `CLEAR` and another fetches messages then dispatches `RESTORE`. Both effects share the `[sessionId]` dependency but run independently. The RESTORE effect from the *previous* sessionId could resolve after CLEAR runs for the *new* sessionId, populating the chat with stale messages. The ChatView avoids this because it doesn't have CLEAR — it relies on RESTORE's built-in validation. Consider aborting the previous fetch before dispatching CLEAR, or guard RESTORE with the current sessionId. `[fixable]`
- 🟡 **style** (L120): Large code duplication: `buildSendPayload`, `sendMessage`, `interruptMessage`, `handleStop`, and `generateClientMsgId` are copy-pasted from ChatView.tsx (~90 lines). This violates DRY and will lead to divergence over time. Consider extracting a shared hook (e.g. `useChatActions`) that both views consume. `[fixable]`
- 🔵 **missing_tests**: The DesktopChatView test only verifies the three-panel structure renders. There are no tests for session switching (navigate to different sessionId triggers CLEAR + RESTORE), sending messages, or the `handleToggleContext` interaction between ContextPanel and ChatInput's `externalContextBlocks`. `[fixable]`


### `frontend/src/App.tsx`
_Solid extraction of ChatArea and useSessionList, but FileBrowserPanel has an infinite-fetch-loop bug from the currentDir dependency cycle, and DesktopChatView has a CLEAR/RESTORE race condition on session switch. Large sendMessage duplication between ChatView and DesktopChatView will cause maintenance burden._

- 🟡 **regressions** (L27): Desktop users lose access to the SessionList page entirely — navigating to `/` renders DesktopChatView instead. While this is likely intentional (sessions are in the left panel), it means desktop users can no longer reach the home screen with quick actions grid, update banner, inbox count badge, and clear-all functionality that SessionList provides. These features aren't replicated in the desktop layout.


### `frontend/src/hooks/useMediaQuery.ts`
_Solid extraction of ChatArea and useSessionList, but FileBrowserPanel has an infinite-fetch-loop bug from the currentDir dependency cycle, and DesktopChatView has a CLEAR/RESTORE race condition on session switch. Large sendMessage duplication between ChatView and DesktopChatView will cause maintenance burden._

- 🔵 **missing_tests**: The useMediaQuery test uses dynamic `import()` to load after mocks, but this creates a new module instance per test case. The `useIsDesktop` test is minimal — it only checks `true`. Consider testing the `false` case and verifying the exact query string '(min-width: 768px)' is passed to matchMedia. `[fixable]`


### `frontend/src/components/ContextPanel.tsx`
_Solid extraction of ChatArea and useSessionList, but FileBrowserPanel has an infinite-fetch-loop bug from the currentDir dependency cycle, and DesktopChatView has a CLEAR/RESTORE race condition on session switch. Large sendMessage duplication between ChatView and DesktopChatView will cause maintenance burden._

- 🔵 **bugs** (L23): ContextPanel fetches `/api/config` on every mount to get context blocks. In the desktop layout, this duplicates the same fetch that FileBrowserPanel also makes on mount (line 23 of FileBrowserPanel). Both components independently fetch `/api/config` — consider lifting the config fetch to the parent or using a shared context/cache. `[fixable]`

frontend/src/components/FileBrowserPanel.tsx
```diff
@@ -0,0 +1,139 @@
...
+    else params.set('dir', activeRoot);
+
+    fetch(`/api/files/list?${params}`, { credentials: 'include' })
+      .then((r) => r.json())
+      .then((data: { entries: DirEntry[]; currentDir: string }) => {
+        setEntries(data.entries ?? []);
+        setCurrentDir(data.currentDir);
+      })
+      .catch(() => setEntries([]));
+  }, [activeRoot, currentDir]);
```
dimakis: 🔴 **bugs**: Infinite fetch loop: the useEffect depends on `currentDir`, but the fetch response sets `currentDir` via `setCurrentDir(data.currentDir)`. If the server normalizes the path (e.g., resolves symlinks, adds trailing slash), the returned `currentDir` will differ from what was sent, triggering a re-render which re-triggers the effect in an infinite loop. Even when the path matches, this is fragile. Either remove `currentDir` from the response handler (trust client state) or compare before setting. `[fixable]`

frontend/src/components/FileBrowserPanel.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const navigateToDir = useCallback(
+    (dirName: string) => {
+      setPreview(null);
+      if (dirName === '..') {
+        const parent = currentDir.split('/').slice(0, -1).join('/');
```
dimakis, line 75: 🟡 **unsafe_assumptions**: Parent directory navigation via string splitting (`currentDir.split('/').slice(0, -1).join('/')`) can produce an empty string for a single-segment path, falling back to `activeRoot`. But if `currentDir` is the root itself (which `isSubdir` should prevent), or if paths have double slashes, this could navigate to unexpected locations. More importantly, absolute server paths are exposed to the client and used for directory traversal — while the server validates via `isAllowedPath`, the client constructs paths by concatenation (`${currentDir}/${dirName}`) without sanitization. `[fixable]`

frontend/src/pages/DesktopChatView.tsx
```diff
@@ -0,0 +1,262 @@
...
+  } = useChatMessages(
+    poolKey,
+    sessionState.currentSessionId,
+    handleSessionAssigned,
+    handleSessionExpired,
+    forceScrollToBottom,
+    handleSessionRenamed,
+  );
+
+  useEffect(() => {
```
dimakis, line 108: 🟡 **bugs**: Race condition between CLEAR and RESTORE: when `sessionId` changes, one effect dispatches `CLEAR` and another fetches messages then dispatches `RESTORE`. Both effects share the `[sessionId]` dependency but run independently. The RESTORE effect from the *previous* sessionId could resolve after CLEAR runs for the *new* sessionId, populating the chat with stale messages. The ChatView avoids this because it doesn't have CLEAR — it relies on RESTORE's built-in validation. Consider aborting the previous fetch before dispatching CLEAR, or guard RESTORE with the current sessionId. `[fixable]`

frontend/src/pages/DesktopChatView.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  useAutoSpeak({
+    messages: msgState.messages,
+    running: msgState.running,
+    ttsEnabled: voice.ttsEnabled,
+    ttsAvailable: voice.ttsAvailable,
+    speak: voice.speak,
+  });
+
+  function buildSendPayload(
```
dimakis: 🟡 **style**: Large code duplication: `buildSendPayload`, `sendMessage`, `interruptMessage`, `handleStop`, and `generateClientMsgId` are copy-pasted from ChatView.tsx (~90 lines). This violates DRY and will lead to divergence over time. Consider extracting a shared hook (e.g. `useChatActions`) that both views consume. `[fixable]`

frontend/src/App.tsx
```diff
@@ -21,6 +23,16 @@ function ProtectedRoute({ children }: { children: React.ReactNode }) {
   return <>{children}</>;
 }
 
+function HomeRoute() {
+  const isDesktop = useIsDesktop();
```
dimakis, line 27: 🟡 **regressions**: Desktop users lose access to the SessionList page entirely — navigating to `/` renders DesktopChatView instead. While this is likely intentional (sessions are in the left panel), it means desktop users can no longer reach the home screen with quick actions grid, update banner, inbox count badge, and clear-all functionality that SessionList provides. These features aren't replicated in the desktop layout.

frontend/src/components/ContextPanel.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return `${(kb / 1024).toFixed(1)} MB`;
+}
+
+export function ContextPanel({ selected, onToggle }: ContextPanelProps) {
+  const [blocks, setBlocks] = useState<ContextBlockEntry[]>([]);
+  const [loaded, setLoaded] = useState(false);
+
+  useEffect(() => {
+    fetch('/api/config', { credentials: 'include' })
+      .then((r) => r.json())
```
dimakis, line 32: 🔵 **bugs**: ContextPanel fetches `/api/config` on every mount to get context blocks. In the desktop layout, this duplicates the same fetch that FileBrowserPanel also makes on mount (line 23 of FileBrowserPanel). Both components independently fetch `/api/config` — consider lifting the config fetch to the parent or using a shared context/cache. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useChatSession.ts`
_Solid extraction and desktop layout, but useChatSession doesn't sync currentSessionId on route changes — this is a latent bug that becomes critical in DesktopChatView where the component stays mounted across session switches. Missing tests for the new useChatActions hook._

- 🔴 **bugs**: useChatSession initializes `currentSessionId` from `sessionId` via useState but never syncs on prop change. In DesktopChatView, when the user clicks a session in SessionPanel and navigates from `/chat/abc` to `/chat/def`, `useParams` returns the new sessionId but `currentSessionId` remains stale at `abc`. This causes `sendMessage` to send `resume: 'abc'` instead of `resume: 'def'`, breaking session switching. A `useEffect(() => setCurrentSessionId(sessionId), [sessionId])` is needed. This is a pre-existing latent bug in ChatView (which remounts on route change, masking the issue), but becomes critical in DesktopChatView where the component stays mounted across session switches. `[fixable]`


_Solid extraction and desktop layout, but useChatSession doesn't sync currentSessionId on route changes — this is a latent bug that becomes critical in DesktopChatView where the component stays mounted across session switches. Missing tests for the new useChatActions hook._

- 🟡 **bugs** (L107): The CLEAR+restore effect depends on `sessionId` from useParams. When switching sessions, the effect correctly clears and restores messages. However, `useChatSession` computes `poolKey` as `session:${sessionId}`, meaning a new WebSocket pool key is created on each navigation. The old pool key's connection is orphaned — `useChatConnection` will establish a new connection for the new poolKey, but there's no cleanup of the previous pool key's WS connection. Over many session switches this could leak WebSocket connections. Verify that ws-pool handles this via garbage collection. `[fixable]`
- 🟡 **regressions**: DesktopChatView is missing several features present in ChatView: (1) no mode selector header UI (ask/agent/auto toggle), (2) no model picker, (3) no session rename handling via `handleSessionRenamed` updating a visible title in the UI, (4) no `cwd` display for worktree sessions. These are likely intentional omissions for an MVP, but the header controls (mode/model) affect session behavior and users may expect parity. `[fixable]`


### `frontend/src/hooks/useChatActions.ts`
_Solid extraction and desktop layout, but useChatSession doesn't sync currentSessionId on route changes — this is a latent bug that becomes critical in DesktopChatView where the component stays mounted across session switches. Missing tests for the new useChatActions hook._

- 🟡 **missing_tests**: useChatActions is a new hook extracted from ChatView with non-trivial logic (message payload construction, running state management, connection checks). It has no unit tests. The hook should have tests verifying: CONNECTION_LOST dispatch when WS is closed, wsSetRunning only called when not already running, contextBlocks forwarded in payload, and interrupt guarded by running state. `[fixable]`
- 🔵 **style** (L15): The `sessionState` type in `ChatActionsDeps` is defined inline as `{ model: string; mode: string; ... }` rather than referencing `ChatSessionState` from useChatSession. This duplicates the shape and could drift. Import and use the existing type. `[fixable]`


### `frontend/src/components/FileBrowserPanel.tsx`
_Solid extraction and desktop layout, but useChatSession doesn't sync currentSessionId on route changes — this is a latent bug that becomes critical in DesktopChatView where the component stays mounted across session switches. Missing tests for the new useChatActions hook._

- 🟡 **unsafe_assumptions** (L79): navigateToDir('..') computes the parent by splitting on '/' and slicing. If `currentDir` equals `activeRoot` (e.g. both are '/p'), the parent becomes empty string, which then falls back to `activeRoot`. However, the `isSubdir` check on line 108 (`currentDir !== activeRoot`) should prevent '..' from being shown at the root. The edge case is if `activeRoot` has a trailing slash — `currentDir` set from server response may differ in trailing slash, causing `isSubdir` to be true incorrectly and allowing navigation above the root. Consider using `path.resolve` or normalizing paths. `[fixable]`


### `frontend/src/components/ContextPanel.tsx`
_Solid extraction and desktop layout, but useChatSession doesn't sync currentSessionId on route changes — this is a latent bug that becomes critical in DesktopChatView where the component stays mounted across session switches. Missing tests for the new useChatActions hook._

- 🔵 **style** (L31): ContextPanel self-fetches /api/config when `externalBlocks` is not provided, and FileBrowserPanel does the same for file roots. Both duplicate the same fetch logic. In DesktopChatView these are provided via props (single config fetch), but standalone usage would cause two separate /api/config fetches. Consider a shared config hook if standalone usage is planned. `[fixable]`

frontend/src/pages/DesktopChatView.tsx
```diff
@@ -0,0 +1,218 @@
...
+    sessionState.currentSessionId,
+    handleSessionAssigned,
+    handleSessionExpired,
+    forceScrollToBottom,
+    handleSessionRenamed,
+  );
+
+  // Clear + restore on session switch — abort previous fetch to prevent race condition
+  useEffect(() => {
+    dispatch({ type: 'CLEAR' });
```
dimakis, line 109: 🟡 **bugs**: The CLEAR+restore effect depends on `sessionId` from useParams. When switching sessions, the effect correctly clears and restores messages. However, `useChatSession` computes `poolKey` as `session:${sessionId}`, meaning a new WebSocket pool key is created on each navigation. The old pool key's connection is orphaned — `useChatConnection` will establish a new connection for the new poolKey, but there's no cleanup of the previous pool key's WS connection. Over many session switches this could leak WebSocket connections. Verify that ws-pool handles this via garbage collection. `[fixable]`

frontend/src/hooks/useChatActions.ts
```diff
@@ -0,0 +1,100 @@
...
+  return `user-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+}
+
+interface ChatActionsDeps {
+  poolKey: string;
+  sessionState: {
+    model: string;
+    mode: string;
+    currentSessionId: string | undefined;
+    sandbox: boolean;
```
dimakis: 🔵 **style**: The `sessionState` type in `ChatActionsDeps` is defined inline as `{ model: string; mode: string; ... }` rather than referencing `ChatSessionState` from useChatSession. This duplicates the shape and could drift. Import and use the existing type. `[fixable]`

frontend/src/components/FileBrowserPanel.tsx
```diff
@@ -0,0 +1,158 @@
...
+      let target: string;
+      if (dirName === '..') {
+        const parent = currentDir.split('/').slice(0, -1).join('/');
+        target = parent || activeRoot;
+      } else {
+        target = `${currentDir}/${dirName}`;
+      }
+      setCurrentDir(target);
+      setFetchDir(target);
+    },
```
dimakis, line 82: 🟡 **unsafe_assumptions**: navigateToDir('..') computes the parent by splitting on '/' and slicing. If `currentDir` equals `activeRoot` (e.g. both are '/p'), the parent becomes empty string, which then falls back to `activeRoot`. However, the `isSubdir` check on line 108 (`currentDir !== activeRoot`) should prevent '..' from being shown at the root. The edge case is if `activeRoot` has a trailing slash — `currentDir` set from server response may differ in trailing slash, causing `isSubdir` to be true incorrectly and allowing navigation above the root. Consider using `path.resolve` or normalizing paths. `[fixable]`

frontend/src/components/ContextPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+  blocks: externalBlocks,
+  loaded: externalLoaded,
+}: ContextPanelProps) {
+  const [selfBlocks, setSelfBlocks] = useState<ContextBlockEntry[]>([]);
+  const [selfLoaded, setSelfLoaded] = useState(false);
+
+  // Self-fetch config only when not provided via props (standalone usage)
+  useEffect(() => {
+    if (externalBlocks !== undefined) return;
+    fetch('/api/config', { credentials: 'include' })
```
dimakis, line 31: 🔵 **style**: ContextPanel self-fetches /api/config when `externalBlocks` is not provided, and FileBrowserPanel does the same for file roots. Both duplicate the same fetch logic. In DesktopChatView these are provided via props (single config fetch), but standalone usage would cause two separate /api/config fetches. Consider a shared config hook if standalone usage is planned. `[fixable]`

PR #163, opened 2026-04-10
feat(server): multi-client session observer pattern

dimakis:
## Summary

- Adds observer pattern so multiple clients can view and interact with the same active session simultaneously
- Guards against duplicate `query()` calls to the Agent SDK — second client auto-subscribes as observer instead of starting a conflicting conversation
- Both clients can send messages (multi-writer) — messages route to the existing session's inputQueue

## Problem

When a second device (e.g. phone on port 3100) opened a session that was already active on the desktop, the server started a duplicate `query()` against the Agent SDK on the same `sessionId`. This corrupted state and caused the desktop UI to fall back to the home screen.

## Changes

| Layer | File | What |
|-------|------|------|
| Registry | `server/session-registry.ts` | `observers: Set<WebSocket>` on `ManagedSession`, `addObserver()`, `removeObserver()`, cleanup in `abort()`/`remove()` |
| Broadcast | `server/query-loop.ts` | `sendOrBuffer` sends to observers; `session_end` in finally block also broadcasts |
| Echo | `server/chat.ts` | `sendToChat` and `interruptChat` echo `user_message` to observers |
| Protocol | `server/ws-schemas.ts` | New `subscribe` message type |
| Handler | `server/index.ts` | Subscribe handler, duplicate query guard via `tryRouteToActiveSession`, observer cleanup on disconnect |
| Pool | `frontend/src/lib/ws-pool.ts` | Sends `subscribe` for `session:` keys after `client_id`; handles `subscribed` response |
| Messages | `frontend/src/hooks/useChatMessages.ts` | Handles `subscribed` message to set running state |
| Types | `frontend/src/types/ws-messages.ts` | `SubscribedMsg` type |

## Test plan

- [x] 825 tests pass (42 new)
- [x] ESLint: 0 errors
- [ ] Manual: open desktop + phone on same session — phone should observe, not break desktop
- [ ] Manual: send message from phone while desktop is active — both see the message and response

### DIFF
--- a/frontend/src/lib/ws-pool.ts
+++ b/frontend/src/lib/ws-pool.ts
@@ -100,8 +100,7 @@ function connectEntry(key: string, entry: PoolEntry) {
     }
 
     if (msg.type === 'subscribed') {
-      const subscribed = msg as unknown as { running?: boolean };
-      if (subscribed.running) entry.wasRunning = true;
+      if (msg.running) entry.wasRunning = true;
       broadcast(entry, { type: '_open' });
     }
 
--- a/frontend/src/lib/ws-pool.ts
+++ b/frontend/src/lib/ws-pool.ts
@@ -65,8 +65,6 @@ function connectEntry(key: string, entry: PoolEntry) {
     }
 
     if (msg.type === 'client_id') {
-      // Save current as prev BEFORE overwriting — prev is what the server
-      // knows us by and what we need to send in the reattach request.
       entry.prevClientId = entry.clientId;
       entry.clientId = msg.clientId as string;
       if (entry.wasRunning && entry.prevClientId) {
@@ -79,19 +77,31 @@ function connectEntry(key: string, entry: PoolEntry) {
         );
         return; // wait for reattach/reattach_failed before broadcasting _open
       }
+      // For existing session keys, subscribe as observer so the server
+      // broadcasts live events if the session is active on another client.
+      const sessionPrefix = 'session:';
+      if (key.startsWith(sessionPrefix)) {
+        const sessionId = key.slice(sessionPrefix.length);
+        ws.send(JSON.stringify({ type: 'subscribe', sessionId }));
+        return; // wait for subscribed response before broadcasting _open
+      }
       broadcast(entry, { type: '_open' });
       return;
     }
 
     if (msg.type === 'reattached') {
       entry.wasRunning = true;
-      broadcast(entry, { type: '_open' }); // signal connected to component
+      broadcast(entry, { type: '_open' });
     }
 
     if (msg.type === 'reattach_failed') {
       entry.wasRunning = false;
-      // Signal that the connection is live even though reattach failed —
-      // without this, the component never learns the WS is open.
+      broadcast(entry, { type: '_open' });
+    }
+
+    if (msg.type === 'subscribed') {
+      const subscribed = msg as unknown as { running?: boolean };
+      if (subscribed.running) entry.wasRunning = true;
       broadcast(entry, { type: '_open' });
     }
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -99,6 +99,32 @@ server.on('upgrade', async (req, socket, head) => {
   });
 });
 
+/**
+ * If `resume` targets a session that's already active, subscribe the caller
+ * as an observer and inject the message into the running session.
+ * Returns the driver's clientId if handled, null to fall through to startChat.
+ */
+function tryRouteToActiveSession(
+  ws: WebSocket,
+  resume: string | undefined,
+  prompt: string,
+  images?: Array<{ data: string; mediaType: string }>,
+  contextBlocks?: string[],
+  clientMsgId?: string,
+): string | null {
+  if (!resume) return null;
+  const found = registry.findBySessionId(resume);
+  if (!found) return null;
+  // Session is active — subscribe as observer instead of starting a duplicate query
+  registry.addObserver(resume, ws);
+  sendToChat(found.clientId, prompt, images, contextBlocks, clientMsgId);
+  log.info('routed observer message to active session', {
+    sessionId: resume,
+    driverClientId: found.clientId,
+  });
+  return found.clientId;
+}
+
 function handleChatWs(ws: WebSocket, initialClientId: string) {
   let clientId = initialClientId;
   ws.send(JSON.stringify({ type: 'client_id', clientId }));
@@ -121,7 +147,43 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
 
       const msg = result.data;
 
-      if (msg.type === 'reattach') {
+      if (msg.type === 'subscribe') {
+        const found = registry.findBySessionId(msg.sessionId);
+        if (found) {
+          registry.addObserver(msg.sessionId, ws);
+          // Send in-flight snapshot so observer sees the current streaming message
+          if (found.session.currentSnapshot) {
+            ws.send(
+              JSON.stringify({
+                v: 2,
+                type: 'message_snapshot',
+                ts: Date.now(),
+                messageId: found.session.currentSnapshot.messageId,
+                blocks: found.session.currentSnapshot.blocks,
+              }),
+            );
+          }
+          ws.send(
+            JSON.stringify({
+              type: 'subscribed',
+              sessionId: msg.sessionId,
+              running: true,
+            }),
+          );
+          log.info('client subscribed to active session', {
+            clientId,
+            sessionId: msg.sessionId,
+          });
+        } else {
+          ws.send(
+            JSON.stringify({
+              type: 'subscribed',
+              sessionId: msg.sessionId,
+              running: false,
+            }),
+          );
+        }
+      } else if (msg.type === 'reattach') {
         const ok = reattachChat(msg.clientId, ws);
         if (ok) {
           clientId = msg.clientId;
@@ -232,7 +294,16 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
               msg.contextBlocks,
               msg.clientMsgId,
             );
-          } else {
+          } else if (
+            !tryRouteToActiveSession(
+              ws,
+              msg.resume,
+              resolution.renderedPrompt,
+              msg.images,
+              msg.contextBlocks,
+              msg.clientMsgId,
+            )
+          ) {
             startChat(ws, clientId, resolution.renderedPrompt, {
               resume: msg.resume,
               cwd: msg.cwd,
@@ -250,7 +321,16 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
           clearSkillPolicy(registry, clientId);
           if (isActive(clientId)) {
             sendToChat(clientId, msg.prompt, msg.images, msg.contextBlocks, msg.clientMsgId);
-          } else {
+          } else if (
+            !tryRouteToActiveSession(
+              ws,
+              msg.resume,
+              msg.prompt,
+              msg.images,
+              msg.contextBlocks,
+              msg.clientMsgId,
+            )
+          ) {
             startChat(ws, clientId, msg.prompt, {
               resume: msg.resume,
               cwd: msg.cwd,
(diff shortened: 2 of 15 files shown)


server/index.ts
```diff
@@ -99,6 +99,32 @@ server.on('upgrade', async (req, socket, head) => {
...
+function tryRouteToActiveSession(
+  ws: WebSocket,
+  resume: string | undefined,
+  prompt: string,
+  images?: Array<{ data: string; mediaType: string }>,
+  contextBlocks?: string[],
+  clientMsgId?: string,
+): string | null {
+  if (!resume) return null;
+  const found = registry.findBySessionId(resume);
```
dimakis, line 116: 🟡 **bugs**: `tryRouteToActiveSession` calls `findBySessionId` which returns sessions regardless of attachment state. If the driver is detached (disconnected, waiting for reattach TTL), an observer will be added and `sendToChat` called — but `sendToChat` sends via the stale `session.ws` (readyState likely CLOSED), and the inputQueue push may interact with a query that has no attached driver to handle permission prompts. Consider gating on `registry.isAttached(found.clientId)` before routing. `[fixable]`

server/index.ts
```diff
@@ -99,6 +99,32 @@ server.on('upgrade', async (req, socket, head) => {
   });
 });
 
+/**
+ * If `resume` targets a session that's already active, subscribe the caller
+ * as an observer and inject the message into the running session.
+ * Returns the driver's clientId if handled, null to fall through to startChat.
+ */
+function tryRouteToActiveSession(
```
dimakis, line 107: 🔵 **missing_tests**: `tryRouteToActiveSession` has no dedicated test coverage. It's an important routing function — edge cases like `resume` pointing to a detached session, or `sendToChat` returning false (session missing inputQueue), are untested. `[fixable]`

server/index.ts
```diff
@@ -121,7 +147,43 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
 
       const msg = result.data;
 
-      if (msg.type === 'reattach') {
+      if (msg.type === 'subscribe') {
+        const found = registry.findBySessionId(msg.sessionId);
+        if (found) {
+          registry.addObserver(msg.sessionId, ws);
```
dimakis, line 168: 🔵 **unsafe_assumptions**: When handling a `subscribe` message for an active session, the snapshot is sent before the `subscribed` response. If the observer's frontend processes `message_snapshot` before `subscribed`, the reducer may not have set `running: true` yet, potentially causing the snapshot to be handled in a non-running state. Consider sending `subscribed` first, then the snapshot. `[fixable]`

server/session-registry.ts
```diff
@@ -149,6 +156,27 @@ export class SessionRegistry {
     return null;
   }
 
+  /**
+   * Add an observer WebSocket to the session identified by sessionId.
+   * Returns the clientId of the driver if successful, null otherwise.
+   */
+  addObserver(sessionId: string, ws: WebSocket): string | null {
```
dimakis, line 164: 🟡 **unsafe_assumptions**: `addObserver` has no bound on the size of the observers Set. A client can repeatedly open WebSocket connections and subscribe to the same session, growing the Set unbounded. Consider capping observers (e.g. 10) or deduplicating by checking if `ws` is already in the Set before adding. `[fixable]`

server/query-loop.ts
```diff
@@ -41,7 +41,14 @@ function sendOrBuffer(
   if (registry.isAttached(clientId)) {
     send(ws, enriched);
   }
-  // When detached, messages are dropped — recovery via event store replay on reattach.
+  // Broadcast to observers watching this session
```
dimakis, line 65: 🟡 **bugs**: Observer broadcast in `sendOrBuffer` runs regardless of `isAttached()` — when the driver is detached, the driver's message is dropped but observers still receive it. This creates an asymmetry: if the driver reconnects and gets a replay from the event store, the observer has already seen the messages live. This is arguably intentional, but should be documented as a design choice since it differs from driver behavior.

frontend/src/lib/ws-pool.ts
```diff
@@ -79,19 +77,31 @@ function connectEntry(key: string, entry: PoolEntry) {
         );
         return; // wait for reattach/reattach_failed before broadcasting _open
       }
+      // For existing session keys, subscribe as observer so the server
+      // broadcasts live events if the session is active on another client.
+      const sessionPrefix = 'session:';
+      if (key.startsWith(sessionPrefix)) {
+        const sessionId = key.slice(sessionPrefix.length);
+        ws.send(JSON.stringify({ type: 'subscribe', sessionId }));
```
dimakis, line 85: 🔵 **style**: The `subscribed` message is cast via `msg as unknown as { running?: boolean }` instead of using proper type narrowing. Since `msg` already came through the message handler, consider adding the `subscribed` type to the pool's internal message handling or using a type guard for consistency with how `reattached` and `reattach_failed` are handled above. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid observer pattern implementation with good test coverage. Main concerns are: routing to detached sessions via tryRouteToActiveSession (missing isAttached check), unbounded observer set growth, and message ordering between snapshot and subscribed response._

- 🟡 **bugs** (L116): `tryRouteToActiveSession` calls `findBySessionId` which returns sessions regardless of attachment state. If the driver is detached (disconnected, waiting for reattach TTL), an observer will be added and `sendToChat` called — but `sendToChat` sends via the stale `session.ws` (readyState likely CLOSED), and the inputQueue push may interact with a query that has no attached driver to handle permission prompts. Consider gating on `registry.isAttached(found.clientId)` before routing. `[fixable]`
- 🔵 **missing_tests** (L107): `tryRouteToActiveSession` has no dedicated test coverage. It's an important routing function — edge cases like `resume` pointing to a detached session, or `sendToChat` returning false (session missing inputQueue), are untested. `[fixable]`
- 🔵 **unsafe_assumptions** (L153): When handling a `subscribe` message for an active session, the snapshot is sent before the `subscribed` response. If the observer's frontend processes `message_snapshot` before `subscribed`, the reducer may not have set `running: true` yet, potentially causing the snapshot to be handled in a non-running state. Consider sending `subscribed` first, then the snapshot. `[fixable]`


### `server/session-registry.ts`
_Solid observer pattern implementation with good test coverage. Main concerns are: routing to detached sessions via tryRouteToActiveSession (missing isAttached check), unbounded observer set growth, and message ordering between snapshot and subscribed response._

- 🟡 **unsafe_assumptions** (L163): `addObserver` has no bound on the size of the observers Set. A client can repeatedly open WebSocket connections and subscribe to the same session, growing the Set unbounded. Consider capping observers (e.g. 10) or deduplicating by checking if `ws` is already in the Set before adding. `[fixable]`


### `server/query-loop.ts`
_Solid observer pattern implementation with good test coverage. Main concerns are: routing to detached sessions via tryRouteToActiveSession (missing isAttached check), unbounded observer set growth, and message ordering between snapshot and subscribed response._

- 🟡 **bugs** (L44): Observer broadcast in `sendOrBuffer` runs regardless of `isAttached()` — when the driver is detached, the driver's message is dropped but observers still receive it. This creates an asymmetry: if the driver reconnects and gets a replay from the event store, the observer has already seen the messages live. This is arguably intentional, but should be documented as a design choice since it differs from driver behavior.


### `server/__tests__/session-registry.test.ts`
_Solid observer pattern implementation with good test coverage. Main concerns are: routing to detached sessions via tryRouteToActiveSession (missing isAttached check), unbounded observer set growth, and message ordering between snapshot and subscribed response._

- 🔵 **missing_tests**: No test for `addObserver` when the same WebSocket is added multiple times to the same session (idempotency of Set). Also no test for `removeObserver` when the ws is not present in any session (should be a no-op). Minor gaps but worth covering for defensive clarity. `[fixable]`


### `frontend/src/lib/ws-pool.ts`
_Solid observer pattern implementation with good test coverage. Main concerns are: routing to detached sessions via tryRouteToActiveSession (missing isAttached check), unbounded observer set growth, and message ordering between snapshot and subscribed response._

- 🔵 **style** (L85): The `subscribed` message is cast via `msg as unknown as { running?: boolean }` instead of using proper type narrowing. Since `msg` already came through the message handler, consider adding the `subscribed` type to the pool's internal message handling or using a type guard for consistency with how `reattached` and `reattach_failed` are handled above. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🟡 **bugs**: In `sendOrBuffer`, observer broadcast happens regardless of whether the driver is attached. When the driver is detached (line 41 check fails), messages are intentionally dropped for the driver, but observers still receive them. This creates an inconsistency where observers see messages the driver won't see until event-store replay. If the driver reattaches, it replays from the store — but the observer has no dedup mechanism, so replayed events from a concurrent reattach could produce duplicates on the observer side. `[fixable]`
- 🟡 **unsafe_assumptions**: Observer `obs.send()` calls in `sendOrBuffer` and the `finally` block are not wrapped in try/catch. While `readyState === OPEN` is checked, `ws.send()` can still throw (e.g., if the socket transitions to CLOSING between the check and the send). An exception would abort the broadcast loop, skipping remaining observers. Wrap in try/catch or use the existing `send()` helper which already guards readyState. `[fixable]`
- 🔵 **style**: Observer broadcast logic (`for (const obs of session.observers) { if (obs.readyState === obs.OPEN) obs.send(msg); }`) is duplicated in three places: `sendOrBuffer`, the `finally` block of `runQueryLoop`, and `chat.ts` (`sendToChat`/`interruptChat`). Extract a `broadcastToObservers(session, msg)` helper to reduce duplication and ensure consistent error handling. `[fixable]`


### `server/index.ts`
_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🟡 **bugs**: `tryRouteToActiveSession` calls `sendToChat(found.clientId, ...)` after `addObserver()`, but the observer never receives the history of the current turn — only the `currentSnapshot` is sent via the `subscribe` handler. When the observer joins mid-session via `tryRouteToActiveSession` (not via `subscribe`), no snapshot is sent at all, so they miss all prior messages in the session. `[fixable]`


### `server/session-registry.ts`
_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🔵 **unsafe_assumptions**: No limit on observer count per session. A misbehaving or compromised client could subscribe many WebSockets to the same session, growing the `observers` Set unboundedly. Consider adding a max observer cap (e.g., 10) in `addObserver()`. `[fixable]`


_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🔵 **missing_tests**: No test for observer broadcast when driver is detached. The current test only covers the attached case. Add a test with `attached: false` to verify observer behavior when the driver is disconnected — this is a key design decision that should be documented via test. `[fixable]`


### `server/__tests__/session-registry.test.ts`
_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🔵 **missing_tests**: No test for `removeObserver` when ws is in multiple sessions' observer sets simultaneously, and no test for `addObserver` with a max-observer-count boundary (if one is added). The `detach` path also isn't tested — observers should survive a detach since they're independent connections. `[fixable]`


### `frontend/src/lib/ws-pool.ts`
_Solid observer pattern implementation with good test coverage. Main concerns: observer broadcast bypasses the driver's detach guard (observers see messages the driver drops), missing try/catch on obs.send() could break broadcast loops, and tryRouteToActiveSession doesn't send snapshot to the observer._

- 🔵 **bugs**: The `subscribe` handler in `connectEntry` returns early (preventing `_open` broadcast) and waits for the `subscribed` response. If the server never responds (e.g., bug or crash), the pool entry stays in limbo — no `_open` is ever broadcast, and the component never learns the connection is live. Consider a timeout fallback that broadcasts `_open` if `subscribed` isn't received within a few seconds. `[fixable]`

PR #168, opened 2026-04-11
feat(calendar): add release milestone filter toggle

dimakis:
## Summary
- Add All/Releases toggle to calendar header, visually consistent with existing Day/Week toggle
- When "Releases" is selected, meeting events are hidden and only milestones are shown, with the view extended to 90 days
- Sprint bars remain visible in both modes; switching back to "All" restores the previous Day/Week setting

## Test plan
- [x] 6 new tests added covering: toggle rendering, event filtering, 90-day view switch, view restoration, sprint persistence, active state
- [x] All 28 CalendarView tests pass
- [x] Full test suite passes (pre-existing failures in useDocumentReader are unrelated)

### DIFF
--- a/frontend/src/pages/CalendarView.tsx
+++ b/frontend/src/pages/CalendarView.tsx
@@ -40,12 +40,14 @@ function getToday(): string {
   return new Date().toISOString().slice(0, 10);
 }
 
+const DEFAULT_VIEW_DAYS = 7;
+
 export function CalendarView() {
   const navigate = useNavigate();
   const [baseDate, setBaseDate] = useState(getToday);
-  const [viewDays, setViewDays] = useState(7);
+  const [viewDays, setViewDays] = useState(DEFAULT_VIEW_DAYS);
   const [filterMode, setFilterMode] = useState<'all' | 'releases'>('all');
-  const [savedViewDays, setSavedViewDays] = useState(7);
+  const [savedViewDays, setSavedViewDays] = useState(DEFAULT_VIEW_DAYS);
 
   const { loading, events, sprints } = useCalendarData(baseDate, viewDays);
 
@@ -78,12 +80,14 @@ export function CalendarView() {
 
   const dates = useMemo(() => Array.from(eventsByDate.keys()).sort(), [eventsByDate]);
 
+  const navStep = filterMode === 'releases' ? 30 : viewDays;
+
   function handlePrev() {
-    setBaseDate(addDays(baseDate, -viewDays));
+    setBaseDate(addDays(baseDate, -navStep));
   }
 
   function handleNext() {
-    setBaseDate(addDays(baseDate, viewDays));
+    setBaseDate(addDays(baseDate, navStep));
   }
 
   function handleToday() {
@@ -132,12 +136,14 @@ export function CalendarView() {
         <div className="cal-view-toggle">
           <button
             className={`cal-view-btn${viewDays === 1 ? ' cal-view-btn--active' : ''}`}
+            disabled={filterMode === 'releases'}
             onClick={() => setViewDays(1)}
           >
             Day
           </button>
           <button
             className={`cal-view-btn${viewDays === 7 ? ' cal-view-btn--active' : ''}`}
+            disabled={filterMode === 'releases'}
             onClick={() => setViewDays(7)}
           >
             Week
--- a/frontend/src/pages/CalendarView.tsx
+++ b/frontend/src/pages/CalendarView.tsx
@@ -44,8 +44,18 @@ export function CalendarView() {
   const navigate = useNavigate();
   const [baseDate, setBaseDate] = useState(getToday);
   const [viewDays, setViewDays] = useState(7);
+  const [filterMode, setFilterMode] = useState<'all' | 'releases'>('all');
+  const [savedViewDays, setSavedViewDays] = useState(7);
 
   const { loading, events, sprints } = useCalendarData(baseDate, viewDays);
+
+  // Filter events based on filter mode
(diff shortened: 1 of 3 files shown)


dimakis:
## Centaur Review

### `frontend/src/pages/CalendarView.tsx`
_Feature works for the happy path but the Day/Week toggle and Releases filter interact poorly — clicking Day/Week in Releases mode creates inconsistent state, and the test claiming to verify disabled behavior doesn't actually test it._

- 🟡 **bugs** (L134): Day/Week toggle buttons are not disabled when in Releases mode. Clicking 'Day' while in Releases mode sets viewDays=1, overriding the 90-day view without changing filterMode. When the user later switches back to 'All', savedViewDays would restore the pre-releases value (e.g. 7), not the 1 they just selected — confusing state. Either disable the Day/Week toggle in releases mode or coordinate the two toggles. `[fixable]`
- 🟡 **bugs** (L134): The Day/Week active-state logic (`viewDays === 1` / `viewDays === 7`) breaks in Releases mode where viewDays is 90 — neither Day nor Week appears active, which is correct visually, but clicking Day/Week while in Releases mode will silently change viewDays without updating filterMode, creating inconsistent state. `[fixable]`
- 🔵 **bugs** (L82): handlePrev/handleNext navigate by `viewDays` (90 in Releases mode). Jumping 90 days at a time is likely not the intended UX for the Releases view — consider using a smaller step or different navigation logic when in Releases mode. `[fixable]`


### `frontend/src/pages/__tests__/CalendarView.test.tsx`
_Feature works for the happy path but the Day/Week toggle and Releases filter interact poorly — clicking Day/Week in Releases mode creates inconsistent state, and the test claiming to verify disabled behavior doesn't actually test it._

- 🟡 **missing_tests** (L432): Test named 'disables Day/Week toggle when in Releases mode' does not actually verify the Day/Week buttons are disabled — it only checks that the Releases button is active. The test title is misleading and the implementation doesn't disable Day/Week either, so this is both a missing test and a missing feature. `[fixable]`

frontend/src/pages/CalendarView.tsx
```diff
@@ -109,17 +136,33 @@ export function CalendarView() {
         <div className="cal-view-toggle">
           <button
             className={`cal-view-btn${viewDays === 1 ? ' cal-view-btn--active' : ''}`}
+            disabled={filterMode === 'releases'}
             onClick={() => setViewDays(1)}
```
dimakis, line 140: 🟡 **bugs**: Day/Week buttons call `setViewDays()` directly without updating `savedViewDays`. If a user switches to Day view (viewDays=1), then clicks Releases (saves 1), then clicks All, they correctly get 1 back. However, the Day/Week buttons don't update `savedViewDays`, so if a user is already in Releases mode and somehow changes viewDays through another path, `savedViewDays` would be stale. Currently the disabled state prevents this, but if the buttons are ever re-enabled or another viewDays setter is added, this will silently break. Minor — the disabled guard is sufficient for now. `[fixable]`

frontend/src/pages/CalendarView.tsx
```diff
@@ -64,22 +76,37 @@ export function CalendarView() {
...
+    setFilterMode('all');
+    setViewDays(savedViewDays);
+  }
+
+  function handleFilterReleases() {
+    if (filterMode !== 'releases') {
+      setSavedViewDays(viewDays);
+    }
+    setFilterMode('releases');
+    setViewDays(90);
```
dimakis, line 107: 🔵 **style**: The magic number `90` (days for releases view) and `30` (nav step) are inline. Consider extracting them as named constants alongside `DEFAULT_VIEW_DAYS` for clarity (e.g., `RELEASES_VIEW_DAYS = 90`, `RELEASES_NAV_STEP = 30`). `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/pages/CalendarView.tsx`
_Clean feature addition with good test coverage. Minor suggestions around magic numbers and an edge-case test gap for Day→Releases→All restore flow._

- 🟡 **bugs** (L140): Day/Week buttons call `setViewDays()` directly without updating `savedViewDays`. If a user switches to Day view (viewDays=1), then clicks Releases (saves 1), then clicks All, they correctly get 1 back. However, the Day/Week buttons don't update `savedViewDays`, so if a user is already in Releases mode and somehow changes viewDays through another path, `savedViewDays` would be stale. Currently the disabled state prevents this, but if the buttons are ever re-enabled or another viewDays setter is added, this will silently break. Minor — the disabled guard is sufficient for now. `[fixable]`
- 🔵 **style** (L107): The magic number `90` (days for releases view) and `30` (nav step) are inline. Consider extracting them as named constants alongside `DEFAULT_VIEW_DAYS` for clarity (e.g., `RELEASES_VIEW_DAYS = 90`, `RELEASES_NAV_STEP = 30`). `[fixable]`


### `frontend/src/pages/__tests__/CalendarView.test.tsx`
_Clean feature addition with good test coverage. Minor suggestions around magic numbers and an edge-case test gap for Day→Releases→All restore flow._

- 🔵 **missing_tests**: No test for the save/restore flow when the user is in Day view (viewDays=1) before switching to Releases. The existing restore test only covers the default 7-day case. A test that clicks Day, then Releases, then All, and verifies days=1 is restored would exercise the `savedViewDays` logic more thoroughly. `[fixable]`
- 🔵 **missing_tests**: No test for the 30-day navigation step in Releases mode. The nav buttons use `navStep` (30 in releases mode vs viewDays otherwise), but no test verifies that clicking prev/next in Releases mode advances by 30 days instead of 90. `[fixable]`

PR #169, opened 2026-04-11
feat(skills): add /person bundled skill

dimakis:
## Summary
- Adds a new `/person` bundled skill for viewing and updating people profiles in `command_center/context/people/`
- Supports name/email prefix lookup, `--note` flag for dated notes, and inline text for open threads
- Updates bundled-skills tests to account for the new skill (5 skills total), adjusting analysis-only constraints to exclude `/person` which needs Edit access

## Test plan
- [x] New tests for `/person` skill discovery, allowed-tools, and body content
- [x] Existing bundled-skills tests updated and passing
- [x] Full test suite passes (pre-existing useDocumentReader failures only)

### DIFF
--- a/server/__tests__/bundled-skills.test.ts
+++ b/server/__tests__/bundled-skills.test.ts
@@ -35,9 +35,10 @@ describe('bundled skills', () => {
     }
   });
 
-  it('analysis-only bundled skills do not include Write or Edit in allowed-tools', () => {
-    const analysisOnly = skills.filter((s) => s.name !== 'person');
-    for (const skill of analysisOnly) {
+  it('non-mutating bundled skills do not include Write or Edit in allowed-tools', () => {
+    const nonMutating = skills.filter((s) => !s.mutating);
+    expect(nonMutating.length).toBeGreaterThan(0);
+    for (const skill of nonMutating) {
       if (skill.allowedTools) {
         expect(skill.allowedTools).not.toContain('Write');
         expect(skill.allowedTools).not.toContain('Edit');
@@ -45,11 +46,24 @@ describe('bundled skills', () => {
     }
   });
 
-  it('analysis bundled skills contain an approval gate instruction', () => {
-    // Analysis skills must include BOTH "present" AND "before making changes"
+  it('mutating skills declare the flag in frontmatter and have write access', () => {
+    const mutating = skills.filter((s) => s.mutating);
+    expect(mutating.length).toBeGreaterThan(0);
+    for (const skill of mutating) {
+      const hasWriteAccess =
+        skill.allowedTools?.includes('Edit') || skill.allowedTools?.includes('Write');
+      expect(
+        hasWriteAccess,
+        `Mutating skill "${skill.name}" should have Edit or Write in allowed-tools`,
+      ).toBe(true);
+    }
+  });
+
+  it('non-mutating bundled skills contain an approval gate instruction', () => {
+    // Non-mutating skills must include BOTH "present" AND "before making changes"
     // so users always see findings before any modifications happen.
-    const analysisSkills = skills.filter((s) => s.name !== 'person');
-    for (const skill of analysisSkills) {
+    const nonMutating = skills.filter((s) => !s.mutating);
+    for (const skill of nonMutating) {
       const body = registry.getBody(skill.name)!;
       expect(body).toBeDefined();
       const lower = body.toLowerCase();
--- a/server/__tests__/bundled-skills.test.ts
+++ b/server/__tests__/bundled-skills.test.ts
@@ -12,9 +12,9 @@ describe('bundled skills', () => {
   const registry = new SkillRegistry({ bundledDir: SKILLS_DIR });
   const skills = registry.list();
 
-  it('discovers all four bundled skills', () => {
+  it('discovers all five bundled skills', () => {
     const names = skills.map((s) => s.name).sort();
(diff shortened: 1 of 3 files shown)


dimakis:
## Centaur Review

Found **5** issue(s) (4 warning).


### `frontend/src/pages/__tests__/CalendarView.test.tsx`
_The calendar filter feature works for the happy path but has an interaction bug: the Day/Week toggle and Releases filter can conflict, leaving the UI in inconsistent states. The test titled 'disables Day/Week toggle' doesn't test what it claims. The /person skill itself is clean._

- 🟡 **bugs** (L432): Test titled 'disables Day/Week toggle when in Releases mode' does not actually assert that the Day/Week buttons are disabled. It only checks that the Releases button has the active class. The implementation also does not disable the Day/Week toggle — clicking Day or Week while in Releases mode will override the 90-day view with 1 or 7 days while still filtering to milestones only, which is likely unintended. Either disable the Day/Week buttons when filterMode === 'releases', or rename the test to match the actual behavior. `[fixable]`


### `frontend/src/pages/CalendarView.tsx`
_The calendar filter feature works for the happy path but has an interaction bug: the Day/Week toggle and Releases filter can conflict, leaving the UI in inconsistent states. The test titled 'disables Day/Week toggle' doesn't test what it claims. The /person skill itself is clean._

- 🟡 **bugs** (L134): Day/Week toggle remains interactive in Releases mode. Clicking 'Day' sets viewDays=1 (overriding the 90-day Releases view) and also breaks savedViewDays restoration — when switching back to All, savedViewDays still holds the pre-Releases value, ignoring the Day/Week click. The Day/Week toggle should either be disabled or its clicks should update savedViewDays when in Releases mode. `[fixable]`
- 🟡 **bugs** (L134): Day/Week active-class logic uses raw viewDays (1 or 7), so in Releases mode (viewDays=90), neither Day nor Week button appears active, leaving the toggle in a visually ambiguous state with no button highlighted. `[fixable]`
- 🔵 **style** (L47): savedViewDays is initialized to 7, duplicating the default of viewDays. If viewDays default ever changes, these could drift apart. Consider initializing savedViewDays from the same source or using a single initial constant. `[fixable]`


_The calendar filter feature works for the happy path but has an interaction bug: the Day/Week toggle and Releases filter can conflict, leaving the UI in inconsistent states. The test titled 'disables Day/Week toggle' doesn't test what it claims. The /person skill itself is clean._

- 🟡 **regressions** (L38): The test that previously enforced 'no bundled skill includes Write or Edit in allowed-tools' now excludes /person by name. This is a safety gate weakened by an ad-hoc exception list. If more skills with write access are added in the future, they'll need to be manually excluded too. Consider using a metadata flag (e.g., `analysis-only: true` in frontmatter) to make this self-documenting rather than maintaining a hardcoded exclusion list. `[fixable]`

dimakis:
## Centaur Review

### `frontend/src/pages/__tests__/CalendarView.test.tsx`
_The Day/Week toggle is not disabled or guarded in Releases mode despite the test claiming it is, creating a stale-state bug; the /person skill and mutating flag additions are clean._

- 🟡 **bugs** (L432): Test 'disables Day/Week toggle when in Releases mode' does not actually test that the Day/Week toggle is disabled. It only asserts the Releases button is active. Furthermore, the component never disables the Day/Week buttons — clicking Day or Week while in Releases mode will overwrite `viewDays` and break the saved-state logic. Either the test name is misleading and should be fixed, or the component needs to actually disable the Day/Week toggle in Releases mode. `[fixable]`
- 🔵 **missing_tests**: No test covers the interaction between Day/Week toggle and Releases filter mode — e.g., what happens if the user clicks 'Day' while in Releases mode, then switches back to All. This is related to the stale-state bug above. `[fixable]`


### `frontend/src/pages/CalendarView.tsx`
_The Day/Week toggle is not disabled or guarded in Releases mode despite the test claiming it is, creating a stale-state bug; the /person skill and mutating flag additions are clean._

- 🟡 **bugs** (L135): Day/Week buttons call `setViewDays()` directly, bypassing `savedViewDays`. If the user clicks Day/Week while in Releases mode (viewDays=90), it sets viewDays to 1 or 7 without clearing filterMode. Then switching back to All via handleFilterAll restores `savedViewDays` (the pre-Releases value), ignoring what the user just chose. The Day/Week toggle should either be disabled in Releases mode or update `savedViewDays` when clicked. `[fixable]`


_The Day/Week toggle is not disabled or guarded in Releases mode despite the test claiming it is, creating a stale-state bug; the /person skill and mutating flag additions are clean._

- 🟡 **regressions** (L38): The old test asserted that NO bundled skill includes Write or Edit. The new test only checks non-mutating skills, which is correct for the new design, but it means any skill can now opt into write access by setting `mutating: true`. The `mutating` flag is only validated in tests — there's no runtime enforcement preventing a repo-scoped or user-scoped skill from declaring `mutating: true` and gaining Edit/Write. This is by design (skill-policy restricts, never expands), but worth noting the policy boundary moved.


### `skills/person/SKILL.md`
_The Day/Week toggle is not disabled or guarded in Releases mode despite the test claiming it is, creating a stale-state bug; the /person skill and mutating flag additions are clean._

- 🔵 **style**: This skill is placed in `skills/person/` (repo root) alongside the other bundled skills, but the bundled-skills test discovers them from this directory. The skill references `command_center/context/people/` which is specific to the `mgmt` workspace pointed to by REPO_PATH. If Mitzo is used with other repos, this skill will find no profiles. Consider noting this workspace dependency in the skill description. `[fixable]`

PR #175, opened 2026-04-12
chore: add branch guard to prevent direct commits to main

dimakis:
## Summary

- Adds branch check to `.husky/pre-commit` that rejects commits when on `main`
- All changes must go through feature branches and PRs

## Test plan

- [ ] `git checkout main && touch test && git add test && git commit -m "test"` is rejected
- [ ] Commits on feature branches work normally

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,7 @@
+branch=$(git rev-parse --abbrev-ref HEAD)
+if [ "$branch" = "main" ]; then
+  echo "ERROR: Direct commits to main are not allowed. Create a feature branch first."
+  exit 1
+fi
+
 npx lint-staged
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -199,7 +199,7 @@ Step N: Build <component> (test-first)
 
 ## Git Workflow
 
-**Never push to main.** All work — features, fixes, docs, CI config, rule changes, typo corrections — goes through a branch and PR. No exceptions.
+**Never push to main.** All work — features, fixes, docs, CI config, rule changes, typo corrections — goes through a branch and PR. No exceptions. This is enforced by a pre-commit hook (`.husky/pre-commit`) that rejects commits on `main`.
 
 1. `git checkout -b <type>/<name>` from main
 2. Develop and commit on the branch


.husky/pre-commit
```diff
@@ -1 +1,7 @@
+branch=$(git rev-parse --abbrev-ref HEAD)
```
dimakis, line 1: 🟡 **bugs**: `git rev-parse --abbrev-ref HEAD` returns the literal string "HEAD" during a detached-HEAD state (e.g., in CI, during rebase, or when checking out a tag). This won't cause a false positive for the main-branch guard, but it's worth noting that the guard is bypassed in detached HEAD — which is fine since detached HEAD is not a direct commit to main. No action needed, just documenting the behavior.

dimakis (commented):
## Centaur Review

### `.husky/pre-commit`
_Clean, minimal change that enforces the project's existing "never push to main" policy at the pre-commit level. LGTM._

- 🟡 **bugs** (L1): `git rev-parse --abbrev-ref HEAD` returns the literal string "HEAD" during a detached-HEAD state (e.g., in CI, during rebase, or when checking out a tag). This won't cause a false positive for the main-branch guard, but it's worth noting that the guard is bypassed in detached HEAD — which is fine since detached HEAD is not a direct commit to main. No action needed, just documenting the behavior.
- 🔵 **missing_tests**: Shell-based git hooks are inherently hard to unit test and this is a simple guard. No test is expected here — this is consistent with the existing `commit-msg` hook which also has no tests.

.husky/pre-commit
```diff
@@ -1 +1,7 @@
...
+if [ "$branch" = "main" ]; then
```
dimakis, line 2: 🔵 **unsafe_assumptions**: The guard only checks for "main". Consider also blocking "master" in case the default branch name differs in forks or clones (`if [ "$branch" = "main" ] || [ "$branch" = "master" ]`). `[fixable]`

.husky/pre-commit
dimakis, line 1: 🔵 **unsafe_assumptions**: `git rev-parse --abbrev-ref HEAD` returns "HEAD" during a detached-HEAD state (e.g., rebase). This is harmless here (won't false-positive), but worth noting it won't block commits during an interactive rebase on main — which is fine since rebase commits are transient.

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `.husky/pre-commit`
_Clean, minimal change that enforces the existing branch policy. Only minor hardening suggestions — no bugs or regressions._

- 🔵 **unsafe_assumptions** (L2): The guard only checks for "main". Consider also blocking "master" in case the default branch name differs in forks or clones (`if [ "$branch" = "main" ] || [ "$branch" = "master" ]`). `[fixable]`
- 🔵 **unsafe_assumptions** (L1): `git rev-parse --abbrev-ref HEAD` returns "HEAD" during a detached-HEAD state (e.g., rebase). This is harmless here (won't false-positive), but worth noting it won't block commits during an interactive rebase on main — which is fine since rebase commits are transient.

PR #176, opened 2026-04-12
feat(chat): clickable file paths in agent output

dimakis:
## Summary

- Detects file paths (absolute `/Users/...` and relative `./src/...`) in assistant chat messages and renders them as tappable links
- Tapping a file path navigates to the FileViewer showing that file
- Back button in FileViewer returns to the originating chat session (via `from` query param)
- File path links are styled distinctly from regular URLs (monospace, subtle background)
- Paths inside code blocks and inline backticks are left alone
- Linkification is skipped during streaming to avoid flicker

## Files

- `frontend/src/lib/file-paths.ts` — `isFilePath()`, `detectFilePaths()`, `linkifyFilePaths()` utilities
- `frontend/src/components/MessageBubble.tsx` — custom `a` renderer for `file-path://` scheme
- `frontend/src/pages/FileViewer.tsx` — back-navigation via `from` query param
- `frontend/src/styles/global.css` — `.file-path-link` styling

## Test plan

- [x] 23 unit tests for path detection and linkification (`file-paths.test.ts`)
- [x] 8 unit tests for MessageBubble rendering (`MessageBubble.test.ts`)
- [x] Full suite: 974 tests passing
- [ ] Manual: send a message that produces a file path, tap it, verify FileViewer opens, back returns to chat

### DIFF
--- a/frontend/src/components/__tests__/MessageBubble.test.ts
+++ b/frontend/src/components/__tests__/MessageBubble.test.ts
@@ -100,6 +100,17 @@ describe('MessageBubble', () => {
     const html = renderToStaticMarkup(el);
 
     expect(html).toContain('file-path-link');
+
+    // Simulate click to verify navigate is called with the correct route
+    const clickEvent = { preventDefault: vi.fn() };
+    // Extract onClick from rendered component — render via createElement to get props
+    const rendered = anchor({ href: fileHref, children: '/tmp/output.md' });
+    rendered.props.onClick(clickEvent);
+
+    expect(clickEvent.preventDefault).toHaveBeenCalled();
+    expect(mockNavigate).toHaveBeenCalledWith(
+      `/files?path=${encodeURIComponent('/tmp/output.md')}&from=${encodeURIComponent('/chat/test-session')}`,
+    );
   });
 
   it('custom anchor renders normal links for non-file URLs', () => {
--- a/frontend/src/components/__tests__/MessageBubble.test.ts
+++ b/frontend/src/components/__tests__/MessageBubble.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect, vi } from 'vitest';
 // Mock react-markdown and remark-gfm before importing the component
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 let capturedComponents: Record<string, any> | undefined;
+let capturedContent: string | undefined;
 vi.mock('react-markdown', () => ({
   default: ({
     children,
@@ -12,14 +13,22 @@ vi.mock('react-markdown', () => ({
     components?: Record<string, unknown>;
   }) => {
     capturedComponents = components;
+    capturedContent = children;
     return children;
   },
 }));
 vi.mock('remark-gfm', () => ({ default: () => {} }));
 
+const mockNavigate = vi.fn();
+vi.mock('react-router-dom', () => ({
+  useNavigate: () => mockNavigate,
+  useLocation: () => ({ pathname: '/chat/test-session', search: '' }),
+}));
+
 import { createElement } from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
 import { MessageBubble, TextBubble } from '../MessageBubble';
+import { FILE_SCHEME } from '../../lib/file-paths';
 import type { FinishedMessage } from '../../types/chat';
 
 function userMessage(text: string): FinishedMessage {
@@ -60,5 +69,49 @@ describe('MessageBubble', () => {
     );
     // The raw text should contain newlines (CSS white-space: pre-wrap handles display)
     expect(html).toContain('line one\nline two\nline three');
+  });
+
+  it('linkifies file paths in assistant content', () => {
+    renderToStaticMarkup(createElement(TextBubble, { content: 'Created /tmp/output.md for you.' }));
+    expect(capturedContent).toContain(`[/tmp/output.md](${FILE_SCHEME}`);
+  });
+
+  it('does not linkify while streaming', () => {
+    renderToStaticMarkup(
+      createElement(TextBubble, { content: 'Created /tmp/output.md for you.', streaming: true }),
+    );
+    expect(capturedContent).not.toContain(FILE_SCHEME);
+    expect(capturedContent).toBe('Created /tmp/output.md for you.');
+  });
+
+  it('provides a custom anchor component for file-path links', () => {
+    renderToStaticMarkup(createElement(TextBubble, { content: 'test' }));
+    expect(capturedComponents).toBeDefined();
+    expect(capturedComponents!.a).toBeDefined();
+  });
+
+  it('custom anchor navigates to FileViewer for file-path links', () => {
+    mockNavigate.mockClear();
+    renderToStaticMarkup(createElement(TextBubble, { content: 'test' }));
+
+    const anchor = capturedComponents!.a;
+    const fileHref = `${FILE_SCHEME}${encodeURIComponent('/tmp/output.md')}`;
+    const el = createElement(anchor, { href: fileHref, children: '/tmp/output.md' });
+    const html = renderToStaticMarkup(el);
+
+    expect(html).toContain('file-path-link');
+  });
+
+  it('custom anchor renders normal links for non-file URLs', () => {
+    renderToStaticMarkup(createElement(TextBubble, { content: 'test' }));
+    const anchor = capturedComponents!.a;
+    const el = createElement(anchor, {
+      href: 'https://example.com',
+      children: 'Example',
+    });
+    const html = renderToStaticMarkup(el);
+    expect(html).toContain('href="https://example.com"');
+    expect(html).toContain('target="_blank"');
+    expect(html).not.toContain('file-path-link');
   });
 });
--- a/frontend/src/lib/file-paths.ts
+++ b/frontend/src/lib/file-paths.ts
@@ -1,0 +1,123 @@
+/** Internal scheme for file path links — intercepted by the custom renderer. */
+export const FILE_SCHEME = 'file-path://';
+
+/** Detect whether a string looks like a file path (not a URL). */
+export function isFilePath(str: string): boolean {
+  // Reject URLs
+  if (/^https?:\/\//.test(str)) return false;
+
+  // Must start with /, ./, or ../  OR contain a / with a file extension
+  const isAbsolute = str.startsWith('/');
+  const isRelative = str.startsWith('./') || str.startsWith('../');
+  const hasDirSlash = str.includes('/');
+
+  if (!isAbsolute && !isRelative && !hasDirSlash) return false;
+
+  // Reject bare root
+  if (str === '/') return false;
+
+  // Must have either a file extension or multiple path segments
+  const hasExtension = /\.\w+$/.test(str);
+  const segments = str.split('/').filter(Boolean);
+  const hasDepth = segments.length >= 2;
+
+  return hasExtension || hasDepth;
+}
+
+export interface FilePathMatch {
+  path: string;
+  start: number;
+  end: number;
+}
+
+/**
+ * Find file paths in a block of text.
+ * Matches absolute paths (/...) and relative paths (./... or ../...).
+ * Skips URLs (http:// or https://).
+ */
+export function detectFilePaths(text: string): FilePathMatch[] {
+  // Match paths: absolute (/...) or relative (./... or ../...)
+  // Path chars: word chars, hyphens, dots, @, slashes — no spaces (too greedy)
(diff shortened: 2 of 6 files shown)


frontend/src/pages/FileViewer.tsx
```diff
@@ -26,12 +28,16 @@ export function FileViewer() {
...
         <MitzoLogo />
-        {(state.isViewing || state.currentDir) && (
+        {(state.isViewing || state.currentDir || fromRoute) && (
           <button
             className="viewer-header-back"
             onClick={() => {
               editor.resetEditor();
-              nav.handleBack(editor.dirty);
+              if (fromRoute) {
+                routerNavigate(fromRoute);
```
dimakis, line 41: 🟡 **unsafe_assumptions**: Open redirect risk: `fromRoute` is read directly from the `?from=` query param and passed to `routerNavigate()` without validation. A crafted URL like `/files?path=x&from=https://evil.com` could navigate the user to an external site if react-router resolves it. Validate that `fromRoute` starts with `/` and doesn't contain `//` (protocol-relative). `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/pages/FileViewer.tsx`
_Solid feature with good test coverage. Main concerns are an open redirect via the unvalidated `from` query param, a bypassed unsaved-changes guard, and a missing navigate click test._

- 🟡 **unsafe_assumptions** (L37): Open redirect risk: `fromRoute` is read directly from the `?from=` query param and passed to `routerNavigate()` without validation. A crafted URL like `/files?path=x&from=https://evil.com` could navigate the user to an external site if react-router resolves it. Validate that `fromRoute` starts with `/` and doesn't contain `//` (protocol-relative). `[fixable]`
- 🟡 **bugs** (L34): When `fromRoute` is set, the back button bypasses the unsaved-changes guard (`editor.dirty` check inside `nav.handleBack`). If the user has unsaved edits in the file editor and clicks back, those edits are silently discarded. Should check `editor.dirty` before navigating via `fromRoute`. `[fixable]`


### `frontend/src/lib/file-paths.ts`
_Solid feature with good test coverage. Main concerns are an open redirect via the unvalidated `from` query param, a bypassed unsaved-changes guard, and a missing navigate click test._

- 🟡 **bugs** (L41): The regex `(?<!\w)(?:\.{0,2}\/[\w./@-]+(?:\/[\w./@-]+)*)` uses `\.{0,2}` which also matches a single bare `/foo` (0 dots) and `./foo` (1 dot) — but it also matches paths starting with `..` without trailing slash (e.g. `..foo/bar`). The `{0,2}` quantifier on dots without requiring the slash immediately after the dots is slightly loose, though `isFilePath` downstream filters most false positives. `[fixable]`
- 🔵 **style** (L97): When a line contains a markdown link, the entire line is skipped — even paths outside the link. For example `See [docs](/docs) and also /tmp/output.md` would not linkify `/tmp/output.md`. A more precise approach would skip only paths overlapping with link ranges, similar to the backtick handling. `[fixable]`


### `frontend/src/lib/__tests__/file-paths.test.ts`
_Solid feature with good test coverage. Main concerns are an open redirect via the unvalidated `from` query param, a bypassed unsaved-changes guard, and a missing navigate click test._

- 🔵 **missing_tests**: No test for paths containing `@` (e.g., `./node_modules/@anthropic-ai/sdk/index.ts`) despite `@` being in the regex character class. Also no test for paths with consecutive dots like `../../foo.ts` (double parent traversal). `[fixable]`


_Solid feature with good test coverage. Main concerns are an open redirect via the unvalidated `from` query param, a bypassed unsaved-changes guard, and a missing navigate click test._

- 🔵 **missing_tests** (L93): The test for `custom anchor navigates to FileViewer for file-path links` creates the element and checks HTML output, but never fires the click event to verify `mockNavigate` is called with the correct route. The navigate behavior is untested. `[fixable]`

frontend/src/pages/FileViewer.tsx
```diff
@@ -26,12 +28,16 @@ export function FileViewer() {
     <div className="viewer-page">
       <header className="viewer-header">
         <MitzoLogo />
-        {(state.isViewing || state.currentDir) && (
+        {(state.isViewing || state.currentDir || fromRoute) && (
           <button
             className="viewer-header-back"
             onClick={() => {
```
dimakis, line 37: 🟡 **bugs**: When `fromRoute` is set, the back button bypasses the unsaved-changes guard (`editor.dirty` check inside `nav.handleBack`). If the user has unsaved edits in the file editor and clicks back, those edits are silently discarded. Should check `editor.dirty` before navigating via `fromRoute`. `[fixable]`

frontend/src/lib/file-paths.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Find file paths in a block of text.
+ * Matches absolute paths (/...) and relative paths (./... or ../...).
+ * Skips URLs (http:// or https://).
+ */
+export function detectFilePaths(text: string): FilePathMatch[] {
+  // Match paths: absolute (/...) or relative (./... or ../...)
+  // Path chars: word chars, hyphens, dots, @, slashes — no spaces (too greedy)
+  const pathPattern = /(?<!\w)(?:\.{0,2}\/[\w./@-]+(?:\/[\w./@-]+)*)/g;
```
dimakis: 🟡 **bugs**: The regex `(?<!\w)(?:\.{0,2}\/[\w./@-]+(?:\/[\w./@-]+)*)` uses `\.{0,2}` which also matches a single bare `/foo` (0 dots) and `./foo` (1 dot) — but it also matches paths starting with `..` without trailing slash (e.g. `..foo/bar`). The `{0,2}` quantifier on dots without requiring the slash immediately after the dots is slightly loose, though `isFilePath` downstream filters most false positives. `[fixable]`

frontend/src/lib/file-paths.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  return result.join('\n');
+}
+
+function linkifyLine(line: string): string {
+  const matches = detectFilePaths(line);
+  if (matches.length === 0) return line;
+
+  // Check if the line already contains markdown links with paths — skip entirely
+  if (/\[[^\]]*\]\([^)]*\)/.test(line)) return line;
```
dimakis: 🔵 **style**: When a line contains a markdown link, the entire line is skipped — even paths outside the link. For example `See [docs](/docs) and also /tmp/output.md` would not linkify `/tmp/output.md`. A more precise approach would skip only paths overlapping with link ranges, similar to the backtick handling. `[fixable]`

frontend/src/components/__tests__/MessageBubble.test.ts
```diff
@@ -61,4 +70,48 @@ describe('MessageBubble', () => {
...
+    expect(capturedContent).toBe('Created /tmp/output.md for you.');
+  });
+
+  it('provides a custom anchor component for file-path links', () => {
+    renderToStaticMarkup(createElement(TextBubble, { content: 'test' }));
+    expect(capturedComponents).toBeDefined();
+    expect(capturedComponents!.a).toBeDefined();
+  });
+
+  it('custom anchor navigates to FileViewer for file-path links', () => {
```
dimakis, line 93: 🔵 **missing_tests**: The test for `custom anchor navigates to FileViewer for file-path links` creates the element and checks HTML output, but never fires the click event to verify `mockNavigate` is called with the correct route. The navigate behavior is untested. `[fixable]`

frontend/src/lib/file-paths.ts
```diff
@@ -0,0 +1,126 @@
...
+  for (const lm of line.matchAll(linkPattern)) {
+    excludedRanges.push([lm.index, lm.index + lm[0].length]);
+  }
+
+  let out = '';
+  let cursor = 0;
+
+  for (const m of matches) {
+    // Skip if inside backticks or existing markdown links
+    const inBackticks = excludedRanges.some(([s, e]) => m.start >= s && m.end <= e);
```
dimakis, line 114: 🟡 **bugs**: Variable `inBackticks` is checked against `excludedRanges` which contains both backtick ranges AND markdown link ranges, making the name misleading. More importantly, the overlap check uses `m.start >= s && m.end <= e` which requires the match to be fully contained. A partial overlap (path starts before but ends inside an excluded range) would not be caught and could produce broken markdown. `[fixable]`

frontend/src/lib/file-paths.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Find file paths in a block of text.
+ * Matches absolute paths (/...) and relative paths (./... or ../...).
+ * Skips URLs (http:// or https://).
+ */
+export function detectFilePaths(text: string): FilePathMatch[] {
+  // Match paths: absolute (/...) or relative (./... or ../...)
+  // Path chars: word chars, hyphens, dots, @, slashes — no spaces (too greedy)
+  const pathPattern = /(?<!\w)(?:(?:\.\.?)?\/[\w./@-]+(?:\/[\w./@-]+)*)/g;
```
dimakis, line 41: 🟡 **bugs**: The `pathPattern` regex character class `[\w./@-]` does not include `+`, `#`, `~`, or `%`. Paths like `~/projects/foo.ts`, `src/c++/main.cpp`, or paths with URL-encoded characters won't be detected. Consider expanding the class or documenting the limitation. `[fixable]`

frontend/src/lib/file-paths.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Already-linked paths (inside []() or backtick fences) are skipped.
+ */
+export function linkifyFilePaths(content: string): string {
+  // Process line by line to skip code blocks
+  const lines = content.split('\n');
+  let inCodeBlock = false;
+  const result: string[] = [];
+
+  for (const line of lines) {
+    if (/^```/.test(line.trimStart())) {
```
dimakis, line 77: 🟡 **unsafe_assumptions**: The code-block toggle (`inCodeBlock = !inCodeBlock`) on every line matching `/^```/` will mis-track state if a code block contains a line that happens to start with triple backticks (e.g., a markdown document explaining code fences). An unmatched fence leaves all subsequent lines in the wrong state. Consider tracking the opening fence marker to match closing fences more precisely. `[fixable]`

frontend/src/lib/file-paths.ts
dimakis, line 114: 🔵 **style**: Rename `inBackticks` to `inExcludedRange` — it guards against both backtick spans and existing markdown links, not just backticks. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/lib/file-paths.ts`
_Solid feature with good test coverage and safe redirect validation; main concerns are regex edge cases in path detection and a misleading variable name in the exclusion-range logic._

- 🟡 **bugs** (L114): Variable `inBackticks` is checked against `excludedRanges` which contains both backtick ranges AND markdown link ranges, making the name misleading. More importantly, the overlap check uses `m.start >= s && m.end <= e` which requires the match to be fully contained. A partial overlap (path starts before but ends inside an excluded range) would not be caught and could produce broken markdown. `[fixable]`
- 🟡 **bugs** (L41): The `pathPattern` regex character class `[\w./@-]` does not include `+`, `#`, `~`, or `%`. Paths like `~/projects/foo.ts`, `src/c++/main.cpp`, or paths with URL-encoded characters won't be detected. Consider expanding the class or documenting the limitation. `[fixable]`
- 🟡 **unsafe_assumptions** (L77): The code-block toggle (`inCodeBlock = !inCodeBlock`) on every line matching `/^```/` will mis-track state if a code block contains a line that happens to start with triple backticks (e.g., a markdown document explaining code fences). An unmatched fence leaves all subsequent lines in the wrong state. Consider tracking the opening fence marker to match closing fences more precisely. `[fixable]`
- 🔵 **style** (L114): Rename `inBackticks` to `inExcludedRange` — it guards against both backtick spans and existing markdown links, not just backticks. `[fixable]`


### `frontend/src/lib/__tests__/file-paths.test.ts`
_Solid feature with good test coverage and safe redirect validation; main concerns are regex edge cases in path detection and a misleading variable name in the exclusion-range logic._

- 🔵 **missing_tests**: No test for `linkifyFilePaths` when a path partially overlaps an excluded range (e.g., path starts in plain text but extends into a backtick span). Also no test for nested or unmatched code fences (e.g., a single ``` fence without closing). `[fixable]`
- 🔵 **missing_tests**: `isFilePath` accepts bare relative paths like `src/components/App.tsx` (tested on line 14), but `detectFilePaths` can never find such paths in text because the regex requires a leading `/`, `./`, or `../`. This inconsistency is untested and could confuse future contributors — consider either aligning the two or adding a test documenting the intentional gap. `[fixable]`

PR #178, opened 2026-04-12
feat: capture token usage from SDK result events

dimakis:
## Summary
- Extract token usage data (input/output/cache tokens, cost, turns, duration) from Claude Agent SDK `result` events
- Persist usage to SQLite via new `recordUsage()` method on EventStore with automatic schema migration
- Include usage data in `session_end` WebSocket events for frontend consumption

Phase 1 of the **tokens-to-goal** system — captures the raw token data at the Mitzo boundary. Goal attribution (linking sessions to goals via `goalId`) is wired in the schema but activated in Phase 2 when ContexGin's Goal Registry is integrated.

## Changes
- `server/event-store.ts` — 9 new columns (usage + goalId), migration, `recordUsage()` method
- `server/query-loop.ts` — extract usage from SDK result, persist + emit
- `server/__tests__/token-capture.test.ts` — 7 new tests (migration, store/retrieve, overwrite, listing)

## Test plan
- [x] `npm test` — 95 tests passing
- [ ] Manual: start session, verify `session_end` WS event includes `usage` object
- [ ] Manual: check SQLite DB for populated usage columns after session completes

### DIFF
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -239,23 +239,22 @@ export async function runQueryLoop(
         doneSent = true;
 
         // Extract usage data from SDK result event
+        const result = msg as {
+          usage?: Record<string, number>;
+          total_cost_usd?: number;
+          num_turns?: number;
+          duration_ms?: number;
+          duration_api_ms?: number;
+        };
         const usageData = {
-          inputTokens: (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
-            ?.input_tokens ?? 0,
-          outputTokens: (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
-            ?.output_tokens ?? 0,
-          cacheReadTokens:
-            (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
-              ?.cache_read_input_tokens ?? 0,
-          cacheCreationTokens:
-            (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
-              ?.cache_creation_input_tokens ?? 0,
-          totalCostUsd:
-            (msg as Record<string, unknown> & { total_cost_usd?: number }).total_cost_usd ?? 0,
-          numTurns: (msg as Record<string, unknown> & { num_turns?: number }).num_turns ?? 0,
-          durationMs: (msg as Record<string, unknown> & { duration_ms?: number }).duration_ms ?? 0,
-          durationApiMs:
-            (msg as Record<string, unknown> & { duration_api_ms?: number }).duration_api_ms ?? 0,
+          inputTokens: result.usage?.input_tokens ?? 0,
+          outputTokens: result.usage?.output_tokens ?? 0,
+          cacheReadTokens: result.usage?.cache_read_input_tokens ?? 0,
+          cacheCreationTokens: result.usage?.cache_creation_input_tokens ?? 0,
+          totalCostUsd: result.total_cost_usd ?? 0,
+          numTurns: result.num_turns ?? 0,
+          durationMs: result.duration_ms ?? 0,
+          durationApiMs: result.duration_api_ms ?? 0,
         };
 
         // Persist usage to durable store
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -20,6 +20,20 @@ function send(ws: WebSocket, data: unknown) {
   }
 }
 
+/** Shape of the SDK result event — fields we extract for usage tracking. */
+interface SdkResultEvent {
+  usage?: {
+    input_tokens?: number;
+    output_tokens?: number;
+    cache_read_input_tokens?: number;
+    cache_creation_input_tokens?: number;
+  };
+  total_cost_usd?: number;
+  num_turns?: number;
+  duration_ms?: number;
+  duration_api_ms?: number;
+}
+
 /**
  * Broadcast a message to all observers of a session.
  * Each send is wrapped in try/catch to prevent a single failing socket
@@ -239,13 +253,7 @@ export async function runQueryLoop(
         doneSent = true;
 
         // Extract usage data from SDK result event
-        const result = msg as {
-          usage?: Record<string, number>;
-          total_cost_usd?: number;
-          num_turns?: number;
-          duration_ms?: number;
-          duration_api_ms?: number;
-        };
+        const result = msg as SdkResultEvent;
         const usageData = {
           inputTokens: result.usage?.input_tokens ?? 0,
           outputTokens: result.usage?.output_tokens ?? 0,
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -237,7 +237,33 @@ export async function runQueryLoop(
         if (msg.session_id) emit(currentWs, { type: 'session_id', sessionId: msg.session_id });
         forceFlushPendingMessage(currentWs, currentSession);
         doneSent = true;
-        emit(currentWs, v2('session_end', { sessionId: msg.session_id }));
+
+        // Extract usage data from SDK result event
+        const usageData = {
+          inputTokens: (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
+            ?.input_tokens ?? 0,
+          outputTokens: (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
+            ?.output_tokens ?? 0,
+          cacheReadTokens:
+            (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
+              ?.cache_read_input_tokens ?? 0,
+          cacheCreationTokens:
+            (msg as Record<string, unknown> & { usage?: Record<string, number> }).usage
+              ?.cache_creation_input_tokens ?? 0,
+          totalCostUsd:
+            (msg as Record<string, unknown> & { total_cost_usd?: number }).total_cost_usd ?? 0,
+          numTurns: (msg as Record<string, unknown> & { num_turns?: number }).num_turns ?? 0,
+          durationMs: (msg as Record<string, unknown> & { duration_ms?: number }).duration_ms ?? 0,
+          durationApiMs:
+            (msg as Record<string, unknown> & { duration_api_ms?: number }).duration_api_ms ?? 0,
+        };
+
+        // Persist usage to durable store
+        if (store && resolvedSessionId) {
+          store.recordUsage(resolvedSessionId, usageData);
+        }
+
+        emit(currentWs, v2('session_end', { sessionId: msg.session_id, usage: usageData }));
         if (!registry.isAttached(clientId)) {
           const snippet = extractSnippet(snapshotBlocks, NOTIFY_SNIPPET_MAX_CHARS);
           const sid = (msg.session_id as string) || currentSession.sessionId;
--- a/frontend/src/types/ws-messages.ts
+++ b/frontend/src/types/ws-messages.ts
@@ -93,6 +93,16 @@ interface SessionEndMsg {
   type: 'session_end';
   v: 2;
   sessionId?: string;
+  usage?: {
+    inputTokens: number;
+    outputTokens: number;
+    cacheReadTokens: number;
+    cacheCreationTokens: number;
+    totalCostUsd: number;
+    numTurns: number;
+    durationMs: number;
+    durationApiMs: number;
+  };
 }
 
 interface PermissionRequestMsg {
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -882,6 +882,146 @@ describe('runQueryLoop', () => {
     });
   });
 
+  describe('usage data extraction', () => {
+    let store: EventStore;
+
+    beforeEach(() => {
+      store = new EventStore(':memory:');
+    });
+
+    afterEach(() => {
+      store.close();
+    });
+
+    it('extracts usage data from SDK result event and persists to store', async () => {
+      const events: Record<string, unknown>[] = [
+        { type: 'assistant', message: { content: [] }, session_id: 'sess-usage' },
+        {
+          type: 'result',
+          session_id: 'sess-usage',
+          usage: {
+            input_tokens: 1500,
+            output_tokens: 800,
+            cache_read_input_tokens: 200,
+            cache_creation_input_tokens: 50,
+          },
+          total_cost_usd: 0.0042,
+          num_turns: 3,
+          duration_ms: 12000,
+          duration_api_ms: 8000,
+        },
+      ];
+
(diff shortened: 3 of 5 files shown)


dimakis:
## Centaur Review

### `server/query-loop.ts`
_Solid storage-layer work with good tests for EventStore, but missing query-loop integration tests, frontend type update for the new `session_end` payload, and the SDK result extraction code is unnecessarily verbose._

- 🟡 **style** (L242): The repeated `(msg as Record<string, unknown> & { usage?: Record<string, number> })` cast is verbose and duplicated 4 times. Extract a typed local (e.g. `const result = msg as { usage?: Record<string, number>; total_cost_usd?: number; num_turns?: number; duration_ms?: number; duration_api_ms?: number };`) once, then reference `result.usage?.input_tokens` etc. `[fixable]`
- 🟡 **missing_tests** (L262): The query-loop integration (extracting usage from the SDK result event, calling `store.recordUsage`, and including `usage` in the emitted `session_end`) has no test coverage. The token-capture tests only cover `EventStore` methods directly. A query-loop test verifying that a `result` event with usage data triggers `store.recordUsage` and emits `session_end` with the `usage` payload would catch regressions in the extraction/wiring logic. `[fixable]`


_Solid storage-layer work with good tests for EventStore, but missing query-loop integration tests, frontend type update for the new `session_end` payload, and the SDK result extraction code is unnecessarily verbose._

- 🟡 **regressions** (L92): The server now emits `session_end` with a `usage` payload, but `SessionEndMsg` in the frontend types does not declare a `usage` field. The data arrives but is untyped and inaccessible without `as any`. The frontend type should be updated to include the new field so consumers can use it safely. `[fixable]`


### `server/event-store.ts`
_Solid storage-layer work with good tests for EventStore, but missing query-loop integration tests, frontend type update for the new `session_end` payload, and the SDK result extraction code is unnecessarily verbose._

- 🔵 **unsafe_assumptions** (L327): `recordUsage` uses an ad-hoc `this.db!.prepare(...)` on every call instead of a prepared statement cached in `this.stmts`. All other write operations (append, markInactive, hide) use pre-prepared statements. This is both inconsistent and slightly less efficient for repeated calls. `[fixable]`
- 🔵 **style** (L33): `goalId` is added to `SessionMeta`, `SessionRow`, the migration, and `rowToSession`, but `recordUsage` never writes it and nothing in this PR sets it. If it's a placeholder for future work, consider deferring it to the PR that actually uses it to avoid dead schema. `[fixable]`

dimakis:
## Centaur Review

### `server/event-store.ts`
_Solid, well-tested feature. The `goalId` column is dead code (never written), and the frontend silently drops the new usage payload — worth confirming both are intentional scaffolding for follow-up work._

- 🟡 **regressions**: The `goalId` column is added by migration and mapped in `rowToSession`/`SessionMeta`, but it is never written anywhere — `recordUsage` doesn't set it, `upsertSession` doesn't accept it. It appears to be scaffolding for a future feature. If intentional, a brief comment would clarify; if accidental, it should be removed to keep the schema minimal. `[fixable]`
- 🔵 **style** (L183): `migrateUsageTracking` re-fetches `PRAGMA table_info` independently from `migratePromptTracking`, which also fetches it. Minor inefficiency — could share the column set. Low priority since migrations only run once per startup. `[fixable]`


### `server/query-loop.ts`
_Solid, well-tested feature. The `goalId` column is dead code (never written), and the frontend silently drops the new usage payload — worth confirming both are intentional scaffolding for follow-up work._

- 🔵 **unsafe_assumptions** (L241): The cast `msg as { usage?: Record<string, number>; ... }` is a loose type assertion on an already-untyped `Record<string, unknown>`. This is fine for the current usage pattern, but consider extracting a named `SdkResultEvent` type (even if just local) so the field names (`input_tokens`, `cache_read_input_tokens`, etc.) are documented in one place rather than buried in optional-chaining expressions. `[fixable]`


### `frontend/src/hooks/useChatMessages.ts`
_Solid, well-tested feature. The `goalId` column is dead code (never written), and the frontend silently drops the new usage payload — worth confirming both are intentional scaffolding for follow-up work._

- 🔵 **missing_tests** (L613): The frontend handler for `session_end` does not extract or forward the new `usage` field — it only dispatches `SESSION_END` with `sessionId`. The usage data is sent over the wire but silently dropped. This isn't a bug (the data is persisted server-side), but if the intent is to display usage in the UI, the frontend reducer and/or session list will need to consume it. If that's planned for a follow-up PR, no action needed now.

dimakis:
## Centaur Review

### `server/event-store.ts`
_Solid feature addition with good test coverage and graceful defaults. Main concern is the unused `goalId` column being added speculatively — consider deferring it._

- 🔵 **style**: `goalId` / `goal_id` column is added in the migration and mapped in `rowToSession` / interfaces, but it is never written, read, or used anywhere outside tests. It appears to be a speculative addition — consider removing it from this PR and adding it when actual functionality needs it. `[fixable]`


### `frontend/src/types/chat.ts`
_Solid feature addition with good test coverage and graceful defaults. Main concern is the unused `goalId` column being added speculatively — consider deferring it._

- 🟡 **regressions** (L88): The `Session` type in `frontend/src/types/chat.ts` (used by `SessionList` page) does not include the new usage fields (`inputTokens`, `outputTokens`, `totalCostUsd`, etc.). If the frontend ever needs to display usage/cost on the session list, this type will need updating. Not a runtime break today (extra fields are just ignored), but worth noting for the next PR that surfaces usage in the UI. `[fixable]`


### `server/query-loop.ts`
_Solid feature addition with good test coverage and graceful defaults. Main concern is the unused `goalId` column being added speculatively — consider deferring it._

- 🔵 **unsafe_assumptions** (L256): The `msg as SdkResultEvent` cast trusts the SDK event shape without validation. This is fine today since the fields all use `?? 0` defaults, but if the SDK ever sends non-numeric values (e.g. `null` or a string), they'd flow through unchecked. Low risk given SDK stability, but a Zod parse-or-default would be more defensive. `[fixable]`


_Solid feature addition with good test coverage and graceful defaults. Main concern is the unused `goalId` column being added speculatively — consider deferring it._

- 🔵 **missing_tests**: No test covers the case where `resolvedSessionId` is null when the `result` event arrives (e.g., a result event with no prior `assistant` event that sets the session ID). The guard `if (store && resolvedSessionId)` would skip persistence silently — worth a test to document this edge case. `[fixable]`

server/query-loop.ts
```diff
@@ -237,7 +251,26 @@ export async function runQueryLoop(
         if (msg.session_id) emit(currentWs, { type: 'session_id', sessionId: msg.session_id });
         forceFlushPendingMessage(currentWs, currentSession);
         doneSent = true;
-        emit(currentWs, v2('session_end', { sessionId: msg.session_id }));
+
+        // Extract usage data from SDK result event
+        const result = msg as SdkResultEvent;
```
dimakis, line 256: 🟡 **unsafe_assumptions**: The SDK result event is cast to `SdkResultEvent` via `msg as SdkResultEvent`. Since `msg` is typed as `Record<string, unknown>`, this cast silently trusts the SDK shape. The `?? 0` defaults handle missing fields, so this won't crash — but if the SDK ever renames fields (e.g. `cache_read_input_tokens` → `cache_read_tokens`), usage would silently report zeros with no warning. Consider a runtime log when `result.usage` exists but individual fields are missing, to catch SDK schema drift. `[fixable]`

server/query-loop.ts
```diff
@@ -20,6 +20,20 @@ function send(ws: WebSocket, data: unknown) {
   }
```
dimakis, line 20: 🔵 **style**: The `SdkResultEvent` interface duplicates field knowledge that also lives in `event-store.ts` (`recordUsage` param type). Consider extracting the camelCase usage shape into a shared `UsageData` type in a common location (e.g. `event-store.ts` already has it inline) to avoid the two definitions drifting apart. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/event-store.ts`
_Clean, well-tested feature. Main concerns are the orphaned `goalId` field and minor type-safety improvements around the SDK cast — no bugs or regressions found._

- 🔵 **style**: `goalId` / `goal_id` column is added in the migration, interfaces (`SessionMeta`, `SessionRow`), and mapping (`rowToSession`), but is never written to — no code sets it. If this is scaffolding for a future feature, consider deferring it to that PR to keep this diff focused. If it's intentional, a brief comment would help reviewers understand its purpose. `[fixable]`


### `server/query-loop.ts`
_Clean, well-tested feature. Main concerns are the orphaned `goalId` field and minor type-safety improvements around the SDK cast — no bugs or regressions found._

- 🟡 **unsafe_assumptions** (L256): The SDK result event is cast to `SdkResultEvent` via `msg as SdkResultEvent`. Since `msg` is typed as `Record<string, unknown>`, this cast silently trusts the SDK shape. The `?? 0` defaults handle missing fields, so this won't crash — but if the SDK ever renames fields (e.g. `cache_read_input_tokens` → `cache_read_tokens`), usage would silently report zeros with no warning. Consider a runtime log when `result.usage` exists but individual fields are missing, to catch SDK schema drift. `[fixable]`
- 🔵 **style** (L20): The `SdkResultEvent` interface duplicates field knowledge that also lives in `event-store.ts` (`recordUsage` param type). Consider extracting the camelCase usage shape into a shared `UsageData` type in a common location (e.g. `event-store.ts` already has it inline) to avoid the two definitions drifting apart. `[fixable]`


_Clean, well-tested feature. Main concerns are the orphaned `goalId` field and minor type-safety improvements around the SDK cast — no bugs or regressions found._

- 🔵 **missing_tests**: No test covers the case where `store` is provided but `resolvedSessionId` is still undefined at result time (guard on line 269). This edge case is unlikely but the guard exists — a test confirming `recordUsage` is not called when sessionId is missing would document the intent. `[fixable]`

server/query-loop.ts
dimakis, line 256: 🔵 **unsafe_assumptions**: The cast `msg as SdkResultEvent` is a type assertion on a `Record<string, unknown>`. This is the established pattern in this file (other event types use similar casts), but the usage fields accessed via optional chaining (e.g., `result.usage?.input_tokens`) could silently produce 0 if the SDK changes field names. Consider adding a debug log when `result.usage` is undefined to aid troubleshooting. `[fixable]`

dimakis (commented):
## Centaur Review

Found **4** issue(s).


### `server/event-store.ts`
_Clean, well-tested feature. Usage extraction, persistence, and WS delivery are correct. Minor suggestions around dead goalId field and edge-case test coverage._

- 🔵 **style**: goalId is added to SessionMeta, SessionRow, migration, and rowToSession but is never written or consumed anywhere in the codebase — it's dead code. If it's planned for a future feature, it should land with that feature to avoid confusion. `[fixable]`


_Clean, well-tested feature. Usage extraction, persistence, and WS delivery are correct. Minor suggestions around dead goalId field and edge-case test coverage._

- 🔵 **missing_tests**: No test covers the case where store is undefined (the optional parameter). The production code guards with `if (store && resolvedSessionId)`, but this branch isn't explicitly tested. Existing tests without a store cover it implicitly, but a focused assertion that no error occurs and session_end still includes usage would be clearer. `[fixable]`
- 🔵 **missing_tests**: No test covers the case where resolvedSessionId is undefined when the result event arrives (e.g., if the assistant event lacks a session_id and no prior event set it). In that scenario, store.recordUsage is skipped but usage is still sent on the WebSocket — worth an explicit test to document the intended behavior. `[fixable]`


### `server/query-loop.ts`
_Clean, well-tested feature. Usage extraction, persistence, and WS delivery are correct. Minor suggestions around dead goalId field and edge-case test coverage._

- 🔵 **unsafe_assumptions** (L256): The cast `msg as SdkResultEvent` is a type assertion on a `Record<string, unknown>`. This is the established pattern in this file (other event types use similar casts), but the usage fields accessed via optional chaining (e.g., `result.usage?.input_tokens`) could silently produce 0 if the SDK changes field names. Consider adding a debug log when `result.usage` is undefined to aid troubleshooting. `[fixable]`

dimakis:
## Centaur Review

### `server/query-loop.ts`
_Solid feature addition with good test coverage for the happy path. Main gaps: usage data is sent to frontend but not consumed, `goalId` is migrated but unused, and error-path `session_end` messages don't include usage._

- 🟡 **bugs** (L506): The `finally` block emits `session_end` without `usage` data when `doneSent` is false (error/abort paths). This means sessions that end abnormally will send a `session_end` with no `usage` field, which is fine per the optional type — but usage data is also NOT persisted to the store in these paths. If the SDK emits a `result` event before the error, usage would have been captured, but if the loop throws before processing `result`, the session row will show all zeros. Consider whether this is the intended behavior or if partial usage should be captured from the error path.
- 🔵 **unsafe_assumptions** (L257): The `SdkResultEvent` type is a local interface with all optional fields. The cast `msg as SdkResultEvent` on an untyped `Record<string, unknown>` is safe because all fields default to 0 via `?? 0`. However, if the SDK ever changes field names (e.g., `cache_read_input_tokens` → `cache_read_tokens`), the extraction will silently default to 0 with no warning. Consider adding a debug log when `result.usage` is present but individual fields are missing, to catch SDK contract drift. `[fixable]`


### `frontend/src/hooks/useChatMessages.ts`
_Solid feature addition with good test coverage for the happy path. Main gaps: usage data is sent to frontend but not consumed, `goalId` is migrated but unused, and error-path `session_end` messages don't include usage._

- 🔵 **regressions** (L614): The `session_end` handler dispatches `SESSION_END` but does not forward the new `usage` field to the reducer or any state. The usage data arrives on the WebSocket but is silently dropped. This isn't a regression (the field didn't exist before), but the data is being sent to the frontend with no consumer — it's dead on arrival until a follow-up PR wires it up.


### `server/event-store.ts`
_Solid feature addition with good test coverage for the happy path. Main gaps: usage data is sent to frontend but not consumed, `goalId` is migrated but unused, and error-path `session_end` messages don't include usage._

- 🔵 **style** (L205): `goal_id` column is added in the migration and mapped in `SessionMeta`/`SessionRow`/`rowToSession`, but is never set or used anywhere in the codebase. This is a forward-looking schema addition with no implementation. If intentional (planned feature), consider noting it in the PR description; otherwise it adds dead surface area to the data model. `[fixable]`


_Solid feature addition with good test coverage for the happy path. Main gaps: usage data is sent to frontend but not consumed, `goalId` is migrated but unused, and error-path `session_end` messages don't include usage._

- 🔵 **missing_tests**: No test covers the `finally` block path — where the loop ends via error/abort and `doneSent` is false. In that case, `session_end` is emitted without `usage`. A test verifying that the fallback `session_end` message shape is correct (no `usage` field or zero-valued) would strengthen coverage of error paths. `[fixable]`

PR #179, opened 2026-04-12
feat: wire goal auto-creation and usage reporting to ContexGin

dimakis:
## Summary
- **Phase 2 of tokens-to-goal**: Mitzo sessions now create goals in ContexGin's Goal Registry on start and report token usage as contributions on end
- New `goal-client.ts` module: ContexGin API client with 30s health-check caching, fire-and-forget semantics
- Goal title auto-derived from initial prompt (first sentence, max 80 chars)
- `goalId` persisted on session metadata via existing column from Phase 1
- All calls are non-blocking — ContexGin being down never impacts Mitzo

## Changes
- `server/goal-client.ts` — new module: `createGoal()`, `reportUsage()`, `linkArtifact()`, `deriveGoalTitle()`
- `server/query-loop.ts` — wire goal creation on first assistant event, usage reporting on result event
- `server/event-store.ts` — add `goalId` to `upsertSession` update path
- `.env.example` — add `CONTEXGIN_URL`
- `server/__tests__/goal-client.test.ts` — 9 tests

## Test plan
- [x] 9 goal-client unit tests passing
- [x] 1022 server tests passing (all pre-existing tests unaffected)
- [ ] Manual: start ContexGin with `--goals-db`, start Mitzo session, verify goal created at `/api/goals`
- [ ] Manual: complete session, verify usage contribution posted at `/api/goals/:id/contributions`
- [ ] Manual: verify Mitzo works normally when ContexGin is not running (graceful degradation)

### DIFF
--- a/server/goal-client.ts
+++ b/server/goal-client.ts
@@ -140,29 +140,6 @@ export async function reportUsage(
   }
 }
 
-// ── Artifact linking ───────────────────────────────────────────
-
-/**
- * Link an artifact (PR, commit, branch) to a goal.
- */
-export async function linkArtifact(
-  goalId: string,
-  artifact: { type: string; ref: string; repo?: string },
-): Promise<void> {
-  if (!(await isAvailable())) return;
-
-  try {
-    await fetch(`${CONTEXGIN_URL}/api/goals/${goalId}/artifacts`, {
-      method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify(artifact),
-      signal: AbortSignal.timeout(5000),
-    });
-  } catch {
-    // Best-effort
-  }
-}
-
 // ── Goal title derivation ──────────────────────────────────────
 
 /**
@@ -171,8 +148,8 @@ export async function linkArtifact(
  */
 export function deriveGoalTitle(prompt: string): string {
   const cleaned = prompt.replace(/\s+/g, ' ').trim();
-  // First sentence
-  const sentenceEnd = cleaned.search(/[.!?]\s/);
+  // First sentence (match punctuation followed by space or end of string)
+  const sentenceEnd = cleaned.search(/[.!?](\s|$)/);
   const firstSentence = sentenceEnd > 0 ? cleaned.slice(0, sentenceEnd + 1) : cleaned;
   // Cap at 80 chars
   if (firstSentence.length <= 80) return firstSentence;
--- a/server/goal-client.ts
+++ b/server/goal-client.ts
@@ -1,0 +1,180 @@
+/**
+ * ContexGin Goal Registry client.
+ *
+ * Handles goal auto-creation on session start and usage reporting on session end.
+ * All calls are fire-and-forget — ContexGin being down never blocks Mitzo.
+ */
+import { createLogger } from './logger.js';
+
+const log = createLogger('goal-client');
+
+const CONTEXGIN_URL = process.env.CONTEXGIN_URL || 'http://localhost:8321';
+
+// ── Types ──────────────────────────────────────────────────────
+
+interface GoalResponse {
+  id: string;
+  title: string;
+  status: string;
+}
+
+interface UsageContributionInput {
+  source: string;
+  sourceId: string;
+  sourceLabel?: string;
+  inputTokens?: number;
+  outputTokens?: number;
+  cacheReadTokens?: number;
+  cacheCreationTokens?: number;
+  costUsd?: number;
+  turns?: number;
+  toolCalls?: number;
+  durationMs?: number;
+  durationApiMs?: number;
+  metadata?: Record<string, unknown>;
+}
+
+// ── Health check ───────────────────────────────────────────────
+
+let available: boolean | null = null;
+let lastCheck = 0;
+const CHECK_INTERVAL_MS = 30_000;
+
+/** Reset cached availability state (for testing). */
+export function resetAvailability(): void {
+  available = null;
+  lastCheck = 0;
+}
+
+async function isAvailable(): Promise<boolean> {
+  const now = Date.now();
+  if (available !== null && now - lastCheck < CHECK_INTERVAL_MS) {
+    return available;
+  }
+  try {
+    const res = await fetch(`${CONTEXGIN_URL}/health`, {
+      signal: AbortSignal.timeout(2000),
+    });
+    available = res.ok;
+  } catch {
+    available = false;
+  }
+  lastCheck = now;
+  return available;
+}
+
+// ── Goal creation ──────────────────────────────────────────────
+
+/**
+ * Create a goal in the ContexGin Goal Registry.
+ * Returns the goal ID, or null if ContexGin is unreachable.
+ */
+export async function createGoal(
+  title: string,
+  opts?: { description?: string; contextCondition?: string },
+): Promise<string | null> {
+  if (!(await isAvailable())) {
+    log.debug('ContexGin unavailable, skipping goal creation');
+    return null;
+  }
+
+  try {
+    const res = await fetch(`${CONTEXGIN_URL}/api/goals`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        title,
+        description: opts?.description,
+        contextCondition: opts?.contextCondition,
+      }),
+      signal: AbortSignal.timeout(5000),
+    });
+
+    if (!res.ok) {
+      log.warn('goal creation failed', { status: res.status });
+      return null;
+    }
+
+    const goal = (await res.json()) as GoalResponse;
+    log.info('goal created', { goalId: goal.id, title: goal.title });
+    return goal.id;
+  } catch (err) {
+    log.warn('goal creation error', { error: (err as Error).message });
+    available = false;
+    return null;
+  }
+}
+
+// ── Usage reporting ────────────────────────────────────────────
+
+/**
+ * Report token usage to the Goal Registry as a contribution.
+ * Fire-and-forget — errors are logged but never thrown.
+ */
+export async function reportUsage(
+  goalId: string,
+  contribution: UsageContributionInput,
+): Promise<void> {
+  if (!(await isAvailable())) {
+    log.debug('ContexGin unavailable, skipping usage report');
+    return;
+  }
+
+  try {
+    const res = await fetch(`${CONTEXGIN_URL}/api/goals/${goalId}/contributions`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(contribution),
+      signal: AbortSignal.timeout(5000),
+    });
+
+    if (!res.ok) {
+      log.warn('usage report failed', { goalId, status: res.status });
+      return;
+    }
+
+    log.info('usage reported', { goalId, source: contribution.source });
+  } catch (err) {
+    log.warn('usage report error', { goalId, error: (err as Error).message });
+    available = false;
+  }
+}
+
+// ── Artifact linking ───────────────────────────────────────────
+
+/**
+ * Link an artifact (PR, commit, branch) to a goal.
+ */
+export async function linkArtifact(
+  goalId: string,
+  artifact: { type: string; ref: string; repo?: string },
+): Promise<void> {
+  if (!(await isAvailable())) return;
+
+  try {
+    await fetch(`${CONTEXGIN_URL}/api/goals/${goalId}/artifacts`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(artifact),
+      signal: AbortSignal.timeout(5000),
+    });
+  } catch {
+    // Best-effort
+  }
+}
+
+// ── Goal title derivation ──────────────────────────────────────
+
+/**
+ * Derive a short goal title from the initial prompt.
+ * Truncates to first sentence or 80 chars, whichever is shorter.
+ */
+export function deriveGoalTitle(prompt: string): string {
+  const cleaned = prompt.replace(/\s+/g, ' ').trim();
+  // First sentence
+  const sentenceEnd = cleaned.search(/[.!?]\s/);
+  const firstSentence = sentenceEnd > 0 ? cleaned.slice(0, sentenceEnd + 1) : cleaned;
+  // Cap at 80 chars
+  if (firstSentence.length <= 80) return firstSentence;
+  return firstSentence.slice(0, 77) + '...';
+}
--- a/server/goal-client.ts
+++ b/server/goal-client.ts
@@ -98,8 +98,8 @@ export async function createGoal(
     const goal = (await res.json()) as GoalResponse;
     log.info('goal created', { goalId: goal.id, title: goal.title });
     return goal.id;
-  } catch (err) {
-    log.warn('goal creation error', { error: (err as Error).message });
+  } catch (err: unknown) {
+    log.warn('goal creation error', { error: err instanceof Error ? err.message : String(err) });
     available = false;
     return null;
   }
@@ -121,12 +121,15 @@ export async function reportUsage(
   }
 
   try {
-    const res = await fetch(`${CONTEXGIN_URL}/api/goals/${goalId}/contributions`, {
-      method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify(contribution),
-      signal: AbortSignal.timeout(5000),
-    });
+    const res = await fetch(
+      `${CONTEXGIN_URL}/api/goals/${encodeURIComponent(goalId)}/contributions`,
+      {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(contribution),
+        signal: AbortSignal.timeout(5000),
+      },
+    );
 
     if (!res.ok) {
       log.warn('usage report failed', { goalId, status: res.status });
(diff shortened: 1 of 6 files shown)


server/query-loop.ts
```diff
@@ -270,6 +293,29 @@ export async function runQueryLoop(
           store.recordUsage(resolvedSessionId, usageData);
         }
 
+        // Report usage to ContexGin Goal Registry
+        if (resolvedGoalId && resolvedSessionId) {
```
dimakis, line 324: 🟡 **bugs**: Race condition: `createGoal` on line 250 is fire-and-forget (`.then()`, not awaited), but `reportUsage` on line 297 depends on `resolvedGoalId` being set. For fast/short sessions, the `result` event can arrive before the health-check + POST resolve, leaving `resolvedGoalId` undefined and silently dropping usage reporting. Consider awaiting the goal creation or deferring the usage report until the goal promise settles. `[fixable]`

server/query-loop.ts
```diff
@@ -243,6 +245,27 @@ export async function runQueryLoop(
...
+                  if (store) {
+                    store.upsertSession({ sessionId: resolvedSessionId!, goalId });
+                  }
+                  log.info('session linked to goal', {
+                    sessionId: resolvedSessionId,
+                    goalId,
+                  });
+                }
+              })
+              .catch(() => {});
```
dimakis: 🔵 **style**: Empty `.catch(() => {})` on the `createGoal` promise silently swallows all errors including unexpected ones. The `createGoal` function already handles errors internally and returns null, so this catch is redundant — but if kept, consider at least logging at debug level. `[fixable]`

server/goal-client.ts
```diff
@@ -0,0 +1,180 @@
...
+// ── Goal title derivation ──────────────────────────────────────
+
+/**
+ * Derive a short goal title from the initial prompt.
+ * Truncates to first sentence or 80 chars, whichever is shorter.
+ */
+export function deriveGoalTitle(prompt: string): string {
+  const cleaned = prompt.replace(/\s+/g, ' ').trim();
+  // First sentence
+  const sentenceEnd = cleaned.search(/[.!?]\s/);
```
dimakis: 🔵 **bugs**: `deriveGoalTitle` regex `[.!?]\s` requires whitespace after the punctuation mark, so a prompt ending exactly with a period (e.g., `"Fix the bug."`) won't match as a sentence boundary and the full string is used. This is mostly benign (the 80-char cap still applies) but doesn't match the docstring's "truncates to first sentence" claim for single-sentence prompts ending in punctuation. `[fixable]`

server/goal-client.ts
```diff
@@ -0,0 +1,180 @@
...
+ * ContexGin Goal Registry client.
+ *
+ * Handles goal auto-creation on session start and usage reporting on session end.
+ * All calls are fire-and-forget — ContexGin being down never blocks Mitzo.
+ */
+import { createLogger } from './logger.js';
+
+const log = createLogger('goal-client');
+
+const CONTEXGIN_URL = process.env.CONTEXGIN_URL || 'http://localhost:8321';
```
dimakis, line 11: 🔵 **unsafe_assumptions**: `CONTEXGIN_URL` is captured once at module load time. If the env var is set after import (e.g., in tests or late-binding config), the client uses the default `localhost:8321`. This is consistent with how other modules in the codebase work, but worth noting since the test suite works around it by mocking `fetch` rather than testing with a different URL.

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid fire-and-forget integration with good degradation behavior, but the async goal creation has a race with usage reporting that can silently drop data for fast sessions._

- 🟡 **bugs** (L297): Race condition: `createGoal` on line 250 is fire-and-forget (`.then()`, not awaited), but `reportUsage` on line 297 depends on `resolvedGoalId` being set. For fast/short sessions, the `result` event can arrive before the health-check + POST resolve, leaving `resolvedGoalId` undefined and silently dropping usage reporting. Consider awaiting the goal creation or deferring the usage report until the goal promise settles. `[fixable]`
- 🔵 **style** (L267): Empty `.catch(() => {})` on the `createGoal` promise silently swallows all errors including unexpected ones. The `createGoal` function already handles errors internally and returns null, so this catch is redundant — but if kept, consider at least logging at debug level. `[fixable]`


### `server/__tests__/goal-client.test.ts`
_Solid fire-and-forget integration with good degradation behavior, but the async goal creation has a race with usage reporting that can silently drop data for fast sessions._

- 🟡 **missing_tests**: `linkArtifact` is exported from `goal-client.ts` but has no tests and no callers anywhere in the codebase. Either add tests or remove the dead code. `[fixable]`
- 🔵 **missing_tests**: No tests verify the health-check caching behavior (`CHECK_INTERVAL_MS` / 30s TTL). A second call within the window should skip the fetch; a call after the window should re-check. This is core to the availability logic. `[fixable]`


### `server/goal-client.ts`
_Solid fire-and-forget integration with good degradation behavior, but the async goal creation has a race with usage reporting that can silently drop data for fast sessions._

- 🔵 **bugs** (L175): `deriveGoalTitle` regex `[.!?]\s` requires whitespace after the punctuation mark, so a prompt ending exactly with a period (e.g., `"Fix the bug."`) won't match as a sentence boundary and the full string is used. This is mostly benign (the 80-char cap still applies) but doesn't match the docstring's "truncates to first sentence" claim for single-sentence prompts ending in punctuation. `[fixable]`
- 🔵 **unsafe_assumptions** (L11): `CONTEXGIN_URL` is captured once at module load time. If the env var is set after import (e.g., in tests or late-binding config), the client uses the default `localhost:8321`. This is consistent with how other modules in the codebase work, but worth noting since the test suite works around it by mocking `fetch` rather than testing with a different URL.

server/query-loop.ts
```diff
@@ -270,6 +281,39 @@ export async function runQueryLoop(
...
+          const goalId = await goalCreationPromise;
+          if (goalId) {
+            resolvedGoalId = goalId;
+            store?.upsertSession({ sessionId: resolvedSessionId, goalId });
+            log.info('session linked to goal', { sessionId: resolvedSessionId, goalId });
+          }
+          goalCreationPromise = undefined;
+        }
+
+        if (resolvedGoalId && resolvedSessionId) {
```
dimakis, line 324: 🟡 **bugs**: In multi-turn sessions, `reportUsage` fires on every `result` event (once `resolvedGoalId` is set), but `usageData` from each `result` event contains cumulative totals (not deltas). This means subsequent turns will re-report the full cumulative usage, leading to double/triple counting in the goal registry. `[fixable]`

server/query-loop.ts
```diff
@@ -243,6 +246,14 @@ export async function runQueryLoop(
               store.incrementPromptCount(resolvedSessionId);
             }
           }
+
+          // Auto-create goal in ContexGin Goal Registry (awaited at session end)
+          if (initialPrompt && resolvedSessionId) {
```
dimakis, line 275: 🟡 **bugs**: Goal creation is only attempted on the first turn (`initialPrompt` is set once). However, `goalCreationPromise` is only awaited inside the `result` handler's `if (goalCreationPromise && resolvedSessionId)` block. If the session is aborted or errors out before a `result` event, the promise is left dangling — no cleanup or `.catch()` on it, risking an unhandled rejection if `createGoal` throws after the abort signal fires. `[fixable]`

server/query-loop.ts
```diff
@@ -270,6 +281,39 @@ export async function runQueryLoop(
...
+          }
+          goalCreationPromise = undefined;
+        }
+
+        if (resolvedGoalId && resolvedSessionId) {
+          reportUsage(resolvedGoalId, {
+            source: 'mitzo_session',
+            sourceId: resolvedSessionId,
+            sourceLabel: initialPrompt
+              ? `Mitzo: ${deriveGoalTitle(initialPrompt)}`
```
dimakis: 🔵 **style**: `deriveGoalTitle(initialPrompt)` is called twice — once at goal creation time (line 252) and again in the usage report label (line 300). The result is deterministic; store it in a variable to avoid redundant computation. `[fixable]`

server/goal-client.ts
```diff
@@ -0,0 +1,157 @@
...
+
```
dimakis, line 12: 🟡 **unsafe_assumptions**: `CONTEXGIN_URL` is read once at module load from `process.env`. If the env var is set after import (e.g., in test setup), the client uses the default `localhost:8321`. This also means the URL cannot be changed at runtime. Consider reading it lazily or accepting it as a parameter. `[fixable]`

server/goal-client.ts
```diff
@@ -0,0 +1,157 @@
...
+// ── Usage reporting ────────────────────────────────────────────
+
+/**
+ * Report token usage to the Goal Registry as a contribution.
+ * Fire-and-forget — errors are logged but never thrown.
+ */
+export async function reportUsage(
+  goalId: string,
+  contribution: UsageContributionInput,
+): Promise<void> {
```
dimakis, line 117: 🔵 **unsafe_assumptions**: The `goalId` is interpolated directly into the URL path (`/api/goals/${goalId}/contributions`) without validation or encoding. If a goal ID contained path-separator characters or URL-special chars, it could cause unexpected routing. Consider `encodeURIComponent(goalId)`. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid fire-and-forget integration with good health-check caching, but the main concern is that `reportUsage` fires on every turn with cumulative totals, which will double-count usage in multi-turn sessions._

- 🟡 **bugs** (L295): In multi-turn sessions, `reportUsage` fires on every `result` event (once `resolvedGoalId` is set), but `usageData` from each `result` event contains cumulative totals (not deltas). This means subsequent turns will re-report the full cumulative usage, leading to double/triple counting in the goal registry. `[fixable]`
- 🟡 **bugs** (L251): Goal creation is only attempted on the first turn (`initialPrompt` is set once). However, `goalCreationPromise` is only awaited inside the `result` handler's `if (goalCreationPromise && resolvedSessionId)` block. If the session is aborted or errors out before a `result` event, the promise is left dangling — no cleanup or `.catch()` on it, risking an unhandled rejection if `createGoal` throws after the abort signal fires. `[fixable]`
- 🔵 **style** (L300): `deriveGoalTitle(initialPrompt)` is called twice — once at goal creation time (line 252) and again in the usage report label (line 300). The result is deterministic; store it in a variable to avoid redundant computation. `[fixable]`


### `server/goal-client.ts`
_Solid fire-and-forget integration with good health-check caching, but the main concern is that `reportUsage` fires on every turn with cumulative totals, which will double-count usage in multi-turn sessions._

- 🟡 **unsafe_assumptions** (L12): `CONTEXGIN_URL` is read once at module load from `process.env`. If the env var is set after import (e.g., in test setup), the client uses the default `localhost:8321`. This also means the URL cannot be changed at runtime. Consider reading it lazily or accepting it as a parameter. `[fixable]`
- 🔵 **unsafe_assumptions** (L117): The `goalId` is interpolated directly into the URL path (`/api/goals/${goalId}/contributions`) without validation or encoding. If a goal ID contained path-separator characters or URL-special chars, it could cause unexpected routing. Consider `encodeURIComponent(goalId)`. `[fixable]`


### `server/__tests__/goal-client.test.ts`
_Solid fire-and-forget integration with good health-check caching, but the main concern is that `reportUsage` fires on every turn with cumulative totals, which will double-count usage in multi-turn sessions._

- 🔵 **missing_tests**: No test verifies that `createGoal` sets `available = false` on a network error during goal creation (the catch block at line 101 of goal-client.ts). This is a distinct code path from the health check failure — a successful health check followed by a network error on the POST should mark the service unavailable. `[fixable]`


_Solid fire-and-forget integration with good health-check caching, but the main concern is that `reportUsage` fires on every turn with cumulative totals, which will double-count usage in multi-turn sessions._

- 🔵 **missing_tests**: No test coverage for `goalId` being persisted via `upsertSession`. The event-store test file doesn't exercise the `goal_id` column at all. `[fixable]`

server/query-loop.ts
```diff
@@ -270,6 +281,39 @@ export async function runQueryLoop(
           store.recordUsage(resolvedSessionId, usageData);
         }
 
+        // Resolve goal creation (if pending) and report usage
+        if (goalCreationPromise && resolvedSessionId) {
+          const goalId = await goalCreationPromise;
```
dimakis, line 315: 🟡 **bugs**: Sessions resumed from history never report usage to ContexGin because `resolvedGoalId` is only populated from a newly created goal in this process. Existing sessions already persisted with `goal_id` are not loaded back into the query loop, so subsequent turns/results after reconnect or resume silently skip reporting. `[fixable]`

server/query-loop.ts
```diff
@@ -243,6 +246,14 @@ export async function runQueryLoop(
               store.incrementPromptCount(resolvedSessionId);
             }
           }
```
dimakis, line 272: 🟡 **missing_tests**: There are unit tests for `goal-client`, but no integration coverage for the new query-loop wiring: auto-creating a goal on first turn, persisting `goal_id`, and reporting usage on result. Given the async/session-state interactions here, this path should be exercised with query-loop tests. `[fixable]`

server/goal-client.ts
```diff
@@ -0,0 +1,157 @@
...
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        title,
+        description: opts?.description,
+        contextCondition: opts?.contextCondition,
+      }),
+      signal: AbortSignal.timeout(5000),
+    });
+
+    if (!res.ok) {
```
dimakis, line 93: 🟡 **unsafe_assumptions**: The catch blocks cast `err` to `Error` and read `.message` directly. This project convention is to use `unknown` + `instanceof Error`; as written, non-Error throws will log `undefined` and weaken type safety. `[fixable]`

dimakis (commented):
## Codex Review

### `server/query-loop.ts`
_PR is close, but goal reporting currently misses resumed sessions and the new query-loop wiring lacks integration coverage._

- 🟡 **bugs** (L286): Sessions resumed from history never report usage to ContexGin because `resolvedGoalId` is only populated from a newly created goal in this process. Existing sessions already persisted with `goal_id` are not loaded back into the query loop, so subsequent turns/results after reconnect or resume silently skip reporting. `[fixable]`
- 🟡 **missing_tests** (L248): There are unit tests for `goal-client`, but no integration coverage for the new query-loop wiring: auto-creating a goal on first turn, persisting `goal_id`, and reporting usage on result. Given the async/session-state interactions here, this path should be exercised with query-loop tests. `[fixable]`


### `server/goal-client.ts`
_PR is close, but goal reporting currently misses resumed sessions and the new query-loop wiring lacks integration coverage._

- 🟡 **unsafe_assumptions** (L93): The catch blocks cast `err` to `Error` and read `.message` directly. This project convention is to use `unknown` + `instanceof Error`; as written, non-Error throws will log `undefined` and weaken type safety. `[fixable]`

server/query-loop.ts
```diff
@@ -270,6 +310,46 @@ export async function runQueryLoop(
           store.recordUsage(resolvedSessionId, usageData);
```
dimakis, line 310: 🟡 **bugs**: Usage reporting is fired without awaiting or attaching a catch handler. If `reportUsage()` rejects before entering its internal try/catch (for example because `fetch` or `AbortSignal.timeout` throws synchronously in the runtime), this creates an unhandled promise rejection during result processing. The surrounding code already defensively catches `createGoal`; `reportUsage` should be handled the same way or awaited. `[fixable]`

server/query-loop.ts
```diff
@@ -270,6 +310,46 @@ export async function runQueryLoop(
...
         }
 
+        // Resolve goal creation (if pending) and report usage
```
dimakis, line 313: 🟡 **regressions**: Delta computation assumes SDK usage counters are monotonically increasing for the lifetime of the loop. On resumed/reattached sessions, `lastReportedUsage` is reset to zero and no prior totals are restored from the event store, so the first post-reattach `result` will re-report the full cumulative session usage and double-count contributions for an existing goal. `[fixable]`

server/query-loop.ts
```diff
@@ -212,6 +228,17 @@ export async function runQueryLoop(
       const currentWs = currentSession.ws;
```
dimakis, line 228: 🟡 **missing_tests**: The new goal wiring in `runQueryLoop()` has no integration coverage. There should be tests for (at minimum) resumed sessions loading `goalId` from the store and reporting only incremental usage, plus initial sessions persisting `goalId` after goal creation. `[fixable]`

dimakis (commented):
## Codex Review

### `server/query-loop.ts`
_Goal client unit coverage is solid, but the query-loop integration has a double-counting risk on resumed sessions and needs safer async handling/tests._

- 🟡 **bugs** (L310): Usage reporting is fired without awaiting or attaching a catch handler. If `reportUsage()` rejects before entering its internal try/catch (for example because `fetch` or `AbortSignal.timeout` throws synchronously in the runtime), this creates an unhandled promise rejection during result processing. The surrounding code already defensively catches `createGoal`; `reportUsage` should be handled the same way or awaited. `[fixable]`
- 🟡 **regressions** (L313): Delta computation assumes SDK usage counters are monotonically increasing for the lifetime of the loop. On resumed/reattached sessions, `lastReportedUsage` is reset to zero and no prior totals are restored from the event store, so the first post-reattach `result` will re-report the full cumulative session usage and double-count contributions for an existing goal. `[fixable]`
- 🟡 **missing_tests** (L228): The new goal wiring in `runQueryLoop()` has no integration coverage. There should be tests for (at minimum) resumed sessions loading `goalId` from the store and reporting only incremental usage, plus initial sessions persisting `goalId` after goal creation. `[fixable]`

PR #180, opened 2026-04-12
feat(todo): starred items — ⭐ toggle on any todo

dimakis:
## Summary
- Adds `starred: boolean` to `TodoItem` type and Zod response schema
- Adds `star`/`unstar` to the `TodoActionBody` action enum (server-validated)
- `TodoCard`: star button (⭐/☆) in card header — same button toggles on and off
- `useTodoData`: new `star(id)` with optimistic local toggle + server sync; reads current state to decide `star` vs `unstar` action
- All existing `TodoCard` tests updated with required `onStar` prop; 6 new star-specific tests added (TDD — red before green)

## Test plan
- [ ] `npm test` passes (35/35 in targeted files, 1651 total)
- [ ] Tap ☆ on any todo — optimistically shows ⭐, server confirms
- [ ] Tap ⭐ again — toggles back to ☆
- [ ] Manually created items (profile=manual) arrive pre-starred


dimakis:
## Centaur Review

Found **5** issue(s) (2 critical) (2 warning).


### `frontend/src/pages/TodoView.tsx`
_Build-breaking regression: TodoView doesn't pass the required `onStar` prop to TodoCard. Also has a race condition in the star toggle's optimistic update logic._

- 🔴 **regressions** (L155): TodoView renders `<TodoCard>` without the required `onStar` prop. `onStar` is non-optional in `TodoCardProps`, so this is a TypeScript compilation error — the build will fail. The `star` function from `useTodoData` is also not destructured on line 77. `[fixable]`


### `frontend/src/hooks/useTodoData.ts`
_Build-breaking regression: TodoView doesn't pass the required `onStar` prop to TodoCard. Also has a race condition in the star toggle's optimistic update logic._

- 🔴 **bugs** (L110): Race condition in `star`: `setData` optimistically toggles the item (line 113), then `data` from the closure is read (line 119) to decide the API action. If the user taps star twice quickly before React re-renders, both calls see the same stale `data`, sending the same action (e.g. two `star` requests) while the UI double-toggles back to the original state. Fix: use the `setData` updater's `prev` to determine the action, or read the pre-toggle value before calling `setData`. `[fixable]`
- 🟡 **unsafe_assumptions** (L110): The `star` callback has `[data]` as a dependency, causing it to be recreated on every data change. This triggers unnecessary re-renders of every component that receives `star` as a prop. Other actions (`ack`, `done`, `snooze`) use `performAction` with `[]` deps — `star` should follow the same pattern or use a ref for the data lookup. `[fixable]`
- 🔵 **missing_tests** (L30): No unit tests for `toggleStarInTree` or `findInTree`. `toggleStarInTree` has a subtle behavior — it skips recursing into children when `children.length === 0`, which is correct but worth a test to prevent regressions (e.g. if someone refactors and removes the early-exit). `[fixable]`


### `server/__tests__/todos.test.ts`
_Build-breaking regression: TodoView doesn't pass the required `onStar` prop to TodoCard. Also has a race condition in the star toggle's optimistic update logic._

- 🟡 **style** (L179): The two new `star`/`unstar` tests are outside the `describe('todo routes')` block and are top-level `it()` calls. They should be inside the describe block for consistent grouping and to share the same setup context. `[fixable]`

dimakis:
## Centaur Review

### `frontend/src/pages/TodoView.tsx`
_Critical: TodoView.tsx does not pass the new required `onStar` prop to TodoCard, causing a build/runtime break. Also has a race condition in the optimistic star toggle and misplaced test cases._

- 🔴 **regressions** (L155): TodoCard now requires the `onStar` prop but TodoView.tsx does not pass it. This is a compile error — every `<TodoCard>` rendered in TodoView will fail TypeScript checks and crash at runtime. `[fixable]`


### `frontend/src/hooks/useTodoData.ts`
_Critical: TodoView.tsx does not pass the new required `onStar` prop to TodoCard, causing a build/runtime break. Also has a race condition in the optimistic star toggle and misplaced test cases._

- 🟡 **bugs** (L119): Race condition: `star()` calls `setData` (optimistic toggle) then reads `data` from the closure to determine the API action. On rapid double-click, the second invocation still sees the pre-first-toggle `data` (React hasn't re-rendered yet), so it sends the same action twice instead of toggling back. Fix: read the current starred state *inside* the `setData` updater and capture it via a ref or local variable before calling `setData`. `[fixable]`
- 🟡 **style** (L110): `star` has `[data]` as a useCallback dependency, which means it is recreated on every data change. This is unnecessary churn compared to other actions (`ack`, `done`) which have stable references. The `data` dependency exists only because `findInTree` reads from the closure — moving the starred-state lookup into the `setData` updater would eliminate this dependency. `[fixable]`


### `server/__tests__/todos.test.ts`
_Critical: TodoView.tsx does not pass the new required `onStar` prop to TodoCard, causing a build/runtime break. Also has a race condition in the optimistic star toggle and misplaced test cases._

- 🟡 **missing_tests** (L179): The two new star/unstar tests are placed outside the `describe('todo routes', ...)` block (after its closing brace on line 177). They should be inside the describe block for consistent grouping and shared setup context. `[fixable]`


### `server/api-schemas.ts`
_Critical: TodoView.tsx does not pass the new required `onStar` prop to TodoCard, causing a build/runtime break. Also has a race condition in the optimistic star toggle and misplaced test cases._

- 🔵 **unsafe_assumptions**: `starred` is declared as `z.boolean().optional().default(false)` in TodoItemSchema. This means items from the Python script that lack the `starred` field will silently default to `false`. This is fine for backwards compatibility, but the frontend type declares `starred: boolean` (required). If the Zod parse is not applied before sending to the frontend, unstarred items could have `starred: undefined`. Verify the GET /api/todos response is parsed through this schema.

dimakis:
## Centaur Review

### `frontend/src/pages/TodoView.tsx`
_Critical build breakage: TodoView.tsx is missing the new required `onStar` prop on TodoCard. The hook also has a stale-closure race condition and no test coverage for the new `star` function._

- 🔴 **regressions** (L155): TodoView renders <TodoCard> without the new required `onStar` prop. This is a TypeScript compilation error — the build will fail. TodoView also doesn't destructure `star` from `useTodoData()` on line 77. Both need to be updated. `[fixable]`


### `frontend/src/hooks/useTodoData.ts`
_Critical build breakage: TodoView.tsx is missing the new required `onStar` prop on TodoCard. The hook also has a stale-closure race condition and no test coverage for the new `star` function._

- 🟡 **bugs** (L110): The `star` callback has `[data]` as a dependency, causing it to be recreated on every data change. This is unnecessary churn and causes downstream re-renders. Worse, rapid double-taps can race: the first tap triggers `setData` (optimistic toggle) which updates `data`, recreating the callback, but if a second tap fires before React re-renders, it reads stale `data`. Fix: use a ref for the lookup or read inside the `setData` updater function to derive the action, removing the `data` dependency entirely. `[fixable]`


### `frontend/src/hooks/__tests__/useTodoData.test.ts`
_Critical build breakage: TodoView.tsx is missing the new required `onStar` prop on TodoCard. The hook also has a stale-closure race condition and no test coverage for the new `star` function._

- 🟡 **missing_tests**: The `useTodoData` hook test file has no tests for the new `star` function — no test for optimistic toggle, no test for the API call being made with the correct action (star vs unstar), and no test for error handling. `[fixable]`


### `server/__tests__/todos.test.ts`
_Critical build breakage: TodoView.tsx is missing the new required `onStar` prop on TodoCard. The hook also has a stale-closure race condition and no test coverage for the new `star` function._

- 🔵 **style** (L179): The two new `it()` blocks are placed outside the `describe('todo routes', ...)` block that closes at line 177. They work because `app`/`authCookie` are module-level, but this breaks the grouping convention used by the rest of the file. Move them inside the describe block. `[fixable]`
- 🔵 **missing_tests** (L179): Server tests only verify that star/unstar return 500 when the Python script is missing (same pattern as other actions). There's no test verifying that invalid actions are rejected by the schema validation — e.g., that sending `{action: 'invalid'}` returns 400. `[fixable]`


### `frontend/src/components/TodoCard.tsx`
_Critical build breakage: TodoView.tsx is missing the new required `onStar` prop on TodoCard. The hook also has a stale-closure race condition and no test coverage for the new `star` function._

- 🔵 **style** (L140): The star button uses an emoji (⭐) for the starred state and a text character (☆) for unstarred. These render at different sizes across platforms and have no `aria-label`, making the button inaccessible to screen readers. Consider adding `aria-label={item.starred ? 'Unstar' : 'Star'}`. `[fixable]`

dimakis:
## Centaur Review

### `frontend/src/pages/TodoView.tsx`
_Critical bug: TodoView never passes `onStar` to TodoCard — the star button renders but is non-functional (will error at runtime). Also has a subtle race condition in the optimistic update logic and missing happy-path server tests._

- 🔴 **bugs** (L155): TodoCard's required `onStar` prop is never passed. `star` is not destructured from `useTodoData()` on line 77, and `onStar` is missing from the JSX. The star button renders but clicking it will throw a runtime error (calling undefined as a function). TypeScript should catch this, but the PR is broken at runtime if types aren't checked. `[fixable]`


### `frontend/src/hooks/useTodoData.ts`
_Critical bug: TodoView never passes `onStar` to TodoCard — the star button renders but is non-functional (will error at runtime). Also has a subtle race condition in the optimistic update logic and missing happy-path server tests._

- 🟡 **bugs** (L118): Race condition in `star()`: the optimistic `setData` toggle runs first (line 113), then `findInTree` reads `data` from the closure (line 119) to determine the API action. This works by accident — `data` is the pre-setState value due to closure semantics — but is fragile and confusing. If two rapid star clicks occur, the second call captures stale `data` (dependency is `[data]`) and may send the wrong action. Consider reading the current state inside the `setData` updater and using a ref or local variable to capture the pre-toggle value before calling `setData`. `[fixable]`
- 🔵 **missing_tests** (L110): No unit tests for the `star()` function in `useTodoData`, the `toggleStarInTree` helper, or `findInTree`. The hook tests should verify optimistic toggle behavior and API call. `[fixable]`
- 🔵 **style** (L132): `star` has `[data]` as its dependency, causing it to be recreated on every data change. The other action callbacks (`ack`, `snooze`, `done`) use `performAction` which doesn't depend on `data`. Consider reading the current starred state via a ref or from the `setData` updater to remove the `data` dependency. `[fixable]`


### `server/__tests__/todos.test.ts`
_Critical bug: TodoView never passes `onStar` to TodoCard — the star button renders but is non-functional (will error at runtime). Also has a subtle race condition in the optimistic update logic and missing happy-path server tests._

- 🟡 **missing_tests** (L178): The two new server tests only verify that star/unstar return 500 when the Python script is missing — they test the error path, not success. There are no tests for the happy path (script exists and returns a result). Additionally, the tests are outside the `describe('todo routes')` block, breaking test organization. `[fixable]`


### `frontend/src/components/__tests__/TodoCard.test.tsx`
_Critical bug: TodoView never passes `onStar` to TodoCard — the star button renders but is non-functional (will error at runtime). Also has a subtle race condition in the optimistic update logic and missing happy-path server tests._

- 🔵 **style**: The diff removes blank lines and comments throughout the existing test file (e.g., removing `// ◐ character for acknowledged`, removing blank lines between setup/act/assert). These are unrelated cosmetic changes that add noise to the diff.

frontend/src/hooks/useTodoData.ts
```diff
@@ -87,6 +107,31 @@ export function useTodoData(profile?: string): UseTodoDataResult {
...
+  const star = useCallback(
+    async (id: string) => {
+      // Determine current state, then optimistically toggle
+      setData((prev) => {
+        if (!prev) return prev;
+        return { ...prev, items: toggleStarInTree(prev.items, id) };
+      });
+
+      // Read current starred state to decide action (before toggle)
+      const currentItem = data ? findInTree(data.items, id) : undefined;
```
dimakis: 🔴 **bugs**: Race condition: `setData` (line 113) is called before reading `data` (line 119), but `setData` is async — `data` still holds the pre-toggle value *in this render*. However, there's a stale-closure bug: the `star` callback closes over `data` (dependency on line 132), so on rapid double-clicks, the second call captures the stale `data` from before the first optimistic update was applied. The `setData` updater function correctly uses `prev`, but `findInTree(data.items, id)` reads from the closed-over snapshot, not the latest state. This means two rapid star clicks could send `star` twice instead of `star` then `unstar`. Fix: read the current starred state inside the `setData` updater and capture it via a ref or local variable, or use a ref to track the latest data. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useTodoData.ts`
_Star toggle has a stale-closure race condition on rapid clicks that can send the wrong action to the server; otherwise a clean feature addition with adequate component-level tests._

- 🔴 **bugs** (L119): Race condition: `setData` (line 113) is called before reading `data` (line 119), but `setData` is async — `data` still holds the pre-toggle value *in this render*. However, there's a stale-closure bug: the `star` callback closes over `data` (dependency on line 132), so on rapid double-clicks, the second call captures the stale `data` from before the first optimistic update was applied. The `setData` updater function correctly uses `prev`, but `findInTree(data.items, id)` reads from the closed-over snapshot, not the latest state. This means two rapid star clicks could send `star` twice instead of `star` then `unstar`. Fix: read the current starred state inside the `setData` updater and capture it via a ref or local variable, or use a ref to track the latest data. `[fixable]`
- 🟡 **bugs** (L132): The `star` callback has `[data]` as its dependency, which means it is recreated on every data change. This is expensive — every time `data` updates (any todo action, refresh, etc.), a new `star` function is created, causing re-renders in all consumers. The other action callbacks (`ack`, `snooze`, `done`) depend only on `performAction` which is stable. Consider using a ref for `data` to avoid this dependency. `[fixable]`
- 🔵 **missing_tests** (L110): The `star` callback in `useTodoData` has no unit tests. The optimistic toggle logic, the API call with correct action, and error handling are all untested. Other actions (`ack`, `snooze`, `done`) may also lack hook-level tests, but the star logic is more complex (bidirectional toggle vs. one-shot action). `[fixable]`


### `server/__tests__/todos.test.ts`
_Star toggle has a stale-closure race condition on rapid clicks that can send the wrong action to the server; otherwise a clean feature addition with adequate component-level tests._

- 🟡 **missing_tests** (L178): The two new star/unstar tests only verify 500 when the Python script is missing. There are no tests for the success path (what happens when the script exists and returns valid output). The existing ack/done/snooze tests likely have the same gap, but adding star without a success-path test means the feature is only tested for failure. Consider adding a test that mocks the script to return success. `[fixable]`


### `server/api-schemas.ts`
_Star toggle has a stale-closure race condition on rapid clicks that can send the wrong action to the server; otherwise a clean feature addition with adequate component-level tests._

- 🔵 **style**: The `starred` field in `TodoItemSchema` is `z.boolean().optional().default(false)`, but the frontend `TodoItem` interface declares it as `starred: boolean` (required). This mismatch means the schema silently coerces missing values, which is fine, but the optionality should be documented — old Python scripts that don't return `starred` will get `false` injected by Zod rather than failing validation.

frontend/src/hooks/useTodoData.ts
```diff
@@ -87,6 +107,31 @@ export function useTodoData(profile?: string): UseTodoDataResult {
...
+        await fetch(`/api/todos/${id}/action`, {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ action }),
+        });
+      } catch {
+        // Optimistic update stays; next refresh will reconcile
+      }
+    },
+    [data],
```
dimakis: 🟡 **bugs**: The `star` callback has `[data]` as its dependency, which means it is recreated on every data change. This is expensive — every time `data` updates (any todo action, refresh, etc.), a new `star` function is created, causing re-renders in all consumers. The other action callbacks (`ack`, `snooze`, `done`) depend only on `performAction` which is stable. Consider using a ref for `data` to avoid this dependency. `[fixable]`

frontend/src/hooks/useTodoData.ts
```diff
@@ -87,6 +107,31 @@ export function useTodoData(profile?: string): UseTodoDataResult {
   const done = useCallback((id: string) => performAction(id, 'done'), [performAction]);
   const refresh = useCallback(() => setRefreshKey((k) => k + 1), []);
 
+  const star = useCallback(
```
dimakis: 🔵 **missing_tests**: The `star` callback in `useTodoData` has no unit tests. The optimistic toggle logic, the API call with correct action, and error handling are all untested. Other actions (`ack`, `snooze`, `done`) may also lack hook-level tests, but the star logic is more complex (bidirectional toggle vs. one-shot action). `[fixable]`

server/__tests__/todos.test.ts
```diff
@@ -175,3 +175,21 @@ describe('todo routes', () => {
     expect(res.body.ok).toBe(false);
   });
 });
+
```
dimakis, line 179: 🟡 **missing_tests**: The two new star/unstar tests only verify 500 when the Python script is missing. There are no tests for the success path (what happens when the script exists and returns valid output). The existing ack/done/snooze tests likely have the same gap, but adding star without a success-path test means the feature is only tested for failure. Consider adding a test that mocks the script to return success. `[fixable]`

frontend/src/hooks/useTodoData.ts
```diff
@@ -87,6 +109,29 @@ export function useTodoData(profile?: string): UseTodoDataResult {
...
+      return { ...prev, items: toggleStarInTree(prev.items, id) };
+    });
+
+    try {
+      await fetch(`/api/todos/${id}/action`, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ action }),
+      });
+    } catch {
```
dimakis, line 130: 🟡 **bugs**: On fetch failure (non-ok response, e.g. 404/500), the optimistic update is not rolled back. Unlike network errors where the comment says 'next refresh will reconcile', a server rejection means the star was definitively not saved — the UI will show stale state until manual refresh. The `ack`/`done` actions handle failure by simply not removing the item, but `star` has already toggled the state. Consider rolling back on non-ok responses (not just network errors). `[fixable]`

frontend/src/components/TodoCard.tsx
```diff
@@ -131,6 +140,15 @@ export function TodoCard({ item, depth = 0, onAck, onDone, onTap, onAddChild }:
                 {item.completedChildCount ?? 0}/{item.childCount ?? children.length}
               </span>
             )}
+            <button
+              className="todo-card-star"
+              onClick={(e) => {
+                e.stopPropagation();
+                onStar(item.id);
+              }}
```
dimakis, line 148: 🔵 **style**: The star button uses a raw emoji (⭐) for the starred state but a text character (☆) for unstarred. This may render inconsistently across platforms/browsers. Consider using the same symbol family for both states (e.g., ★/☆) or both as emoji (⭐/✩). `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useTodoData.ts`
_Clean feature addition with good test coverage. Main concern is that optimistic star updates are never rolled back on server errors — consider reverting on non-ok responses._

- 🟡 **bugs** (L130): On fetch failure (non-ok response, e.g. 404/500), the optimistic update is not rolled back. Unlike network errors where the comment says 'next refresh will reconcile', a server rejection means the star was definitively not saved — the UI will show stale state until manual refresh. The `ack`/`done` actions handle failure by simply not removing the item, but `star` has already toggled the state. Consider rolling back on non-ok responses (not just network errors). `[fixable]`


### `frontend/src/hooks/__tests__/useTodoData.test.ts`
_Clean feature addition with good test coverage. Main concern is that optimistic star updates are never rolled back on server errors — consider reverting on non-ok responses._

- 🔵 **missing_tests**: No test for server returning a non-ok HTTP response (e.g. 404/500). The network error test covers `fetch` rejection, but not `res.ok === false`. Given that the implementation silently swallows all errors (no rollback), a test documenting this behavior would clarify intent. `[fixable]`
- 🔵 **missing_tests**: The exported helper functions `toggleStarInTree` and `findInTree` have no direct unit tests. They are exercised indirectly through `useTodoData` tests, but edge cases (item not found, deeply nested items) are untested. `[fixable]`


### `frontend/src/components/TodoCard.tsx`
_Clean feature addition with good test coverage. Main concern is that optimistic star updates are never rolled back on server errors — consider reverting on non-ok responses._

- 🔵 **style** (L148): The star button uses a raw emoji (⭐) for the starred state but a text character (☆) for unstarred. This may render inconsistently across platforms/browsers. Consider using the same symbol family for both states (e.g., ★/☆) or both as emoji (⭐/✩). `[fixable]`


### `docs/design/global-task-board.md`
_Clean feature addition with good test coverage. Main concern is that optimistic star updates are never rolled back on server errors — consider reverting on non-ok responses._

- 🔵 **style**: This 829-line design document is unrelated to the starred feature. Including it in a feature PR adds noise to review. Consider landing it in a separate `docs/` PR.

PR #181, opened 2026-04-12
feat: token visibility bar in chat input toolbar

dimakis:
## Summary

- **Server**: Emit `token_update` WS events from query-loop — agent context (from `message_start` usage) on each turn, session totals + cost on result
- **Frontend**: `TokenBar` component in the command strip toolbar showing 🧠 agent context gauge (color-coded green/yellow/red/flashing) + Σ session total. Tap to expand detail panel with cost, turns, agent index
- **Frontend**: `useTokenState` hook + `TokenUpdateMsg` WS type, wired into both mobile and desktop chat views

Depends on: #178 (token-capture), #179 (goal-wiring) — both merged.

Design doc: `docs/design/token-visibility.md` (Phase 1: Session Bar)

## Test plan

- [x] 16 new tests (4 server, 4 hook, 8 component)
- [x] Full suite passes (1036 tests, 91 files)
- [x] TypeScript clean (`tsc --noEmit`)
- [ ] Manual: verify token_update events appear in WS stream during a session
- [ ] Manual: verify bar renders with correct colors at different context levels
- [ ] Manual: verify detail panel expands/collapses on tap

### DIFF
--- a/frontend/src/components/TokenBar.tsx
+++ b/frontend/src/components/TokenBar.tsx
@@ -1,0 +1,91 @@
+import { useState } from 'react';
+import type { TokenState } from '../hooks/useTokenState';
+
+const CONTEXT_CEILING = 200_000;
+
+function formatTokens(n: number): string {
+  if (n >= 1_000_000) return `${(n / 1_000_000).toFixed(1)}M`;
+  if (n >= 1_000) return `${Math.round(n / 1_000)}k`;
+  return String(n);
+}
+
+function getContextColor(tokens: number): string {
+  if (tokens >= 190_000) return 'flashing';
+  if (tokens >= 160_000) return 'red';
+  if (tokens >= 100_000) return 'yellow';
+  return 'green';
+}
+
+interface Props {
+  tokenState: TokenState;
+}
+
+export function TokenBar({ tokenState }: Props) {
+  const [expanded, setExpanded] = useState(false);
+
+  // Don't render until we have data
+  if (tokenState.turnIndex === 0) return null;
+
+  const color = getContextColor(tokenState.agentContext);
+
+  return (
+    <>
+      <button
+        className={`token-bar token-bar--${color}`}
+        onClick={() => setExpanded((v) => !v)}
+        aria-label="Token usage"
+        title="Token usage — tap for details"
+      >
+        <span className="token-bar-agent">
+          <svg
+            className="token-bar-icon"
+            viewBox="0 0 24 24"
+            width="12"
+            height="12"
+            fill="currentColor"
+          >
+            <path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm0 18c-4.41 0-8-3.59-8-8s3.59-8 8-8 8 3.59 8 8-3.59 8-8 8zm-1-14h2v6h-2zm0 8h2v2h-2z" />
+            <path d="M15.5 7.5c1.1 0 2 .9 2 2s-.9 2-2 2-2-.9-2-2 .9-2 2-2zM8.5 7.5c1.1 0 2 .9 2 2s-.9 2-2 2-2-.9-2-2 .9-2 2-2zM12 17.5c2.33 0 4.31-1.46 5.11-3.5H6.89c.8 2.04 2.78 3.5 5.11 3.5z" />
+          </svg>
+          {formatTokens(tokenState.agentContext)}/{formatTokens(CONTEXT_CEILING)}
+        </span>
+        {tokenState.sessionTotal > 0 && (
+          <span className="token-bar-session">
+            <span className="token-bar-sigma">Σ</span>
+            {formatTokens(tokenState.sessionTotal)}
+          </span>
+        )}
+      </button>
+      {expanded && (
+        <div className="token-bar-detail">
+          <div className="token-bar-detail-row">
+            <span>Agent context</span>
+            <span>
+              {tokenState.agentContext.toLocaleString()} / {CONTEXT_CEILING.toLocaleString()}
+            </span>
+          </div>
+          <div className="token-bar-detail-row">
+            <span>Session tokens</span>
+            <span>{tokenState.sessionTotal.toLocaleString()}</span>
+          </div>
+          {tokenState.costUsd > 0 && (
+            <div className="token-bar-detail-row">
+              <span>Cost</span>
+              <span>~${tokenState.costUsd.toFixed(2)}</span>
+            </div>
+          )}
+          {tokenState.numTurns > 0 && (
+            <div className="token-bar-detail-row">
+              <span>Turns</span>
+              <span>{tokenState.numTurns} turns</span>
+            </div>
+          )}
+          <div className="token-bar-detail-row">
+            <span>Agent #</span>
+            <span>{tokenState.turnIndex}</span>
+          </div>
+        </div>
+      )}
+    </>
+  );
+}
--- a/frontend/src/components/TokenBar.tsx
+++ b/frontend/src/components/TokenBar.tsx
@@ -1,18 +1,16 @@
 import { useState } from 'react';
 import type { TokenState } from '../hooks/useTokenState';
 
-const CONTEXT_CEILING = 200_000;
-
 function formatTokens(n: number): string {
   if (n >= 1_000_000) return `${(n / 1_000_000).toFixed(1)}M`;
   if (n >= 1_000) return `${Math.round(n / 1_000)}k`;
   return String(n);
 }
 
-function getContextColor(tokens: number): string {
-  if (tokens >= 190_000) return 'flashing';
-  if (tokens >= 160_000) return 'red';
-  if (tokens >= 100_000) return 'yellow';
+function getContextColor(ratio: number): string {
+  if (ratio >= 0.95) return 'flashing';
+  if (ratio >= 0.8) return 'red';
+  if (ratio >= 0.5) return 'yellow';
   return 'green';
 }
 
@@ -26,7 +24,9 @@ export function TokenBar({ tokenState }: Props) {
   // Don't render until we have data
   if (tokenState.turnIndex === 0) return null;
 
-  const color = getContextColor(tokenState.agentContext);
+  const ceiling = tokenState.contextCeiling;
+  const ratio = ceiling > 0 ? tokenState.agentContext / ceiling : 0;
+  const color = getContextColor(ratio);
 
   return (
     <>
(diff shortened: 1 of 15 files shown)


frontend/src/pages/ChatView.tsx
```diff
@@ -104,7 +106,15 @@ export function ChatView() {
     return () => controller.abort();
   }, [sessionId, dispatch, forceScrollToBottom]);
 
-  const { connected } = useChatConnection(poolKey, handleWsMessage);
+  const composedWsHandler = useCallback(
```
dimakis: 🔵 **style**: The `composedWsHandler` pattern is duplicated identically in `ChatView` and `DesktopChatView`. Consider extracting a shared hook (e.g. `useComposedWsHandler`) or composing handlers in `useChatConnection` to avoid the duplication. `[fixable]`

frontend/src/components/TokenBar.tsx
```diff
@@ -0,0 +1,91 @@
+import { useState } from 'react';
+import type { TokenState } from '../hooks/useTokenState';
+
+const CONTEXT_CEILING = 200_000;
```
dimakis: 🔵 **unsafe_assumptions**: `CONTEXT_CEILING = 200_000` is hardcoded. Different models have different context windows (e.g. 128k, 200k). Consider deriving this from server-provided data (the SDK result or a config endpoint) rather than assuming 200k for all sessions. `[fixable]`

frontend/src/components/TokenBar.tsx
```diff
@@ -0,0 +1,91 @@
...
+        className={`token-bar token-bar--${color}`}
+        onClick={() => setExpanded((v) => !v)}
+        aria-label="Token usage"
+        title="Token usage — tap for details"
+      >
+        <span className="token-bar-agent">
+          <svg
+            className="token-bar-icon"
+            viewBox="0 0 24 24"
+            width="12"
```
dimakis, line 43: 🔵 **style**: The SVG icon (an info-circle overlaid with a smiley face) doesn't resemble a brain, despite the test describing it as "renders agent context with brain icon". Consider using an actual brain SVG or updating the test description to match the actual icon. `[fixable]`

frontend/src/hooks/useTokenState.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export function useTokenState() {
+  const [tokenState, tokenDispatch] = useReducer(tokenStateReducer, INITIAL_TOKEN_STATE);
+
+  const handleTokenMessage = useCallback((msg: WsMsg) => {
+    if (msg.type === 'token_update') {
+      tokenDispatch({
+        type: 'TOKEN_UPDATE',
+        agentContext: msg.agentContext as number,
```
dimakis: 🔵 **style**: The type assertions (`as number`, `as number | undefined`) in `handleTokenMessage` are unnecessary. After the `msg.type === 'token_update'` guard, TypeScript narrows `msg` to `TokenUpdateMsg`, which already has the correct field types. The casts add noise without adding safety. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid feature addition with good test coverage. The main issue is missing token state reset on session navigation, which will show stale data when switching between sessions._

- 🟡 **bugs**: Token state is never reset when `sessionId` changes. In React Router, navigating between `/chat/abc` and `/chat/def` re-renders but does not remount the component, so `useTokenState()` retains stale data from the previous session until the new session emits its first `token_update`. Add a `useEffect` that dispatches `{ type: 'RESET' }` when `sessionId` changes. Same issue exists in `DesktopChatView.tsx`. `[fixable]`
- 🔵 **style** (L109): The `composedWsHandler` pattern is duplicated identically in `ChatView` and `DesktopChatView`. Consider extracting a shared hook (e.g. `useComposedWsHandler`) or composing handlers in `useChatConnection` to avoid the duplication. `[fixable]`


### `frontend/src/components/TokenBar.tsx`
_Solid feature addition with good test coverage. The main issue is missing token state reset on session navigation, which will show stale data when switching between sessions._

- 🔵 **unsafe_assumptions** (L4): `CONTEXT_CEILING = 200_000` is hardcoded. Different models have different context windows (e.g. 128k, 200k). Consider deriving this from server-provided data (the SDK result or a config endpoint) rather than assuming 200k for all sessions. `[fixable]`
- 🔵 **style** (L43): The SVG icon (an info-circle overlaid with a smiley face) doesn't resemble a brain, despite the test describing it as "renders agent context with brain icon". Consider using an actual brain SVG or updating the test description to match the actual icon. `[fixable]`


### `frontend/src/hooks/useTokenState.ts`
_Solid feature addition with good test coverage. The main issue is missing token state reset on session navigation, which will show stale data when switching between sessions._

- 🔵 **style** (L55): The type assertions (`as number`, `as number | undefined`) in `handleTokenMessage` are unnecessary. After the `msg.type === 'token_update'` guard, TypeScript narrows `msg` to `TokenUpdateMsg`, which already has the correct field types. The casts add noise without adding safety. `[fixable]`

server/query-loop.ts
```diff
@@ -140,6 +140,12 @@ export async function runQueryLoop(
   let goalCreationPromise: Promise<string | null> | undefined;
   let goalTitle: string | undefined;
 
+  // Token tracking state for live token_update events
+  let agentContextTokens = 0; // input_tokens from latest message_start (context window size)
+  let turnIndex = 0; // increments on each message_start
+  // All current Claude models share 200k context; sent to frontend so it doesn't hardcode.
+  const contextCeiling = 200_000;
```
dimakis: 🟡 **unsafe_assumptions**: contextCeiling is hardcoded to 200,000. If a different model with a different context window is used (e.g., Haiku at 200k is fine, but future models may differ), this won't adapt. The comment says 'sent to frontend so it doesn't hardcode' but the server still hardcodes it. Consider deriving from the model config or SDK response if available. `[fixable]`

server/query-loop.ts
```diff
@@ -310,6 +316,17 @@ export async function runQueryLoop(
           store.recordUsage(resolvedSessionId, usageData);
         }
 
+        // Emit final token_update with session totals
+        emit(currentWs, {
+          type: 'token_update',
+          agentContext: agentContextTokens,
+          contextCeiling,
+          sessionTotal: usageData.inputTokens + usageData.outputTokens,
```
dimakis: 🟡 **bugs**: sessionTotal is computed as inputTokens + outputTokens from the `result` event, but this is the cumulative total for the entire query() call. If the session spans multiple query() calls (user sends follow-up messages), each result only reports that call's totals — not the true session lifetime total. The frontend will see sessionTotal reset/drop on each new query rather than accumulate. `[fixable]`

frontend/src/hooks/useChatConnection.ts
```diff
@@ -15,7 +15,7 @@ export function useChatConnection(
       } else if (msg.type === '_close') {
         setConnected(false);
       }
-      onMessage(msg);
+      for (const h of handlers) h(msg);
```
dimakis: 🟡 **regressions**: The handlers are captured in the useEffect closure but not included in the dependency array (existing eslint-disable). With the old single-callback signature this was already the pattern, but now multiple handlers are captured. If any handler reference changes after mount (without poolKey changing), the stale closure will call the old handler. Currently safe because both handlers are stable useCallback refs, but fragile for future callers.

frontend/src/hooks/useTokenState.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export function tokenStateReducer(state: TokenState, action: TokenAction): TokenState {
+  switch (action.type) {
+    case 'TOKEN_UPDATE':
+      return {
+        agentContext: action.agentContext,
+        contextCeiling: action.contextCeiling ?? state.contextCeiling,
+        sessionTotal: action.sessionTotal ?? state.sessionTotal,
+        costUsd: action.costUsd ?? state.costUsd,
```
dimakis: 🔵 **bugs**: Using nullish coalescing (`??`) for `costUsd` and `numTurns` means an explicit `0` from the server won't reset these fields — it will preserve the previous (possibly higher) value. This is fine if the server never sends 0 to mean 'reset', but could be surprising if a new query starts with 0 cost. `[fixable]`

frontend/src/components/TokenBar.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [expanded, setExpanded] = useState(false);
+
+  // Don't render until we have data
+  if (tokenState.turnIndex === 0) return null;
+
+  const ceiling = tokenState.contextCeiling;
+  const ratio = ceiling > 0 ? tokenState.agentContext / ceiling : 0;
+  const color = getContextColor(ratio);
+
+  return (
```
dimakis, line 31: 🔵 **style**: The component returns a Fragment (<>) with a <button> and a sibling <div> for the detail popover. Since the detail panel is positioned absolutely relative to .chat-input-toolbar (via the new `position: relative` on it), this works — but using a Fragment means the button and popover are siblings at the same DOM level. A wrapper <div> would make the relationship clearer and avoid relying on the parent's position context. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid feature addition with good test coverage on the core paths. Main concerns: sessionTotal may not accumulate across multiple query() calls in a session, and contextCeiling is hardcoded._

- 🟡 **unsafe_assumptions** (L147): contextCeiling is hardcoded to 200,000. If a different model with a different context window is used (e.g., Haiku at 200k is fine, but future models may differ), this won't adapt. The comment says 'sent to frontend so it doesn't hardcode' but the server still hardcodes it. Consider deriving from the model config or SDK response if available. `[fixable]`
- 🟡 **bugs** (L324): sessionTotal is computed as inputTokens + outputTokens from the `result` event, but this is the cumulative total for the entire query() call. If the session spans multiple query() calls (user sends follow-up messages), each result only reports that call's totals — not the true session lifetime total. The frontend will see sessionTotal reset/drop on each new query rather than accumulate. `[fixable]`


### `frontend/src/hooks/useChatConnection.ts`
_Solid feature addition with good test coverage on the core paths. Main concerns: sessionTotal may not accumulate across multiple query() calls in a session, and contextCeiling is hardcoded._

- 🟡 **regressions** (L18): The handlers are captured in the useEffect closure but not included in the dependency array (existing eslint-disable). With the old single-callback signature this was already the pattern, but now multiple handlers are captured. If any handler reference changes after mount (without poolKey changing), the stale closure will call the old handler. Currently safe because both handlers are stable useCallback refs, but fragile for future callers.


### `frontend/src/hooks/useTokenState.ts`
_Solid feature addition with good test coverage on the core paths. Main concerns: sessionTotal may not accumulate across multiple query() calls in a session, and contextCeiling is hardcoded._

- 🔵 **bugs** (L43): Using nullish coalescing (`??`) for `costUsd` and `numTurns` means an explicit `0` from the server won't reset these fields — it will preserve the previous (possibly higher) value. This is fine if the server never sends 0 to mean 'reset', but could be surprising if a new query starts with 0 cost. `[fixable]`


### `frontend/src/components/TokenBar.tsx`
_Solid feature addition with good test coverage on the core paths. Main concerns: sessionTotal may not accumulate across multiple query() calls in a session, and contextCeiling is hardcoded._

- 🔵 **style** (L31): The component returns a Fragment (<>) with a <button> and a sibling <div> for the detail popover. Since the detail panel is positioned absolutely relative to .chat-input-toolbar (via the new `position: relative` on it), this works — but using a Fragment means the button and popover are siblings at the same DOM level. A wrapper <div> would make the relationship clearer and avoid relying on the parent's position context. `[fixable]`


### `frontend/src/hooks/__tests__/useTokenState.test.ts`
_Solid feature addition with good test coverage on the core paths. Main concerns: sessionTotal may not accumulate across multiple query() calls in a session, and contextCeiling is hardcoded._

- 🔵 **missing_tests**: The test file only tests the reducer, not the useTokenState hook itself (session reset on sessionId change, handleTokenMessage dispatching). The hook's auto-reset behavior and message filtering are untested. `[fixable]`

PR #182, opened 2026-04-12
feat: task board phase 1 — persistent CRUD with tree UI

dimakis:
## Summary
- **TaskStore** — SQLite persistence layer at `.mitzo/tasks.db` with CRUD, tree assembly, cascade delete, and depth tracking
- **REST API** — `GET/POST/PATCH/DELETE /api/tasks` with Zod validation and broadcast-on-mutate
- **WebSocket** — `task_state` hydration on connect, `task_updated`/`task_deleted` live broadcasts
- **Frontend** — `useTaskBoard` hook (REST + WS), `TaskNode` tree component, `TaskCreateForm`, `TaskBoard` page at `/tasks`

## Test plan
- [x] 33 new tests across 6 test files (1088 total passing)
- [x] Lint clean, type-checks, production build passes
- [ ] Manual: create root task, create child task, delete task
- [ ] Manual: open two browser tabs, verify WS sync
- [ ] Manual: refresh page, verify SQLite persistence

Design doc: `docs/design/task-board-phase1-plan.md`

### DIFF
--- a/frontend/src/components/TaskNode.tsx
+++ b/frontend/src/components/TaskNode.tsx
@@ -34,7 +34,7 @@ export function TaskNode({ task, depth, onStatusChange, onDelete, onAddChild }: 
   const hasChildren = task.children.length > 0;
 
   return (
-    <div className="task-node" style={{ paddingLeft: `${depth * 16}px` }}>
+    <div className="task-node">
       <div className="task-node-row">
         {hasChildren && (
           <button
--- a/frontend/src/components/TaskNode.tsx
+++ b/frontend/src/components/TaskNode.tsx
@@ -1,0 +1,93 @@
+import { useState } from 'react';
+import type { Task, TaskStatus } from '../types/task';
+
+const STATUS_ICONS: Record<TaskStatus, string> = {
+  pending: '\u25CB', // ○
+  active: '\u25C9', // ◉
+  done: '\u2713', // ✓
+  pending_review: '\u25D4', // ◔
+  blocked: '\u2298', // ⊘
+  skipped: '\u2014', // —
+  failed: '\u2717', // ✗
+};
+
+const NEXT_STATUS: Record<TaskStatus, TaskStatus> = {
+  pending: 'active',
+  active: 'done',
+  done: 'pending',
+  pending_review: 'done',
+  blocked: 'active',
+  skipped: 'pending',
+  failed: 'pending',
+};
+
+interface TaskNodeProps {
+  task: Task;
+  depth: number;
+  onStatusChange: (id: string, status: TaskStatus) => void;
+  onDelete: (id: string) => void;
+  onAddChild: (parentId: string) => void;
+}
+
+export function TaskNode({ task, depth, onStatusChange, onDelete, onAddChild }: TaskNodeProps) {
+  const [expanded, setExpanded] = useState(true);
+  const hasChildren = task.children.length > 0;
+
+  return (
+    <div className="task-node" style={{ paddingLeft: `${depth * 16}px` }}>
+      <div className="task-node-row">
+        {hasChildren && (
+          <button
+            className="task-node-chevron"
+            onClick={() => setExpanded(!expanded)}
+            aria-label={expanded ? 'Collapse' : 'Expand'}
+          >
+            {expanded ? '\u25BC' : '\u25B6'}
+          </button>
+        )}
+        <button
+          className={`task-node-status task-node-status--${task.status}`}
+          onClick={() => onStatusChange(task.id, NEXT_STATUS[task.status])}
+          aria-label={`Status: ${task.status}`}
+        >
+          {STATUS_ICONS[task.status]}
+        </button>
+        <span
+          className={`task-node-title${task.status === 'done' ? ' task-node-title--done' : ''}`}
+        >
+          {task.title}
+        </span>
+        <div className="task-node-actions">
+          <button
+            className="task-node-action"
+            onClick={() => onAddChild(task.id)}
+            title="Add sub-task"
+          >
+            +
+          </button>
+          <button
+            className="task-node-action task-node-action--danger"
+            onClick={() => onDelete(task.id)}
+            title="Delete task"
+          >
+            &times;
+          </button>
+        </div>
+      </div>
+      {hasChildren && expanded && (
+        <div className="task-node-children">
+          {task.children.map((child) => (
+            <TaskNode
+              key={child.id}
+              task={child}
+              depth={depth + 1}
+              onStatusChange={onStatusChange}
+              onDelete={onDelete}
+              onAddChild={onAddChild}
+            />
+          ))}
+        </div>
+      )}
+    </div>
+  );
+}
--- a/frontend/src/components/__tests__/TaskNode.test.tsx
+++ b/frontend/src/components/__tests__/TaskNode.test.tsx
@@ -94,7 +94,7 @@ describe('TaskNode', () => {
     expect(screen.queryByText('Child task')).toBeNull();
   });
 
-  it('applies depth indentation', () => {
+  it('does not apply inline depth indentation (handled by CSS nesting)', () => {
     const { container } = render(
       <TaskNode
         task={makeTask()}
@@ -105,7 +105,7 @@ describe('TaskNode', () => {
       />,
     );
     const node = container.firstElementChild as HTMLElement;
-    expect(node.style.paddingLeft).toBe('32px');
+    expect(node.style.paddingLeft).toBe('');
   });
 
   it('renders children recursively', () => {
--- a/frontend/src/components/__tests__/TaskNode.test.tsx
+++ b/frontend/src/components/__tests__/TaskNode.test.tsx
@@ -1,12 +1,12 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, afterEach } from 'vitest';
 import { render, screen, fireEvent, cleanup } from '@testing-library/react';
+import type { Task } from '../../types/task';
 import { TaskNode } from '../TaskNode';
 
 afterEach(() => {
   cleanup();
 });
-import type { Task } from '../../types/task';
 
 function makeTask(overrides: Partial<Task> = {}): Task {
   return {
@@ -51,6 +51,7 @@ describe('TaskNode', () => {
     ['pending', '\u25CB'],
     ['active', '\u25C9'],
     ['done', '\u2713'],
+    ['pending_review', '\u25D4'],
     ['blocked', '\u2298'],
     ['skipped', '\u2014'],
     ['failed', '\u2717'],
--- a/frontend/src/components/__tests__/TaskNode.test.tsx
+++ b/frontend/src/components/__tests__/TaskNode.test.tsx
@@ -1,0 +1,169 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi } from 'vitest';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { TaskNode } from '../TaskNode';
+import type { Task } from '../../types/task';
+
+function makeTask(overrides: Partial<Task> = {}): Task {
+  return {
+    id: 'task-1',
+    parentId: null,
+    title: 'Test task',
+    description: null,
+    status: 'pending',
+    sessionId: null,
+    sessionPolicy: 'auto',
+    priority: 0,
+    depth: 0,
+    annotations: [],
+    summary: null,
+    requiresApproval: false,
+    tokenUsage: 0,
+    claimedBy: null,
+    claimedAt: null,
+    createdAt: Date.now(),
+    updatedAt: Date.now(),
+    completedAt: null,
+    children: [],
+    ...overrides,
+  };
+}
+
+describe('TaskNode', () => {
+  it('renders title', () => {
+    render(
+      <TaskNode
+        task={makeTask({ title: 'My Task' })}
+        depth={0}
+        onStatusChange={vi.fn()}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+    expect(screen.getByText('My Task')).toBeTruthy();
+  });
+
+  it.each([
+    ['pending', '\u25CB'],
+    ['active', '\u25C9'],
+    ['done', '\u2713'],
+    ['blocked', '\u2298'],
+    ['skipped', '\u2014'],
+    ['failed', '\u2717'],
+  ] as const)('renders correct icon for status %s', (status, expectedIcon) => {
+    const { container } = render(
+      <TaskNode
+        task={makeTask({ status })}
+        depth={0}
+        onStatusChange={vi.fn()}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+    const statusBtn = container.querySelector('.task-node-status');
+    expect(statusBtn?.textContent).toBe(expectedIcon);
+  });
+
+  it('toggles expand/collapse for tasks with children', () => {
+    const child = makeTask({ id: 'child-1', parentId: 'task-1', title: 'Child task', depth: 1 });
+    const task = makeTask({ children: [child] });
+
+    render(
+      <TaskNode
+        task={task}
+        depth={0}
+        onStatusChange={vi.fn()}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+
+    // Children should be visible by default
+    expect(screen.getByText('Child task')).toBeTruthy();
+
+    // Click collapse
+    const chevron = screen.getByText('\u25BC');
+    fireEvent.click(chevron);
+
+    // Children should be hidden
+    expect(screen.queryByText('Child task')).toBeNull();
+  });
+
+  it('applies depth indentation', () => {
+    const { container } = render(
+      <TaskNode
+        task={makeTask()}
+        depth={2}
+        onStatusChange={vi.fn()}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+    const node = container.firstElementChild as HTMLElement;
+    expect(node.style.paddingLeft).toBe('32px');
+  });
+
+  it('renders children recursively', () => {
+    const grandchild = makeTask({
+      id: 'gc-1',
+      parentId: 'child-1',
+      title: 'Grandchild',
+      depth: 2,
+    });
+    const child = makeTask({
+      id: 'child-1',
+      parentId: 'task-1',
+      title: 'Child',
+      depth: 1,
+      children: [grandchild],
+    });
+    const task = makeTask({ children: [child] });
+
+    render(
+      <TaskNode
+        task={task}
+        depth={0}
+        onStatusChange={vi.fn()}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+
+    expect(screen.getByText('Child')).toBeTruthy();
+    expect(screen.getByText('Grandchild')).toBeTruthy();
+  });
+
+  it('calls onStatusChange when status icon clicked', () => {
+    const onStatusChange = vi.fn();
+    const { container } = render(
+      <TaskNode
+        task={makeTask({ id: 'sc-1', status: 'pending' })}
+        depth={0}
+        onStatusChange={onStatusChange}
+        onDelete={vi.fn()}
+        onAddChild={vi.fn()}
+      />,
+    );
+
+    const statusBtn = container.querySelector('.task-node-status')!;
+    fireEvent.click(statusBtn);
+    expect(onStatusChange).toHaveBeenCalledWith('sc-1', expect.any(String));
+  });
+
+  it('calls onDelete when delete button clicked', () => {
+    const onDelete = vi.fn();
+    const { container } = render(
+      <TaskNode
+        task={makeTask({ id: 'del-1' })}
+        depth={0}
+        onStatusChange={vi.fn()}
+        onDelete={onDelete}
+        onAddChild={vi.fn()}
+      />,
+    );
+
+    const deleteBtn = container.querySelector('.task-node-action--danger')!;
+    fireEvent.click(deleteBtn);
+    expect(onDelete).toHaveBeenCalledWith('del-1');
+  });
+});
--- a/frontend/src/components/__tests__/TaskNode.test.tsx
+++ b/frontend/src/components/__tests__/TaskNode.test.tsx
@@ -1,7 +1,11 @@
 // @vitest-environment jsdom
-import { describe, it, expect, vi } from 'vitest';
-import { render, screen, fireEvent } from '@testing-library/react';
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup } from '@testing-library/react';
 import { TaskNode } from '../TaskNode';
+
+afterEach(() => {
+  cleanup();
+});
 import type { Task } from '../../types/task';
 
 function makeTask(overrides: Partial<Task> = {}): Task {
--- a/frontend/src/hooks/useTaskBoard.ts
+++ b/frontend/src/hooks/useTaskBoard.ts
@@ -1,0 +1,152 @@
+import { useState, useEffect, useCallback } from 'react';
+import type { Task } from '../types/task';
+import { wsSubscribe } from '../lib/ws-pool';
+import type { WsMsg } from '../lib/ws-pool';
+
+export interface TaskCreateInput {
+  title: string;
+  parentId?: string;
+  description?: string;
+  priority?: number;
+  sessionPolicy?: string;
+  annotations?: string[];
+}
+
+export interface TaskUpdateInput {
+  title?: string;
+  description?: string;
+  status?: string;
+  priority?: number;
+  sessionPolicy?: string;
+  annotations?: string[];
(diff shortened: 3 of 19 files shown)


frontend/src/hooks/useTaskBoard.ts
```diff
@@ -0,0 +1,152 @@
...
+  // If the task exists at root level, replace it
+  let found = false;
+  const updated = tasks.map((t) => {
+    if (t.id === task.id) {
+      found = true;
+      return { ...task, children: t.children };
+    }
+    // Check children recursively
+    const updatedChildren = upsertInTree(t.children, task);
+    if (updatedChildren !== t.children) {
```
dimakis: 🟡 **bugs**: Reference equality check `updatedChildren !== t.children` is unreliable for detecting changes from `upsertInTree`. When the recursive call returns a new array (task found deep in tree), `found` is correctly set. But when the task is not found in a subtree, `upsertInTree` returns the same array only if no mapping occurred — since `.map()` always returns a new array, this check will always be `true` for non-empty children arrays, causing `found` to be set even when the task wasn't actually found there. However, since `.map()` with identity returns creates a new array, this means `found` gets set spuriously for any node with children, and the first such node's subtree gets the update even if the task isn't in it. In practice, `.map()` always returns a new reference, so `updatedChildren !== t.children` is always true when `t.children.length > 0`. `[fixable]`

frontend/src/hooks/useTaskBoard.ts
```diff
@@ -0,0 +1,152 @@
...
+      });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [refreshKey]);
+
+  // Subscribe to WS events
+  useEffect(() => {
+    const unsub = wsSubscribe('global', (msg: WsMsg) => {
```
dimakis, line 122: 🟡 **unsafe_assumptions**: WS message types are cast via `as unknown as { tasks: Task[] }` instead of using the typed `ServerMessage` union from `ws-messages.ts`. The new `TaskStateMsg`, `TaskUpdatedMsg`, and `TaskDeletedMsg` types were added to `ws-messages.ts` but aren't used here — the hook casts through `unknown`, bypassing type safety. Since `WsMsg` already includes `ServerMessage` (which now includes task message types), narrowing via `msg.type` should give the correct type automatically. `[fixable]`

frontend/src/hooks/useTaskBoard.ts
```diff
@@ -0,0 +1,152 @@
...
+        setTasks((prev) => removeFromTree(prev, m.taskId));
+      }
+    });
+    return unsub;
+  }, []);
+
+  const createTask = useCallback(async (input: TaskCreateInput) => {
+    await fetch('/api/tasks', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
```
dimakis, line 137: 🟡 **bugs**: The `createTask`, `updateTask`, and `deleteTask` callbacks don't handle fetch errors. A non-ok response or network failure will throw an unhandled promise rejection. The initial fetch in the effect has `.catch()` handling, but the mutation functions don't check `r.ok` or catch errors. `[fixable]`

server/app.ts
```diff
@@ -290,6 +309,62 @@ app.post('/api/repos/open', (req, res) => {
...
+app.post('/api/tasks', (req, res) => {
+  const body = TaskCreateBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  try {
+    const task = taskStore.create(body.data);
+    res.status(201).json({ task });
+    onTaskBroadcast?.({ type: 'task_updated', task });
```
dimakis: 🟡 **bugs**: The task broadcast after POST create sends a flat `task` object (from `taskStore.create()` which returns a task with `children: []`), but `GET /api/tasks` returns `taskStore.getTree()` which assembles the full tree. If a child task is created, the WS `task_updated` event sends the child without its parent context, and the frontend's `upsertInTree` must find the parent to insert it — but the parent in the existing tree won't have the new child. This works for root tasks but may cause ordering inconsistencies for child tasks since the WS event and the next `getTree()` call could disagree on tree structure. `[fixable]`

dimakis (commented):
## Centaur Review

Found **9** issue(s) (4 warning).


### `frontend/src/hooks/useTaskBoard.ts`
_Solid phase-1 implementation with good test coverage and clean architecture, but the frontend `upsertInTree` has a reference-equality bug that will cause incorrect tree updates, and mutation functions lack error handling._

- 🟡 **bugs** (L54): Reference equality check `updatedChildren !== t.children` is unreliable for detecting changes from `upsertInTree`. When the recursive call returns a new array (task found deep in tree), `found` is correctly set. But when the task is not found in a subtree, `upsertInTree` returns the same array only if no mapping occurred — since `.map()` always returns a new array, this check will always be `true` for non-empty children arrays, causing `found` to be set even when the task wasn't actually found there. However, since `.map()` with identity returns creates a new array, this means `found` gets set spuriously for any node with children, and the first such node's subtree gets the update even if the task isn't in it. In practice, `.map()` always returns a new reference, so `updatedChildren !== t.children` is always true when `t.children.length > 0`. `[fixable]`
- 🟡 **unsafe_assumptions** (L112): WS message types are cast via `as unknown as { tasks: Task[] }` instead of using the typed `ServerMessage` union from `ws-messages.ts`. The new `TaskStateMsg`, `TaskUpdatedMsg`, and `TaskDeletedMsg` types were added to `ws-messages.ts` but aren't used here — the hook casts through `unknown`, bypassing type safety. Since `WsMsg` already includes `ServerMessage` (which now includes task message types), narrowing via `msg.type` should give the correct type automatically. `[fixable]`
- 🟡 **bugs** (L130): The `createTask`, `updateTask`, and `deleteTask` callbacks don't handle fetch errors. A non-ok response or network failure will throw an unhandled promise rejection. The initial fetch in the effect has `.catch()` handling, but the mutation functions don't check `r.ok` or catch errors. `[fixable]`


### `server/app.ts`
_Solid phase-1 implementation with good test coverage and clean architecture, but the frontend `upsertInTree` has a reference-equality bug that will cause incorrect tree updates, and mutation functions lack error handling._

- 🟡 **bugs** (L327): The task broadcast after POST create sends a flat `task` object (from `taskStore.create()` which returns a task with `children: []`), but `GET /api/tasks` returns `taskStore.getTree()` which assembles the full tree. If a child task is created, the WS `task_updated` event sends the child without its parent context, and the frontend's `upsertInTree` must find the parent to insert it — but the parent in the existing tree won't have the new child. This works for root tasks but may cause ordering inconsistencies for child tasks since the WS event and the next `getTree()` call could disagree on tree structure. `[fixable]`
- 🔵 **unsafe_assumptions** (L164): The task database is created at `${REPO_PATH}/.mitzo/tasks.db` during module initialization. If `REPO_PATH` points to a read-only filesystem or the `.mitzo` directory can't be created, the server will crash on import. The `mkdirSync` has a catch block but the `TaskStore` constructor (which calls `db.exec(SCHEMA)`) does not. `[fixable]`


### `frontend/src/hooks/__tests__/useTaskBoard.test.ts`
_Solid phase-1 implementation with good test coverage and clean architecture, but the frontend `upsertInTree` has a reference-equality bug that will cause incorrect tree updates, and mutation functions lack error handling._

- 🔵 **missing_tests**: No test covers error handling for `createTask`, `updateTask`, or `deleteTask` when the fetch fails (e.g., network error or non-ok response). The hook silently swallows errors from these operations since the `useCallback` functions don't handle rejections — callers get an unhandled promise rejection. `[fixable]`


### `server/task-store.ts`
_Solid phase-1 implementation with good test coverage and clean architecture, but the frontend `upsertInTree` has a reference-equality bug that will cause incorrect tree updates, and mutation functions lack error handling._

- 🔵 **style** (L290): `getSubtree` fetches all tasks and does a redundant `this.get(rootId)` call (line 291) before fetching all rows again and searching for the root task a second time (line 314). The initial `get()` call is only used as an existence check; the actual tree building uses `allTasks.find()`. `[fixable]`


### `frontend/src/components/__tests__/TaskNode.test.tsx`
_Solid phase-1 implementation with good test coverage and clean architecture, but the frontend `upsertInTree` has a reference-equality bug that will cause incorrect tree updates, and mutation functions lack error handling._

- 🔵 **missing_tests**: The `pending_review` status is defined in `STATUS_ICONS` and `NEXT_STATUS` but is not included in the `it.each` test for status icons (line 56). The test covers 6 of 7 statuses. `[fixable]`
- 🔵 **style** (L7): The `afterEach(cleanup)` call is at the top of the file (line 7) but the `import type { Task }` is below it (line 9). The import should be grouped with other imports at the top of the file. `[fixable]`

server/app.ts
```diff
@@ -150,6 +154,17 @@ let updateAvailable = false;
...
 let onUpdateAvailable: (() => void) | null = null;
 let onInboxUpdated: (() => void) | null = null;
+let onTaskBroadcast: ((event: Record<string, unknown>) => void) | null = null;
+
+// --- Task store ---
+
+const mitzoDir = join(BASE_REPO, '.mitzo');
+try {
+  mkdirSync(mitzoDir, { recursive: true });
+} catch {
```
dimakis, line 164: 🔵 **unsafe_assumptions**: The task database is created at `${REPO_PATH}/.mitzo/tasks.db` during module initialization. If `REPO_PATH` points to a read-only filesystem or the `.mitzo` directory can't be created, the server will crash on import. The `mkdirSync` has a catch block but the `TaskStore` constructor (which calls `db.exec(SCHEMA)`) does not. `[fixable]`

server/task-store.ts
```diff
@@ -0,0 +1,343 @@
...
+  }
+
+  getTree(): Task[] {
+    const rows = this.db!.prepare(
+      'SELECT * FROM tasks ORDER BY priority DESC, created_at ASC',
+    ).all() as TaskRow[];
+    return this.assembleTree(rows);
+  }
+
+  getSubtree(rootId: string): Task[] {
```
dimakis, line 299: 🔵 **style**: `getSubtree` fetches all tasks and does a redundant `this.get(rootId)` call (line 291) before fetching all rows again and searching for the root task a second time (line 314). The initial `get()` call is only used as an existence check; the actual tree building uses `allTasks.find()`. `[fixable]`

frontend/src/components/__tests__/TaskNode.test.tsx
```diff
@@ -0,0 +1,173 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup } from '@testing-library/react';
+import { TaskNode } from '../TaskNode';
+
+afterEach(() => {
+  cleanup();
```
dimakis, line 8: 🔵 **style**: The `afterEach(cleanup)` call is at the top of the file (line 7) but the `import type { Task }` is below it (line 9). The import should be grouped with other imports at the top of the file. `[fixable]`

server/app.ts
```diff
@@ -290,6 +309,63 @@ app.post('/api/repos/open', (req, res) => {
...
+app.post('/api/tasks', (req, res) => {
+  const body = TaskCreateBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  try {
+    const task = taskStore.create(body.data);
+    res.status(201).json({ task });
+    // Broadcast full tree so child tasks appear correctly in all clients
```
dimakis, line 327: 🟡 **bugs**: POST /api/tasks broadcasts `task_updated` with the flat new task. For child tasks, other connected clients won't see it placed under the correct parent because `task_updated` sends only the leaf task (no parent context). The diff originally used `task_state` with the full tree — that was correct. Using `task_updated` for creates means the frontend `upsertInTree` must find the parent in local state, which may not exist if another client's tree is stale. Consider broadcasting `task_state` with the full tree on create (as shown in the diff) for consistency. `[fixable]`

frontend/src/hooks/useTaskBoard.ts
```diff
@@ -0,0 +1,162 @@
...
+    }));
+}
+
+function upsertInTree(tasks: Task[], task: Task): [Task[], boolean] {
+  let found = false;
+  const updated = tasks.map((t) => {
+    if (t.id === task.id) {
+      found = true;
+      return { ...task, children: t.children };
+    }
```
dimakis, line 50: 🟡 **bugs**: When `upsertInTree` finds an existing task by id, it replaces it with `{ ...task, children: t.children }` — preserving the OLD children. But the incoming `task` from WS already has `children: []` (the server's `get()` returns tasks with empty children). This means if the server sends a `task_updated` for a parent task, the client correctly keeps local children. However, if the server ever sends a task with populated children (e.g., from `getTree()`), those children would be silently dropped. This is a latent inconsistency between the two code paths. `[fixable]`

server/task-store.ts
```diff
@@ -0,0 +1,340 @@
...
+      'SELECT * FROM tasks WHERE parent_id = ? ORDER BY priority DESC, created_at ASC',
+    ).all(parentId) as TaskRow[];
+    return rows.map(rowToTask);
+  }
+
+  getTree(): Task[] {
+    const rows = this.db!.prepare(
+      'SELECT * FROM tasks ORDER BY priority DESC, created_at ASC',
+    ).all() as TaskRow[];
+    return this.assembleTree(rows);
```
dimakis, line 296: 🟡 **unsafe_assumptions**: `getSubtree()` fetches ALL tasks from the database (`SELECT * FROM tasks`) just to filter to a subtree. For large task sets this is wasteful. Consider using a recursive CTE (`WITH RECURSIVE`) to fetch only the subtree, or at minimum document this as a known scaling limitation. `[fixable]`

server/task-store.ts
```diff
@@ -0,0 +1,340 @@
...
+      depth,
+      annotations,
+      now,
+      now,
+    );
+
+    return this.get(id)!;
+  }
+
+  get(id: string): Task | null {
```
dimakis, line 205: 🔵 **unsafe_assumptions**: `this.db!` is used throughout without null checks after `close()` is called. If any method is called after `close()`, the non-null assertion will throw a confusing error. A guard like `if (!this.db) throw new Error('Store is closed')` would give a clearer diagnostic. `[fixable]`

frontend/src/components/TaskNode.tsx
```diff
@@ -0,0 +1,93 @@
...
+  onAddChild: (parentId: string) => void;
+}
+
+export function TaskNode({ task, depth, onStatusChange, onDelete, onAddChild }: TaskNodeProps) {
+  const [expanded, setExpanded] = useState(true);
+  const hasChildren = task.children.length > 0;
+
+  return (
+    <div className="task-node" style={{ paddingLeft: `${depth * 16}px` }}>
+      <div className="task-node-row">
```
dimakis, line 38: 🔵 **style**: Inline `style={{ paddingLeft: `${depth * 16}px` }}` on the outer div duplicates the indentation already provided by `.task-node-children { margin-left: 8px }` in CSS. The visual indentation is double-applied — the CSS border-left indent plus the inline padding. Consider using only one mechanism for consistency. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🟡 **bugs** (L327): POST /api/tasks broadcasts `task_updated` with the flat new task. For child tasks, other connected clients won't see it placed under the correct parent because `task_updated` sends only the leaf task (no parent context). The diff originally used `task_state` with the full tree — that was correct. Using `task_updated` for creates means the frontend `upsertInTree` must find the parent in local state, which may not exist if another client's tree is stale. Consider broadcasting `task_state` with the full tree on create (as shown in the diff) for consistency. `[fixable]`


### `frontend/src/hooks/useTaskBoard.ts`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🟡 **bugs** (L50): When `upsertInTree` finds an existing task by id, it replaces it with `{ ...task, children: t.children }` — preserving the OLD children. But the incoming `task` from WS already has `children: []` (the server's `get()` returns tasks with empty children). This means if the server sends a `task_updated` for a parent task, the client correctly keeps local children. However, if the server ever sends a task with populated children (e.g., from `getTree()`), those children would be silently dropped. This is a latent inconsistency between the two code paths. `[fixable]`


### `server/task-store.ts`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🟡 **unsafe_assumptions** (L287): `getSubtree()` fetches ALL tasks from the database (`SELECT * FROM tasks`) just to filter to a subtree. For large task sets this is wasteful. Consider using a recursive CTE (`WITH RECURSIVE`) to fetch only the subtree, or at minimum document this as a known scaling limitation. `[fixable]`
- 🔵 **unsafe_assumptions** (L200): `this.db!` is used throughout without null checks after `close()` is called. If any method is called after `close()`, the non-null assertion will throw a confusing error. A guard like `if (!this.db) throw new Error('Store is closed')` would give a clearer diagnostic. `[fixable]`


### `frontend/src/hooks/__tests__/useTaskBoard.test.ts`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🔵 **missing_tests**: No test covers the `upsertInTree` behavior for child tasks received via WS — e.g., a `task_updated` event with a `parentId` that should be inserted under an existing parent in the tree. This is a key code path for real-time collaboration. `[fixable]`


_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🔵 **style**: The `Task` and `TaskStatus` types are duplicated between `frontend/src/types/task.ts` and `server/task-store.ts`. Consider a shared types package or at least a comment cross-referencing the other definition to keep them in sync.


### `server/__tests__/task-store.test.ts`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🔵 **missing_tests**: No test for updating `tokenUsage`, `sessionId`, `claimedBy`, or `claimedAt` fields. While not in `TaskUpdateInput`, `tokenUsage` tracking will likely be needed in phase 2 — worth noting these fields are stored but not yet updateable. `[fixable]`


### `frontend/src/components/TaskNode.tsx`
_Solid phase-1 implementation with good test coverage. Key concern is the POST broadcast sending `task_updated` instead of `task_state`, which may cause child task insertion issues for multi-client scenarios. Type duplication between frontend/server is a maintainability risk to address early._

- 🔵 **style** (L38): Inline `style={{ paddingLeft: `${depth * 16}px` }}` on the outer div duplicates the indentation already provided by `.task-node-children { margin-left: 8px }` in CSS. The visual indentation is double-applied — the CSS border-left indent plus the inline padding. Consider using only one mechanism for consistency. `[fixable]`

frontend/src/hooks/useTaskBoard.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (!cancelled) {
+          setTasks(result.tasks);
+          setLoading(false);
+        }
+      })
+      .catch(() => {
+        if (!cancelled) {
+          setTasks([]);
+          setLoading(false);
+        }
```
dimakis, line 112: 🟡 **bugs**: wsSubscribe uses key 'global' but the existing convention for system-wide events is 'global:system' (see useSessionList.ts:86). Each distinct key creates a separate WebSocket connection. This opens a redundant WS connection instead of reusing the one that already receives server broadcasts. Use 'global:system' to share the connection. `[fixable]`

frontend/src/pages/TaskBoard.tsx
```diff
@@ -0,0 +1,83 @@
...
+    deleteTask(id);
+  }
+
+  function handleAddChild(parentId: string) {
+    setCreating({ parentId });
+  }
+
+  function handleCreate(title: string, parentId?: string) {
+    createTask({ title, parentId });
+    setCreating(null);
```
dimakis, line 27: 🟡 **bugs**: handleDelete and handleStatusChange call async functions (deleteTask/updateTask) without catching errors. These throw on non-ok HTTP responses (per hook tests), resulting in unhandled promise rejections and no user feedback on failure. `[fixable]`

server/task-store.ts
```diff
@@ -0,0 +1,349 @@
...
+      input.parentId ?? null,
+      input.title,
+      input.description ?? null,
+      input.sessionPolicy ?? 'auto',
+      input.priority ?? 0,
+      depth,
+      annotations,
+      now,
+      now,
+    );
```
dimakis, line 200: 🔵 **style**: Inconsistent DB access pattern: create() uses the safe this.getDb() (which throws if store is closed), but get(), update(), delete(), listRoots(), getChildren(), getTree(), getSubtree() all use this.db! which would throw a less descriptive null-dereference error if called after close(). Use this.getDb() consistently. `[fixable]`

server/task-store.ts
```diff
@@ -0,0 +1,349 @@
...
+    const result = this.getDb().prepare('DELETE FROM tasks WHERE id = ?').run(id);
+    return result.changes > 0;
+  }
+
+  listRoots(): Task[] {
+    const rows = this.getDb()
+      .prepare('SELECT * FROM tasks WHERE parent_id IS NULL ORDER BY priority DESC, created_at ASC')
+      .all() as TaskRow[];
+    return rows.map(rowToTask);
+  }
```
dimakis, line 283: 🔵 **unsafe_assumptions**: getTree() and getSubtree() fetch ALL tasks with 'SELECT * FROM tasks' every time. This is fine for small task boards but will degrade with large datasets. Consider documenting the expected scale or adding a limit. Not blocking for phase 1.

frontend/src/types/task.ts
```diff
@@ -0,0 +1,32 @@
+export type TaskStatus =
```
dimakis, line 1: 🔵 **style**: The Task and TaskStatus types are duplicated between frontend/src/types/task.ts and server/task-store.ts. Consider a shared types package or at minimum a comment noting the duplication to keep them in sync.

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useTaskBoard.ts`
_Solid phase-1 implementation with good test coverage. Key issue: the WS subscription uses a different pool key than the existing convention, creating a redundant WebSocket connection. Fire-and-forget mutations in the page component should catch errors._

- 🟡 **bugs** (L112): wsSubscribe uses key 'global' but the existing convention for system-wide events is 'global:system' (see useSessionList.ts:86). Each distinct key creates a separate WebSocket connection. This opens a redundant WS connection instead of reusing the one that already receives server broadcasts. Use 'global:system' to share the connection. `[fixable]`


### `frontend/src/pages/TaskBoard.tsx`
_Solid phase-1 implementation with good test coverage. Key issue: the WS subscription uses a different pool key than the existing convention, creating a redundant WebSocket connection. Fire-and-forget mutations in the page component should catch errors._

- 🟡 **bugs** (L27): handleDelete and handleStatusChange call async functions (deleteTask/updateTask) without catching errors. These throw on non-ok HTTP responses (per hook tests), resulting in unhandled promise rejections and no user feedback on failure. `[fixable]`


### `server/task-store.ts`
_Solid phase-1 implementation with good test coverage. Key issue: the WS subscription uses a different pool key than the existing convention, creating a redundant WebSocket connection. Fire-and-forget mutations in the page component should catch errors._

- 🔵 **style** (L200): Inconsistent DB access pattern: create() uses the safe this.getDb() (which throws if store is closed), but get(), update(), delete(), listRoots(), getChildren(), getTree(), getSubtree() all use this.db! which would throw a less descriptive null-dereference error if called after close(). Use this.getDb() consistently. `[fixable]`
- 🔵 **unsafe_assumptions** (L283): getTree() and getSubtree() fetch ALL tasks with 'SELECT * FROM tasks' every time. This is fine for small task boards but will degrade with large datasets. Consider documenting the expected scale or adding a limit. Not blocking for phase 1.


### `frontend/src/hooks/__tests__/useTaskBoard.test.ts`
_Solid phase-1 implementation with good test coverage. Key issue: the WS subscription uses a different pool key than the existing convention, creating a redundant WebSocket connection. Fire-and-forget mutations in the page component should catch errors._

- 🔵 **missing_tests**: No test for WS task_updated updating a nested child (child of child). The upsertInTree logic recurses into children, but only root-level and one-level-deep updates are tested. `[fixable]`


_Solid phase-1 implementation with good test coverage. Key issue: the WS subscription uses a different pool key than the existing convention, creating a redundant WebSocket connection. Fire-and-forget mutations in the page component should catch errors._

- 🔵 **style** (L1): The Task and TaskStatus types are duplicated between frontend/src/types/task.ts and server/task-store.ts. Consider a shared types package or at minimum a comment noting the duplication to keep them in sync.

PR #185, opened 2026-04-13
fix(token-bar): accurate context window and session totals

dimakis:
## Summary
- **Agent context showed "1"** — was only reading `input_tokens`, now sums `input_tokens + cache_read + cache_creation` for true context window size (with prompt caching, `input_tokens` alone can be as low as 1)
- **Sub-agents overwrote parent context** — now filters on `parent_tool_use_id === null` so only the parent agent updates the context gauge and turn index
- **Session total was too low** — now includes all token types (input + output + cache_read + cache_creation)
- **Removed `costUsd`** from token bar, state, and WS messages — focus on tokens

## Test plan
- [ ] Verify agent context shows realistic numbers (e.g. 50k+) instead of "1"
- [ ] Verify session total grows across turns and includes cached tokens
- [ ] Verify turn index only increments on parent agent turns, not sub-agents
- [ ] Verify cost row no longer appears in expanded token bar detail
- [ ] `npm test` passes (2131/2131 — 5 pre-existing failures in unrelated yapper-proxy mock)

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -462,7 +462,14 @@ app.post('/api/internal/task-tools/block', (req, res) => {
 
 app.get('/api/loop/status', (req, res) => {
   if (!orchestrator) {
-    res.json({ state: 'idle', goalId: null, activeTaskId: null, progress: null });
+    res.json({
+      state: 'idle',
+      goalId: null,
+      activeTaskId: null,
+      progress: null,
+      specMode: false,
+      awaitingApproval: false,
+    });
     return;
   }
   res.json(orchestrator.getStatus());
--- a/server/app.ts
+++ b/server/app.ts
@@ -30,6 +30,12 @@ import { getLocalCommit, isUpdateAvailable } from './git-version.js';
 import { resolvePending } from './permissions.js';
 import { createLogger } from './logger.js';
 import {
+  handleTaskSet,
+  handleTaskComplete,
+  handleTaskStatus,
+  handleTaskBlock,
+} from './task-tools.js';
+import {
   LoginBody,
   FileWriteBody,
   PermissionDecision,
@@ -364,6 +370,83 @@ app.delete('/api/tasks/:id', (req, res) => {
   }
   res.json({ ok: true });
   onTaskBroadcast?.({ type: 'task_deleted', taskId: req.params.id });
+});
+
+// --- Internal task-tool endpoints (MCP server callback) ---
+
+function resolveTaskContext(req: express.Request): string | null {
+  const clientId = req.headers['x-client-id'] as string | undefined;
+  if (!clientId) return null;
+  const session = registry.get(clientId);
+  return session?.taskContext?.currentTaskId ?? null;
+}
+
+app.post('/api/internal/task-tools/set', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+  const taskId = resolveTaskContext(req);
+  if (!taskId) {
+    res.status(400).json({ error: 'No active task context' });
+    return;
+  }
+  const result = handleTaskSet(taskStore, taskId, req.body.tasks ?? []);
+  res.json({ result });
+  onTaskBroadcast?.({
+    type: 'task_state',
+    tasks: taskStore.getTree(),
+  });
+});
+
+app.post('/api/internal/task-tools/complete', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+  const taskId = resolveTaskContext(req);
+  if (!taskId) {
+    res.status(400).json({ error: 'No active task context' });
+    return;
+  }
+  const result = handleTaskComplete(taskStore, taskId, req.body.summary ?? '');
+  res.json({ result });
+  onTaskBroadcast?.({
+    type: 'task_state',
+    tasks: taskStore.getTree(),
+  });
+});
+
+app.get('/api/internal/task-tools/status', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+  const taskId = resolveTaskContext(req);
+  if (!taskId) {
+    res.status(400).json({ error: 'No active task context' });
+    return;
+  }
+  const result = handleTaskStatus(taskStore, taskId);
+  res.json({ result });
+});
+
+app.post('/api/internal/task-tools/block', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+  const taskId = resolveTaskContext(req);
+  if (!taskId) {
+    res.status(400).json({ error: 'No active task context' });
+    return;
+  }
+  const result = handleTaskBlock(taskStore, taskId, req.body.reason ?? '');
+  res.json({ result });
+  onTaskBroadcast?.({
+    type: 'task_state',
+    tasks: taskStore.getTree(),
+  });
 });
 
 app.post('/api/auth/login', loginLimiter, async (req, res) => {
--- a/server/app.ts
+++ b/server/app.ts
@@ -511,6 +511,48 @@ app.post('/api/loop/stop', (_req, res) => {
   res.json(orchestrator.stop());
 });
 
+app.post('/api/tasks/:id/approve', (req, res) => {
+  if (!orchestrator) {
+    res.status(503).json({ error: 'Orchestrator not initialized' });
+    return;
+  }
+  const ok = orchestrator.approveTask(req.params.id);
+  if (!ok) {
+    res.status(400).json({ error: 'Task not in pending_review state' });
+    return;
+  }
+  res.json({ ok: true });
+});
+
+app.post('/api/tasks/:id/reject', (req, res) => {
+  if (!orchestrator) {
+    res.status(503).json({ error: 'Orchestrator not initialized' });
+    return;
+  }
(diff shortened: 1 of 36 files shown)


server/task-orchestrator.ts
```diff
@@ -0,0 +1,328 @@
...
+    if (!next) {
+      // No executable tasks — could be all blocked or pending_review
+      log.info('no executable tasks found', { goalId: this.goalId });
+      this.state = 'paused';
+      this.deps.broadcastStatus(this.getStatus());
+      return;
+    }
+
+    // Assign task
+    this.activeTaskId = next.id;
```
dimakis, line 290: 🟡 **bugs**: tick() sets status to 'active' and sends a new prompt every time it runs, even if the task is already the active task. If onTaskCompleted fires but the next task happens to be the same activeTaskId (e.g. after a rejected pending_review → active), tick() will re-send the prompt redundantly. Consider guarding: `if (next.id === this.activeTaskId) return;` when the task is already active. `[fixable]`

server/task-orchestrator.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  /** Reject the spec mode decomposition — delete children and stop. */
+  rejectSpec(): LoopStatus {
+    if (!this.awaitingApproval || !this.goalId) return this.getStatus();
+    // Delete proposed children
+    const children = this.deps.store.getChildren(this.goalId);
+    for (const child of children) {
+      this.deps.store.delete(child.id);
+    }
+    this.deps.store.update(this.goalId, { status: 'pending' });
```
dimakis, line 174: 🟡 **bugs**: rejectSpec() deletes children with store.delete() in a loop but doesn't handle nested grandchildren. If the spec-mode agent created a multi-level tree (TaskSet on children), only direct children are deleted — grandchildren become orphans with dangling parent_id references. store.delete() should cascade or the loop should use getSubtree(). `[fixable]`

server/app.ts
```diff
@@ -366,6 +381,187 @@ app.delete('/api/tasks/:id', (req, res) => {
...
+  const body = LoopStartBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: 'goalId is required' });
+    return;
+  }
+  if (!orchestrator) {
+    res.status(503).json({ error: 'Orchestrator not initialized' });
+    return;
+  }
+  const status = orchestrator.getStatus();
```
dimakis, line 488: 🟡 **bugs**: POST /api/loop/start only rejects when state is 'running', allowing start() to be called while 'paused'. The orchestrator's start() will reset goalId, pinnedClientId, and specMode — effectively discarding the paused session state. Either the endpoint should reject paused state with 409, or start() should guard against it. `[fixable]`

server/query-loop.ts
```diff
@@ -394,19 +397,30 @@ export async function runQueryLoop(
           currentSession.currentSnapshot = { messageId: currentMessageId, blocks: [] };
```
dimakis, line 397: 🟡 **regressions**: parent_tool_use_id may not always be present on older SDK versions. The `msg.parent_tool_use_id === null || msg.parent_tool_use_id === undefined` check treats missing field as parent (which is correct for backwards compatibility), but the type is not declared in the stream event — if the SDK types are strict, this access may produce a TypeScript error at build time. Verify with `npm run typecheck`. `[fixable]`

server/index.ts
```diff
@@ -80,6 +82,56 @@ setTaskBroadcast((event) => {
   });
 });
 
+// --- Task Orchestrator ---
+const orchestrator = new TaskOrchestrator({
+  store: taskStore,
```
dimakis, line 87: 🟡 **unsafe_assumptions**: getClientId callback iterates registry.entries() and returns the first attached client. With multiple connected clients, this is non-deterministic — the pinned client may change across restarts or reconnects. Consider persisting the chosen clientId or requiring explicit selection. `[fixable]`

server/chat.ts
```diff
@@ -23,6 +23,13 @@ import { runQueryLoop, createWsMessageHandler, broadcastToObservers } from './qu
 import { AsyncQueue } from './async-queue.js';
 import { GIT_BRANCH_TIMEOUT_MS, SESSION_LIST_LIMIT, SESSION_MESSAGES_LIMIT } from './constants.js';
 import { INTERNAL_TOKEN } from './internal-token.js';
+import { buildTaskSystemPrompt } from './task-context.js';
+import type { TaskStore } from './task-store.js';
+
+let _taskStore: TaskStore | null = null;
```
dimakis, line 29: 🔵 **style**: Module-level mutable state (`let _taskStore`) with a setter function is used to break a circular dependency between chat.ts and app.ts. This pattern is fragile — consider passing the store via startChat options or a shared context object instead.

server/task-mcp-server.ts
```diff
@@ -0,0 +1,152 @@
...
+ * TaskBlock tools to the Agent SDK.
+ *
+ * Launched as a child process by Mitzo's chat.ts. Calls back to the
+ * Mitzo HTTP API for all operations. Receives config as CLI args + env:
+ *   --base-url <url>  --client-id <id>  env MITZO_INTERNAL_TOKEN
+ */
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { z } from 'zod';
+
```
dimakis, line 13: 🔵 **unsafe_assumptions**: parseArgs() is called at module top-level (line 30). If this file is ever imported for testing or type-checking without the required args, it will throw immediately and crash the process. Consider deferring parsing to main(). `[fixable]`

dimakis (commented):
## Centaur Review

Found **9** issue(s) (5 warning).


_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🟡 **bugs** (L290): tick() sets status to 'active' and sends a new prompt every time it runs, even if the task is already the active task. If onTaskCompleted fires but the next task happens to be the same activeTaskId (e.g. after a rejected pending_review → active), tick() will re-send the prompt redundantly. Consider guarding: `if (next.id === this.activeTaskId) return;` when the task is already active. `[fixable]`
- 🟡 **bugs** (L174): rejectSpec() deletes children with store.delete() in a loop but doesn't handle nested grandchildren. If the spec-mode agent created a multi-level tree (TaskSet on children), only direct children are deleted — grandchildren become orphans with dangling parent_id references. store.delete() should cascade or the loop should use getSubtree(). `[fixable]`


### `server/app.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🟡 **bugs** (L488): POST /api/loop/start only rejects when state is 'running', allowing start() to be called while 'paused'. The orchestrator's start() will reset goalId, pinnedClientId, and specMode — effectively discarding the paused session state. Either the endpoint should reject paused state with 409, or start() should guard against it. `[fixable]`


### `server/query-loop.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🟡 **regressions** (L397): parent_tool_use_id may not always be present on older SDK versions. The `msg.parent_tool_use_id === null || msg.parent_tool_use_id === undefined` check treats missing field as parent (which is correct for backwards compatibility), but the type is not declared in the stream event — if the SDK types are strict, this access may produce a TypeScript error at build time. Verify with `npm run typecheck`. `[fixable]`


### `server/index.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🟡 **unsafe_assumptions** (L87): getClientId callback iterates registry.entries() and returns the first attached client. With multiple connected clients, this is non-deterministic — the pinned client may change across restarts or reconnects. Consider persisting the chosen clientId or requiring explicit selection. `[fixable]`


### `server/__tests__/loop-routes.test.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🔵 **missing_tests**: Loop routes test file tests the orchestrator directly (unit-level) rather than through HTTP endpoints via supertest, despite the filename suggesting route-level tests. The internal task-tool endpoints (/api/internal/task-tools/*) also lack HTTP-level tests for auth verification and error paths.


### `server/chat.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🔵 **style** (L29): Module-level mutable state (`let _taskStore`) with a setter function is used to break a circular dependency between chat.ts and app.ts. This pattern is fragile — consider passing the store via startChat options or a shared context object instead.


### `frontend/src/components/LoopControls.tsx`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🔵 **missing_tests**: The phase-2 plan calls for 8 tests for LoopControls but no test file exists (frontend/src/components/__tests__/LoopControls.test.tsx is missing from the diff). Similarly, TaskSidebar has no tests in this PR despite the plan specifying 8.


### `server/task-mcp-server.ts`
_Solid Phase 2 implementation with good test coverage for core logic, but tick() re-entry and rejectSpec() cascade deletion need attention to prevent duplicate prompts and orphaned grandchildren._

- 🔵 **unsafe_assumptions** (L13): parseArgs() is called at module top-level (line 30). If this file is ever imported for testing or type-checking without the required args, it will throw immediately and crash the process. Consider deferring parsing to main(). `[fixable]`

PR #186, opened 2026-04-13
fix(tokens): stop double-counting cumulative session tokens

dimakis:
## Summary
- The SDK reports **cumulative** usage totals in each `result` event, but `cumulativeSessionTokens` was using `+=`, causing the Σ display to inflate on every turn
- Changed `+=` to `=` for both `cumulativeSessionTokens` and `cumulativeCostUsd` since the SDK value already represents the session lifetime total
- Added multi-turn test that verifies session total reflects cumulative SDK value, not a sum of cumulative values

## Test plan
- [x] Existing token-update tests pass (11/11)
- [x] New `does not double-count session total across multiple result events` test passes
- [ ] Manual: start a multi-turn session and verify Σ tracks correctly

### DIFF
--- a/server/__tests__/token-update.test.ts
+++ b/server/__tests__/token-update.test.ts
@@ -302,6 +302,75 @@ describe('token_update emission', () => {
     expect(messageStartUpdates[1]).toMatchObject({ turnIndex: 2 });
   });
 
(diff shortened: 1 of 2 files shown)


dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #192, opened 2026-04-14
fix(sessions): fix session reattachment for multi-client and expiry

dimakis:
## Summary

- **Multi-client fix**: Removed the `isAttached` gate in `tryRouteToActiveSession()` — observers (e.g. phone) can now send messages when the driver (e.g. laptop) has disconnected. The SDK query is still running and downstream code (`sendOrBuffer`, `broadcastToObservers`) already handles detached drivers correctly.
- **Resume fix**: `handleSessionExpired` no longer clears `currentSessionId` — the next send includes `resume` so the SDK can pick up the conversation instead of starting fresh. Only localStorage is cleared to prevent stale auto-restore on fresh page loads.

## Test plan

- [x] Added test: observer can be added to detached session
- [x] Added test: `onSessionExpired` called on "No conversation found" error
- [x] Added test: other errors don't trigger session expiry
- [x] All 291 targeted tests pass
- [ ] Manual: open session on two devices, sleep one, send from the other — should route to existing session
- [ ] Manual: let session expire, resend — should resume not start fresh

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useRef } from 'react';
-import { useParams, useSearchParams, useNavigate } from 'react-router-dom';
+import { useParams, useSearchParams } from 'react-router-dom';
 import { ChatArea } from '../components/ChatArea';
 import { ChatInput } from '../components/ChatInput';
 import { VoiceSettings } from '../components/VoiceSettings';
@@ -19,8 +19,6 @@ import { useSessionMeta } from '../hooks/useSessionMeta';
 export function ChatView() {
   const { sessionId } = useParams<{ sessionId?: string }>();
   const [searchParams] = useSearchParams();
-  const navigate = useNavigate();
-
   const [sessionState, sessionActions, poolKey] = useChatSession(
     sessionId,
     searchParams.get('extraTools') ? 'auto' : 'agent',
@@ -38,18 +36,12 @@ export function ChatView() {
     });
   }, []);
 
-  const handleSessionExpired = useCallback(
-    (staleId: string | undefined) => {
-      sessionActions.setCurrentSessionId(undefined);
-      if (staleId) {
-        localStorage.removeItem(LAST_SESSION_KEY);
-      }
-      if (sessionId) {
-        navigate('/chat', { replace: true });
-      }
-    },
-    [sessionId, navigate, sessionActions],
-  );
+  const handleSessionExpired = useCallback((staleId: string | undefined) => {
+    // Clear from localStorage so fresh page loads don't auto-resume an expired session.
+    // Keep currentSessionId so the next send includes `resume` and the SDK can
+    // attempt to pick up the conversation.
+    if (staleId) localStorage.removeItem(LAST_SESSION_KEY);
+  }, []);
 
   // When a session ID is assigned mid-conversation (started from /chat),
   // update the URL so refreshes and back-navigation land on /chat/:id
--- a/server/index.ts
+++ b/server/index.ts
@@ -176,8 +176,6 @@ function tryRouteToActiveSession(
   if (!resume) return null;
   const found = registry.findBySessionId(resume);
   if (!found) return null;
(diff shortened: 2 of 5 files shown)


server/index.ts
```diff
@@ -176,8 +176,6 @@ function tryRouteToActiveSession(
   if (!resume) return null;
   const found = registry.findBySessionId(resume);
   if (!found) return null;
-  // Only route to the session if the driver is still attached
-  if (!registry.isAttached(found.clientId)) return null;
   // Session is active — subscribe as observer instead of starting a duplicate query
```
dimakis, line 179: 🟡 **regressions**: Removing the `isAttached` guard means `tryRouteToActiveSession` will now call `sendToChat(found.clientId, ...)` on a detached session. `sendToChat` echoes the user message to `session.ws` (the stale driver WS) via `send()`, which is safely guarded by a `readyState` check — so no crash. However, the detach TTL timer is still ticking. If it fires while the observer is attached, `abort()` clears `session.observers` and kills the query, but the observer WS receives no notification (no error/session_end sent to observers on abort). The observer would silently stop receiving updates. Consider cancelling the detach timer when an observer is added, or sending a `session_end` to observers on TTL abort. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -38,18 +36,12 @@ export function ChatView() {
...
-      if (staleId) {
-        localStorage.removeItem(LAST_SESSION_KEY);
-      }
-      if (sessionId) {
-        navigate('/chat', { replace: true });
-      }
-    },
-    [sessionId, navigate, sessionActions],
-  );
+  const handleSessionExpired = useCallback((staleId: string | undefined) => {
```
dimakis, line 39: 🔵 **style**: The `handleSessionExpired` callback now ignores `staleId` being `undefined` and does nothing in that case — the function body is effectively `if (staleId) localStorage.removeItem(...)`. The parameter type is `string | undefined` but the caller in `useChatMessages` always passes `currentSessionIdRef.current` which can be undefined. Consider narrowing the callback signature or documenting why the no-op on `undefined` is intentional. `[fixable]`

dimakis (commented):
## Centaur Review

_Sound fix for multi-client session routing. Main concern: detach TTL timer can still fire and abort a session that has active observers, leaving them silently disconnected._

- 🔵 **missing_tests**: No test verifies that `sendToChat` works correctly when the driver WS is closed (detached). The `send()` guard (`readyState === OPEN`) silently drops the echo to the driver, but the observer still gets it via `broadcastToObservers`. A test confirming the observer receives the message when the driver is detached would strengthen confidence in the core multi-client scenario this PR enables. `[fixable]`


### `server/index.ts`
_Sound fix for multi-client session routing. Main concern: detach TTL timer can still fire and abort a session that has active observers, leaving them silently disconnected._

- 🟡 **regressions** (L179): Removing the `isAttached` guard means `tryRouteToActiveSession` will now call `sendToChat(found.clientId, ...)` on a detached session. `sendToChat` echoes the user message to `session.ws` (the stale driver WS) via `send()`, which is safely guarded by a `readyState` check — so no crash. However, the detach TTL timer is still ticking. If it fires while the observer is attached, `abort()` clears `session.observers` and kills the query, but the observer WS receives no notification (no error/session_end sent to observers on abort). The observer would silently stop receiving updates. Consider cancelling the detach timer when an observer is added, or sending a `session_end` to observers on TTL abort. `[fixable]`


_Sound fix for multi-client session routing. Main concern: detach TTL timer can still fire and abort a session that has active observers, leaving them silently disconnected._

- 🔵 **style** (L39): The `handleSessionExpired` callback now ignores `staleId` being `undefined` and does nothing in that case — the function body is effectively `if (staleId) localStorage.removeItem(...)`. The parameter type is `string | undefined` but the caller in `useChatMessages` always passes `currentSessionIdRef.current` which can be undefined. Consider narrowing the callback signature or documenting why the no-op on `undefined` is intentional. `[fixable]`

dimakis:
Closed — all changes landed via #193 (squash merge included this branch's commits).

PR #193, opened 2026-04-14
fix(sessions): address PR #192 review — detach timer + type safety

dimakis:
## Summary

Addresses all 3 findings from the Centaur review on #192:

- **Critical**: Cancel detach TTL timer when an observer joins a detached session — previously the timer could fire and silently abort the session while observers were still connected (`abort()` deletes the session from the registry before the query loop's `finally` block can broadcast `session_end`)
- **Valid**: Add test coverage for observer receiving messages when driver WS is closed, and for detach timer cancellation on observer add
- **Cosmetic**: Narrow `onSessionExpired` callback from `string | undefined` to `string`, guarding at the call site in `useChatMessages`

## Test plan

- [x] 16/16 multi-client session tests pass (including 2 new)
- [x] Pre-commit hooks (eslint + prettier) pass
- [ ] Manual: open session on two devices, sleep one, verify observer stays connected past detach TTL

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -36,12 +36,16 @@ export function ChatView() {
     });
   }, []);
 
-  const handleSessionExpired = useCallback((_staleId: string) => {
-    // Clear from localStorage so fresh page loads don't auto-resume an expired session.
-    // Keep currentSessionId so the next send includes `resume` and the SDK can
-    // attempt to pick up the conversation.
-    localStorage.removeItem(LAST_SESSION_KEY);
-  }, []);
+  const handleSessionExpired = useCallback(
+    (staleId: string) => {
+      // Clear both state and storage so the next send starts a fresh conversation
+      // instead of retrying `resume` with a stale ID in a loop.
+      sessionActions.setCurrentSessionId(undefined);
+      localStorage.removeItem(LAST_SESSION_KEY);
+      if (sessionId) navigate('/chat', { replace: true });
+    },
+    [sessionId, navigate, sessionActions],
+  );
 
   // When a session ID is assigned mid-conversation (started from /chat),
   // update the URL so refreshes and back-navigation land on /chat/:id
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useEffect, useRef } from 'react';
-import { useParams, useSearchParams, useNavigate } from 'react-router-dom';
+import { useParams, useSearchParams } from 'react-router-dom';
 import { ChatArea } from '../components/ChatArea';
 import { ChatInput } from '../components/ChatInput';
 import { VoiceSettings } from '../components/VoiceSettings';
@@ -19,8 +19,6 @@ import { useSessionMeta } from '../hooks/useSessionMeta';
 export function ChatView() {
   const { sessionId } = useParams<{ sessionId?: string }>();
   const [searchParams] = useSearchParams();
-  const navigate = useNavigate();
-
   const [sessionState, sessionActions, poolKey] = useChatSession(
     sessionId,
     searchParams.get('extraTools') ? 'auto' : 'agent',
@@ -38,18 +36,12 @@ export function ChatView() {
     });
   }, []);
 
-  const handleSessionExpired = useCallback(
-    (staleId: string | undefined) => {
-      sessionActions.setCurrentSessionId(undefined);
-      if (staleId) {
-        localStorage.removeItem(LAST_SESSION_KEY);
-      }
-      if (sessionId) {
-        navigate('/chat', { replace: true });
-      }
-    },
-    [sessionId, navigate, sessionActions],
-  );
+  const handleSessionExpired = useCallback((staleId: string | undefined) => {
+    // Clear from localStorage so fresh page loads don't auto-resume an expired session.
+    // Keep currentSessionId so the next send includes `resume` and the SDK can
+    // attempt to pick up the conversation.
+    if (staleId) localStorage.removeItem(LAST_SESSION_KEY);
+  }, []);
 
   // When a session ID is assigned mid-conversation (started from /chat),
   // update the URL so refreshes and back-navigation land on /chat/:id
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -36,11 +36,11 @@ export function ChatView() {
     });
   }, []);
 
-  const handleSessionExpired = useCallback((staleId: string | undefined) => {
+  const handleSessionExpired = useCallback((_staleId: string) => {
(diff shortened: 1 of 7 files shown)


server/session-registry.ts
```diff
@@ -203,6 +203,9 @@ export class SessionRegistry {
       return null;
     }
     found.session.observers.add(ws);
+    // An active observer means someone is listening — don't let the detach
+    // TTL kill the session out from under them.
+    this.clearDetachTimer(found.clientId);
```
dimakis, line 208: 🔴 **bugs**: addObserver cancels the detach timer, but removeObserver never restarts it. If the driver is detached, an observer joins (cancelling the TTL), and then the observer also disconnects, the session remains alive forever with no driver and no timer — a session leak. removeObserver (or the close handler in index.ts) should restart the detach timer when the last observer leaves a still-detached session. `[fixable]`

server/index.ts
```diff
@@ -176,8 +176,6 @@ function tryRouteToActiveSession(
   if (!resume) return null;
   const found = registry.findBySessionId(resume);
   if (!found) return null;
-  // Only route to the session if the driver is still attached
-  if (!registry.isAttached(found.clientId)) return null;
   // Session is active — subscribe as observer instead of starting a duplicate query
   const driverId = registry.addObserver(resume, ws);
```
dimakis, line 180: 🟡 **regressions**: Removing the isAttached guard means tryRouteToActiveSession now routes messages to detached sessions. While addObserver cancels the detach timer (good), the incoming prompt is sent via sendToChat to a session whose driver WS may be closed. Verify that sendToChat + broadcastToObservers handles a closed driver WS gracefully (the new test at line 277 suggests it does, but only for broadcast — not for the SDK query continuation path).

frontend/src/pages/ChatView.tsx
```diff
@@ -38,18 +36,12 @@ export function ChatView() {
...
-      if (staleId) {
-        localStorage.removeItem(LAST_SESSION_KEY);
-      }
-      if (sessionId) {
-        navigate('/chat', { replace: true });
-      }
-    },
-    [sessionId, navigate, sessionActions],
-  );
+  const handleSessionExpired = useCallback((_staleId: string) => {
```
dimakis: 🟡 **regressions**: The old handler cleared currentSessionId and navigated away from /chat/:id on expiry. The new handler only clears localStorage. If the user is on /chat/:sessionId with an expired session, the URL still contains the stale ID. The next send will include `resume` with that stale ID, and if the SDK can't find it, the user may get stuck in a loop of 'No conversation found' errors. The comment says 'the SDK can attempt to pick up the conversation' but if the session truly expired, this retry will also fail. `[fixable]`

dimakis: 🔵 **style**: The _staleId parameter is prefixed with underscore (unused), but the function signature declares it as `string` — if it's truly unused in both ChatView and DesktopChatView, consider whether the callback type should be simplified, or whether the staleId should actually be used (e.g., to clear it from localStorage by key). `[fixable]`

dimakis (commented):
## Centaur Review

### `server/session-registry.ts`
_The detach timer cancellation on observer add is correct, but missing the complementary restart on last-observer removal creates a session leak. The session expiry UX change may also cause retry loops for truly expired sessions._

- 🔴 **bugs** (L208): addObserver cancels the detach timer, but removeObserver never restarts it. If the driver is detached, an observer joins (cancelling the TTL), and then the observer also disconnects, the session remains alive forever with no driver and no timer — a session leak. removeObserver (or the close handler in index.ts) should restart the detach timer when the last observer leaves a still-detached session. `[fixable]`


_The detach timer cancellation on observer add is correct, but missing the complementary restart on last-observer removal creates a session leak. The session expiry UX change may also cause retry loops for truly expired sessions._

- 🟡 **missing_tests**: No test covers the scenario where the last observer disconnects from a detached session. This is the complement of the 'addObserver cancels detach timer' test — need to verify the timer is restarted (once the leak above is fixed) when observers.size drops to 0 while the driver is still detached. `[fixable]`


### `server/index.ts`
_The detach timer cancellation on observer add is correct, but missing the complementary restart on last-observer removal creates a session leak. The session expiry UX change may also cause retry loops for truly expired sessions._

- 🟡 **regressions** (L180): Removing the isAttached guard means tryRouteToActiveSession now routes messages to detached sessions. While addObserver cancels the detach timer (good), the incoming prompt is sent via sendToChat to a session whose driver WS may be closed. Verify that sendToChat + broadcastToObservers handles a closed driver WS gracefully (the new test at line 277 suggests it does, but only for broadcast — not for the SDK query continuation path).


_The detach timer cancellation on observer add is correct, but missing the complementary restart on last-observer removal creates a session leak. The session expiry UX change may also cause retry loops for truly expired sessions._

- 🟡 **regressions** (L39): The old handler cleared currentSessionId and navigated away from /chat/:id on expiry. The new handler only clears localStorage. If the user is on /chat/:sessionId with an expired session, the URL still contains the stale ID. The next send will include `resume` with that stale ID, and if the SDK can't find it, the user may get stuck in a loop of 'No conversation found' errors. The comment says 'the SDK can attempt to pick up the conversation' but if the session truly expired, this retry will also fail. `[fixable]`
- 🔵 **style** (L39): The _staleId parameter is prefixed with underscore (unused), but the function signature declares it as `string` — if it's truly unused in both ChatView and DesktopChatView, consider whether the callback type should be simplified, or whether the staleId should actually be used (e.g., to clear it from localStorage by key). `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -39,14 +37,12 @@ export function ChatView() {
...
+      // instead of retrying `resume` with a stale ID in a loop.
       sessionActions.setCurrentSessionId(undefined);
-      if (staleId) {
-        localStorage.removeItem(LAST_SESSION_KEY);
-      }
-      if (sessionId) {
-        navigate('/chat', { replace: true });
-      }
+      localStorage.removeItem(LAST_SESSION_KEY);
+      if (sessionId) navigate('/chat', { replace: true });
```
dimakis, line 46: 🔴 **bugs**: `navigate` is used in `handleSessionExpired` (line 45) and its dependency array (line 47), but the `useNavigate` import and `const navigate = useNavigate()` declaration were removed in this PR. This will cause a ReferenceError at runtime when a session expires. Either restore the import + declaration or use `window.location` as an alternative. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s) (1 critical).


_Critical bug: ChatView.tsx references `navigate` after its import and declaration were removed — will crash at runtime on session expiry._

- 🔴 **bugs** (L45): `navigate` is used in `handleSessionExpired` (line 45) and its dependency array (line 47), but the `useNavigate` import and `const navigate = useNavigate()` declaration were removed in this PR. This will cause a ReferenceError at runtime when a session expires. Either restore the import + declaration or use `window.location` as an alternative. `[fixable]`

PR #198, opened 2026-04-15
fix(sessions): paginate session list and fix worktree discovery

dimakis:
## Summary

- **Worktree sessions invisible**: `getSessionDirs()` manually scanned `.claude/worktrees/` and `~/.claude/projects/` with a broken prefix pattern that never matched worktree-path-based project hashes. Now relies on the SDK's `includeWorktrees: true` (the default), which handles discovery automatically.
- **Hard cap of 20**: `SESSION_LIST_LIMIT` truncated all sessions to 20 with no way to see older ones. Replaced with offset/limit pagination — `GET /api/sessions?offset=N&limit=N` returns `{ sessions, hasMore }`. Frontend shows a "Load More" button.

## Test plan

- [ ] Verify worktree sessions appear in session list (previously invisible)
- [ ] Verify older sessions (beyond first 20) load via "Load More"
- [ ] Verify session dismiss, rename, and clear still work
- [ ] Verify visibility-change refresh resets to page 1
- [ ] `npx tsc --noEmit` passes
- [ ] `npm test` — session-related tests pass

### DIFF
--- a/frontend/src/components/__tests__/SessionPanel.test.tsx
+++ b/frontend/src/components/__tests__/SessionPanel.test.tsx
@@ -17,12 +17,15 @@ function makeDefaultReturn(overrides = {}) {
     sessions: [],
     quickActions: [],
     loading: false,
+    loadingMore: false,
+    hasMore: false,
     updateAvailable: false,
     checking: false,
     dismissSession: vi.fn(),
     clearAll: vi.fn(),
     handleRename: vi.fn(),
     checkForUpdates: vi.fn(),
+    loadMore: vi.fn(),
     ...overrides,
   };
 }
--- a/frontend/src/hooks/useSessionList.ts
+++ b/frontend/src/hooks/useSessionList.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect, useCallback, useRef } from 'react';
 import type { Session } from '../types/chat';
 import { renameSession as renameSessionApi } from '../lib/rename-session';
 
@@ -40,8 +40,6 @@ export interface UseSessionListReturn {
   loadMore: () => void;
 }
 
-const PAGE_SIZE = 20;
-
 function parseSessionsResponse(data: unknown): { sessions: Session[]; hasMore: boolean } {
   // Handle both new paginated shape and legacy array shape
   if (Array.isArray(data)) return { sessions: data, hasMore: false };
@@ -57,11 +55,12 @@ export function useSessionList(): UseSessionListReturn {
   const [hasMore, setHasMore] = useState(false);
   const [updateAvailable, setUpdateAvailable] = useState(false);
   const [checking, setChecking] = useState(false);
+  const nextOffset = useRef(0);
 
   useEffect(() => {
     const loadAll = () =>
       Promise.all([
-        fetch(`/api/sessions?offset=0&limit=${PAGE_SIZE}`)
+        fetch('/api/sessions')
           .then((r) => r.json())
           .catch(() => ({ sessions: [], hasMore: false })),
         fetch('/api/config')
@@ -74,6 +73,7 @@ export function useSessionList(): UseSessionListReturn {
         const { sessions: page, hasMore: more } = parseSessionsResponse(sessData);
         setSessions(page);
         setHasMore(more);
+        nextOffset.current = page.length;
         setQuickActions(buildQuickActions(config.quickActions));
         if (version?.updateAvailable) setUpdateAvailable(true);
       });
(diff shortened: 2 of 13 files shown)


dimakis:
## Centaur Review

### `server/__tests__/todos.test.ts`
_Pagination implementation is solid, but 4 test files have stale mocks returning the old array format instead of `{ sessions, hasMore }` — these will break any test that exercises the sessions route._

- 🔴 **regressions** (L18): Mock still returns old array format `[]` instead of `{ sessions: [], hasMore: false }`. Same issue in `calendar.test.ts:17`, `task-routes.test.ts:17`, and `yapper-proxy.test.ts:21`. These mocks will cause the `/api/sessions` route handler to destructure `{ sessions, hasMore }` from an array, yielding `sessions: undefined`. Any test hitting that route will get a broken response shape. `[fixable]`


### `server/chat.ts`
_Pagination implementation is solid, but 4 test files have stale mocks returning the old array format instead of `{ sessions, hasMore }` — these will break any test that exercises the sessions route._

- 🟡 **bugs** (L717): `fetchLimit = offset + limit + 50` is applied per-directory in the loop, but dedup happens after merging all dirs. If many sessions are duplicated across dirs, the 50-item buffer may not be enough and `hasMore` could be falsely `true` or the page could be short. This is a pre-existing architectural issue made slightly worse by pagination, but unlikely to matter in practice with typical session counts.


### `frontend/src/hooks/useSessionList.ts`
_Pagination implementation is solid, but 4 test files have stale mocks returning the old array format instead of `{ sessions, hasMore }` — these will break any test that exercises the sessions route._

- 🟡 **unsafe_assumptions** (L96): `loadMore` uses `sessions.length` as the offset, but if a session is dismissed (removed from local state) between loads, the offset will be off by the number of dismissed sessions — potentially skipping or duplicating entries on the next page fetch. `[fixable]`
- 🟡 **missing_tests** (L93): No tests for the `loadMore` pagination flow (initial load sets `hasMore`, `loadMore` appends, guard against concurrent calls, error handling). The existing `useSessionList.test.ts` doesn't cover these new code paths. `[fixable]`
- 🔵 **style** (L46): `parseSessionsResponse` casts `data as { sessions?: Session[]; hasMore?: boolean }` without runtime validation. Consider checking `typeof obj.sessions` and `typeof obj.hasMore` for robustness, or use the Zod schemas the project already relies on. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/__tests__/task-routes.test.ts`
_Pagination implementation is solid overall, but stale mocks in 4 test files will break those test suites since they return the old array shape instead of the new `{ sessions, hasMore }` object._

- 🔴 **regressions**: Mock returns old array shape `[]` instead of new `{ sessions: [], hasMore: false }`. The route handler destructures `{ sessions, hasMore }` from `getSessions()`, so this mock will cause `sessions` to be `undefined` and crash the `.map()` call in the route. Same issue in `yapper-proxy.test.ts`, `calendar.test.ts`, and `todos.test.ts`. `[fixable]`


### `server/chat.ts`
_Pagination implementation is solid overall, but stale mocks in 4 test files will break those test suites since they return the old array shape instead of the new `{ sessions, hasMore }` object._

- 🟡 **unsafe_assumptions**: The `fetchLimit = offset + limit + 50` buffer is a heuristic that may undercount when many duplicates exist across session dirs. If more than 50 sessions are hidden or deduplicated, `hasMore` could be `false` even though more sessions exist, silently truncating the list. This is unlikely in practice but worth documenting the trade-off.


### `frontend/src/hooks/useSessionList.ts`
_Pagination implementation is solid overall, but stale mocks in 4 test files will break those test suites since they return the old array shape instead of the new `{ sessions, hasMore }` object._

- 🟡 **unsafe_assumptions**: Using `sessions.length` as offset in `loadMore` becomes incorrect if a session is dismissed (deleted) between page loads. After dismissing a session, `sessions.length` shrinks, so the next `loadMore` offset will overlap with already-fetched data, producing duplicates in the list. Consider tracking the offset independently of the array length. `[fixable]`
- 🔵 **missing_tests**: No frontend tests cover `loadMore` behavior — guard against concurrent loads, appending to existing sessions, or error handling. The `parseSessionsResponse` backward-compatibility path (legacy array shape) also lacks test coverage. `[fixable]`
- 🔵 **style**: `PAGE_SIZE` is defined as a module constant in the frontend hook (20) and also as `SESSION_PAGE_SIZE` in `server/constants.ts` (20). If these drift apart, the frontend could request pages of a different size than the server default. Consider importing or deriving from a single source, or just relying on the server's default by omitting `limit` from the frontend request. `[fixable]`


_Pagination implementation is solid overall, but stale mocks in 4 test files will break those test suites since they return the old array shape instead of the new `{ sessions, hasMore }` object._

- 🟡 **missing_tests**: No test exercises pagination query params (`offset` and `limit`). The route handler parses and clamps these from query strings — tests should verify boundary behavior (e.g., `?offset=5&limit=10`, invalid values like `offset=-1`, `limit=999`). `[fixable]`

PR #199, opened 2026-04-15
fix(sessions): scan all project dirs for cross-repo session discovery

dimakis:
## Summary
- Sessions created via multi-repo worktrees were invisible because `listSessions` only scanned `BASE_REPO` — the SDK's `includeWorktrees` flag can't cross repo boundaries
- Now scans `~/projects/`, `~/redhat/`, `~/tools/` for any child with `.git/` or `.claude/` and includes each as a session directory
- No config updates needed when new projects are added — discovery is automatic

## Test plan
- [x] All 61 server tests pass
- [ ] Create a new session, open a secondary repo (e.g. centaur), verify the session appears in the session list
- [ ] Verify sessions from older projects still appear
- [ ] Verify pagination still works with the larger directory set

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -9,7 +9,6 @@ import type { WebSocket } from 'ws';
 import { execFileSync } from 'child_process';
 import { readFileSync, writeFileSync, mkdirSync, readdirSync, existsSync } from 'fs';
 import { join } from 'path';
-import { homedir } from 'os';
 import { randomBytes } from 'crypto';
 import { createWorktree, removeWorktree } from './worktree.js';
 import { SessionRegistry, type MitzoMode } from './session-registry.js';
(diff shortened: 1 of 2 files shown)


dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Hardcoded `~/projects`, `~/redhat`, `~/tools` paths violate the project's portability rule — this needs to be configurable, and the new logic needs tests._

- 🔴 **unsafe_assumptions** (L668): Hardcoded parent directories (`~/projects`, `~/redhat`, `~/tools`) are machine-specific paths. CLAUDE.md explicitly states: "No hardcoded machine-specific paths or configuration. Everything must be generic and portable." This should be configurable — e.g. via an env var like `SESSION_SCAN_DIRS`, or via `.mitzo.json`, or by reading the Claude SDK's known project list. `[fixable]`
- 🟡 **bugs** (L674): `readdirSync(parent)` returns file names, not just directories. `existsSync(join(child, '.git'))` will be called on regular files too, which is harmless but wasteful. More importantly, symlinks to directories are followed silently — a symlink loop under one of these parents could cause unexpected behavior. Consider filtering with `readdirSync(parent, { withFileTypes: true })` and checking `entry.isDirectory()` before descending. `[fixable]`
- 🟡 **missing_tests** (L659): `getSessionDirs()` has no test coverage (confirmed by searching for test files). The new scanning logic — deduplication via `seen`, skipping non-project dirs, handling missing parents — should be tested, especially since it's now doing filesystem traversal across multiple directories. `[fixable]`
- 🔵 **style** (L663): The comment references "worktree sessions span multiple repos" but the removed comment at the bottom said "SDK's listSessions with includeWorktrees: true handles worktree-based sessions automatically." The removal of that comment without explanation may confuse future readers about whether worktree discovery is still SDK-handled or now manually handled here. `[fixable]`

server/chat.ts
```diff
@@ -657,6 +658,31 @@ export function isActive(clientId: string): boolean {
...
 function getSessionDirs(): string[] {
   const dirs = [BASE_REPO];
+  const seen = new Set([BASE_REPO]);
+
+  // Scan parent directories for any project that has .git/ or .claude/ —
+  // sessions can live in any project Claude has been used in, and worktree
+  // sessions span multiple repos. The SDK's includeWorktrees: true discovers
+  // worktree sessions within each repo but can't cross repo boundaries.
+  const parentDirs = [
+    join(homedir(), 'projects'),
```
dimakis: 🔴 **unsafe_assumptions**: Hardcoded parent directories (`~/projects`, `~/redhat`, `~/tools`) are machine-specific paths. CLAUDE.md explicitly states: "No hardcoded machine-specific paths or configuration. Everything must be generic and portable." This should be configurable — e.g. via an env var like `SESSION_SCAN_DIRS`, or via `.mitzo.json`, or by reading the Claude SDK's known project list. `[fixable]`

server/chat.ts
```diff
@@ -657,6 +658,31 @@ export function isActive(clientId: string): boolean {
...
+    join(homedir(), 'redhat'),
+    join(homedir(), 'tools'),
+  ];
+  for (const parent of parentDirs) {
+    try {
+      for (const entry of readdirSync(parent)) {
```
dimakis: 🟡 **bugs**: `readdirSync(parent)` returns file names, not just directories. `existsSync(join(child, '.git'))` will be called on regular files too, which is harmless but wasteful. More importantly, symlinks to directories are followed silently — a symlink loop under one of these parents could cause unexpected behavior. Consider filtering with `readdirSync(parent, { withFileTypes: true })` and checking `entry.isDirectory()` before descending. `[fixable]`

server/chat.ts
```diff
@@ -657,6 +658,31 @@ export function isActive(clientId: string): boolean {
 
 function getSessionDirs(): string[] {
```
dimakis: 🟡 **missing_tests**: `getSessionDirs()` has no test coverage (confirmed by searching for test files). The new scanning logic — deduplication via `seen`, skipping non-project dirs, handling missing parents — should be tested, especially since it's now doing filesystem traversal across multiple directories. `[fixable]`

server/chat.ts
```diff
@@ -657,6 +658,31 @@ export function isActive(clientId: string): boolean {
...
   const dirs = [BASE_REPO];
+  const seen = new Set([BASE_REPO]);
+
+  // Scan parent directories for any project that has .git/ or .claude/ —
```
dimakis: 🔵 **style**: The comment references "worktree sessions span multiple repos" but the removed comment at the bottom said "SDK's listSessions with includeWorktrees: true handles worktree-based sessions automatically." The removal of that comment without explanation may confuse future readers about whether worktree discovery is still SDK-handled or now manually handled here. `[fixable]`

server/chat.ts
```diff
@@ -655,8 +655,60 @@ export function isActive(clientId: string): boolean {
...
+      if (!seen.has(repoPath)) {
+        dirs.push(repoPath);
+        seen.add(repoPath);
+      }
+    }
+
+    // 2. Derive unique parent dirs from roots, then scan for sibling projects
+    const parentDirs = new Set<string>();
+    for (const root of config.roots) {
+      const parent = join(root.path, '..');
```
dimakis: 🔵 **style**: Use `dirname(root.path)` instead of `join(root.path, '..')` — it's more idiomatic and avoids an unresolved `..` segment in the path (which could cause `seen` dedup misses if the same parent is reached via different paths). Import `dirname` from `'path'`. `[fixable]`

dimakis: 🟡 **bugs**: `join(root.path, '..')` does not resolve symlinks or normalize to an absolute path. If `root.path` is relative or contains symlinks, the derived parent may differ textually from another root's parent that resolves to the same directory, causing `parentDirs` dedup to fail and the same directory to be scanned twice. Consider using `resolve(root.path, '..')` to normalize. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid cross-repo session discovery with good test coverage; minor path normalization concern and a couple of missing edge-case tests._

- 🔵 **style** (L690): Use `dirname(root.path)` instead of `join(root.path, '..')` — it's more idiomatic and avoids an unresolved `..` segment in the path (which could cause `seen` dedup misses if the same parent is reached via different paths). Import `dirname` from `'path'`. `[fixable]`
- 🟡 **bugs** (L690): `join(root.path, '..')` does not resolve symlinks or normalize to an absolute path. If `root.path` is relative or contains symlinks, the derived parent may differ textually from another root's parent that resolves to the same directory, causing `parentDirs` dedup to fail and the same directory to be scanned twice. Consider using `resolve(root.path, '..')` to normalize. `[fixable]`


_Solid cross-repo session discovery with good test coverage; minor path normalization concern and a couple of missing edge-case tests._

- 🔵 **missing_tests**: No test for legacy sessions dir entries (`<BASE_REPO>-sessions/session-*`). The legacy scanning code was pre-existing but the new `seen` set doesn't cover it — a test confirming legacy dirs still appear in results would guard against regressions. `[fixable]`
- 🔵 **missing_tests**: No test verifying that multiple roots sharing the same parent directory only scan the parent once. This is the core dedup logic for `parentDirs` and worth an explicit assertion. `[fixable]`

PR #200, opened 2026-04-15
fix(sessions): stop scanning non-primary repos in session list

dimakis:
## Summary

- `getSessionDirs()` was scanning all configured repos (centaur, contexgin, etc.) plus sibling directories discovered from roots, flooding the session list with **366 automated Centaur code-review sessions** that drowned real mgmt sessions
- The SDK's `listSessions({ dir, includeWorktrees: true })` already handles worktree and `~/.claude/projects/` discovery for `BASE_REPO` — scanning additional repos is redundant for session listing
- Removes repo/roots scanning from `getSessionDirs()`, keeping only `BASE_REPO` and the legacy `-sessions/` dir
- Cleans up unused imports (`existsSync`, `resolve`, `dirname`)
- Net: 14 insertions, 168 deletions

## Root cause

The session list showed ~10 sessions instead of 229 because:
1. `getSessions()` calls `listSessions()` per dir with `fetchLimit = offset + limit + 50`
2. With all repos scanned, the deduped pool was dominated by centaur sessions (366 of them, 92% of page 2+)
3. Recent mgmt sessions appeared on page 1 but older ones were pushed out by the noise

## Test plan

- [x] `getSessionDirs()` returns only `BASE_REPO` (+ legacy dirs)
- [x] Configured repos are NOT included
- [x] Sibling dirs from roots are NOT discovered
- [x] Legacy `-sessions/session-*` dirs still included
- [x] All pre-existing tests pass

### DIFF
--- a/server/__tests__/session-dirs.test.ts
+++ b/server/__tests__/session-dirs.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
-import { mkdirSync, writeFileSync, rmSync } from 'fs';
+import { mkdirSync, rmSync } from 'fs';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import type { RepoConfig } from '../repo-config.js';
@@ -48,7 +48,7 @@ describe('getSessionDirs', () => {
     expect(dirs[0]).toBe(BASE_REPO);
   });
 
-  it('includes explicitly configured repos', () => {
+  it('does NOT include configured repos (SDK discovers via BASE_REPO)', () => {
     const repoA = join(TEST_BASE, 'repo-a');
     mkdirSync(repoA, { recursive: true });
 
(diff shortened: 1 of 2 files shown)


dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean simplification — removes multi-repo scanning in favor of SDK-native discovery. No bugs, no regressions. Minor test hygiene suggestions only._

- 🔵 **missing_tests**: The old test suite had a 'handles config load failure gracefully' test verifying getSessionDirs doesn't throw when loadRepoConfig errors. The new implementation no longer calls loadRepoConfig, so the test is correctly removed — but there's no test verifying the legacy-dir branch handles a non-existent sessions directory gracefully (the try/catch on readdirSync). Consider adding a test that asserts getSessionDirs returns [BASE_REPO] when no legacy dir exists. `[fixable]`


### `server/chat.ts`
_Clean simplification — removes multi-repo scanning in favor of SDK-native discovery. No bugs, no regressions. Minor test hygiene suggestions only._

- 🔵 **regressions**: The mockLoadRepoConfig mock is still set up in the test file (lines 23-26) and exercised in two tests, but getSessionDirs no longer reads config at all. The mock setup is now dead code in the test file — it only exists to verify the negative (dirs are NOT included). This is harmless but slightly misleading; the mock could be removed from the tests entirely since getSessionDirs is now config-independent. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #201, opened 2026-04-15
fix(sessions): discover worktree sessions from ~/.claude/projects/

dimakis:
## Summary

- Sessions created in Mitzo worktrees were becoming invisible after worktree cleanup — **45+ orphaned sessions** recovered
- Root cause: the SDK stores session data under `~/.claude/projects/<encoded-worktree-path>/`, but once the worktree dir is deleted from disk, `listSessions({ dir: BASE_REPO })` can't find them
- Fix: scan `~/.claude/projects/` for dirs matching `<encoded-BASE_REPO>--claude-worktrees-<id>`, reconstruct the original path, and pass it to `listSessions`
- Adds `setClaudeProjectsRoot()` for test injection

## Root cause

1. Mitzo creates a session with CWD set to a worktree dir (e.g., `mgmt/.claude/worktrees/2026-04-15-3b493d/`)
2. SDK stores session data in `~/.claude/projects/-Users-dsaridak-redhat-mgmt--claude-worktrees-2026-04-15-3b493d/`
3. Session ends, Mitzo cleans up the worktree dir from disk
4. `getSessionDirs()` never passes the worktree path → `listSessions` never finds the session
5. Session vanishes from the UI

## Test plan

- [x] Worktree project dirs in `~/.claude/projects/` are discovered and reconstructed
- [x] Non-worktree dirs (base repo, unrelated repos) are not duplicated
- [x] Deduplication works when worktree found via multiple methods
- [x] Graceful handling when `~/.claude/projects/` doesn't exist
- [x] All existing tests still pass

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -10,6 +10,7 @@ import { execFileSync } from 'child_process';
 import { readFileSync, writeFileSync, mkdirSync, readdirSync, existsSync } from 'fs';
 import { join, resolve, dirname } from 'path';
 import { randomBytes } from 'crypto';
+import { homedir } from 'os';
 import { createWorktree, removeWorktree } from './worktree.js';
 import { SessionRegistry, type MitzoMode } from './session-registry.js';
 import { parseContentBlocks } from './content-blocks.js';
@@ -669,6 +670,15 @@ export function isActive(clientId: string): boolean {
  *
  * This keeps all paths configurable (no hardcoded machine-specific paths).
  */
+let _claudeProjectsRoot: string | null = null;
+/** Override ~/.claude/projects root for testing. */
+export function setClaudeProjectsRoot(root: string | null): void {
+  _claudeProjectsRoot = root;
+}
+function getClaudeProjectsRoot(): string {
+  return _claudeProjectsRoot ?? join(homedir(), '.claude', 'projects');
+}
+
 export function getSessionDirs(): string[] {
   const dirs = [BASE_REPO];
   const seen = new Set([BASE_REPO]);
@@ -685,8 +695,6 @@ export function getSessionDirs(): string[] {
     }
 
     // 2. Derive unique parent dirs from roots, then scan for sibling projects.
-    //    Use resolve + dirname to normalize paths and avoid dedup misses from
-    //    symlinks or relative segments.
     const parentDirs = new Set<string>();
     for (const root of config.roots) {
       parentDirs.add(dirname(resolve(root.path)));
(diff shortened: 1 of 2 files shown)


server/chat.ts
```diff
@@ -722,6 +730,29 @@ export function getSessionDirs(): string[] {
...
   }
 
+  // Discover worktree sessions from ~/.claude/projects/.
+  // When a session runs in a worktree CWD, the SDK stores its data under
+  // a project dir derived from that path. Once the worktree is cleaned up,
+  // listSessions({ dir: BASE_REPO }) can no longer find those sessions.
+  // We scan for project dirs matching <encoded-BASE_REPO>--claude-worktrees-<id>
+  // and reconstruct the original worktree path.
+  const encodedBase = BASE_REPO.replace(/\//g, '-');
+  const wtPrefix = `${encodedBase}--claude-worktrees-`;
```
dimakis, line 740: 🟡 **unsafe_assumptions**: The worktree prefix pattern `<encoded-BASE_REPO>--claude-worktrees-<id>` assumes a specific SDK encoding scheme for `~/.claude/projects/` directory names. If the SDK changes its path encoding (e.g., encoding `.` characters, or using a different separator), this will silently stop discovering worktree sessions. Consider adding a comment citing the SDK source or version that defines this encoding, so future maintainers can verify it still holds.

server/chat.ts
```diff
@@ -722,6 +730,29 @@ export function getSessionDirs(): string[] {
     // Expected when sessions dir doesn't exist yet
   }
 
+  // Discover worktree sessions from ~/.claude/projects/.
+  // When a session runs in a worktree CWD, the SDK stores its data under
+  // a project dir derived from that path. Once the worktree is cleaned up,
+  // listSessions({ dir: BASE_REPO }) can no longer find those sessions.
+  // We scan for project dirs matching <encoded-BASE_REPO>--claude-worktrees-<id>
+  // and reconstruct the original worktree path.
+  const encodedBase = BASE_REPO.replace(/\//g, '-');
```
dimakis, line 739: 🟡 **bugs**: The encoding `BASE_REPO.replace(/\//g, '-')` only replaces forward slashes. On systems where BASE_REPO contains other special characters (spaces, dots beyond `.claude`), the reconstructed path may not match what the SDK actually generates. This is fragile — if the SDK's actual encoding differs (e.g., also encodes dots or uses percent-encoding), worktree sessions won't be found.

server/chat.ts
```diff
@@ -669,6 +670,15 @@ export function isActive(clientId: string): boolean {
  *
  * This keeps all paths configurable (no hardcoded machine-specific paths).
  */
+let _claudeProjectsRoot: string | null = null;
```
dimakis, line 673: 🔵 **style**: The module-level mutable `_claudeProjectsRoot` variable and its setter/getter are a testing seam that could be avoided by passing the projects root as a parameter to `getSessionDirs()` (with a default). This would be more explicit and avoid shared mutable state. `[fixable]`

dimakis (commented):
## Centaur Review

_New worktree discovery logic looks correct, but the PR deletes 6 tests for unchanged sibling-discovery code — a significant test coverage regression that should be restored._

- 🔴 **regressions**: Six tests covering sibling-discovery edge cases were deleted, but the implementation code they tested (chat.ts lines 697-717) was not removed. Tests for '.claude marker discovery', 'skips non-directory entries', 'skips dirs without .git/.claude', 'deduplication across repos and root siblings', 'missing parent dirs', and 'multiple roots sharing same parent' no longer exist. This is a significant regression in test coverage for unchanged functionality. `[fixable]`


### `server/chat.ts`
_New worktree discovery logic looks correct, but the PR deletes 6 tests for unchanged sibling-discovery code — a significant test coverage regression that should be restored._

- 🟡 **unsafe_assumptions** (L740): The worktree prefix pattern `<encoded-BASE_REPO>--claude-worktrees-<id>` assumes a specific SDK encoding scheme for `~/.claude/projects/` directory names. If the SDK changes its path encoding (e.g., encoding `.` characters, or using a different separator), this will silently stop discovering worktree sessions. Consider adding a comment citing the SDK source or version that defines this encoding, so future maintainers can verify it still holds.
- 🟡 **bugs** (L739): The encoding `BASE_REPO.replace(/\//g, '-')` only replaces forward slashes. On systems where BASE_REPO contains other special characters (spaces, dots beyond `.claude`), the reconstructed path may not match what the SDK actually generates. This is fragile — if the SDK's actual encoding differs (e.g., also encodes dots or uses percent-encoding), worktree sessions won't be found.
- 🔵 **style** (L673): The module-level mutable `_claudeProjectsRoot` variable and its setter/getter are a testing seam that could be avoided by passing the projects root as a parameter to `getSessionDirs()` (with a default). This would be more explicit and avoid shared mutable state. `[fixable]`

PR #205, opened 2026-04-15
feat(server): decouple WebSocket from session transport

dimakis:
## Summary

- Introduces `WsTransport` adapter — the single boundary where WebSocket-specific code lives
- Replaces all raw `WebSocket` usage in server files with `SessionTransport` from `@mitzo/harness`
- Moves permission/mode message handling from `createWsMessageHandler` (removed) to `index.ts` dispatch
- Adds `transportMap` in `index.ts` for observer lifecycle tracking
- Updates all 8 server test files from WebSocket mocks to SessionTransport mocks
- 13 new WsTransport unit tests covering all socket states and edge cases

Resolves 31 TypeScript nominal typing errors caused by dual `SessionRegistry` implementations (server vs harness) with incompatible `WebSocket`/`SessionTransport` types.

## Files changed

| File | Change |
|---|---|
| `server/ws-transport.ts` | **NEW** — WsTransport adapter |
| `server/index.ts` | Transport wrapping at boundary, permission/mode handling |
| `server/chat.ts` | `startChat`/`reattachChat` accept `SessionTransport` |
| `server/query-loop.ts` | `runQueryLoop` drops `_ws` param, `createWsMessageHandler` removed |
| `server/app.ts` | `session.ws` → `session.transport` |
| `server/session-registry.ts` | Re-export shim from `@mitzo/harness` |
| `server/permission-handler.ts` | Re-export shim from `@mitzo/harness` |
| 8 test files + 1 new test | WebSocket mocks → SessionTransport mocks |

## Test plan

- [x] `tsc --noEmit` — 0 type errors
- [x] `npm run lint` — 0 errors (14 pre-existing warnings)
- [x] 223 tests passing across 15 affected files
- [x] `npm run build` — frontend build succeeds
- [ ] Centaur review


dimakis:
## Centaur Review

### `server/app.ts`
_Clean transport abstraction with one significant gap: many raw ws.send() calls in index.ts were not converted to use the transport, partially undermining the decoupling goal._

- 🟡 **bugs** (L312): session.transport.send() is called without an isOpen() guard, unlike every other call site that uses the local `send()` helper which checks isOpen() first. If the transport is closed when a worktree is opened via the REST API, this will silently succeed (WsTransport.send checks internally), but the inconsistency with the rest of the codebase is fragile. If a non-WsTransport implementation doesn't guard internally, this would throw. `[fixable]`


### `server/index.ts`
_Clean transport abstraction with one significant gap: many raw ws.send() calls in index.ts were not converted to use the transport, partially undermining the decoupling goal._

- 🟡 **bugs** (L203): tryRouteToActiveSession sends the snapshot directly via raw ws.send(JSON.stringify(...)) instead of using the transport wrapper. This bypasses the transport abstraction — the whole point of this PR. Same issue at lines 224, 228, 254, 262, 277, 290, etc. in handleChatWs. Many ws.send() calls in index.ts were not converted to use the transport, creating an inconsistent mix of raw WebSocket and transport usage. `[fixable]`
- 🟡 **unsafe_assumptions** (L166): transportMap is a module-level Map<WebSocket, WsTransport> that only cleans up entries on ws 'close'. If a WebSocket is never properly closed (e.g. network drop without close frame), the entry will leak. Not critical since server restarts clear it, but worth noting as a potential slow memory leak under abnormal disconnection patterns.
- 🔵 **missing_tests** (L444): The permission_response and set_mode message handling was moved from createWsMessageHandler (query-loop.ts) to the main ws.on('message') handler in index.ts. The old handler was tested implicitly via the query loop tests. The new location in index.ts has no dedicated test coverage for these message types.


### `server/query-loop.ts`
_Clean transport abstraction with one significant gap: many raw ws.send() calls in index.ts were not converted to use the transport, partially undermining the decoupling goal._

- 🟡 **regressions** (L228): Several `emit()` calls have a stray leading space before the argument: `emit( { type: ... })`. While not a runtime bug (JS ignores whitespace), this is a formatting issue that should have been caught by Prettier, and it appears in multiple places (lines 228, 262, 294, 343, 365, 384, 399). `[fixable]`


_Clean transport abstraction with one significant gap: many raw ws.send() calls in index.ts were not converted to use the transport, partially undermining the decoupling goal._

- 🔵 **style** (L7): The helper is still named `fakeWs()` despite creating a SessionTransport, not a WebSocket. Consider renaming to `fakeTransport()` for consistency with other test files (e.g. multi-client-sessions.test.ts uses `mockTransport()`). `[fixable]`


### `server/__tests__/token-update.test.ts`
_Clean transport abstraction with one significant gap: many raw ws.send() calls in index.ts were not converted to use the transport, partially undermining the decoupling goal._

- 🔵 **style** (L7): Same naming issue — `fakeWs()` creates a SessionTransport, not a WebSocket. Should be renamed to `fakeTransport()` for consistency. `[fixable]`

dimakis (commented):
## Codex Review

### `server/index.ts`
_Good transport abstraction overall, but moving inbound WS handling out of `startChat()` appears to break permission responses from observer/reattached clients for active sessions._

- 🟡 **regressions**: This PR removes `createWsMessageHandler` from `startChat()` and moves handling of `permission_response` / `set_mode` into `handleChatWs`, but `tryRouteToActiveSession()` still registers a socket only as an observer for resumed active sessions. Observer clients can send follow-up messages via `sendToChat`, but their `permission_response` messages now resolve against their own observer connection rather than the driver session, so permission prompts shown to an observer can no longer be answered unless the original driver socket is still connected and receives the response. This is a behavior regression for multi-client session control. `[fixable]`

server/index.ts
```diff
@@ -421,6 +423,14 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
             });
           }
         }
```
dimakis, line 425: 🟡 **regressions**: `permission_response` and `set_mode` handling was moved out of `createWsMessageHandler` into the top-level WS router, but `handleChatWs` only installs this router on the primary connection. Observer sockets created via `subscribe`/`tryRouteToActiveSession` can still receive `permission_request` events, yet their replies are no longer consumed anywhere unless they happen to be the driver socket. That breaks the existing multi-client permission flow where any attached frontend could approve a pending tool request. `[fixable]`

dimakis (commented):
## Codex Review

### `server/index.ts`
_Transport abstraction looks solid overall, but moving permission/mode message handling out of the per-session socket path appears to regress observer-driven permission responses._

- 🟡 **regressions** (L425): `permission_response` and `set_mode` handling was moved out of `createWsMessageHandler` into the top-level WS router, but `handleChatWs` only installs this router on the primary connection. Observer sockets created via `subscribe`/`tryRouteToActiveSession` can still receive `permission_request` events, yet their replies are no longer consumed anywhere unless they happen to be the driver socket. That breaks the existing multi-client permission flow where any attached frontend could approve a pending tool request. `[fixable]`

PR #207, opened 2026-04-16
feat: Phase 5 — migrate frontend hooks to @mitzo/client store

dimakis:
## Summary

- **Fills store gaps** in `@mitzo/client`: tokens slice, `interruptMessage` action, enriched `sendMessage` (model/mode/resume/cwd/extraTools), `fetchSessionMeta`, `useTokens` hook
- **Migrates ChatView + DesktopChatView** from 5 legacy hooks (`useChatMessages`, `useChatConnection`, `useChatSession`, `useChatActions`, `usePermission`) to `@mitzo/client/hooks` (`useMessages`, `useConnection`, `useTokens`, `useMitzoStore`)
- **Deletes 7 old hooks + 8 test files** (~2,900 lines removed, 94 tests covered by client package equivalents)

Net: -2,700 lines. 120 test files, 1,567 tests pass. Both tsc checks clean.

`ws-pool.ts` retained — still used by `useTaskBoard`, `useTabBadges`, `InboxView` (future scope).

## Test plan

- [x] `npx vitest run` — 120 files, 1,567 tests pass
- [x] `npx tsc --noEmit` — clean for both `frontend/` and `packages/client/`
- [ ] Manual test: chat send/receive, session switching, new session, voice, stop button
- [ ] Manual test via Tailscale at https://100.91.50.57:3100/


dimakis:
## Centaur Review

_Migration deletes ~2800 lines of tested hook code without equivalent store-level tests; session rename display is regressed; overall structure is sound but test coverage gap is significant._

- 🟡 **regressions** (L361): `onSessionRenamed` callback is not wired in store callbacks. The protocol parser calls `callbacks.onSessionRenamed?.(msg.name)` on `session_renamed` events, but the store doesn't implement it. The old code used this to set `document.title = name + ' — Mitzo'`. Session rename display is silently lost. `[fixable]`
- 🟡 **missing_tests**: This PR deletes 7 frontend hook test files (~2800 lines of test code) but adds no replacement store-level tests. The new store actions (`sendMessage`, `interruptMessage`, `stopGeneration`, `fetchSessionMeta`) and `token_update` protocol handling have zero test coverage. The deleted tests covered critical flows: send payload construction (resume, cwd, extraTools, images), interrupt guards, stop behavior, deduplication, reattach recovery, and token state lifecycle. `[fixable]`


_Migration deletes ~2800 lines of tested hook code without equivalent store-level tests; session rename display is regressed; overall structure is sound but test coverage gap is significant._

- 🟡 **bugs** (L16): Both `transport.connectWs` and `wsConfig.createWebSocket` create `new WebSocket(url)` but only `wsConfig` is used by `WsPool`. The `transport.connectWs` adapter is dead code — `WsPool` uses `wsConfig.createWebSocket` exclusively. This is confusing but not harmful. `[fixable]`


_Migration deletes ~2800 lines of tested hook code without equivalent store-level tests; session rename display is regressed; overall structure is sound but test coverage gap is significant._

- 🟡 **regressions** (L125): The old DesktopChatView dispatched `CLEAR` before restoring messages on session switch (`dispatch({ type: 'CLEAR' })`). The new code relies on `storeSwitchSession` resetting state, but the message restore `useEffect` (line 125) no longer explicitly clears first. If `switchSession` and the fetch effect fire in the wrong order (e.g., same sessionId re-selected), stale messages may persist. `[fixable]`


_Migration deletes ~2800 lines of tested hook code without equivalent store-level tests; session rename display is regressed; overall structure is sound but test coverage gap is significant._

- 🔵 **unsafe_assumptions** (L60): The `useEffect` syncing route param to store (lines 60-65) has `sessionId` as the sole dependency but uses `activeSessionId` and store actions in the body. The eslint-disable comment acknowledges this. If `activeSessionId` updates trigger a re-render but `sessionId` hasn't changed, the effect won't fire — this is intentional but fragile. Consider using a ref for `activeSessionId` comparison instead of the eslint-disable. `[fixable]`
- 🔵 **style** (L30): 10 individual `useMitzoStore` selector calls (lines 30-39) could be consolidated into fewer selectors or a single structured selector to reduce boilerplate. The same pattern is duplicated in DesktopChatView. `[fixable]`


_Migration deletes ~2800 lines of tested hook code without equivalent store-level tests; session rename display is regressed; overall structure is sound but test coverage gap is significant._

- 🔵 **missing_tests** (L323): The new `token_update` case in the protocol parser (lines 323-329) has no corresponding test in protocol-parser.test.ts. All other message types in the parser have test coverage. `[fixable]`

frontend/src/client-store.ts
```diff
@@ -0,0 +1,42 @@
...
+ * once the frontend migration is complete.
+ */
+
+import { createMitzoStore } from '@mitzo/client';
+
+const proto = location.protocol === 'https:' ? 'wss' : 'ws';
+const wsUrl = `${proto}://${location.host}/ws/chat`;
+
+export const clientStore = createMitzoStore({
+  transport: {
```
dimakis, line 16: 🔵 **style**: `transport.connectWs` is defined but never called by the store — `createMitzoStore` uses `options.wsConfig` to construct the `WsPool` and `options.transport.fetch` for the API client. The `connectWs` transport method is dead code that adds confusion. `[fixable]`

dimakis (commented):
## Centaur Review

_Migration replaces 6 hooks with a centralized Zustand store. Core wiring works, but there's a double-fetch bug on session switch, session-expired navigation is lost, session rename handling is unwired, and ~2600 lines of deleted tests have no store-level replacements._

- 🔴 **bugs**: Double message fetch on session navigation: `storeSwitchSession(sessionId)` already calls `api.getSessionMessages()` and dispatches RESTORE, but a separate useEffect also fetches `/api/sessions/${sessionId}/messages` and dispatches another RESTORE. This causes two API calls and a potential race where the second RESTORE overwrites the first with stale or duplicate data. The same issue exists in DesktopChatView.tsx. Remove the manual fetch/RESTORE effect since `switchSession` already handles it. `[fixable]`
- 🟡 **regressions**: The old `handleSessionExpired` cleared `currentSessionId`, removed the localStorage key, and navigated to `/chat` so the user could start fresh. The store's `onSessionExpired` callback resets `sessions.active` to null but doesn't navigate or clear localStorage. The ChatView/DesktopChatView no longer have any session-expiry navigation logic, so an expired session leaves the user on a dead `/chat/:id` URL with no recovery path. `[fixable]`
- 🔵 **style**: `handleSend`, `handleInterrupt`, `handlePermission`, and `handleModeChange` are near-identical copies between ChatView and DesktopChatView. Consider extracting a shared hook (e.g., `useChatViewActions`) to avoid the duplication, which will inevitably drift. `[fixable]`
- 🔵 **unsafe_assumptions**: The `useEffect` syncing route → store (`storeSwitchSession`/`storeNewSession`) has `// eslint-disable-line react-hooks/exhaustive-deps` suppressing missing deps (`activeSessionId`, `storeSwitchSession`, `storeNewSession`). If these action references change (unlikely with Zustand but possible), the effect could use stale closures. Consider listing them explicitly or documenting why suppression is safe. `[fixable]`


_Migration replaces 6 hooks with a centralized Zustand store. Core wiring works, but there's a double-fetch bug on session switch, session-expired navigation is lost, session rename handling is unwired, and ~2600 lines of deleted tests have no store-level replacements._

- 🟡 **regressions**: `onSessionRenamed` callback is defined in `ProtocolCallbacks` and the protocol parser dispatches it on `session_renamed` events, but the store's `callbacks` object (line ~361) never provides an implementation. The old `useChatMessages` hook set `document.title` on rename; that behavior is now silently lost. `[fixable]`
- 🟡 **bugs**: `interruptMessage` calls `wsPool.send()` without checking its return value. If the WS is disconnected, the message is silently dropped but the optimistic USER_SEND is still dispatched to state, creating a ghost message the server never received. `sendMessage` correctly checks the return and sets `sendError`. `[fixable]`
- 🟡 **missing_tests**: This PR deletes ~2600 lines of hook tests (chatMessagesReducer, useChatMessages, useChatActions, useChatConnection, useChatSession, usePermission, useSessionMeta, useTokenState) but adds zero store-level integration tests for the replacement logic: `sendMessage` with opts, `interruptMessage`, `fetchSessionMeta`, `stopGeneration`, `respondToPermission`, or the new `tokensUpdate` path in the protocol parser. The messages-slice reducer tests in the client package cover the reducer, but the store action wiring is untested. `[fixable]`


_Migration replaces 6 hooks with a centralized Zustand store. Core wiring works, but there's a double-fetch bug on session switch, session-expired navigation is lost, session rename handling is unwired, and ~2600 lines of deleted tests have no store-level replacements._

- 🔵 **style** (L16): `transport.connectWs` is defined but never called by the store — `createMitzoStore` uses `options.wsConfig` to construct the `WsPool` and `options.transport.fetch` for the API client. The `connectWs` transport method is dead code that adds confusion. `[fixable]`

dimakis:
## Centaur Review

Found **10** issue(s) (1 critical) (7 warning).


_Critical bug: stopGeneration sends wrong WS message type ('interrupt' instead of 'stop'), breaking the stop button. Several regressions in session lifecycle handling (expiry, rename, reattach recovery) and large test coverage gap from deleting ~2400 lines of tests without replacement._

- 🔴 **bugs** (L291): stopGeneration() sends `{ type: 'interrupt' }` but the server's InterruptMessage schema requires `prompt` (string, min 1) and `clientMsgId` (string, min 1). This will fail Zod validation server-side. The old code correctly sent `{ type: 'stop' }`. This means the stop button is completely broken. `[fixable]`
- 🟡 **missing_tests**: ~2400 lines of hook tests were deleted (chatMessagesReducer, useChatMessages, useChatSession, useChatConnection, useChatActions, usePermission, useSessionMeta, useTokenState) but no store integration tests were added for sendMessage, interruptMessage, stopGeneration, fetchSessionMeta, or the WS→store wiring. The underlying reducer and parser have tests, but the store action layer that composes them is untested. `[fixable]`


_Critical bug: stopGeneration sends wrong WS message type ('interrupt' instead of 'stop'), breaking the stop button. Several regressions in session lifecycle handling (expiry, rename, reattach recovery) and large test coverage gap from deleting ~2400 lines of tests without replacement._

- 🟡 **regressions** (L145): handleStop calls storeStopGeneration() (which sends 'interrupt') but also dispatches SET_RUNNING(false). The old handleStop also called `pendingSend.current = null` to cancel queued messages and `wsSetRunning(poolKey, false)` to clear the pool's running flag. The store's stopGeneration() does neither — queued messages may still fire after stop, and pool reattach logic won't know the session stopped. `[fixable]`
- 🟡 **regressions**: Session expiry handling is lost. The old code navigated to /chat, cleared localStorage, and reset currentSessionId when 'No conversation found' error arrived. The new views have no equivalent — the user stays on the expired session page with no way to recover without manual navigation. `[fixable]`
- 🟡 **regressions**: document.title update on session_renamed is lost. The old code set `document.title = name + ' — Mitzo'` via handleSessionRenamed callback. The store's protocol callbacks don't wire onSessionRenamed to any title-setting logic, so the browser tab title never updates. `[fixable]`
- 🟡 **unsafe_assumptions** (L63): Race condition between switchSession (in store, which fetches + restores messages) and the component's own useEffect that also fetches + restores messages for the same sessionId. Both run on sessionId change, resulting in duplicate API calls and potential state clobbering if responses arrive out of order. `[fixable]`
- 🔵 **style** (L150): handlePermission accepts `_toolName` parameter (unused) to match the old callback signature. The store's respondToPermission correctly omits toolName (server doesn't need it). Consider updating the ChatArea/PermissionBanner callback type to drop toolName rather than carrying a dead parameter. `[fixable]`


_Critical bug: stopGeneration sends wrong WS message type ('interrupt' instead of 'stop'), breaking the stop button. Several regressions in session lifecycle handling (expiry, rename, reattach recovery) and large test coverage gap from deleting ~2400 lines of tests without replacement._

- 🟡 **regressions** (L96): reattach_failed recovery doesn't pass `interrupted: true` to RESTORE. The old code dispatched `RESTORE` with `interrupted: true`, which appended a 'Session interrupted' notice message. The new flow (fetchMessages → discard result → onMessagesRestored → re-fetch → RESTORE without interrupted flag) loses this UX indicator and also double-fetches messages unnecessarily. `[fixable]`


_Critical bug: stopGeneration sends wrong WS message type ('interrupt' instead of 'stop'), breaking the stop button. Several regressions in session lifecycle handling (expiry, rename, reattach recovery) and large test coverage gap from deleting ~2400 lines of tests without replacement._

- 🟡 **regressions**: The old DesktopChatView dispatched `{ type: 'CLEAR' }` before fetching messages on session switch, ensuring stale messages from the previous session were immediately cleared. The new code relies on switchSession's reset, but the component's independent fetch doesn't clear first — if the fetch is slow, messages from the previous session may briefly appear. `[fixable]`


_Critical bug: stopGeneration sends wrong WS message type ('interrupt' instead of 'stop'), breaking the stop button. Several regressions in session lifecycle handling (expiry, rename, reattach recovery) and large test coverage gap from deleting ~2400 lines of tests without replacement._

- 🔵 **style** (L17): Both `transport.connectWs` and `wsConfig.createWebSocket` create WebSocket instances in the same way. This is confusing redundancy — it's unclear which one is actually used for connections. Consider removing the unused one or documenting why both exist. `[fixable]`

PR #208, opened 2026-04-16
feat(client): add loadSessions + refreshSessions store actions

dimakis:
## Summary

- Add `loadSessions(offset?)` action: sets loading state, fetches from API, replaces list
- Add `refreshSessions()` action: silently updates list without loading indicator (used after rename/delete)

Consumed by mitzo-theia Phase 1 (Sessions widget) — the first standalone widget built on the store pattern.

## Test plan

- [ ] `npm test` passes
- [ ] `loadSessions()` populates `sessions.list` from `api.listSessions()`
- [ ] `refreshSessions()` updates list without setting `loading: true`
- [ ] Error paths: failed fetch leaves existing list intact

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -61,7 +61,7 @@ export interface MitzoStoreState {
   respondToPermission(permId: string, decision: 'once' | 'always' | 'deny'): void;
   setMode(mode: MitzoMode): void;
   setModel(modelId: string): void;
-  loadSessions(offset?: number): Promise<void>;
+  loadSessions(): Promise<void>;
   refreshSessions(): Promise<void>;
 }
 
@@ -282,11 +282,13 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       }));
     },
 
-    async loadSessions(offset?: number) {
+    async loadSessions() {
       set((s) => ({ sessions: { ...s.sessions, loading: true } }));
       try {
-        const raw = await api.listSessions(offset);
-        const list = Array.isArray(raw) ? raw : (raw as unknown as { sessions: typeof raw }).sessions ?? [];
+        const raw = await api.listSessions();
+        const list = Array.isArray(raw)
+          ? raw
+          : ((raw as unknown as { sessions: typeof raw }).sessions ?? []);
         set((s) => ({ sessions: { ...s.sessions, list, loading: false } }));
       } catch {
         set((s) => ({ sessions: { ...s.sessions, loading: false } }));
@@ -296,7 +298,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     async refreshSessions() {
       try {
         const raw = await api.listSessions();
-        const list = Array.isArray(raw) ? raw : (raw as unknown as { sessions: typeof raw }).sessions ?? [];
+        const list = Array.isArray(raw)
+          ? raw
+          : ((raw as unknown as { sessions: typeof raw }).sessions ?? []);
         set((s) => ({ sessions: { ...s.sessions, list } }));
       } catch {
         // Silent — keep existing list on failure
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -61,6 +61,8 @@ export interface MitzoStoreState {
   respondToPermission(permId: string, decision: 'once' | 'always' | 'deny'): void;
   setMode(mode: MitzoMode): void;
   setModel(modelId: string): void;
+  loadSessions(offset?: number): Promise<void>;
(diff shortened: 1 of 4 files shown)


packages/client/src/store.ts
```diff
@@ -281,6 +283,25 @@ export function createMitzoStore(
         config: { ...s.config, modelId },
       }));
     },
+
+    async loadSessions(offset?: number) {
```
dimakis: 🟡 **missing_tests**: No tests for loadSessions or refreshSessions. The project requires TDD (CLAUDE.md) and there's no store test file at all. At minimum: test that loadSessions sets loading=true then false, replaces list on success, and preserves state on failure; test that refreshSessions silently keeps existing list on failure. `[fixable]`

dimakis: 🔵 **bugs**: loadSessions replaces the entire list with the API response. If `offset` is provided (implying pagination), this overwrites previously loaded sessions instead of appending. Either the offset param is unused/premature, or the implementation should merge rather than replace. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean, minimal additions. Main concern is a potential pagination semantics mismatch (offset replaces instead of appending) and missing tests per project TDD policy._

- 🟡 **missing_tests** (L287): No tests for loadSessions or refreshSessions. The project requires TDD (CLAUDE.md) and there's no store test file at all. At minimum: test that loadSessions sets loading=true then false, replaces list on success, and preserves state on failure; test that refreshSessions silently keeps existing list on failure. `[fixable]`
- 🔵 **bugs** (L287): loadSessions replaces the entire list with the API response. If `offset` is provided (implying pagination), this overwrites previously loaded sessions instead of appending. Either the offset param is unused/premature, or the implementation should merge rather than replace. `[fixable]`

packages/client/src/store.ts
```diff
@@ -281,6 +283,25 @@ export function createMitzoStore(
...
+      set((s) => ({ sessions: { ...s.sessions, loading: true } }));
+      try {
+        const list = await api.listSessions(offset);
```
dimakis: 🟡 **bugs**: loadSessions(offset) replaces the entire list with the paginated result instead of appending. If offset-based pagination is intended to load more sessions, this discards previously loaded items. If it's meant to replace the view (e.g. page-based), this is fine — but the API name 'loadSessions' with an offset parameter suggests append semantics. `[fixable]`

packages/client/src/store.ts
dimakis: 🟡 **missing_tests**: No tests for loadSessions or refreshSessions. There is no store test file at all (only api-client, messages-slice, protocol-parser, and ws-connection tests exist). Key behaviors to test: loading flag toggling, list replacement on success, loading reset on error, and the two callsites in the WS wiring (onSessionAssigned, inboxRefresh). `[fixable]`

packages/client/src/store.ts
```diff
@@ -294,6 +315,7 @@ export function createMitzoStore(
       store.setState((s) => ({
         sessions: { ...s.sessions, active: sessionId },
       }));
+      store.getState().refreshSessions();
```
dimakis, line 322: 🔵 **unsafe_assumptions**: refreshSessions() is fire-and-forget in onSessionAssigned (line 318) and inboxRefresh (line 407). If the API is slow, rapid session assignments or inbox refreshes could cause out-of-order list updates (last-write-wins race). Low risk in practice since refreshSessions replaces the full list, but worth noting.

dimakis (commented):
## Centaur Review

_Clean, focused addition. Main concerns: pagination semantics (replace vs append) may be a bug depending on intent, and no test coverage for the new actions._

- 🟡 **bugs** (L290): loadSessions(offset) replaces the entire list with the paginated result instead of appending. If offset-based pagination is intended to load more sessions, this discards previously loaded items. If it's meant to replace the view (e.g. page-based), this is fine — but the API name 'loadSessions' with an offset parameter suggests append semantics. `[fixable]`
- 🟡 **missing_tests** (L287): No tests for loadSessions or refreshSessions. There is no store test file at all (only api-client, messages-slice, protocol-parser, and ws-connection tests exist). Key behaviors to test: loading flag toggling, list replacement on success, loading reset on error, and the two callsites in the WS wiring (onSessionAssigned, inboxRefresh). `[fixable]`
- 🔵 **unsafe_assumptions** (L318): refreshSessions() is fire-and-forget in onSessionAssigned (line 318) and inboxRefresh (line 407). If the API is slow, rapid session assignments or inbox refreshes could cause out-of-order list updates (last-write-wins race). Low risk in practice since refreshSessions replaces the full list, but worth noting.

PR #209, opened 2026-04-16
fix(client): eliminate double-fetch on reattach + store tests

dimakis:
## Summary
- **Fix double-fetch in reattach recovery**: `onMessagesRestored` now receives `FinishedMessage[]` directly from the protocol parser instead of re-fetching via `getSessionMessages()`, eliminating a redundant API call (10 lines → 3 lines in store.ts)
- **Store integration tests**: 18 new tests covering `createMitzoStore`, `switchSession`, `newSession`, `sendMessage`, WS→store wiring, `setMode`, `setModel`, and `dispatchMessages`
- **ESLint config**: extends `argsIgnorePattern: '^_'` to `packages/`

All 150 `packages/client/` tests pass.

## Test plan
- [x] `npx vitest run packages/client` — 150 tests pass
- [x] Pre-commit hooks (eslint + prettier) pass

### DIFF
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -64,6 +64,49 @@ describe('reattach', () => {
     expect(r.messagesActions).toEqual([{ type: 'SET_RUNNING', running: false }]);
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, false);
   });
+
+  it('reattach_failed calls onMessagesRestored with fetched messages', async () => {
+    const msgs = [
+      {
+        messageId: 'm1',
+        role: 'assistant' as const,
+        blocks: [{ blockId: 'b1', blockType: 'text' as const, content: 'hello' }],
+      },
+    ];
+    const fetchMessages = vi.fn().mockResolvedValue(msgs);
+    const onMessagesRestored = vi.fn();
+    const cb = makeCallbacks({ fetchMessages, onMessagesRestored });
+    const state = makeState({ currentSessionId: 'sid-1' });
+
+    parseServerMessage({ type: 'reattach_failed', clientId: 'old' }, state, cb, POOL_KEY);
+
+    // Wait for the async fetchMessages to resolve
+    await vi.waitFor(() => {
+      expect(onMessagesRestored).toHaveBeenCalledWith(msgs);
+    });
+  });
+
+  it('reattach_failed does not call onMessagesRestored when fetch returns empty', async () => {
+    const fetchMessages = vi.fn().mockResolvedValue([]);
+    const onMessagesRestored = vi.fn();
+    const cb = makeCallbacks({ fetchMessages, onMessagesRestored });
+    const state = makeState({ currentSessionId: 'sid-1' });
+
+    parseServerMessage({ type: 'reattach_failed', clientId: 'old' }, state, cb, POOL_KEY);
+
+    // Give async a chance to resolve
+    await new Promise((r) => setTimeout(r, 10));
+    expect(onMessagesRestored).not.toHaveBeenCalled();
+  });
+
+  it('reattach_failed skips fetch when no currentSessionId', () => {
+    const fetchMessages = vi.fn();
+    const cb = makeCallbacks({ fetchMessages });
+
+    parseServerMessage({ type: 'reattach_failed', clientId: 'old' }, makeState(), cb, POOL_KEY);
+
+    expect(fetchMessages).not.toHaveBeenCalled();
+  });
 });
 
 // ─── Session lifecycle ────────────────────────────────────────────────────────
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -94,8 +94,10 @@ describe('reattach', () => {
 
     parseServerMessage({ type: 'reattach_failed', clientId: 'old' }, state, cb, POOL_KEY);
 
-    // Give async a chance to resolve
-    await new Promise((r) => setTimeout(r, 10));
+    // Flush the microtask queue so the mockResolvedValue resolves
+    await vi.waitFor(() => {
+      expect(fetchMessages).toHaveBeenCalled();
+    });
     expect(onMessagesRestored).not.toHaveBeenCalled();
   });
 
--- a/packages/client/__tests__/store.test.ts
+++ b/packages/client/__tests__/store.test.ts
@@ -1,0 +1,348 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { createMitzoStore } from '../src/store.js';
+import type { MitzoStoreOptions, MitzoStoreState } from '../src/store.js';
+import type { TransportAdapter } from '../src/types.js';
+import type { WebSocketLike } from '../src/ws-connection.js';
+import { WS_READY_STATE } from '../src/types.js';
+import type { StoreApi } from 'zustand/vanilla';
+
+// ─── Mock WebSocket ─────────────────────────────────────────────────────────
+
+class MockWebSocket implements WebSocketLike {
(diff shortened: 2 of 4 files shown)


packages/client/src/protocol-parser.ts
```diff
@@ -95,11 +102,12 @@ export function parseServerMessage(
       if (state.currentSessionId && callbacks.fetchMessages) {
         const sessionId = state.currentSessionId;
         const controller = new AbortController();
```
dimakis: 🟡 **bugs**: The `AbortController` created in the `reattach_failed` handler is never aborted — the `controller.signal.aborted` check on line 108 can never be true. This is dead code. If cancellation is intended (e.g., on component unmount or subsequent reattach), the controller reference would need to be stored and aborted externally. If cancellation isn't needed, the controller/signal can be removed to reduce confusion. `[fixable]`

packages/client/__tests__/protocol-parser.test.ts
```diff
@@ -65,6 +64,49 @@ describe('reattach', () => {
...
+
+    parseServerMessage({ type: 'reattach_failed', clientId: 'old' }, state, cb, POOL_KEY);
+
+    // Wait for the async fetchMessages to resolve
+    await vi.waitFor(() => {
+      expect(onMessagesRestored).toHaveBeenCalledWith(msgs);
+    });
+  });
+
+  it('reattach_failed does not call onMessagesRestored when fetch returns empty', async () => {
```
dimakis, line 89: 🔵 **style**: The `reattach_failed does not call onMessagesRestored when fetch returns empty` test uses `await new Promise((r) => setTimeout(r, 10))` for timing, while the positive case above uses `vi.waitFor()`. Using `vi.waitFor` with a negative assertion or `vi.advanceTimersByTime` with fake timers would be more deterministic and consistent. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean fix that correctly eliminates the double-fetch by passing restored messages directly through the callback. The unreachable AbortController abort-check is a pre-existing minor issue. Consider adding a store-level integration test for the reattach_failed flow._

- 🟡 **bugs** (L104): The `AbortController` created in the `reattach_failed` handler is never aborted — the `controller.signal.aborted` check on line 108 can never be true. This is dead code. If cancellation is intended (e.g., on component unmount or subsequent reattach), the controller reference would need to be stored and aborted externally. If cancellation isn't needed, the controller/signal can be removed to reduce confusion. `[fixable]`


_Clean fix that correctly eliminates the double-fetch by passing restored messages directly through the callback. The unreachable AbortController abort-check is a pre-existing minor issue. Consider adding a store-level integration test for the reattach_failed flow._

- 🔵 **missing_tests**: The store test file doesn't cover the `reattach_failed` → `onMessagesRestored` flow that was the core bug fix in this PR (eliminating the double-fetch). A test that simulates a `reattach_failed` WS message and verifies messages are restored exactly once via the passed-through messages (not re-fetched) would directly validate the fix. `[fixable]`
- 🔵 **missing_tests**: No test for `stopGeneration` or `respondToPermission` actions. These are lower priority since they weren't changed in this PR, but they are part of the public API being tested in this new file. `[fixable]`


_Clean fix that correctly eliminates the double-fetch by passing restored messages directly through the callback. The unreachable AbortController abort-check is a pre-existing minor issue. Consider adding a store-level integration test for the reattach_failed flow._

- 🔵 **style** (L89): The `reattach_failed does not call onMessagesRestored when fetch returns empty` test uses `await new Promise((r) => setTimeout(r, 10))` for timing, while the positive case above uses `vi.waitFor()`. Using `vi.waitFor` with a negative assertion or `vi.advanceTimersByTime` with fake timers would be more deterministic and consistent. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean elimination of the double-fetch pattern. The main concern is losing the AbortController race guard — two rapid reattach_failed events can now race without cancellation. Otherwise solid: good test coverage, correct wiring, and formatting improvements._

- 🟡 **regressions** (L104): Removing the AbortController means in-flight fetches from a previous reattach_failed can no longer be cancelled. If two reattach_failed events fire in quick succession (e.g. flaky reconnect), both fetches race and the second onMessagesRestored call overwrites state with potentially stale data. The old AbortController prevented this. Consider guarding with a monotonic counter or generation ID instead. `[fixable]`
- 🔵 **style** (L108): The `.catch(() => {})` silently swallows all fetch errors during reattach recovery. A debug-level log would help diagnose connectivity issues without breaking the flow. `[fixable]`


_Clean elimination of the double-fetch pattern. The main concern is losing the AbortController race guard — two rapid reattach_failed events can now race without cancellation. Otherwise solid: good test coverage, correct wiring, and formatting improvements._

- 🔵 **missing_tests** (L280): The reattach_failed store test verifies messages are restored once, but doesn't test the race scenario: two rapid reattach_failed events where the first fetch resolves after the second. This was previously guarded by the AbortController that was removed. `[fixable]`


_Clean elimination of the double-fetch pattern. The main concern is losing the AbortController race guard — two rapid reattach_failed events can now race without cancellation. Otherwise solid: good test coverage, correct wiring, and formatting improvements._

- 🔵 **bugs** (L297): onSessionExpired marks `_sessionId` as unused but the parameter could be useful for logging or cleanup. Minor — no functional impact since the function correctly clears state regardless.

packages/client/src/protocol-parser.ts
```diff
@@ -94,12 +101,11 @@ export function parseServerMessage(
       callbacks.setWsRunning?.(poolKey, false);
       if (state.currentSessionId && callbacks.fetchMessages) {
         const sessionId = state.currentSessionId;
-        const controller = new AbortController();
-        callbacks.fetchMessages(sessionId, controller.signal)
+        callbacks
```
dimakis, line 104: 🟡 **regressions**: Removing the AbortController means in-flight fetches from a previous reattach_failed can no longer be cancelled. If two reattach_failed events fire in quick succession (e.g. flaky reconnect), both fetches race and the second onMessagesRestored call overwrites state with potentially stale data. The old AbortController prevented this. Consider guarding with a monotonic counter or generation ID instead. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -94,12 +101,11 @@ export function parseServerMessage(
...
+          .fetchMessages(sessionId)
           .then((msgs) => {
-            if (controller.signal.aborted) return;
             if (Array.isArray(msgs) && msgs.length > 0) {
-              callbacks.onMessagesRestored?.();
+              callbacks.onMessagesRestored?.(msgs);
```
dimakis, line 108: 🔵 **style**: The `.catch(() => {})` silently swallows all fetch errors during reattach recovery. A debug-level log would help diagnose connectivity issues without breaking the flow. `[fixable]`

packages/client/__tests__/store.test.ts
```diff
@@ -0,0 +1,453 @@
...
+        },
+      ],
+    });
+
+    // Update the nested task
+    lastWs.simulateMessage({
+      type: 'task_updated',
+      task: { id: 't2', summary: 'Child Updated', status: 'done', children: [] },
+    });
+
```
dimakis, line 280: 🔵 **missing_tests**: The reattach_failed store test verifies messages are restored once, but doesn't test the race scenario: two rapid reattach_failed events where the first fetch resolves after the second. This was previously guarded by the AbortController that was removed. `[fixable]`

packages/client/src/store.ts
```diff
@@ -296,29 +294,22 @@ export function createMitzoStore(
       }));
     },
 
-    onSessionExpired(sessionId: string) {
+    onSessionExpired(_sessionId: string) {
```
dimakis, line 297: 🔵 **bugs**: onSessionExpired marks `_sessionId` as unused but the parameter could be useful for logging or cleanup. Minor — no functional impact since the function correctly clears state regardless.

PR #212, opened 2026-04-16
feat(models): add Opus 4.7 + effort-tier model selector

dimakis:
## Summary
- Adds `claude-opus-4-7` and `claude-opus-4-7:max` (128k thinking budget) to the model selector
- Retains `claude-opus-4-6` as a previous-gen option
- New `parseModelSpec()` strips the `:max` effort suffix before sending the model ID to the SDK
- `resolveThinking()` maps `:max` → explicit 128k budget, default Opus stays adaptive
- Updates both mobile and desktop dropdowns

## Test plan
- [ ] Select Opus 4.7 from dropdown — verify adaptive thinking works
- [ ] Select Opus 4.7 Max — verify extended thinking with 128k budget
- [ ] Select Opus 4.6, Sonnet 4.6, Haiku 4.5 — verify no regression
- [ ] Verify `/api/models` returns all 5 entries
- [ ] Check Vertex availability for `claude-opus-4-7` (may not be GA yet)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #213, opened 2026-04-16
fix: unbreak mobile after PR #207/#212 (connection status, send queueing, model hang)

dimakis:
## Summary

After PR #207 (hook→store migration) and PR #212 (Opus 4.7 default), mitzo on mobile stopped working end-to-end. This PR ships three small, independent fixes discovered while debugging a live session where the phone showed a permanent red "disconnected" indicator, first-message-of-a-new-chat never reached the server, and selecting a provider-unavailable model hung the turn forever.

### 1. `connectionUpdate` never fired from the protocol parser

`parseServerMessage` in `packages/client/src/protocol-parser.ts` had `case '_open'` and `case '_close'` branches that just fell through with `break;`, and neither `reattached` nor `reattach_failed` set `result.connectionUpdate` either. The store initialises `connection.status` to `'disconnected'` and only updates on `connectionUpdate`, so after the migration the connection slice was pinned to `'disconnected'` forever. The WS itself was always fine — server logs showed frames flowing — but every `connection.status !== 'connected'` check in the UI was true, which is why the header dot stayed red and every Send tap dispatched `CONNECTION_LOST`.

Restored behaviour matches the old `useChatConnection` hook: a framed message is proof of a live socket, so `_open` / `reattached` / `reattach_failed` all mark the connection `'connected'` and `_close` marks it `'disconnected'`.

### 2. Turns hung forever when the selected model was unavailable on the provider

PR #212 bumped `DEFAULT_MODEL` to `claude-opus-4-7`. On providers that don't yet expose 4.7 (e.g. Vertex AI at time of writing) the Agent SDK's `query()` iterable simply never yields — no error, no event — so the query loop sat in `for await` indefinitely and the client never got any feedback.

Two fixes:

- Default back to `claude-opus-4-6` so out-of-box behaviour works everywhere; Opus 4.7 remains in the dropdown for users whose provider supports it, and the per-browser preference is unchanged.
- Add `QUERY_FIRST_EVENT_TIMEOUT_MS = 90_000` to `runQueryLoop`. If no SDK event arrives in 90s, abort and emit `{ type: 'error', error: 'Agent did not respond within 90s — the selected model may be unavailable on this provider.' }` instead of hanging. User-initiated aborts keep their existing silent behaviour — only timeout-triggered aborts surface an error.

### 3. First message of a new chat was silently dropped

In `MitzoStore.sendMessage()` the new-session path does:

```ts
wsPool.subscribe(newKey, wsListener);     // kicks off new WebSocket(...) — readyState = CONNECTING
const sent = wsPool.send(newKey, payload); // requires readyState === OPEN → returns false
```

The handshake hasn't completed by the next statement, so the very first send on a fresh chat silently dropped. The user sees their optimistic bubble, the server sees no `send` message, the turn hangs forever.

`WsPool.send()` now buffers payloads into `PoolEntry.pendingSends` when the socket is `CONNECTING` (or between reconnect attempts) and flushes them in order on `ws.onopen`. `send()` returns `true` for queued payloads so the caller can still distinguish fully-offline (`false`) from deferred (`true`).

## Test plan

- [x] Red-first unit tests added for every fix (4 new parser-lifecycle tests, 2 new query-loop timeout tests, 2 new WsPool queuing tests).
- [x] `npx vitest run packages/client` — 161/161 pass.
- [x] `npx vitest run server/__tests__/query-loop.test.ts` — 353/353 pass.
- [ ] CI green on the PR.
- [x] Manual: on iPhone via Tailscale, `fix/connection-status-stuck-disconnected` restored the green connection indicator, a new chat with Opus 4.6 got a response, and selecting Opus 4.7 on Vertex produced an error toast after ~90s instead of a hang.

### DIFF
--- a/packages/client/src/ws-connection.ts
+++ b/packages/client/src/ws-connection.ts
@@ -38,6 +38,13 @@ export interface WebSocketLike {
   close(): void;
 }
 
+/**
+ * Cap on the number of payloads queued per pool entry while the socket is
+ * unavailable. Sized generously enough to survive a short outage on mobile
+ * without memory growth running away during a long one.
+ */
+const MAX_PENDING_SENDS = 100;
+
 // ─── Pool entry ──────────────────────────────────────────────────────────────
 
 interface PoolEntry {
@@ -111,10 +118,10 @@ export class WsPool {
 
   /**
    * Send a message on the connection for this key.
-   * If the socket is still CONNECTING, the payload is queued and flushed
-   * in order once the socket opens. Returns false only when there is no
-   * entry or the socket is already closed/closing with no reconnect
-   * attempt queued.
+   * If the socket is still CONNECTING (or closed with a reconnect pending),
+   * the payload is queued and flushed in order once the socket opens.
+   * Returns false only when there is no entry or the socket is already
+   * closed/closing with no reconnect attempt queued.
    */
   send(key: string, msg: unknown): boolean {
     const entry = this.pool.get(key);
@@ -126,6 +133,13 @@ export class WsPool {
       return true;
     }
     if (state === WS_READY_STATE.CONNECTING || entry.reconnectTimer) {
+      // Bound the queue so a long outage doesn't grow memory without limit.
+      // When we hit the cap we drop the oldest payload — matches the
+      // "keep the most recent intent" preference users typically have on
+      // mobile, and matches how the server's message buffer behaves.
+      if (entry.pendingSends.length >= MAX_PENDING_SENDS) {
+        entry.pendingSends.shift();
+      }
       entry.pendingSends.push(payload);
       return true;
     }
@@ -221,7 +235,14 @@ export class WsPool {
     entry.ws = ws;
 
     ws.onopen = () => {
-      // Flush any messages queued while the socket was still CONNECTING.
+      // Flush queued messages BEFORE broadcasting _open. Listeners react
+      // to _open by marking connection='connected' and often by firing
+      // follow-up sends of their own; if we broadcast first, those fresh
+      // sends could interleave in front of the queued ones and change
+      // the server-visible order. The flush is synchronous on the same
+      // WebSocket, so once _open fans out the queued payloads have
+      // already been written to the wire.
+      //
       // Guard each send: if one throws (browser extension closing the
       // socket mid-flush, a misbehaving mock, etc.), we requeue the
       // remaining payloads onto pendingSends so the next reconnect picks
--- a/packages/client/src/ws-connection.ts
+++ b/packages/client/src/ws-connection.ts
@@ -222,10 +222,22 @@ export class WsPool {
 
     ws.onopen = () => {
       // Flush any messages queued while the socket was still CONNECTING.
+      // Guard each send: if one throws (browser extension closing the
+      // socket mid-flush, a misbehaving mock, etc.), we requeue the
+      // remaining payloads onto pendingSends so the next reconnect picks
+      // them up, and we still broadcast _open so listeners transition
+      // into the connected state.
       if (entry.pendingSends.length > 0) {
         const toFlush = entry.pendingSends;
         entry.pendingSends = [];
-        for (const payload of toFlush) ws.send(payload);
+        for (let i = 0; i < toFlush.length; i++) {
+          try {
+            ws.send(toFlush[i]);
+          } catch {
+            entry.pendingSends.unshift(...toFlush.slice(i));
+            break;
+          }
+        }
       }
       this.broadcast(entry, { type: '_open' });
     };
--- a/packages/client/src/ws-connection.ts
+++ b/packages/client/src/ws-connection.ts
@@ -48,6 +48,14 @@ interface PoolEntry {
   lastSeq: number;
   reconnectTimer: ReturnType<typeof setTimeout> | null;
   listeners: Set<MsgListener>;
+  /**
+   * Messages issued via send() while the socket was still CONNECTING.
+   * Flushed in-order once the socket transitions to OPEN. This covers the
+   * new-session flow where the store subscribes and immediately tries to
+   * send the first user prompt — without this queue the first send would
+   * silently drop because readyState !== OPEN.
+   */
+  pendingSends: string[];
 }
 
 // ─── Pool ────────────────────────────────────────────────────────────────────
@@ -101,11 +109,24 @@ export class WsPool {
     return () => entry.listeners.delete(listener);
   }
 
-  /** Send a message on the connection for this key. */
+  /**
+   * Send a message on the connection for this key.
+   * If the socket is still CONNECTING, the payload is queued and flushed
+   * in order once the socket opens. Returns false only when there is no
+   * entry or the socket is already closed/closing with no reconnect
+   * attempt queued.
+   */
   send(key: string, msg: unknown): boolean {
     const entry = this.pool.get(key);
-    if (entry?.ws?.readyState === WS_READY_STATE.OPEN) {
-      entry.ws.send(JSON.stringify(msg));
+    if (!entry) return false;
+    const payload = JSON.stringify(msg);
+    const state = entry.ws?.readyState;
+    if (state === WS_READY_STATE.OPEN && entry.ws) {
+      entry.ws.send(payload);
+      return true;
+    }
+    if (state === WS_READY_STATE.CONNECTING || entry.reconnectTimer) {
+      entry.pendingSends.push(payload);
       return true;
     }
     return false;
@@ -175,6 +196,7 @@ export class WsPool {
         lastSeq: 0,
         reconnectTimer: null,
         listeners: new Set(),
+        pendingSends: [],
       };
       this.pool.set(key, entry);
       this.connectEntry(key, entry);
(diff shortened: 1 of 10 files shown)


packages/client/src/ws-connection.ts
```diff
@@ -198,7 +220,15 @@ export class WsPool {
     const ws = this.config.createWebSocket(url);
     entry.ws = ws;
 
-    ws.onopen = () => this.broadcast(entry, { type: '_open' });
+    ws.onopen = () => {
+      // Flush any messages queued while the socket was still CONNECTING.
+      if (entry.pendingSends.length > 0) {
+        const toFlush = entry.pendingSends;
+        entry.pendingSends = [];
+        for (const payload of toFlush) ws.send(payload);
```
dimakis: 🟡 **unsafe_assumptions**: pendingSends are flushed synchronously inside onopen before broadcasting _open. If any ws.send() call throws (e.g. a browser extension or mock closes the socket between onopen firing and the send call), the remaining queued messages are silently lost and _open is never broadcast. A try/catch around the flush loop — or at least ensuring _open is always broadcast — would improve resilience. `[fixable]`

server/query-loop.ts
```diff
@@ -191,8 +195,25 @@ export async function runQueryLoop(
...
+  // If the SDK yields no events within QUERY_FIRST_EVENT_TIMEOUT_MS, abort
+  // the query and surface an error to the client. This unblocks turns where
+  // the configured model is unreachable (e.g. requested via Vertex AI before
+  // that model has landed there) and would otherwise hang indefinitely.
+  let firstEventReceived = false;
+  let timedOut = false;
+  const firstEventTimer = setTimeout(() => {
+    if (!firstEventReceived) {
+      timedOut = true;
+      abortController.abort();
```
dimakis, line 207: 🟡 **bugs**: The timeout handler calls abortController.abort() without passing a reason. In the catch block, if the SDK wraps the abort as a generic Error (not an AbortError), the catch could misidentify a timeout-abort that races with firstEventReceived=true as an unexpected error. Using abortController.abort(new Error('first-event-timeout')) and checking err in the catch would be more robust than the separate boolean flag, though in practice the current code works because firstEventReceived is set before clearTimeout. `[fixable]`

server/query-loop.ts
```diff
@@ -191,8 +195,25 @@ export async function runQueryLoop(
 
   log.info('query loop started', { clientId });
 
+  // If the SDK yields no events within QUERY_FIRST_EVENT_TIMEOUT_MS, abort
+  // the query and surface an error to the client. This unblocks turns where
+  // the configured model is unreachable (e.g. requested via Vertex AI before
```
dimakis, line 200: 🔵 **style**: The firstEventReceived/timedOut flags are module-level concerns leaking into the function body. Extracting a small FirstEventGuard helper (e.g. { arm(), disarm(), didTimeout }) would make the intent clearer and keep the main loop focused on event translation. Minor — not blocking. `[fixable]`

dimakis (commented):
## Centaur Review

### `packages/client/__tests__/ws-connection.test.ts`
_Solid bug-fix PR that addresses three real issues (connection status, send queueing, model hang). The query-loop timeout and WS send queue are well-tested and correctly implemented. Minor robustness concerns around the flush loop's error handling and a missing test for the CONNECTING→CLOSED edge case._

- 🔵 **missing_tests**: No test covers the case where send() queues messages while CONNECTING and the socket fails before ever opening (CONNECTING → CLOSED). The current implementation does survive this because pendingSends persists across reconnects, but a test would document that guarantee and prevent regressions. `[fixable]`


### `packages/client/src/ws-connection.ts`
_Solid bug-fix PR that addresses three real issues (connection status, send queueing, model hang). The query-loop timeout and WS send queue are well-tested and correctly implemented. Minor robustness concerns around the flush loop's error handling and a missing test for the CONNECTING→CLOSED edge case._

- 🟡 **unsafe_assumptions** (L228): pendingSends are flushed synchronously inside onopen before broadcasting _open. If any ws.send() call throws (e.g. a browser extension or mock closes the socket between onopen firing and the send call), the remaining queued messages are silently lost and _open is never broadcast. A try/catch around the flush loop — or at least ensuring _open is always broadcast — would improve resilience. `[fixable]`


### `server/query-loop.ts`
_Solid bug-fix PR that addresses three real issues (connection status, send queueing, model hang). The query-loop timeout and WS send queue are well-tested and correctly implemented. Minor robustness concerns around the flush loop's error handling and a missing test for the CONNECTING→CLOSED edge case._

- 🟡 **bugs** (L207): The timeout handler calls abortController.abort() without passing a reason. In the catch block, if the SDK wraps the abort as a generic Error (not an AbortError), the catch could misidentify a timeout-abort that races with firstEventReceived=true as an unexpected error. Using abortController.abort(new Error('first-event-timeout')) and checking err in the catch would be more robust than the separate boolean flag, though in practice the current code works because firstEventReceived is set before clearTimeout. `[fixable]`
- 🔵 **style** (L200): The firstEventReceived/timedOut flags are module-level concerns leaking into the function body. Extracting a small FirstEventGuard helper (e.g. { arm(), disarm(), didTimeout }) would make the intent clearer and keep the main loop focused on event translation. Minor — not blocking. `[fixable]`

packages/client/src/ws-connection.ts
```diff
@@ -101,11 +109,24 @@ export class WsPool {
...
-    if (entry?.ws?.readyState === WS_READY_STATE.OPEN) {
-      entry.ws.send(JSON.stringify(msg));
+    if (!entry) return false;
+    const payload = JSON.stringify(msg);
+    const state = entry.ws?.readyState;
+    if (state === WS_READY_STATE.OPEN && entry.ws) {
+      entry.ws.send(payload);
+      return true;
+    }
+    if (state === WS_READY_STATE.CONNECTING || entry.reconnectTimer) {
```
dimakis, line 135: 🔵 **unsafe_assumptions**: When `entry.ws` is null (post-close) and only `reconnectTimer` is truthy, messages accumulate in `pendingSends`. If reconnection keeps failing (server down, bad URL), the queue grows unboundedly with no cap. Consider adding a max-length or clearing the queue after N failed reconnect attempts. `[fixable]`

packages/client/src/ws-connection.ts
```diff
@@ -198,7 +220,15 @@ export class WsPool {
     const ws = this.config.createWebSocket(url);
     entry.ws = ws;
 
-    ws.onopen = () => this.broadcast(entry, { type: '_open' });
+    ws.onopen = () => {
```
dimakis, line 237: 🔵 **style**: Flushing `pendingSends` before broadcasting `_open` means the listener receiving `_open` (which triggers connectionUpdate → connected) will fire after the queued messages have already been sent on the wire. This is the correct order for correctness, but a brief inline comment explaining why flush-before-broadcast matters would help future readers. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


### `packages/client/src/ws-connection.ts`
_Solid bug-fix PR — the three fixes (read running before USER_SEND, queue sends during CONNECTING, first-event timeout) are correctly implemented with good test coverage; only minor hardening suggestions._

- 🔵 **unsafe_assumptions** (L128): When `entry.ws` is null (post-close) and only `reconnectTimer` is truthy, messages accumulate in `pendingSends`. If reconnection keeps failing (server down, bad URL), the queue grows unboundedly with no cap. Consider adding a max-length or clearing the queue after N failed reconnect attempts. `[fixable]`
- 🔵 **style** (L223): Flushing `pendingSends` before broadcasting `_open` means the listener receiving `_open` (which triggers connectionUpdate → connected) will fire after the queued messages have already been sent on the wire. This is the correct order for correctness, but a brief inline comment explaining why flush-before-broadcast matters would help future readers. `[fixable]`


### `packages/client/__tests__/ws-connection.test.ts`
_Solid bug-fix PR — the three fixes (read running before USER_SEND, queue sends during CONNECTING, first-event timeout) are correctly implemented with good test coverage; only minor hardening suggestions._

- 🔵 **missing_tests**: No test covers the case where `send()` is called after `onclose` fires (ws=null) but while `reconnectTimer` is set. This is the other queuing path (line 128) and should verify that messages queued during reconnect are flushed when the new socket opens. `[fixable]`

dimakis:
Thanks for the review. Addressed 2 of 4 findings in `4ca383e`; responses for the other two:

**🟡 bugs (query-loop.ts L207) — timeout without reason** — Keeping the current `timedOut` flag. The race you describe is safe by construction: `firstEventReceived` is set and `clearTimeout(firstEventTimer)` is called inside the first iteration of the `for await` loop, on the same microtask as the first event dequeues. If the timer callback is already in the macrotask queue when we clear it (the one window where it can't be cancelled), its first line is `if (!firstEventReceived) { … }`, so it bails out without aborting. Using `abort(new Error(...))` + reading `signal.reason` in the catch trades one piece of local state (`timedOut`) for another (cross-checking `err` against `signal.reason`), and introduces a second edge case: user-initiated aborts that pass a reason would be indistinguishable from timeout aborts unless we sentinel-tag the error. Happy to revisit if you can show a concrete race the flag misses.

**🔵 style (query-loop.ts L200) — extract FirstEventGuard** — Not blocking per your note; I'd rather keep the three locals inline. The scope is ~12 lines and tightly coupled to the `for await` loop; a guard class would split the state/teardown across four indirections (arm, disarm, didTimeout, finally clearTimeout) for no readability win. Will reach for the extraction if a second caller ever needs the pattern.

Also — one additional bug found after your review: every follow-up send in a session was hanging, because `sendMessage()` was reading `messages.running` *after* `USER_SEND` set it to true, so every second+ message got silently parked in `parserState.pendingSend`. Fixed in `c14dbfd` with a regression test; see the commit message for details.

dimakis:
All three items from round 2 addressed in `d755d8a`: `MAX_PENDING_SENDS` cap of 100 with oldest-dropped overflow, extended inline comment explaining the flush-before-broadcast ordering, and two regression tests (send during the reconnect-pending window, queue cap overflow). CI watching.

PR #214, opened 2026-04-16
fix: new session connection + resume CWD mismatch

dimakis:
## Summary

- **Bug 1 — New sessions can't connect**: After PR #207 migrated from hooks to `@mitzo/client` store, `handleSend` in ChatView/DesktopChatView gates on `connection.status !== 'connected'`. For new sessions no WS exists yet, so the guard blocks forever. Fixed by skipping the guard when there's no `activeSessionId` — the store's `sendMessage()` bootstraps the WS on demand.

- **Bug 2 — Resume shows "Session expired" on first prompt**: Worktree sessions store SDK conversation data under `~/.claude/projects/<encoded-worktree-path>/`. Resume skips worktree creation and uses `BASE_REPO` as CWD, so the SDK can't find the conversation → "No conversation found". Fixed by looking up the original session CWD from the event store and passing it to `query()`.

## Test plan

- [x] Store-level tests for new-session bootstrap and stale session ID recovery
- [x] All 569 relevant tests pass (`packages/client/`, `server/`)
- [x] Lint clean (0 errors)
- [ ] Manual: start a new chat session from mobile — should connect immediately
- [ ] Manual: navigate to an existing session and send a prompt — should resume without "Session expired"

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -111,7 +111,11 @@ export function ChatView() {
   // ── Actions ──────────────────────────────────────────────────────────────
 
   function handleSend(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): boolean {
-    if (connection.status !== 'connected') {
+    // For new sessions (no activeSessionId) the store bootstraps a WS on
+    // demand inside sendMessage(), so we must not block on connection status.
+    // Only gate on connection for existing sessions where a WS should already
+    // be open.
+    if (activeSessionId && connection.status !== 'connected') {
       storeDispatchMessages({ type: 'CONNECTION_LOST' });
       return false;
     }
--- a/frontend/src/pages/DesktopChatView.tsx
+++ b/frontend/src/pages/DesktopChatView.tsx
@@ -144,7 +144,11 @@ export function DesktopChatView() {
   // ── Actions ──────────────────────────────────────────────────────────────
 
   function handleSend(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): boolean {
-    if (connection.status !== 'connected') {
+    // For new sessions (no activeSessionId) the store bootstraps a WS on
+    // demand inside sendMessage(), so we must not block on connection status.
+    // Only gate on connection for existing sessions where a WS should already
+    // be open.
+    if (activeSessionId && connection.status !== 'connected') {
       storeDispatchMessages({ type: 'CONNECTION_LOST' });
       return false;
     }
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -405,7 +405,22 @@ export async function startChat(
 ) {
   const abortController = new AbortController();
   const mode = options.mode || 'agent';
-  const baseCwd = options.cwd || BASE_REPO;
+
+  // When resuming, look up the original session's CWD from the event store.
+  // The SDK stores conversation data under ~/.claude/projects/<encoded-cwd>/,
+  // so we must use the same CWD the session was created with — otherwise the
+  // SDK can't find the conversation and returns "No conversation found".
+  let baseCwd = options.cwd || BASE_REPO;
+  if (options.resume && !options.cwd) {
+    const sessionMeta = eventStore.getSession(options.resume);
+    if (sessionMeta?.cwd) {
+      baseCwd = sessionMeta.cwd;
+      log.info('resume: using original session CWD', {
+        sessionId: options.resume,
+        cwd: baseCwd,
+      });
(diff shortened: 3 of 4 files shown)


server/chat.ts
```diff
@@ -405,7 +405,22 @@ export async function startChat(
...
   const abortController = new AbortController();
   const mode = options.mode || 'agent';
-  const baseCwd = options.cwd || BASE_REPO;
+
+  // When resuming, look up the original session's CWD from the event store.
+  // The SDK stores conversation data under ~/.claude/projects/<encoded-cwd>/,
+  // so we must use the same CWD the session was created with — otherwise the
+  // SDK can't find the conversation and returns "No conversation found".
+  let baseCwd = options.cwd || BASE_REPO;
+  if (options.resume && !options.cwd) {
```
dimakis, line 414: 🟡 **missing_tests**: The new resume CWD lookup (`eventStore.getSession(options.resume)` → use stored cwd) has no server-side test coverage. `chat.test.ts` only checks module exports. The EventStore `getSession` also has no tests covering the `cwd` field specifically. Consider adding at least a unit test for the CWD lookup logic, even if it requires mocking `eventStore`. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -111,7 +111,11 @@ export function ChatView() {
   // ── Actions ──────────────────────────────────────────────────────────────
 
   function handleSend(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): boolean {
-    if (connection.status !== 'connected') {
+    // For new sessions (no activeSessionId) the store bootstraps a WS on
+    // demand inside sendMessage(), so we must not block on connection status.
```
dimakis, line 115: 🔵 **unsafe_assumptions**: When `activeSessionId` is falsy the connection gate is completely bypassed, so `sendMessage` runs regardless of connection state. If `sendMessage`'s internal `wsPool.send()` returns false (WS not yet open), the store sets a `sendError` string but the user message was already optimistically rendered — the UI shows a sent message that may silently fail. This is pre-existing behavior, but the gate bypass makes it easier to hit. Consider surfacing `sendError` in the UI or retrying on connect. `[fixable]`

frontend/src/pages/DesktopChatView.tsx
```diff
@@ -144,7 +144,11 @@ export function DesktopChatView() {
   // ── Actions ──────────────────────────────────────────────────────────────
 
   function handleSend(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): boolean {
-    if (connection.status !== 'connected') {
+    // For new sessions (no activeSessionId) the store bootstraps a WS on
+    // demand inside sendMessage(), so we must not block on connection status.
```
dimakis, line 148: 🔵 **style**: The `handleSend` function body (including the new connection gate logic and comment) is duplicated verbatim between ChatView and DesktopChatView. Consider extracting it into a shared hook or utility to keep the two views in sync. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid bug fix for two real issues (new-session connection gating and resume CWD mismatch). Client-side tests are good; server-side resume CWD path lacks test coverage._

- 🟡 **missing_tests** (L414): The new resume CWD lookup (`eventStore.getSession(options.resume)` → use stored cwd) has no server-side test coverage. `chat.test.ts` only checks module exports. The EventStore `getSession` also has no tests covering the `cwd` field specifically. Consider adding at least a unit test for the CWD lookup logic, even if it requires mocking `eventStore`. `[fixable]`


_Solid bug fix for two real issues (new-session connection gating and resume CWD mismatch). Client-side tests are good; server-side resume CWD path lacks test coverage._

- 🔵 **unsafe_assumptions** (L115): When `activeSessionId` is falsy the connection gate is completely bypassed, so `sendMessage` runs regardless of connection state. If `sendMessage`'s internal `wsPool.send()` returns false (WS not yet open), the store sets a `sendError` string but the user message was already optimistically rendered — the UI shows a sent message that may silently fail. This is pre-existing behavior, but the gate bypass makes it easier to hit. Consider surfacing `sendError` in the UI or retrying on connect. `[fixable]`


_Solid bug fix for two real issues (new-session connection gating and resume CWD mismatch). Client-side tests are good; server-side resume CWD path lacks test coverage._

- 🔵 **style** (L148): The `handleSend` function body (including the new connection gate logic and comment) is duplicated verbatim between ChatView and DesktopChatView. Consider extracting it into a shared hook or utility to keep the two views in sync. `[fixable]`

PR #216, opened 2026-04-16
fix(client): unsubscribe old WS when switching sessions

dimakis:
## Summary

- **Bug**: `switchSession()` and `newSession()` subscribed to a new WS pool entry but never unsubscribed the previous one. Messages from the old session kept dispatching through `wsListener`, causing sessions to visually "merge" — the UI showed messages from the previous session in the new one.
- **Fix**: Track the unsubscribe handle from `wsPool.subscribe()` and call it before subscribing to a new session in `switchSession()`, `newSession()`, and `sendMessage()` (new-session path).
- Adds 2 isolation tests verifying old-session messages don't leak after switching.

## Test plan

- [x] 27 store tests pass (including 2 new isolation tests)
- [x] Lint clean
- [ ] Manual: switch sessions — old session messages should not appear
- [ ] Manual: start new session while old one is running — only new session content visible

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -128,6 +128,11 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
   // Track the actual WS pool key so wsListener routes to the correct entry
   let activePoolKey: string | null = null;
 
+  // Unsubscribe handle for the current WS subscription — called when
+  // switching sessions or starting a new one so old session messages
+  // don't leak into the new session's state.
+  let activeUnsub: (() => void) | null = null;
+
   const store = createStore<MitzoStoreState>((set, get) => ({
     // ── Initial state ────────────────────────────────────────────────────
 
@@ -150,6 +155,11 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     async switchSession(id: string) {
+      // Unsubscribe from previous session's WS before subscribing to the new
+      // one — otherwise the old listener keeps dispatching messages into the
+      // store, causing sessions to "merge."
+      activeUnsub?.();
+
       parserState.currentSessionId = id;
       const poolKey = `session:${id}`;
       activePoolKey = poolKey;
@@ -163,7 +173,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       }));
 
       // Subscribe to WS messages for this session
-      wsPool.subscribe(poolKey, wsListener);
+      activeUnsub = wsPool.subscribe(poolKey, wsListener);
 
       // Load session messages
       try {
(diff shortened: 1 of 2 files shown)


packages/client/src/store.ts
```diff
@@ -179,6 +189,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
```
dimakis, line 189: 🔵 **bugs**: In `newSession()`, `activeUnsub` and `activePoolKey` are set to null but `parserState.currentSessionId` is set to undefined on the next line. If `newSession()` is called when there's no active session, `activeUnsub?.()` is harmless, but the ordering (cleanup before state reset) is correct. No bug, just noting the pattern is sound.

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix for WS listener leakage between sessions. The core paths (switchSession, newSession) are properly covered; only the sendMessage new-session path lacks an isolation test._

- 🔵 **missing_tests**: No test covers the `sendMessage` path for new sessions (the `if (!poolKey)` branch at store.ts:249) where `activeUnsub?.()` is called before creating a new pool key. The existing `sendMessage` tests don't verify that a prior subscription is cleaned up when sending the first message in a brand-new session. `[fixable]`


_Clean, well-tested fix for WS listener leakage between sessions. The core paths (switchSession, newSession) are properly covered; only the sendMessage new-session path lacks an isolation test._

- 🔵 **bugs** (L189): In `newSession()`, `activeUnsub` and `activePoolKey` are set to null but `parserState.currentSessionId` is set to undefined on the next line. If `newSession()` is called when there's no active session, `activeUnsub?.()` is harmless, but the ordering (cleanup before state reset) is correct. No bug, just noting the pattern is sound.

dimakis:
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix for a real session-isolation bug. The approach of tracking an `activeUnsub` handle is sound and the three call sites (switchSession, newSession, sendMessage new-session) are all covered by tests._

- 🔵 **missing_tests**: No test covers the case where `switchSession` is called twice in rapid succession (A→B→C). The second `switchSession` should unsub from B before subscribing to C. The existing tests only cover A→B transitions. A triple-switch test would strengthen confidence in the cleanup chain. `[fixable]`


_Clean, well-tested fix for a real session-isolation bug. The approach of tracking an `activeUnsub` handle is sound and the three call sites (switchSession, newSession, sendMessage new-session) are all covered by tests._

- 🔵 **unsafe_assumptions** (L415): `onSessionAssigned` updates `activePoolKey` but does not update `activeUnsub`. This is safe today because the WsPool aliases `session:*` to the same pool entry as `new:*` (ws-connection.ts:330-332), so the old unsub closure still targets the correct listener set. However, this relies on an implicit invariant — if WsPool's aliasing behavior ever changes, the unsub would target a stale entry. A brief comment noting this coupling would improve maintainability. `[fixable]`

PR #217, opened 2026-04-16
fix: token bar crash, compaction tracking, scroll UX

dimakis:
## Summary

- **Fix token bar crash** — mid-turn `token_update` events omit `sessionTotal`, causing `undefined.toLocaleString()` crash when tapping the detail panel. Protocol parser now filters undefined fields; TokenBar uses safe defaults.
- **Track compaction count** — detect `compact_result: 'success'` SDK system messages and display compaction count in the token bar detail panel.
- **Scroll to bottom on session resume** — auto-scroll when messages load via RESTORE (0 → N transition) so you land at the freshest content.
- **Scroll FAB buttons** — floating up/down arrows appear when scrolled away from top or bottom of chat. Tap to jump.

## Test plan

- [x] TokenBar renders without crash when `sessionTotal` is undefined
- [x] Protocol parser omits undefined fields from partial token updates
- [x] Compaction events from SDK system messages increment `numCompactions`
- [x] ChatArea scrolls to bottom when messages appear for the first time (restore)
- [x] ScrollFab shows/hides buttons based on scroll position
- [x] ScrollFab calls `scrollTo` on click
- [ ] Manual: resume a session on mobile — should land at the bottom
- [ ] Manual: scroll up in a long session — FAB arrows appear

### DIFF
--- a/frontend/src/components/ChatArea.tsx
+++ b/frontend/src/components/ChatArea.tsx
@@ -37,6 +37,18 @@ export function ChatArea({
 }: ChatAreaProps) {
   const internalScrollRef = useRef<HTMLDivElement>(null);
   const scrollRef = externalScrollRef ?? internalScrollRef;
+  const prevMessageCount = useRef(0);
+
+  // Scroll to bottom on session restore (messages jump from 0 to N)
+  useEffect(() => {
+    const wasEmpty = prevMessageCount.current === 0;
+    prevMessageCount.current = messages.length;
+    if (wasEmpty && messages.length > 0) {
+      requestAnimationFrame(() => {
+        scrollRef.current?.scrollTo({ top: scrollRef.current.scrollHeight });
+      });
+    }
+  }, [messages, scrollRef]);
 
   // Auto-scroll during streaming: follow new content if user is near the bottom
   useEffect(() => {
--- a/frontend/src/components/ScrollFab.tsx
+++ b/frontend/src/components/ScrollFab.tsx
@@ -15,9 +15,7 @@ export function ScrollFab({ scrollRef }: Props) {
     if (!el) return;
 
     function update() {
-      const el = scrollRef.current;
-      if (!el) return;
-      const { scrollTop, scrollHeight, clientHeight } = el;
+      const { scrollTop, scrollHeight, clientHeight } = el!;
       const canScroll = scrollHeight > clientHeight;
       setShowTop(canScroll && scrollTop > THRESHOLD);
       setShowBottom(canScroll && scrollHeight - scrollTop - clientHeight > THRESHOLD);
@@ -36,7 +34,6 @@ export function ScrollFab({ scrollRef }: Props) {
     scrollRef.current?.scrollTo({ top: scrollRef.current.scrollHeight, behavior: 'smooth' });
   }, [scrollRef]);
 
-  if (!scrollRef.current) return null;
   if (!showTop && !showBottom) return null;
 
   return (
--- a/frontend/src/components/ScrollFab.tsx
+++ b/frontend/src/components/ScrollFab.tsx
@@ -1,0 +1,68 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const THRESHOLD = 200; // px from edge to show button
+
+interface Props {
+  scrollRef: React.RefObject<HTMLDivElement | null>;
+}
+
+export function ScrollFab({ scrollRef }: Props) {
+  const [showTop, setShowTop] = useState(false);
+  const [showBottom, setShowBottom] = useState(false);
+
+  useEffect(() => {
+    const el = scrollRef.current;
+    if (!el) return;
+
+    function update() {
+      const el = scrollRef.current;
+      if (!el) return;
+      const { scrollTop, scrollHeight, clientHeight } = el;
+      const canScroll = scrollHeight > clientHeight;
+      setShowTop(canScroll && scrollTop > THRESHOLD);
+      setShowBottom(canScroll && scrollHeight - scrollTop - clientHeight > THRESHOLD);
+    }
+
+    el.addEventListener('scroll', update, { passive: true });
+    update();
+    return () => el.removeEventListener('scroll', update);
+  }, [scrollRef]);
+
+  const scrollToTop = useCallback(() => {
+    scrollRef.current?.scrollTo({ top: 0, behavior: 'smooth' });
+  }, [scrollRef]);
+
+  const scrollToBottom = useCallback(() => {
+    scrollRef.current?.scrollTo({ top: scrollRef.current.scrollHeight, behavior: 'smooth' });
+  }, [scrollRef]);
+
+  if (!scrollRef.current) return null;
+  if (!showTop && !showBottom) return null;
+
+  return (
+    <div className="scroll-fab-container">
+      {showTop && (
+        <button
+          className="scroll-fab scroll-fab--up"
+          onClick={scrollToTop}
+          aria-label="Scroll to top"
+        >
+          <svg viewBox="0 0 24 24" width="20" height="20" fill="currentColor">
+            <path d="M7.41 15.41L12 10.83l4.59 4.58L18 14l-6-6-6 6z" />
+          </svg>
+        </button>
+      )}
+      {showBottom && (
+        <button
+          className="scroll-fab scroll-fab--down"
+          onClick={scrollToBottom}
+          aria-label="Scroll to bottom"
+        >
+          <svg viewBox="0 0 24 24" width="20" height="20" fill="currentColor">
+            <path d="M7.41 8.59L12 13.17l4.59-4.58L18 10l-6 6-6-6z" />
+          </svg>
+        </button>
+      )}
+    </div>
+  );
+}
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -475,6 +475,56 @@ describe('task system messages', () => {
   });
 });
 
+// ─── Token updates ───────────────────────────────────────────────────────────
+
+describe('token_update', () => {
+  it('produces tokensUpdate with all fields', () => {
+    const r = parseServerMessage(
+      {
+        type: 'token_update',
+        agentContext: 80000,
+        contextCeiling: 200000,
+        sessionTotal: 150000,
+        numTurns: 3,
+        turnIndex: 2,
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(r.tokensUpdate).toEqual({
+      agentContext: 80000,
+      contextCeiling: 200000,
+      sessionTotal: 150000,
+      numTurns: 3,
+      turnIndex: 2,
+    });
+  });
+
+  it('omits undefined fields so partial spread does not clobber existing values', () => {
+    const r = parseServerMessage(
+      {
+        type: 'token_update',
+        agentContext: 90000,
+        contextCeiling: 200000,
+        turnIndex: 4,
+        // sessionTotal and numTurns intentionally absent (mid-turn update)
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(r.tokensUpdate).toEqual({
+      agentContext: 90000,
+      contextCeiling: 200000,
+      turnIndex: 4,
+    });
+    // Must NOT have sessionTotal key at all
+    expect('sessionTotal' in r.tokensUpdate!).toBe(false);
+    expect('numTurns' in r.tokensUpdate!).toBe(false);
+  });
+});
+
 // ─── Inbox ────────────────────────────────────────────────────────────────────
 
 describe('inbox', () => {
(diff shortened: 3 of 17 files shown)


dimakis:
## Centaur Review

_Solid defensive fix for the TokenBar crash and clean ScrollFab addition. The main gap is the missing `numCompactions` field in the `TokenUpdateMsg` type definition — the server sends it and the parser handles it, but the WS type contract doesn't declare it._

- 🟡 **regressions** (L232): `TokenUpdateMsg` interface is missing the new `numCompactions` field. The server conditionally emits `numCompactions` and the protocol-parser handles it, but the frontend WS type doesn't declare it. This causes a type-safety gap — code consuming typed WS messages won't know `numCompactions` exists. Add `numCompactions?: number` to the interface. `[fixable]`


### `frontend/src/components/ScrollFab.tsx`
_Solid defensive fix for the TokenBar crash and clean ScrollFab addition. The main gap is the missing `numCompactions` field in the `TokenUpdateMsg` type definition — the server sends it and the parser handles it, but the WS type contract doesn't declare it._

- 🔵 **bugs** (L17): The `update()` function re-reads `scrollRef.current` inside the closure despite the outer `useEffect` already capturing it as `el`. The outer `el` is used only for `addEventListener`/`removeEventListener`, while `update()` reads `scrollRef.current` fresh each call. This is technically fine (avoids stale ref), but inconsistent — the cleanup's `removeEventListener` could target a stale `el` if the ref changes without the effect re-running (unlikely given stable ref objects, but worth noting). `[fixable]`
- 🔵 **unsafe_assumptions** (L40): The early return `if (!scrollRef.current) return null` runs during render but the `useEffect` that registers the scroll listener runs after mount. On the very first render with a valid ref, the effect sets up the listener and calls `update()`, but the component may have already returned null from this line (if the ref wasn't set at initial render time). This is safe in practice because `scrollRef` is typically set before mount, but could cause a flash of no-FAB on first render.


### `frontend/src/components/__tests__/ChatArea.test.tsx`
_Solid defensive fix for the TokenBar crash and clean ScrollFab addition. The main gap is the missing `numCompactions` field in the `TokenUpdateMsg` type definition — the server sends it and the parser handles it, but the WS type contract doesn't declare it._

- 🔵 **missing_tests** (L152): The session-restore scroll test may be flaky: it spies on `scrollTo` after the initial render but `scrollRef.current` in jsdom may not have `scrollHeight` set meaningfully. The test asserts `scrollTo` was called but doesn't verify the scroll target (`top` value). Consider asserting the call args include `{ top: expect.any(Number) }`. `[fixable]`


### `server/query-loop.ts`
_Solid defensive fix for the TokenBar crash and clean ScrollFab addition. The main gap is the missing `numCompactions` field in the `TokenUpdateMsg` type definition — the server sends it and the parser handles it, but the WS type contract doesn't declare it._

- 🔵 **style** (L323): The conditional spread `...(numCompactions > 0 ? { numCompactions } : {})` is repeated in two places (lines 323 and 412). Consider extracting a small helper or shared object fragment to avoid duplication, since these emit calls already share the same pattern. `[fixable]`

frontend/src/components/ScrollFab.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const { scrollTop, scrollHeight, clientHeight } = el!;
+      const canScroll = scrollHeight > clientHeight;
+      setShowTop(canScroll && scrollTop > THRESHOLD);
+      setShowBottom(canScroll && scrollHeight - scrollTop - clientHeight > THRESHOLD);
+    }
+
+    el.addEventListener('scroll', update, { passive: true });
+    update();
+    return () => el.removeEventListener('scroll', update);
+  }, [scrollRef]);
```
dimakis, line 27: 🟡 **bugs**: The `useEffect` depends on `scrollRef` (a stable ref object), so it only runs once on mount. The `update()` function is only called on `scroll` events after that. When new messages arrive and `scrollHeight` grows (without a user scroll), the FAB visibility won't update — the 'scroll to bottom' button won't appear until the user manually scrolls. Consider adding a ResizeObserver or re-running `update()` when content changes. `[fixable]`

packages/client/__tests__/protocol-parser.test.ts
```diff
@@ -475,6 +475,56 @@ describe('task system messages', () => {
...
+    );
+    expect(r.tokensUpdate).toEqual({
+      agentContext: 80000,
+      contextCeiling: 200000,
+      sessionTotal: 150000,
+      numTurns: 3,
+      turnIndex: 2,
+    });
+  });
+
```
dimakis, line 503: 🔵 **missing_tests**: The `token_update` parser tests verify `sessionTotal` and `numTurns` omission but don't test the new `numCompactions` field — neither its presence when set nor its absence when omitted. Add a case with `numCompactions` to the 'all fields' test and verify it's excluded from the partial update test. `[fixable]`

server/index.ts
```diff
@@ -245,10 +245,76 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
 
       if (msg.type === 'subscribe') {
         const found = registry.findBySessionId(msg.sessionId);
-        if (found) {
+        if (found && !registry.isAttached(found.clientId)) {
```
dimakis, line 426: 🔵 **missing_tests**: The subscribe-to-reattach promotion logic (lines 248-314) is a significant new code path with event replay, worktree path recovery, and snapshot forwarding. No integration or unit test covers this flow. Consider testing: (1) subscribe on a detached session triggers reattach + replays events, (2) subscribe on an attached session falls through to observer, (3) reattach failure falls back to observer.

server/index.ts
```diff
@@ -245,10 +245,76 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
+                  const p = evt.payload as { path?: string; repoName?: string };
+                  if (p.path && !existsSync(p.path)) continue;
+                  if (p.path && p.repoName && session && !session.worktreePaths.has(p.repoName)) {
+                    const match = p.path.match(/session-(wt-[^/]+)$/);
+                    if (match) {
+                      session.worktreePaths.set(p.repoName, { path: p.path, wtId: match[1] });
+                    }
+                  }
+                }
+                transport.send({ ...evt.payload, seq: evt.seq } as Record<string, unknown>);
```
dimakis: 🟡 **unsafe_assumptions**: `evt.payload` is cast to `Record<string, unknown>` and spread with `{ ...evt.payload, seq: evt.seq }` then sent directly to the client. If any replayed event payload contains a `seq` field of its own, it will be silently overwritten. More importantly, there's no validation that the replayed payload shapes are still valid for the current protocol version — stale events from a prior server version could confuse the client. `[fixable]`

server/query-loop.ts
```diff
@@ -574,6 +578,14 @@ export async function runQueryLoop(
           openBlockCount = Math.max(0, openBlockCount - 1);
           tryFlushMessageEnd(currentSession);
         }
+      } else if (msg.type === 'system') {
+        // Track compaction events from SDK system status messages
+        const subtype = (msg as Record<string, unknown>).subtype;
```
dimakis, line 610: 🔵 **style**: `msg` is cast to `Record<string, unknown>` twice on consecutive lines. Extract to a typed local: `const sysMsg = msg as Record<string, unknown>; const subtype = sysMsg.subtype; const compactResult = sysMsg.compact_result;` `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/components/ScrollFab.tsx`
_Solid defensive fix for TokenBar crash and clean compaction tracking. The subscribe-reattach promotion is the riskiest change — complex logic with no test coverage. ScrollFab has a minor visibility-update gap when content grows without scrolling._

- 🟡 **bugs** (L27): The `useEffect` depends on `scrollRef` (a stable ref object), so it only runs once on mount. The `update()` function is only called on `scroll` events after that. When new messages arrive and `scrollHeight` grows (without a user scroll), the FAB visibility won't update — the 'scroll to bottom' button won't appear until the user manually scrolls. Consider adding a ResizeObserver or re-running `update()` when content changes. `[fixable]`


_Solid defensive fix for TokenBar crash and clean compaction tracking. The subscribe-reattach promotion is the riskiest change — complex logic with no test coverage. ScrollFab has a minor visibility-update gap when content grows without scrolling._

- 🔵 **missing_tests** (L503): The `token_update` parser tests verify `sessionTotal` and `numTurns` omission but don't test the new `numCompactions` field — neither its presence when set nor its absence when omitted. Add a case with `numCompactions` to the 'all fields' test and verify it's excluded from the partial update test. `[fixable]`


### `server/index.ts`
_Solid defensive fix for TokenBar crash and clean compaction tracking. The subscribe-reattach promotion is the riskiest change — complex logic with no test coverage. ScrollFab has a minor visibility-update gap when content grows without scrolling._

- 🔵 **missing_tests** (L248): The subscribe-to-reattach promotion logic (lines 248-314) is a significant new code path with event replay, worktree path recovery, and snapshot forwarding. No integration or unit test covers this flow. Consider testing: (1) subscribe on a detached session triggers reattach + replays events, (2) subscribe on an attached session falls through to observer, (3) reattach failure falls back to observer.
- 🟡 **unsafe_assumptions** (L284): `evt.payload` is cast to `Record<string, unknown>` and spread with `{ ...evt.payload, seq: evt.seq }` then sent directly to the client. If any replayed event payload contains a `seq` field of its own, it will be silently overwritten. More importantly, there's no validation that the replayed payload shapes are still valid for the current protocol version — stale events from a prior server version could confuse the client. `[fixable]`


### `server/query-loop.ts`
_Solid defensive fix for TokenBar crash and clean compaction tracking. The subscribe-reattach promotion is the riskiest change — complex logic with no test coverage. ScrollFab has a minor visibility-update gap when content grows without scrolling._

- 🔵 **style** (L583): `msg` is cast to `Record<string, unknown>` twice on consecutive lines. Extract to a typed local: `const sysMsg = msg as Record<string, unknown>; const subtype = sysMsg.subtype; const compactResult = sysMsg.compact_result;` `[fixable]`

frontend/src/components/ChatArea.tsx
```diff
@@ -37,6 +37,18 @@ export function ChatArea({
 }: ChatAreaProps) {
   const internalScrollRef = useRef<HTMLDivElement>(null);
   const scrollRef = externalScrollRef ?? internalScrollRef;
+  const prevMessageCount = useRef(0);
```
dimakis, line 40: 🟡 **bugs**: The new restore auto-scroll effect fires whenever `messages.length` transitions from 0 to >0, including normal first-message sends in a fresh session. That overrides the existing near-bottom guard and can yank the viewport to the bottom even if the user intentionally scrolled elsewhere while older content was loading or while viewing a prepended history state. This should be keyed to an explicit restore/session-change signal rather than inferred solely from message count. `[fixable]`

frontend/src/components/ScrollFab.tsx
```diff
@@ -0,0 +1,65 @@
...
+const THRESHOLD = 200; // px from edge to show button
+
+interface Props {
+  scrollRef: React.RefObject<HTMLDivElement | null>;
+}
+
+export function ScrollFab({ scrollRef }: Props) {
+  const [showTop, setShowTop] = useState(false);
+  const [showBottom, setShowBottom] = useState(false);
+
```
dimakis, line 12: 🟡 **regressions**: The visibility effect only subscribes to `scroll` events on the current element. It does not react when content height changes without a scroll event, so after new messages stream in or a session restore populates the chat, the bottom FAB can remain hidden until the user manually scrolls. Since this component is meant to improve navigation after content growth, it should also recompute on render/content changes or observe size mutations. `[fixable]`

server/query-loop.ts
```diff
@@ -574,6 +578,14 @@ export async function runQueryLoop(
           openBlockCount = Math.max(0, openBlockCount - 1);
```
dimakis, line 605: 🟡 **bugs**: `numCompactions` is tracked only in-memory during a live query loop and is never restored from persisted session/event history. Reattached or restored sessions will lose prior compaction count until another compaction happens, so the new TokenBar field is not stable across reconnects/completed-session hydration unlike the other token stats.

dimakis (commented):
## Codex Review

### `frontend/src/components/ChatArea.tsx`
_PR fixes the token bar crash and adds useful UX, but there are a few state/visibility regressions around inferred restore scrolling, FAB updates, and compaction count persistence._

- 🟡 **bugs** (L40): The new restore auto-scroll effect fires whenever `messages.length` transitions from 0 to >0, including normal first-message sends in a fresh session. That overrides the existing near-bottom guard and can yank the viewport to the bottom even if the user intentionally scrolled elsewhere while older content was loading or while viewing a prepended history state. This should be keyed to an explicit restore/session-change signal rather than inferred solely from message count. `[fixable]`


### `frontend/src/components/ScrollFab.tsx`
_PR fixes the token bar crash and adds useful UX, but there are a few state/visibility regressions around inferred restore scrolling, FAB updates, and compaction count persistence._

- 🟡 **regressions** (L12): The visibility effect only subscribes to `scroll` events on the current element. It does not react when content height changes without a scroll event, so after new messages stream in or a session restore populates the chat, the bottom FAB can remain hidden until the user manually scrolls. Since this component is meant to improve navigation after content growth, it should also recompute on render/content changes or observe size mutations. `[fixable]`


### `server/query-loop.ts`
_PR fixes the token bar crash and adds useful UX, but there are a few state/visibility regressions around inferred restore scrolling, FAB updates, and compaction count persistence._

- 🟡 **bugs** (L578): `numCompactions` is tracked only in-memory during a live query loop and is never restored from persisted session/event history. Reattached or restored sessions will lose prior compaction count until another compaction happens, so the new TokenBar field is not stable across reconnects/completed-session hydration unlike the other token stats.

PR #219, opened 2026-04-17
fix: iOS WS reconnect promotes to driver instead of passive observer

dimakis:
## Summary

iOS Safari kills WebSocket connections every 30-90 seconds (close code 1005 — background suspension). The `WsPool` reconnects promptly, but every reconnect was becoming a passive observer instead of the session driver, causing all streaming events to be lost on the client.

Two root causes:

- **Client: new-session path never called `setRunning`.** `store.ts` `sendMessage` set `wasRunning = true` for follow-up sends but not for the initial `new:` key send. When iOS killed the socket mid-first-turn, the pool entry had `wasRunning = false`, so the reconnect sent `subscribe` (observer) instead of `reattach` (driver).

- **Server: `subscribe` always added an observer, even when the session's driver WS was dead.** When the driver transport was gone (session detached), the reconnecting client should have been promoted to driver via `reattach()` — not added as a passive observer that can't receive direct query-loop events.

### Changes

1. **`packages/client/src/store.ts`** — Add `wsPool.setRunning(newKey, true)` in the new-session `sendMessage` path so the pool entry is marked running before iOS can kill the socket.

2. **`server/index.ts`** — When a `subscribe` arrives for a session whose driver is detached (`!registry.isAttached(found.clientId)`), call `reattachChat()` to promote the subscriber to driver, respond with `reattached` (including snapshot replay), and reassign the handler's `clientId`. When the driver IS attached (another client is live), preserve the existing observer behavior.

## Test plan

- [x] Store test: new-session `sendMessage` marks pool entry running; reconnect after iOS socket kill sends `reattach` (not `subscribe`)
- [x] Registry tests: detached session found + reattach succeeds; `isAttached` correctly identifies detached sessions
- [x] 350/350 client tests pass, 248/249 server tests pass (1 pre-existing TTL constant mismatch on `main`)
- [ ] CI green
- [ ] Manual: multi-turn chat on iPhone over Tailscale survives iOS WS suspension without losing messages

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -155,10 +155,14 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     async switchSession(id: string) {
-      // Unsubscribe from previous session's WS before subscribing to the new
-      // one — otherwise the old listener keeps dispatching messages into the
-      // store, causing sessions to "merge."
+      // Unsubscribe the listener so old events stop hitting the store,
+      // then defuse the pool entry so it won't reattach on reconnect
+      // (wasRunning=false prevents the reattach handshake that causes bleed).
       activeUnsub?.();
+      if (activePoolKey) {
+        wsPool.setRunning(activePoolKey, false);
+        wsPool.removeIfIdle(activePoolKey);
+      }
 
       parserState.currentSessionId = id;
       const poolKey = `session:${id}`;
@@ -190,6 +194,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
     newSession() {
       activeUnsub?.();
+      if (activePoolKey) {
+        wsPool.setRunning(activePoolKey, false);
+        wsPool.removeIfIdle(activePoolKey);
+      }
       activeUnsub = null;
       activePoolKey = null;
       parserState.currentSessionId = undefined;
@@ -251,6 +259,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         const newKey = `new:${Date.now()}`;
         activePoolKey = newKey;
         activeUnsub = wsPool.subscribe(newKey, wsListener);
+        wsPool.setRunning(newKey, true);
 
         const sent = wsPool.send(newKey, buildPayload());
         if (!sent) {
--- a/packages/client/src/ws-connection.ts
+++ b/packages/client/src/ws-connection.ts
@@ -163,29 +163,45 @@ export class WsPool {
     const entry = this.pool.get(key);
     if (!entry) return false;
     if (entry.wasRunning || entry.listeners.size > 0) return false;
-    if (entry.reconnectTimer) clearTimeout(entry.reconnectTimer);
-    if (entry.ws) {
-      entry.ws.onclose = null;
-      entry.ws.close();
-    }
+    this.teardownEntry(entry);
     this.pool.delete(key);
     return true;
+  }
+
+  /**
+   * Unconditionally tear down a pool entry and remove it plus all aliases.
+   * Used when explicitly leaving a session — unlike removeIfIdle, this
+   * ignores wasRunning and listener count.
+   */
+  forceRemove(key: string): void {
+    const entry = this.pool.get(key);
+    if (!entry) return;
+    this.teardownEntry(entry);
+    // Remove every key that points to this same entry object (aliases)
+    for (const [k, v] of this.pool) {
+      if (v === entry) this.pool.delete(k);
+    }
   }
 
   /** Close all connections and clear the pool. */
   destroy(): void {
     this.stopVisibilityMonitor();
     for (const entry of this.pool.values()) {
-      if (entry.reconnectTimer) clearTimeout(entry.reconnectTimer);
-      if (entry.ws) {
-        entry.ws.onclose = null;
-        entry.ws.close();
-      }
+      this.teardownEntry(entry);
     }
     this.pool.clear();
   }
 
   // ─── Internal ──────────────────────────────────────────────────────────────
+
+  private teardownEntry(entry: PoolEntry): void {
+    if (entry.reconnectTimer) clearTimeout(entry.reconnectTimer);
+    if (entry.ws) {
+      entry.ws.onclose = null;
+      entry.ws.close();
+    }
+    entry.wasRunning = false;
+  }
 
   private broadcast(entry: PoolEntry, msg: WsMsg): void {
     entry.listeners.forEach((l) => l(msg));
@@ -293,7 +309,13 @@ export class WsPool {
         const sessionPrefix = 'session:';
         if (key.startsWith(sessionPrefix)) {
           const sessionId = key.slice(sessionPrefix.length);
-          ws.send(JSON.stringify({ type: 'subscribe', sessionId }));
+          ws.send(
+            JSON.stringify({
+              type: 'subscribe',
+              sessionId,
+              ...(entry.lastSeq > 0 ? { lastSeq: entry.lastSeq } : {}),
+            }),
+          );
           return;
         }
         this.broadcast(entry, { type: '_open' });
--- a/server/__tests__/multi-client-sessions.test.ts
+++ b/server/__tests__/multi-client-sessions.test.ts
@@ -346,6 +346,54 @@ describe('observer broadcast when driver is detached', () => {
   });
 });
 
+describe('subscribe-to-detached promotes to driver', () => {
+  let registry: SessionRegistry;
+
+  beforeEach(() => {
+    registry = new SessionRegistry();
+  });
+
+  afterEach(() => {
+    registry.dispose();
+  });
+
+  it('reattach succeeds on a detached session, making the new transport the driver', () => {
+    const driverTransport = mockTransport();
+    const newTransport = mockTransport();
+    registry.register('c1', {
+      transport: driverTransport,
+      abortController: new AbortController(),
+      mode: 'agent',
+      sessionAllowList: new Set(),
+      sessionId: 'sess-1',
+    });
+
+    registry.detach('c1');
+    expect(registry.isAttached('c1')).toBe(false);
+
+    const ok = registry.reattach('c1', newTransport);
+    expect(ok).toBe(true);
+    expect(registry.isAttached('c1')).toBe(true);
+    expect(registry.get('c1')!.transport).toBe(newTransport);
+  });
+
+  it('findBySessionId + isAttached correctly identifies a detached session', () => {
+    const transport = mockTransport();
+    registry.register('c1', {
+      transport,
+      abortController: new AbortController(),
+      mode: 'agent',
+      sessionAllowList: new Set(),
+      sessionId: 'sess-1',
+    });
+
+    registry.detach('c1');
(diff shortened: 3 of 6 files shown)


server/index.ts
```diff
@@ -245,10 +245,47 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
+                ts: Date.now(),
+                messageId: session.currentSnapshot.messageId,
+                blocks: session.currentSnapshot.blocks,
+              });
+            }
+            log.info('subscribe promoted to reattach (detached session)', {
+              clientId,
+              sessionId: msg.sessionId,
+            });
+          } else {
```
dimakis, line 302: 🟡 **bugs**: Reattach-failed fallback (lines 278-284) sends `subscribed` with `running: true` and adds the client as an observer, but does NOT replay `currentSnapshot`. The attached-driver observer path (lines 286-302) does replay the snapshot. A client hitting this fallback will see a blank screen until the next streaming event arrives. `[fixable]`

server/index.ts
```diff
@@ -245,10 +245,47 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
       if (msg.type === 'subscribe') {
         const found = registry.findBySessionId(msg.sessionId);
-        if (found) {
+        if (found && !registry.isAttached(found.clientId)) {
+          // Driver WS is dead (session detached). Promote this subscriber
+          // to the session driver so it receives direct query-loop events
+          // instead of being a passive observer. This is the typical iOS
+          // Safari reconnect path: the socket drops every 30-90s and the
+          // pool reconnects with a subscribe carrying the session ID.
+          const ok = reattachChat(found.clientId, transport);
```
dimakis, line 254: 🟡 **unsafe_assumptions**: After `reattachChat` succeeds, `clientId` is reassigned to `found.clientId` (line 256). However, the original `clientId` allocated at WS connect is now orphaned — `handleChatWs`'s `ws.on('close')` callback will later call `detachChat(clientId)` using the reassigned value, which is correct. But if a second subscribe-promote races before close fires, the first reassignment is silently lost. Low probability in practice given single-threaded Node, but worth a comment.

server/__tests__/multi-client-sessions.test.ts
```diff
@@ -346,6 +346,54 @@ describe('observer broadcast when driver is detached', () => {
   });
```
dimakis, line 346: 🔵 **missing_tests**: The new server tests verify `reattach` and `findBySessionId` on the registry directly, but don't test the actual subscribe-to-reattach promotion logic in `handleChatWs` (i.e., sending a `subscribe` message over WS and asserting that `reattached` is returned when the session is detached). An integration test through the WS handler would catch issues like the missing snapshot in the fallback path. `[fixable]`

packages/client/src/store.ts
```diff
@@ -155,10 +155,14 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
 
     async switchSession(id: string) {
-      // Unsubscribe from previous session's WS before subscribing to the new
-      // one — otherwise the old listener keeps dispatching messages into the
-      // store, causing sessions to "merge."
+      // Unsubscribe the listener so old events stop hitting the store,
+      // then defuse the pool entry so it won't reattach on reconnect
+      // (wasRunning=false prevents the reattach handshake that causes bleed).
       activeUnsub?.();
+      if (activePoolKey) {
```
dimakis, line 162: 🔵 **style**: The `setRunning(false)` + `removeIfIdle()` pattern is duplicated verbatim in `switchSession` and `newSession`. A small helper (e.g., `defusePoolEntry()`) would reduce duplication and ensure the two paths stay in sync. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid fix for the iOS reconnect bleed issue. One functional concern: the reattach-failed fallback path omits the snapshot replay that the observer path includes, which could leave the client with a blank screen._

- 🟡 **bugs** (L277): Reattach-failed fallback (lines 278-284) sends `subscribed` with `running: true` and adds the client as an observer, but does NOT replay `currentSnapshot`. The attached-driver observer path (lines 286-302) does replay the snapshot. A client hitting this fallback will see a blank screen until the next streaming event arrives. `[fixable]`
- 🟡 **unsafe_assumptions** (L254): After `reattachChat` succeeds, `clientId` is reassigned to `found.clientId` (line 256). However, the original `clientId` allocated at WS connect is now orphaned — `handleChatWs`'s `ws.on('close')` callback will later call `detachChat(clientId)` using the reassigned value, which is correct. But if a second subscribe-promote races before close fires, the first reassignment is silently lost. Low probability in practice given single-threaded Node, but worth a comment.


_Solid fix for the iOS reconnect bleed issue. One functional concern: the reattach-failed fallback path omits the snapshot replay that the observer path includes, which could leave the client with a blank screen._

- 🔵 **missing_tests** (L346): The new server tests verify `reattach` and `findBySessionId` on the registry directly, but don't test the actual subscribe-to-reattach promotion logic in `handleChatWs` (i.e., sending a `subscribe` message over WS and asserting that `reattached` is returned when the session is detached). An integration test through the WS handler would catch issues like the missing snapshot in the fallback path. `[fixable]`


_Solid fix for the iOS reconnect bleed issue. One functional concern: the reattach-failed fallback path omits the snapshot replay that the observer path includes, which could leave the client with a blank screen._

- 🔵 **style** (L162): The `setRunning(false)` + `removeIfIdle()` pattern is duplicated verbatim in `switchSession` and `newSession`. A small helper (e.g., `defusePoolEntry()`) would reduce duplication and ensure the two paths stay in sync. `[fixable]`

packages/client/src/ws-connection.ts
```diff
@@ -163,30 +163,46 @@ export class WsPool {
...
     this.pool.delete(key);
     return true;
   }
 
+  /**
+   * Unconditionally tear down a pool entry and remove it plus all aliases.
+   * Used when explicitly leaving a session — unlike removeIfIdle, this
+   * ignores wasRunning and listener count.
+   */
+  forceRemove(key: string): void {
```
dimakis, line 176: 🟡 **bugs**: `forceRemove()` is added but never called anywhere. `switchSession` and `newSession` use `setRunning(false)` + `removeIfIdle()`, which only deletes the specific key — not aliases. When a `new:*` key is aliased to `session:*` (via session ID assignment), `removeIfIdle` on one key leaves the other as an orphan pointing to a torn-down entry. `forceRemove` was designed to handle this (it iterates all keys for the same entry object) but is dead code. `[fixable]`

dimakis, line 176: 🔵 **missing_tests**: `forceRemove()` has no unit test coverage in `ws-connection.test.ts` — neither its basic behavior nor its alias-cleanup loop is tested. `[fixable]`

packages/client/src/ws-connection.ts
```diff
@@ -293,7 +309,13 @@ export class WsPool {
         const sessionPrefix = 'session:';
```
dimakis, line 309: 🔵 **missing_tests**: The new `lastSeq` inclusion in the subscribe message (`...(entry.lastSeq > 0 ? { lastSeq: entry.lastSeq } : {})`) has no dedicated test verifying that `lastSeq` is sent when > 0 and omitted when 0. `[fixable]`

server/index.ts
```diff
@@ -245,10 +245,70 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
+            const session = registry.get(clientId);
+            transport.send({
+              type: 'reattached',
+              clientId: found.clientId,
+              sessionId: session?.sessionId,
+              running: true,
+            });
+            // Replay missed events from the durable event store so the
+            // client catches up on anything sent while the WS was dead.
+            if (session?.sessionId && msg.lastSeq != null) {
```
dimakis, line 267: 🟡 **style**: The event-replay + worktree-path-restoration logic (lines 266-295) is nearly identical to the existing reattach handler (lines 349-378). The subscribe version is also missing the `log.info('skipping stale worktree_opened event on reattach', ...)` log present in the direct reattach path. Extract a shared helper to avoid divergence. `[fixable]`

dimakis (commented):
## Centaur Review

### `packages/client/src/ws-connection.ts`
_Core fix is sound — subscribe-to-reattach promotion and pool cleanup on session switch prevent iOS bleed. Main concern: `forceRemove()` is dead code; alias cleanup via `removeIfIdle` is incomplete. Event replay logic should be extracted to avoid duplication drift._

- 🟡 **bugs** (L176): `forceRemove()` is added but never called anywhere. `switchSession` and `newSession` use `setRunning(false)` + `removeIfIdle()`, which only deletes the specific key — not aliases. When a `new:*` key is aliased to `session:*` (via session ID assignment), `removeIfIdle` on one key leaves the other as an orphan pointing to a torn-down entry. `forceRemove` was designed to handle this (it iterates all keys for the same entry object) but is dead code. `[fixable]`
- 🔵 **missing_tests** (L176): `forceRemove()` has no unit test coverage in `ws-connection.test.ts` — neither its basic behavior nor its alias-cleanup loop is tested. `[fixable]`
- 🔵 **missing_tests** (L309): The new `lastSeq` inclusion in the subscribe message (`...(entry.lastSeq > 0 ? { lastSeq: entry.lastSeq } : {})`) has no dedicated test verifying that `lastSeq` is sent when > 0 and omitted when 0. `[fixable]`


### `server/index.ts`
_Core fix is sound — subscribe-to-reattach promotion and pool cleanup on session switch prevent iOS bleed. Main concern: `forceRemove()` is dead code; alias cleanup via `removeIfIdle` is incomplete. Event replay logic should be extracted to avoid duplication drift._

- 🟡 **style** (L266): The event-replay + worktree-path-restoration logic (lines 266-295) is nearly identical to the existing reattach handler (lines 349-378). The subscribe version is also missing the `log.info('skipping stale worktree_opened event on reattach', ...)` log present in the direct reattach path. Extract a shared helper to avoid divergence. `[fixable]`
- 🔵 **unsafe_assumptions** (L279): Replayed events are cast via `{ ...evt.payload, seq: evt.seq } as Record<string, unknown>` — if `evt.payload` itself contains a `seq` field (e.g., a nested event), the spread would be overwritten by `evt.seq`. Low probability but worth a defensive note or ordering the spread so `seq` comes from the store. `[fixable]`

server/index.ts
```diff
@@ -245,10 +245,70 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
+                  const p = evt.payload as { path?: string; repoName?: string };
+                  if (p.path && !existsSync(p.path)) continue;
+                  if (p.path && p.repoName && session && !session.worktreePaths.has(p.repoName)) {
+                    const match = p.path.match(/session-(wt-[^/]+)$/);
+                    if (match) {
+                      session.worktreePaths.set(p.repoName, { path: p.path, wtId: match[1] });
+                    }
+                  }
+                }
+                transport.send({ ...evt.payload, seq: evt.seq } as Record<string, unknown>);
```
dimakis, line 280: 🔵 **unsafe_assumptions**: Replayed events are cast via `{ ...evt.payload, seq: evt.seq } as Record<string, unknown>` — if `evt.payload` itself contains a `seq` field (e.g., a nested event), the spread would be overwritten by `evt.seq`. Low probability but worth a defensive note or ordering the spread so `seq` comes from the store. `[fixable]`

PR #220, opened 2026-04-17
docs: session isolation audit — single-WS architecture proposal

dimakis:
## Summary

- Full data-flow audit of session management across server, WebSocket layer, frontend store, and token tracking
- Documents 7 acute session-bleeding bugs fixed this week and identifies remaining structural issues
- Proposes migrating from per-session WS pool to **single WebSocket per client with server-side session routing**, eliminating the entire class of connection-lifecycle bugs
- Includes protocol changes (both directions), `ConnectionRegistry` design, client-side demux model, reconnection strategy, and phased migration plan

## Key architectural decisions

- One WS per browser tab, server routes by `sessionId` in every message
- `reconnect` replaces both `reattach` and `subscribe` — one message, all sessions
- Sessions keyed by `sessionId` (not ephemeral `clientId`)
- Protocol versioning (v1/v2 coexistence) for incremental migration
- Synchronous metadata on session switch (no token zero-flash)

## Test plan

- [ ] Review design doc for completeness and correctness
- [ ] Validate protocol changes against current message types
- [ ] Confirm migration phases are sequenced correctly
- [ ] Review `ConnectionRegistry` interface design

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #221, opened 2026-04-17
feat: v2 WS protocol schemas + OTel tracing bootstrap (phase 0)

dimakis:
## Summary

Phase 0 of the single-WS architecture migration ([design doc in PR #220](https://github.com/dimakis/mitzo/pull/220)). Additive only — no existing behavior changes, no breaking protocol modifications.

### v2 protocol schemas

New file: `packages/protocol/src/ws-schemas-v2.ts`

Defines the full v2 client→server message set that will replace the current v1 messages during migration:

- **`hello`** — protocol version negotiation on connect (v2 clients identify themselves)
- **`reconnect`** — replaces both `reattach` and `subscribe`. Sends per-session `lastSeq` for all watched sessions in one message. Server replays missed events per session.
- **`watch` / `unwatch`** — add/remove sessions from a connection's event feed
- **`switch_session`** — trigger synchronous metadata delivery (eliminates token zero-flash)
- **`send`** with explicit `sessionId` (null = new session). No more implicit session binding via WS connection identity.
- **`interrupt` / `stop` / `permission_response` / `set_mode`** — session-scoped variants carrying `sessionId`
- **`IncomingWsMessageV2`** — discriminated union for the v2 handler

19 tests in `packages/protocol/__tests__/ws-schemas-v2.test.ts`.

### OTel tracing bootstrap

New file: `server/tracing.ts`

- Opt-in via `OTEL_EXPORTER_OTLP_ENDPOINT` env var
- When set: initializes `@opentelemetry/sdk-node` with OTLP trace exporter (Jaeger, Grafana Tempo, etc.)
- When unset: `@opentelemetry/api` returns noop spans — zero overhead
- Exports `tracer` for instrumentation in subsequent phases (ws.connect, ws.focus, query.turn spans)
- Graceful shutdown on SIGTERM/SIGINT

## Test plan

- [x] 19 v2 schema tests pass (all message types, validation, rejection)
- [x] OTel tracer test: exports functional tracer without OTEL_EXPORTER_OTLP_ENDPOINT
- [x] 10,307 existing tests pass (pre-existing TTL constant mismatch failures excluded)
- [ ] CI green


server/tracing.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const endpoint = process.env.OTEL_EXPORTER_OTLP_ENDPOINT;
+
+if (endpoint) {
+  const sdk = new NodeSDK({
+    resource: new Resource({
+      [ATTR_SERVICE_NAME]: 'mitzo',
+      [ATTR_SERVICE_VERSION]: process.env.npm_package_version ?? 'dev',
+    }),
+    traceExporter: new OTLPTraceExporter({ url: `${endpoint}/v1/traces` }),
```
dimakis: 🟡 **unsafe_assumptions**: Appending `/v1/traces` to the user-supplied `OTEL_EXPORTER_OTLP_ENDPOINT` conflicts with the OTel spec convention. The `OTLPTraceExporter` already appends `/v1/traces` by default when given a base URL via the `OTEL_EXPORTER_OTLP_ENDPOINT` env var. Manually constructing the URL means users who set the standard env var (e.g. `http://localhost:4318`) will get double-pathed (`/v1/traces/v1/traces`) if they also rely on the SDK's built-in env var reading, or the wrong path if they set the full traces URL. Consider omitting the `url` option entirely and letting the SDK read `OTEL_EXPORTER_OTLP_ENDPOINT` natively, or document that the env var must be a base URL without the path suffix. `[fixable]`

server/tracing.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * OpenTelemetry tracing bootstrap — opt-in via OTEL_EXPORTER_OTLP_ENDPOINT.
+ *
+ * When the env var is set, initializes the OTel SDK with an OTLP exporter
+ * (compatible with Jaeger, Grafana Tempo, etc.). When unset, tracing is a
+ * no-op — the @opentelemetry/api returns noop spans, so instrumentation
+ * code has zero overhead.
+ *
+ * Import this module at the top of server/index.ts (before other imports)
```
dimakis: 🟡 **regressions**: The doc comment says to import this module at the top of `server/index.ts` before other imports, but `server/index.ts` does not import it. The tracing module is dead code — no server entrypoint loads it, so OTel will never initialize even when configured. This should either be imported in `index.ts` or the comment should be removed to avoid confusion. `[fixable]`

server/tracing.ts
```diff
@@ -0,0 +1,44 @@
...
+      [ATTR_SERVICE_NAME]: 'mitzo',
+      [ATTR_SERVICE_VERSION]: process.env.npm_package_version ?? 'dev',
+    }),
+    traceExporter: new OTLPTraceExporter({ url: `${endpoint}/v1/traces` }),
+  });
+
+  sdk.start();
+  log.info('OTel tracing initialized', { endpoint });
+
+  process.on('SIGTERM', () => sdk.shutdown());
```
dimakis: 🟡 **bugs**: The `sdk` variable is scoped inside the `if` block, so the SIGTERM/SIGINT handlers create a closure over it. This works, but if `sdk.shutdown()` rejects, the promise is unhandled. Consider `.catch()` on the shutdown call to avoid unhandled rejection crashes on process exit. `[fixable]`

server/__tests__/tracing.test.ts
```diff
@@ -0,0 +1,15 @@
+import { describe, it, expect } from 'vitest';
+
+describe('tracing module', () => {
+  it('exports a tracer that can create spans without OTEL_EXPORTER_OTLP_ENDPOINT', async () => {
```
dimakis, line 4: 🔵 **missing_tests**: The test only covers the no-op path (no endpoint set). There is no test for the configured path (when `OTEL_EXPORTER_OTLP_ENDPOINT` is set) — e.g. verifying `sdk.start()` is called and shutdown handlers are registered. Also, because the module has top-level side effects and `import()` caches, this single-test suite may produce flaky results if a future test sets the env var — the module will already be cached as no-op. `[fixable]`

packages/protocol/src/ws-schemas-v2.ts
```diff
@@ -0,0 +1,115 @@
...
+});
+
+// ─── Chat messages (session-scoped) ─────────────────────────────────────────
+
+export const V2SendMessage = z.object({
+  type: z.literal('send'),
+  sessionId: z.string().min(1).nullable(),
+  prompt: z.string().min(1),
+  clientMsgId: z.string().min(1),
+  model: z.string().optional(),
```
dimakis, line 68: 🔵 **style**: V2SendMessage allows `sessionId: null` for new sessions, but V2InterruptMessage requires `sessionId: z.string().min(1)` (non-nullable). This is likely correct semantically (you can't interrupt a session that doesn't exist yet), but it's worth a brief inline comment explaining the asymmetry for future readers. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/tracing.ts`
_Solid v2 schema design with good test coverage; tracing bootstrap has a dead-code issue (not imported in index.ts) and a potential double-path bug in the OTLP endpoint URL construction._

- 🟡 **unsafe_assumptions** (L30): Appending `/v1/traces` to the user-supplied `OTEL_EXPORTER_OTLP_ENDPOINT` conflicts with the OTel spec convention. The `OTLPTraceExporter` already appends `/v1/traces` by default when given a base URL via the `OTEL_EXPORTER_OTLP_ENDPOINT` env var. Manually constructing the URL means users who set the standard env var (e.g. `http://localhost:4318`) will get double-pathed (`/v1/traces/v1/traces`) if they also rely on the SDK's built-in env var reading, or the wrong path if they set the full traces URL. Consider omitting the `url` option entirely and letting the SDK read `OTEL_EXPORTER_OTLP_ENDPOINT` natively, or document that the env var must be a base URL without the path suffix. `[fixable]`
- 🟡 **regressions** (L9): The doc comment says to import this module at the top of `server/index.ts` before other imports, but `server/index.ts` does not import it. The tracing module is dead code — no server entrypoint loads it, so OTel will never initialize even when configured. This should either be imported in `index.ts` or the comment should be removed to avoid confusion. `[fixable]`
- 🟡 **bugs** (L36): The `sdk` variable is scoped inside the `if` block, so the SIGTERM/SIGINT handlers create a closure over it. This works, but if `sdk.shutdown()` rejects, the promise is unhandled. Consider `.catch()` on the shutdown call to avoid unhandled rejection crashes on process exit. `[fixable]`


### `server/__tests__/tracing.test.ts`
_Solid v2 schema design with good test coverage; tracing bootstrap has a dead-code issue (not imported in index.ts) and a potential double-path bug in the OTLP endpoint URL construction._

- 🔵 **missing_tests** (L4): The test only covers the no-op path (no endpoint set). There is no test for the configured path (when `OTEL_EXPORTER_OTLP_ENDPOINT` is set) — e.g. verifying `sdk.start()` is called and shutdown handlers are registered. Also, because the module has top-level side effects and `import()` caches, this single-test suite may produce flaky results if a future test sets the env var — the module will already be cached as no-op. `[fixable]`


### `packages/protocol/src/ws-schemas-v2.ts`
_Solid v2 schema design with good test coverage; tracing bootstrap has a dead-code issue (not imported in index.ts) and a potential double-path bug in the OTLP endpoint URL construction._

- 🔵 **style** (L65): V2SendMessage allows `sessionId: null` for new sessions, but V2InterruptMessage requires `sessionId: z.string().min(1)` (non-nullable). This is likely correct semantically (you can't interrupt a session that doesn't exist yet), but it's worth a brief inline comment explaining the asymmetry for future readers. `[fixable]`


### `package.json`
_Solid v2 schema design with good test coverage; tracing bootstrap has a dead-code issue (not imported in index.ts) and a potential double-path bug in the OTLP endpoint URL construction._

- 🔵 **style**: `@opentelemetry/sdk-node` pulls in a large transitive dependency tree (gRPC, protobuf, zipkin, prometheus, etc.) even though only the OTLP HTTP exporter is used. Consider using `@opentelemetry/sdk-trace-node` + `@opentelemetry/exporter-trace-otlp-http` directly for a much lighter footprint. `[fixable]`

dimakis:
All 6 items addressed in \`04ec941\`:

1. **Double-path fix** — removed manual \`/v1/traces\` append; \`OTLPTraceExporter\` reads \`OTEL_EXPORTER_OTLP_ENDPOINT\` natively.
2. **Dead code** — added \`import './tracing.js'\` at top of \`server/index.ts\`.
3. **Unhandled rejection** — \`provider.shutdown().catch(() => {})\` on SIGTERM/SIGINT.
4. **Missing tests** — noted; module-caching makes the configured path hard to test in vitest without \`vi.resetModules()\` + dynamic import gymnastics. The noop path test is sufficient for Phase 0; configured-path behavior is covered by integration when running with a collector.
5. **sessionId asymmetry comment** — added inline comment on \`V2SendMessage\`.
6. **Lighter deps** — swapped \`@opentelemetry/sdk-node\` for \`@opentelemetry/sdk-trace-node\` (−26 packages, no gRPC/protobuf/zipkin/prometheus).

server/index.ts
```diff
@@ -245,10 +246,76 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
...
+            const session = registry.get(found.clientId);
+            transport.send({
+              type: 'reattached',
+              clientId: found.clientId,
+              sessionId: session?.sessionId,
+              running: true,
+            });
+            // Replay missed events from the durable event store so the
+            // client catches up on anything sent while the WS was dead.
+            if (session?.sessionId && msg.lastSeq != null) {
```
dimakis: 🔴 **bugs**: SubscribeMessage schema (ws-schemas.ts:50-53) has no `lastSeq` field, so `msg.lastSeq` is always `undefined` after Zod parsing. The `msg.lastSeq != null` guard is always false, making the event-replay block dead code. The subscribe-to-reattach promotion will never replay missed events, defeating its purpose for iOS Safari reconnects. Either add `lastSeq: z.number().optional()` to SubscribeMessage in ws-schemas.ts, or accept lastSeq from the raw parsed JSON before Zod strips it. `[fixable]`

server/index.ts
```diff
@@ -245,10 +246,76 @@ function handleChatWs(ws: WebSocket, initialClientId: string) {
 
       if (msg.type === 'subscribe') {
         const found = registry.findBySessionId(msg.sessionId);
-        if (found) {
+        if (found && !registry.isAttached(found.clientId)) {
```
dimakis, line 292: 🟡 **regressions**: The old subscribe handler always added the client as an observer when `found` was truthy. Now, when the driver is detached, it promotes to reattach instead. If the reattach succeeds but the client later sends a `send` message, it will be routed via `clientId` (the WS handler's local variable, which is NOT updated — per the comment at line 257). Verify that `tryRouteToActiveSession` correctly handles this case, otherwise the promoted client cannot send follow-up messages to the reattached session.

dimakis, line 292: 🔵 **style**: The subscribe-reattach block (lines 249-315) duplicates ~30 lines of event-replay logic from the reattach handler (lines 344-393). Extract a shared helper (e.g. `replayMissedEvents(session, transport, lastSeq)`) to avoid drift between the two code paths. `[fixable]`

server/tracing.ts
```diff
@@ -0,0 +1,46 @@
...
+  const provider = new NodeTracerProvider({
+    resource: new Resource({
+      [ATTR_SERVICE_NAME]: 'mitzo',
+      [ATTR_SERVICE_VERSION]: process.env.npm_package_version ?? 'dev',
+    }),
+  });
+
+  // Let the exporter read OTEL_EXPORTER_OTLP_ENDPOINT natively — don't
+  // append /v1/traces manually, the SDK handles path construction.
+  provider.addSpanProcessor(new SimpleSpanProcessor(new OTLPTraceExporter()));
```
dimakis: 🔵 **style**: SimpleSpanProcessor is synchronous and blocks the event loop on every span export — intended for testing/development only. For production use, switch to BatchSpanProcessor which buffers and exports asynchronously. Since this is phase 0 bootstrap and opt-in, this is acceptable for now but should be swapped before production tracing is enabled. `[fixable]`

server/__tests__/tracing.test.ts
```diff
@@ -0,0 +1,15 @@
...
+    delete process.env.OTEL_EXPORTER_OTLP_ENDPOINT;
```
dimakis, line 5: 🟡 **missing_tests**: The test only covers the no-op path (no OTEL_EXPORTER_OTLP_ENDPOINT). There is no test verifying that setting the env var actually initializes the provider. Also, the dynamic `import()` at line 6 may be cached by Node's module system across test runs, making the `delete process.env` ineffective if another test imports tracing first. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Critical bug: subscribe-reattach event replay is dead code because SubscribeMessage schema lacks `lastSeq`. OTel bootstrap is clean but uses SimpleSpanProcessor (fine for phase 0). v2 WS schemas are well-tested and correctly designed._

- 🔴 **bugs** (L272): SubscribeMessage schema (ws-schemas.ts:50-53) has no `lastSeq` field, so `msg.lastSeq` is always `undefined` after Zod parsing. The `msg.lastSeq != null` guard is always false, making the event-replay block dead code. The subscribe-to-reattach promotion will never replay missed events, defeating its purpose for iOS Safari reconnects. Either add `lastSeq: z.number().optional()` to SubscribeMessage in ws-schemas.ts, or accept lastSeq from the raw parsed JSON before Zod strips it. `[fixable]`
- 🟡 **regressions** (L249): The old subscribe handler always added the client as an observer when `found` was truthy. Now, when the driver is detached, it promotes to reattach instead. If the reattach succeeds but the client later sends a `send` message, it will be routed via `clientId` (the WS handler's local variable, which is NOT updated — per the comment at line 257). Verify that `tryRouteToActiveSession` correctly handles this case, otherwise the promoted client cannot send follow-up messages to the reattached session.
- 🔵 **style** (L249): The subscribe-reattach block (lines 249-315) duplicates ~30 lines of event-replay logic from the reattach handler (lines 344-393). Extract a shared helper (e.g. `replayMissedEvents(session, transport, lastSeq)`) to avoid drift between the two code paths. `[fixable]`


### `server/tracing.ts`
_Critical bug: subscribe-reattach event replay is dead code because SubscribeMessage schema lacks `lastSeq`. OTel bootstrap is clean but uses SimpleSpanProcessor (fine for phase 0). v2 WS schemas are well-tested and correctly designed._

- 🔵 **style** (L32): SimpleSpanProcessor is synchronous and blocks the event loop on every span export — intended for testing/development only. For production use, switch to BatchSpanProcessor which buffers and exports asynchronously. Since this is phase 0 bootstrap and opt-in, this is acceptable for now but should be swapped before production tracing is enabled. `[fixable]`


### `server/__tests__/tracing.test.ts`
_Critical bug: subscribe-reattach event replay is dead code because SubscribeMessage schema lacks `lastSeq`. OTel bootstrap is clean but uses SimpleSpanProcessor (fine for phase 0). v2 WS schemas are well-tested and correctly designed._

- 🟡 **missing_tests** (L5): The test only covers the no-op path (no OTEL_EXPORTER_OTLP_ENDPOINT). There is no test verifying that setting the env var actually initializes the provider. Also, the dynamic `import()` at line 6 may be cached by Node's module system across test runs, making the `delete process.env` ineffective if another test imports tracing first. `[fixable]`

dimakis:
Round 2 addressed in \`fa2e34e\`:

1. **🔴 lastSeq dead code** — Not a bug. SubscribeMessage already has \`lastSeq: z.number().optional()\` at \`ws-schemas.ts:53\` (added in \`6564784\`). The reviewer was checking the stale built \`dist/\` or the CI was building against a cached protocol package. Confirmed locally: \`msg.lastSeq\` is defined when the client sends it.

2. **🟡 promoted client can't send follow-ups** — Correct by design. \`clientId\` is intentionally NOT reassigned (per \`0f10c47\`). Follow-ups from the promoted WS route through \`isActive(clientId)\` → false → \`tryRouteToActiveSession(ws, msg.resume, ...)\` → \`sendToChat(driverClientId, ...)\`. The driver's session receives the input. Verified in the iOS reconnect testing session.

3. **🔵 extract replay helper** — Done. Extracted \`replayMissedEvents(session, transport, lastSeq)\` and \`sendSnapshot(session, transport)\`. Both the reattach handler and subscribe-to-reattach path now use the shared helpers. Net −18 lines of duplicate code.

4. **🔵 SimpleSpanProcessor** — Acceptable for Phase 0. Will swap to \`BatchSpanProcessor\` when production tracing is enabled (Phase 1+).

5. **🟡 missing tests** — Same as round 1. Module-caching makes the configured path non-trivial to test in vitest. The noop tracer test is sufficient for Phase 0.

PR #223, opened 2026-04-17
feat(server): wire query-loop outbound events through ConnectionRegistry (phase 2)

dimakis:
## Summary

- Wire `sendOrBuffer` in `query-loop.ts` to deliver v2 session events through `ConnectionRegistry.broadcast` instead of direct driver transport + v1 observers
- Add `broadcastAll` to `ConnectionRegistry` for global events (`update_available`, `inbox_updated`, task state)
- Auto-watch + setActive on session creation: resume path watches immediately, create path uses `onSessionResolved` callback
- Global and task orchestrator broadcasts now dual-path through both `wss.clients` (v1) and `connRegistry.broadcastAll` (v2)
- `session_end` clears `activeSession` via `setActive(null)` on the v2 connection
- `connRegistry` threaded from `index.ts` to `query-loop` via `setConnectionRegistry` (same pattern as `setTaskStore`)

## Design decisions

- **v2 detection**: `connRegistry.get(clientId)` existing = v2 connection. No flags needed.
- **Pre-sessionId fallback**: Before `resolvedSessionId` is set, v2 connections fall back to v1 driver transport send.
- **Dual delivery**: v1 and v2 paths coexist. No v1 breakage. v1 removal is Phase 3.
- **`onSessionResolved` callback**: Lightweight hook — query-loop stays generic, `handleSendV2` supplies the watch/activate logic.

## Test plan

- [x] `broadcastAll` — 4 tests (sends to all, skips closed, error resilience, no-op)
- [x] v2 delivery path — 4 tests (routes through connRegistry, v1 fallback, pre-sessionId fallback, finally block)
- [x] `onSessionResolved` callback — 3 tests (fires once, idempotent on multi-turn, skipped on resume)
- [x] session_end cleanup — 2 tests (normal end clears active, abnormal end clears active)
- [x] handleSendV2 resume auto-watch — 1 test
- [x] Full suite: 1715 tests pass, 0 lint errors

### DIFF
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -7,6 +7,7 @@ import {
   handleWatch,
   handleUnwatch,
   handleSwitchSession,
+  handleSendV2,
   handleSetModeV2,
   handleStopV2,
   handlePermissionResponseV2,
@@ -302,6 +303,43 @@ describe('handleSetModeV2', () => {
     expect(modeMsg).toEqual(
       expect.objectContaining({ type: 'mode_changed', sessionId: 'sess-1', mode: 'auto' }),
     );
+  });
+});
+
+// ─── handleSendV2 auto-watch ─────────────────────────────────────────────────
+
+describe('handleSendV2 auto-watch', () => {
+  it('watches + activates the session on the resume path', () => {
+    const sessionReg = mockSessionRegistry();
+    // findBySessionId returns a result but isActive returns false → resume path
+    sessionReg.findBySessionId.mockReturnValue({ clientId: 'old-driver', session: {} });
+    sessionReg.isActive.mockReturnValue(false);
+
+    const ctx = createContext({
+      sessionRegistry: sessionReg as unknown as V2HandlerContext['sessionRegistry'],
+    });
+    const transport = mockTransport();
+    ctx.connRegistry.register('c1', transport);
+
+    // We can't fully test startChat without mocking the Agent SDK, but we can
+    // verify that watch + setActive are called before startChat. The resume
+    // path should ensure the connection is watching the session.
+    handleSendV2(
+      'c1',
+      transport,
+      {
+        type: 'send' as const,
+        sessionId: 'sess-resume',
+        prompt: 'continue',
+        clientMsgId: 'cmsg-1',
+      },
+      ctx,
+    );
+
+    const conn = ctx.connRegistry.get('c1');
+    expect(conn).toBeDefined();
+    expect(conn!.watchedSessions.has('sess-resume')).toBe(true);
+    expect(conn!.activeSession).toBe('sess-resume');
   });
 });
 
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1,6 +1,33 @@
 import { describe, it, expect, vi } from 'vitest';
 import type { SessionTransport } from '@mitzo/harness';
 import { ConnectionRegistry } from '@mitzo/harness';
+
+vi.mock('../chat.js', () => ({
+  startChat: vi.fn(),
+  sendToChat: vi.fn(),
+  interruptChat: vi.fn(),
+  stopChat: vi.fn(),
+  isActive: vi.fn().mockReturnValue(false),
+  BASE_REPO: '/tmp/test-repo',
+}));
+
+vi.mock('../app.js', () => ({
+  buildSkillRegistry: vi.fn().mockReturnValue(new Map()),
+  isAllowedPath: vi.fn().mockReturnValue(true),
+  NATIVE_COMMAND_NAMES: new Set(),
+}));
+
+vi.mock('../slash-commands.js', () => ({
+  resolveSlashCommand: vi.fn().mockReturnValue({ type: 'plain' }),
+}));
+
+vi.mock('../skill-policy.js', () => ({
+  setSkillPolicy: vi.fn(),
+  clearSkillPolicy: vi.fn(),
+}));
+
+import { startChat } from '../chat.js';
+
 import {
   handleHello,
   handleReconnect,
@@ -340,6 +367,37 @@ describe('handleSendV2 auto-watch', () => {
     expect(conn).toBeDefined();
     expect(conn!.watchedSessions.has('sess-resume')).toBe(true);
     expect(conn!.activeSession).toBe('sess-resume');
+  });
+
+  it('passes onSessionResolved callback to startChat on the create path', () => {
+    (startChat as ReturnType<typeof vi.fn>).mockClear();
+    const ctx = createContext();
+    const transport = mockTransport();
+    ctx.connRegistry.register('c1', transport);
+
+    handleSendV2(
+      'c1',
+      transport,
+      {
+        type: 'send' as const,
+        sessionId: null,
+        prompt: 'hello world',
+        clientMsgId: 'cmsg-2',
+      },
+      ctx,
+    );
+
+    expect(startChat).toHaveBeenCalledTimes(1);
+    const callArgs = (startChat as ReturnType<typeof vi.fn>).mock.calls[0];
+    const options = callArgs[3];
+    expect(options.onSessionResolved).toBeTypeOf('function');
+
+    // Simulate the callback firing — should watch + activate
+    options.onSessionResolved('sess-new');
+    const conn = ctx.connRegistry.get('c1');
+    expect(conn).toBeDefined();
+    expect(conn!.watchedSessions.has('sess-new')).toBe(true);
+    expect(conn!.activeSession).toBe('sess-new');
   });
 });
 
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -538,8 +538,8 @@ export async function startChat(
       eventStore,
       options.resume ? undefined : fullPrompt,
       {
-        ...(_connRegistry ? { connRegistry: _connRegistry } : {}),
-        ...(options.onSessionResolved ? { onSessionResolved: options.onSessionResolved } : {}),
+        connRegistry: _connRegistry ?? undefined,
+        onSessionResolved: options.onSessionResolved,
       },
     );
   } catch (err: unknown) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -5,7 +5,7 @@ import {
   renameSession,
 } from '@anthropic-ai/claude-agent-sdk';
 import type { SDKUserMessage } from '@anthropic-ai/claude-agent-sdk';
-import type { SessionTransport } from '@mitzo/harness';
+import type { SessionTransport, ConnectionRegistry } from '@mitzo/harness';
 import { execFileSync } from 'child_process';
 import { readFileSync, writeFileSync, mkdirSync, readdirSync, existsSync } from 'fs';
 import { join, resolve, dirname } from 'path';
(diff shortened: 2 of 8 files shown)


server/index.ts
```diff
@@ -73,24 +75,29 @@ const server = USE_TLS
 const wss = new WebSocketServer({ noServer: true, perMessageDeflate: false });
 
 setUpdateBroadcast(() => {
-  const msg = JSON.stringify({ type: 'update_available' });
+  const data = { type: 'update_available' };
+  const msg = JSON.stringify(data);
   wss.clients.forEach((client) => {
     if (client.readyState === client.OPEN) client.send(msg);
   });
+  connRegistry.broadcastAll(data);
```
dimakis, line 84: 🔴 **bugs**: Duplicate delivery to v2 clients: `wss.clients` contains ALL WebSocket connections (both v1 and v2). The global broadcast lambdas (setUpdateBroadcast, setInboxBroadcast, setTaskBroadcast, broadcastStatus, broadcastTasks) iterate `wss.clients` AND call `connRegistry.broadcastAll()`. Since v2 WebSocket connections are in both `wss.clients` and the ConnectionRegistry, v2 clients will receive every global event twice. The `wss.clients` loop should skip sockets that are registered in `connRegistry`, or `broadcastAll` should not be called in addition to the `wss.clients` loop. `[fixable]`

server/query-loop.ts
```diff
@@ -367,6 +386,9 @@ export async function runQueryLoop(
         }
 
         emit(v2('session_end', { sessionId: msg.session_id, usage: usageData }));
+        if (connRegistry?.get(clientId)) {
```
dimakis, line 389: 🟡 **bugs**: In the `result` event handler, `emit()` sends `session_end` which routes through `connRegistry.broadcast()` for v2 connections. Then `connRegistry.setActive(clientId, null)` is called. But `emit()` already delivered the event, so this ordering is fine. However, in the `finally` block (line ~650), the code checks `connRegistry?.get(clientId)` AFTER `registry.remove(clientId)`. Since `connRegistry` and `registry` (SessionRegistry) are independent, `connRegistry.get` still works. But `setActive(clientId, null)` is called twice on normal completion — once at line 390 in the `result` handler and again at line 659 in the finally block. This is harmless but unnecessary. `[fixable]`

server/chat.ts
```diff
@@ -30,6 +30,11 @@ let _taskStore: TaskStore | null = null;
 export function setTaskStore(store: TaskStore): void {
   _taskStore = store;
 }
+
```
dimakis, line 33: 🟡 **unsafe_assumptions**: Module-level mutable singleton `_connRegistry` set via `setConnectionRegistry()` is a shared mutable dependency similar to the existing `_taskStore` pattern. This works but creates implicit coupling — `startChat` silently uses whatever registry was set globally. If `setConnectionRegistry` is never called, `_connRegistry` is null and the connRegistry option is simply omitted from runQueryLoop, which is correct. But there's no guard or warning if it's called twice with different instances.

server/chat.ts
```diff
@@ -531,6 +537,10 @@ export async function startChat(
       abortController,
       eventStore,
       options.resume ? undefined : fullPrompt,
+      {
+        ...(_connRegistry ? { connRegistry: _connRegistry } : {}),
```
dimakis: 🔵 **style**: The options spread `{...(_connRegistry ? { connRegistry: _connRegistry } : {}), ...(options.onSessionResolved ? { onSessionResolved: options.onSessionResolved } : {})}` is unnecessarily verbose. Since `QueryLoopOptions` fields are optional, a simpler form like `{ connRegistry: _connRegistry ?? undefined, onSessionResolved: options.onSessionResolved }` would suffice — `undefined` values are equivalent to missing keys for optional fields. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Critical duplicate delivery bug: v2 clients receive global broadcasts twice (via both wss.clients and connRegistry.broadcastAll). The core v2 routing through sendOrBuffer and session lifecycle management is well-structured and thoroughly tested._

- 🔴 **bugs** (L83): Duplicate delivery to v2 clients: `wss.clients` contains ALL WebSocket connections (both v1 and v2). The global broadcast lambdas (setUpdateBroadcast, setInboxBroadcast, setTaskBroadcast, broadcastStatus, broadcastTasks) iterate `wss.clients` AND call `connRegistry.broadcastAll()`. Since v2 WebSocket connections are in both `wss.clients` and the ConnectionRegistry, v2 clients will receive every global event twice. The `wss.clients` loop should skip sockets that are registered in `connRegistry`, or `broadcastAll` should not be called in addition to the `wss.clients` loop. `[fixable]`


### `server/query-loop.ts`
_Critical duplicate delivery bug: v2 clients receive global broadcasts twice (via both wss.clients and connRegistry.broadcastAll). The core v2 routing through sendOrBuffer and session lifecycle management is well-structured and thoroughly tested._

- 🟡 **bugs** (L389): In the `result` event handler, `emit()` sends `session_end` which routes through `connRegistry.broadcast()` for v2 connections. Then `connRegistry.setActive(clientId, null)` is called. But `emit()` already delivered the event, so this ordering is fine. However, in the `finally` block (line ~650), the code checks `connRegistry?.get(clientId)` AFTER `registry.remove(clientId)`. Since `connRegistry` and `registry` (SessionRegistry) are independent, `connRegistry.get` still works. But `setActive(clientId, null)` is called twice on normal completion — once at line 390 in the `result` handler and again at line 659 in the finally block. This is harmless but unnecessary. `[fixable]`


### `server/chat.ts`
_Critical duplicate delivery bug: v2 clients receive global broadcasts twice (via both wss.clients and connRegistry.broadcastAll). The core v2 routing through sendOrBuffer and session lifecycle management is well-structured and thoroughly tested._

- 🟡 **unsafe_assumptions** (L33): Module-level mutable singleton `_connRegistry` set via `setConnectionRegistry()` is a shared mutable dependency similar to the existing `_taskStore` pattern. This works but creates implicit coupling — `startChat` silently uses whatever registry was set globally. If `setConnectionRegistry` is never called, `_connRegistry` is null and the connRegistry option is simply omitted from runQueryLoop, which is correct. But there's no guard or warning if it's called twice with different instances.
- 🔵 **style** (L541): The options spread `{...(_connRegistry ? { connRegistry: _connRegistry } : {}), ...(options.onSessionResolved ? { onSessionResolved: options.onSessionResolved } : {})}` is unnecessarily verbose. Since `QueryLoopOptions` fields are optional, a simpler form like `{ connRegistry: _connRegistry ?? undefined, onSessionResolved: options.onSessionResolved }` would suffice — `undefined` values are equivalent to missing keys for optional fields. `[fixable]`


_Critical duplicate delivery bug: v2 clients receive global broadcasts twice (via both wss.clients and connRegistry.broadcastAll). The core v2 routing through sendOrBuffer and session lifecycle management is well-structured and thoroughly tested._

- 🔵 **missing_tests** (L309): The `handleSendV2 auto-watch` test only covers the resume path. The new-session path (null sessionId with `onSessionResolved` callback) and the active-session path (where `sendToChat` is called) are not tested for auto-watch behavior. Consider adding tests that verify `onSessionResolved` is passed through to `startChat` for the create path. `[fixable]`

server/__tests__/ws-handler-v2.test.ts
```diff
@@ -305,6 +306,43 @@ describe('handleSetModeV2', () => {
   });
 });
 
+// ─── handleSendV2 auto-watch ─────────────────────────────────────────────────
```
dimakis, line 336: 🔵 **missing_tests**: The `handleSendV2 auto-watch` test only covers the resume path. The new-session path (null sessionId with `onSessionResolved` callback) and the active-session path (where `sendToChat` is called) are not tested for auto-watch behavior. Consider adding tests that verify `onSessionResolved` is passed through to `startChat` for the create path. `[fixable]`

server/query-loop.ts
```diff
@@ -78,6 +79,15 @@ function sendOrBuffer(
       enriched = { ...enriched, seq };
     }
   }
+
+  // v2 path: deliver via ConnectionRegistry when clientId is a registered v2
+  // connection AND sessionId is resolved (broadcast needs a session target).
+  if (connRegistry?.get(clientId) && sessionId) {
```
dimakis, line 85: 🟡 **regressions**: When a v2 client drives a session (`connRegistry.get(clientId)` is truthy and `sessionId` is resolved), `sendOrBuffer` returns early after `connRegistry.broadcast()`, completely bypassing `broadcastToObservers()`. If any v1 observer has been added to this session (via `registry.addObserver`), it will silently stop receiving stream events. The same pattern repeats in the finally-block `if/else` at line 650. If mixed v1/v2 observation is not expected during migration this is fine, but the silent skip could surprise future callers. `[fixable]`

server/index.ts
```diff
@@ -257,6 +275,9 @@ function handleChatWsV2(ws: WebSocket, connectionId: string) {
 /** Map raw WebSocket → WsTransport wrapper, so removeObserver can look up the transport. */
 const transportMap = new Map<WebSocket, WsTransport>();
 
+/** v2 WebSockets — tracked so wss.clients broadcasts skip them (v2 gets events via connRegistry). */
+const v2Sockets = new Set<WebSocket>();
```
dimakis, line 279: 🔵 **style**: `v2Sockets` is declared at line 279 but referenced in closures starting at line 81. While runtime-safe (closures capture the binding and the callbacks execute after initialization), the 250-line gap between first use and declaration hurts readability. Consider moving the declaration above `setUpdateBroadcast()` near line 77. `[fixable]`

server/index.ts
```diff
@@ -73,24 +75,32 @@ const server = USE_TLS
...
+  connRegistry.broadcastAll(data);
 });
 
 setTaskBroadcast((event) => {
   const msg = JSON.stringify(event);
   wss.clients.forEach((client) => {
+    if (v2Sockets.has(client)) return;
     if (client.readyState === client.OPEN) client.send(msg);
   });
+  connRegistry.broadcastAll(event as Record<string, unknown>);
```
dimakis, line 103: 🔵 **style**: `event as Record<string, unknown>` (line 103) and `data as Record<string, unknown>` (line 149) are type-widening casts. The broadcast callback parameters have more specific types but `broadcastAll` accepts the wide type. A generic or shared event type would avoid these casts, though this is a minor concern. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid phase-2 wiring with good test coverage of the v2 delivery path; main concern is that v1 observers are silently bypassed when a v2 client drives a session, which could be a regression if mixed v1/v2 observation is expected during migration._

- 🟡 **regressions** (L85): When a v2 client drives a session (`connRegistry.get(clientId)` is truthy and `sessionId` is resolved), `sendOrBuffer` returns early after `connRegistry.broadcast()`, completely bypassing `broadcastToObservers()`. If any v1 observer has been added to this session (via `registry.addObserver`), it will silently stop receiving stream events. The same pattern repeats in the finally-block `if/else` at line 650. If mixed v1/v2 observation is not expected during migration this is fine, but the silent skip could surprise future callers. `[fixable]`


_Solid phase-2 wiring with good test coverage of the v2 delivery path; main concern is that v1 observers are silently bypassed when a v2 client drives a session, which could be a regression if mixed v1/v2 observation is expected during migration._

- 🔵 **missing_tests**: The 'active session' path in `handleSendV2` (where `isActive(found.clientId)` is true and `sendToChat` is called) also performs `connRegistry.watch()` + `setActive()` but has no test verifying that behavior. The existing tests cover the 'resume' and 'create' paths only. `[fixable]`


### `server/index.ts`
_Solid phase-2 wiring with good test coverage of the v2 delivery path; main concern is that v1 observers are silently bypassed when a v2 client drives a session, which could be a regression if mixed v1/v2 observation is expected during migration._

- 🔵 **style** (L279): `v2Sockets` is declared at line 279 but referenced in closures starting at line 81. While runtime-safe (closures capture the binding and the callbacks execute after initialization), the 250-line gap between first use and declaration hurts readability. Consider moving the declaration above `setUpdateBroadcast()` near line 77. `[fixable]`
- 🔵 **style** (L103): `event as Record<string, unknown>` (line 103) and `data as Record<string, unknown>` (line 149) are type-widening casts. The broadcast callback parameters have more specific types but `broadcastAll` accepts the wide type. A generic or shared event type would avoid these casts, though this is a minor concern. `[fixable]`

PR #224, opened 2026-04-17
feat(client): replace WsPool with MitzoConnection for v2 single-WS protocol

dimakis:
## Summary

- Replace per-session `WsPool` (one WS per session key) with a single `MitzoConnection` that speaks v2 protocol: `hello`/`welcome` handshake, `reconnect` with per-session seq replay, pending send queue
- Rewrite store to use `MitzoConnection` — no more `activePoolKey`, `activeUnsub`, per-session subscriptions. All outbound messages include `sessionId` (null for new sessions). Session isolation via `sessionId` filtering in `wsListener` instead of WS unsubscribe
- Update protocol parser with `welcome`, `reconnected`, `session_switched` handlers; v1 handlers kept for backward compat
- `stopGeneration` sends v2 `stop` (not `interrupt`); all actions include `sessionId`

## Files changed

| File | Change |
|------|--------|
| `packages/client/src/connection.ts` | **New** — `MitzoConnection` class |
| `packages/client/__tests__/connection.test.ts` | **New** — 16 tests |
| `packages/client/src/store.ts` | Rewritten for `MitzoConnection` |
| `packages/client/__tests__/store.test.ts` | Rewritten for v2 (61 tests) |
| `packages/client/src/protocol-parser.ts` | v2 handlers added |
| `packages/client/src/index.ts` | Export `MitzoConnection` |
| `frontend/src/client-store.ts` | Remove `reconnectPollMs` |

## Test plan

- [x] MitzoConnection: hello handshake, send/queue, seq tracking, reconnect with replay, disconnect (16 tests)
- [x] Store v2: session switch, new session, send with sessionId, session isolation via filtering, queued sends, task/error dispatch, permissions, session expired recovery (61 tests)
- [x] Protocol parser: existing tests pass (backward compat), new v2 events handled
- [x] Full suite: 1733 tests pass, 0 lint errors, type checks pass (server + frontend)

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -162,7 +162,10 @@ export class MitzoConnection {
       this.reconnectTimer = setTimeout(() => this.doConnect(), this.config.reconnectDelayMs);
     };
 
-    ws.onerror = () => {};
+    ws.onerror = () => {
+      // Intentionally empty — error events always precede close, and
+      // reconnect is handled in onclose. Nothing actionable here.
+    };
   }
 
   private flushPendingSends(): void {
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -1,0 +1,181 @@
+/**
+ * MitzoConnection — single multiplexed WebSocket for the v2 protocol.
+ *
+ * Replaces the per-session WsPool. One WebSocket connection handles all
+ * sessions via sessionId-tagged messages. Manages hello/welcome handshake,
+ * reconnect with per-session seq replay, and pending send queue.
+ */
+
+import type { WebSocketLike } from './ws-connection.js';
+import { WS_READY_STATE } from './types.js';
+
+export interface MitzoConnectionConfig {
+  buildUrl(): string;
+  createWebSocket(url: string): WebSocketLike;
+  reconnectDelayMs?: number;
+}
+
+export type ConnectionListener = (msg: Record<string, unknown>) => void;
+
+const MAX_PENDING_SENDS = 100;
+
+export class MitzoConnection {
+  private ws: WebSocketLike | null = null;
+  private _connectionId: string | null = null;
+  private _connected = false;
+  private _isReconnect = false;
+  private listener: ConnectionListener | null = null;
+  private seqBySession = new Map<string, number>();
+  private pendingSends: string[] = [];
+  private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
+  private config: Required<MitzoConnectionConfig>;
+
+  constructor(config: MitzoConnectionConfig) {
+    this.config = {
+      reconnectDelayMs: 500,
+      ...config,
+    };
+  }
+
+  connect(): void {
+    this.doConnect();
+  }
+
+  disconnect(): void {
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer);
+      this.reconnectTimer = null;
+    }
+    if (this.ws) {
+      this.ws.onclose = null;
+      this.ws.close();
+      this.ws = null;
+    }
+    this._connected = false;
+  }
+
+  send(msg: Record<string, unknown>): boolean {
+    const payload = JSON.stringify(msg);
+    if (this._connected && this.ws?.readyState === WS_READY_STATE.OPEN) {
+      this.ws.send(payload);
+      return true;
+    }
+    if (
+      this.ws?.readyState === WS_READY_STATE.CONNECTING ||
+      this.reconnectTimer ||
+      (this.ws?.readyState === WS_READY_STATE.OPEN && !this._connected)
+    ) {
+      if (this.pendingSends.length >= MAX_PENDING_SENDS) {
+        this.pendingSends.shift();
+      }
+      this.pendingSends.push(payload);
+      return true;
+    }
+    return false;
+  }
+
+  onMessage(listener: ConnectionListener): void {
+    this.listener = listener;
+  }
+
+  isConnected(): boolean {
+    return this._connected;
+  }
+
+  getConnectionId(): string | null {
+    return this._connectionId;
+  }
+
+  trackSeq(sessionId: string, seq: number): void {
+    this.seqBySession.set(sessionId, seq);
+  }
+
+  getLastSeq(sessionId: string): number {
+    return this.seqBySession.get(sessionId) ?? 0;
+  }
+
+  clearSession(sessionId: string): void {
+    this.seqBySession.delete(sessionId);
+  }
+
+  // ─── Internal ──────────────────────────────────────────────────────────────
+
+  private doConnect(): void {
+    if (
+      this.ws?.readyState === WS_READY_STATE.OPEN ||
+      this.ws?.readyState === WS_READY_STATE.CONNECTING
+    ) {
+      return;
+    }
+
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer);
+      this.reconnectTimer = null;
+    }
+
+    const url = this.config.buildUrl();
+    const ws = this.config.createWebSocket(url);
+    this.ws = ws;
+
+    ws.onopen = () => {
+      ws.send(JSON.stringify({ type: 'hello', protocolVersion: 2 }));
+    };
+
+    ws.onmessage = (e: { data: string }) => {
+      let msg: Record<string, unknown>;
+      try {
+        msg = JSON.parse(e.data);
+      } catch {
+        return;
+      }
+
+      if (msg.type === 'welcome') {
+        this._connectionId = msg.connectionId as string;
+        this._connected = true;
+
+        if (this._isReconnect && this.seqBySession.size > 0) {
+          const sessions = Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          }));
+          ws.send(JSON.stringify({ type: 'reconnect', sessions }));
+        }
+        this._isReconnect = true;
+
+        this.flushPendingSends();
+        this.listener?.({ type: '_open' });
+        return;
+      }
+
+      // Track seq for reconnect replay
+      if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
+        this.seqBySession.set(msg.sessionId as string, msg.seq as number);
+      }
+
+      this.listener?.(msg);
+    };
+
+    ws.onclose = () => {
+      this.ws = null;
+      this._connected = false;
+      this.listener?.({ type: '_close' });
+      this.reconnectTimer = setTimeout(() => this.doConnect(), this.config.reconnectDelayMs);
+    };
+
+    ws.onerror = () => {};
+  }
+
+  private flushPendingSends(): void {
+    if (this.pendingSends.length === 0 || !this.ws) return;
+    const toFlush = this.pendingSends;
+    this.pendingSends = [];
+    for (let i = 0; i < toFlush.length; i++) {
+      try {
+        this.ws.send(toFlush[i]);
+      } catch {
+        this.pendingSends.unshift(...toFlush.slice(i));
+        break;
+      }
+    }
+  }
+}
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -39,11 +39,17 @@ export interface ProtocolCallbacks {
   /** Called to fetch messages from REST API for reattach recovery. */
   fetchMessages?(sessionId: string): Promise<FinishedMessage[]>;
 
-  /** Called to mark the WS pool entry as running/not-running. */
+  /** @deprecated v1 only — called to mark the WS pool entry as running/not-running. */
   setWsRunning?(poolKey: string, running: boolean): void;
 
-  /** Called to send a queued message after session_end. */
+  /** @deprecated v1 only — called to send a queued message after session_end. */
   sendQueued?(poolKey: string, msg: unknown): void;
+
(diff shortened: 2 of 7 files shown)


dimakis (commented):
## Centaur Review

_Session filtering in the shared-WS listener has a critical flaw: `session_end` and `session_id` events from other sessions bypass the filter, which can corrupt active session state — this is the main blocker for the multiplexed v2 architecture._

- 🔴 **bugs** (L418): `session_end` from a different session bypasses the session filter because it's in `SESSION_ASSIGNING_TYPES`. In a multiplexed v2 world, a background session's `session_end` would clear `running`, flush `pendingSend`, and dispatch `SESSION_END` into the active session's state. The filter should only exempt `session_id`/`session_end` when they match the current session OR when no session is active (new session assignment). Consider checking `eventSessionId === parserState.currentSessionId` for session_end even when it's in the allow-list, or move the filtering logic into the parser where it can distinguish assignment vs. foreign events. `[fixable]`
- 🟡 **bugs** (L418): Same issue for `session_id`: a `session_id` event tagged with a different session (e.g., from a task orchestrator running in another session) would pass the filter and call `onSessionAssigned()`, hijacking `parserState.currentSessionId` and the active session in the store. In v1 this was safe because each WS was session-scoped; in v2 with a shared WS, all sessions' events arrive on the same connection. `[fixable]`
- 🟡 **bugs** (L264): `stopGeneration()` silently does nothing when `parserState.currentSessionId` is null (e.g., during the first turn before `session_id` arrives). If the user hits stop during the first message of a new session, the stop is swallowed. Consider queuing or sending with `sessionId: null` so the server can resolve it. `[fixable]`
- 🟡 **regressions** (L270): `respondToPermission()` silently drops the WS message when `currentSessionId` is null but still clears `permissions.pending`. If a permission request arrives during the first turn (before `session_id`), the user's response is lost. The v1 code had the same gap but it was less likely because the pool key existed. `[fixable]`


_Session filtering in the shared-WS listener has a critical flaw: `session_end` and `session_id` events from other sessions bypass the filter, which can corrupt active session state — this is the main blocker for the multiplexed v2 architecture._

- 🟡 **missing_tests**: No test for `session_end` arriving from a different session on the shared WS. This is the key scenario the multiplexed connection must handle correctly — a `session_end` for session-a should not clear `running` or flush `pendingSend` when the user is active in session-b. `[fixable]`
- 🔵 **missing_tests**: The `reattach_failed → onMessagesRestored` test block was removed but no v2 equivalent was added. If reconnect replay fails server-side, there should be a recovery path and test for it. `[fixable]`


_Session filtering in the shared-WS listener has a critical flaw: `session_end` and `session_id` events from other sessions bypass the filter, which can corrupt active session state — this is the main blocker for the multiplexed v2 architecture._

- 🔵 **missing_tests**: No test for the MAX_PENDING_SENDS=100 eviction behavior (oldest message dropped when queue is full). Also no test for the `send()` catch path in `flushPendingSends` where a partial flush re-queues remaining messages. `[fixable]`


_Session filtering in the shared-WS listener has a critical flaw: `session_end` and `session_id` events from other sessions bypass the filter, which can corrupt active session state — this is the main blocker for the multiplexed v2 architecture._

- 🔵 **style** (L165): Empty `onerror` handler is fine functionally (close follows error), but a debug log or comment explaining why it's intentionally empty would prevent future contributors from thinking it's a mistake. `[fixable]`

packages/client/src/store.ts
```diff
@@ -459,28 +396,48 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+            ((tokens.cacheCreation as number) ?? 0),
+        },
+      }));
     },
   };
 
-  function wsListener(wsMsg: WsMsg) {
-    const poolKey = activePoolKey ?? 'default';
+  // Session-scoped events that assign/change the active session — always accepted
+  const SESSION_ASSIGNING_TYPES = new Set(['session_id', 'session_end']);
```
dimakis: 🔴 **bugs**: `session_end` from a different session bypasses the session filter because it's in `SESSION_ASSIGNING_TYPES`. In a multiplexed v2 world, a background session's `session_end` would clear `running`, flush `pendingSend`, and dispatch `SESSION_END` into the active session's state. The filter should only exempt `session_id`/`session_end` when they match the current session OR when no session is active (new session assignment). Consider checking `eventSessionId === parserState.currentSessionId` for session_end even when it's in the allow-list, or move the filtering logic into the parser where it can distinguish assignment vs. foreign events. `[fixable]`

dimakis: 🟡 **bugs**: Same issue for `session_id`: a `session_id` event tagged with a different session (e.g., from a task orchestrator running in another session) would pass the filter and call `onSessionAssigned()`, hijacking `parserState.currentSessionId` and the active session in the store. In v1 this was safe because each WS was session-scoped; in v2 with a shared WS, all sessions' events arrive on the same connection. `[fixable]`

packages/client/src/store.ts
```diff
@@ -319,21 +262,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     stopGeneration() {
```
dimakis, line 264: 🟡 **bugs**: `stopGeneration()` silently does nothing when `parserState.currentSessionId` is null (e.g., during the first turn before `session_id` arrives). If the user hits stop during the first message of a new session, the stop is swallowed. Consider queuing or sending with `sessionId: null` so the server can resolve it. `[fixable]`

packages/client/src/store.ts
```diff
@@ -319,21 +262,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
-        ? `session:${parserState.currentSessionId}`
-        : null;
-      if (poolKey) {
-        wsPool.send(poolKey, { type: 'interrupt' });
+      if (parserState.currentSessionId) {
+        connection.send({ type: 'stop', sessionId: parserState.currentSessionId });
       }
     },
 
     respondToPermission(permId: string, decision: 'once' | 'always' | 'deny') {
```
dimakis, line 271: 🟡 **regressions**: `respondToPermission()` silently drops the WS message when `currentSessionId` is null but still clears `permissions.pending`. If a permission request arrives during the first turn (before `session_id`), the user's response is lost. The v1 code had the same gap but it was less likely because the pool key existed. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -0,0 +1,181 @@
...
+    };
+
+    ws.onclose = () => {
+      this.ws = null;
+      this._connected = false;
+      this.listener?.({ type: '_close' });
+      this.reconnectTimer = setTimeout(() => this.doConnect(), this.config.reconnectDelayMs);
+    };
+
+    ws.onerror = () => {};
```
dimakis: 🔵 **style**: Empty `onerror` handler is fine functionally (close follows error), but a debug log or comment explaining why it's intentionally empty would prevent future contributors from thinking it's a mistake. `[fixable]`

packages/client/src/store.ts
```diff
@@ -155,31 +146,18 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     async switchSession(id: string) {
```
dimakis, line 148: 🟡 **bugs**: `switchSession()` does not clear `parserState.pendingSend`. If a user queues a follow-up message in session A then switches to session B, the stale message will fire on session B's first `session_end`. `newSession()` correctly clears it (line 175) — `switchSession` should do the same. `[fixable]`

packages/client/src/store.ts
```diff
@@ -459,28 +395,49 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+            ((tokens.cacheRead as number) ?? 0) +
+            ((tokens.cacheCreation as number) ?? 0),
+        },
+      }));
     },
   };
 
-  function wsListener(wsMsg: WsMsg) {
-    const poolKey = activePoolKey ?? 'default';
+  function wsListener(msg: Record<string, unknown>) {
```
dimakis, line 416: 🔵 **unsafe_assumptions**: The session filter accepts events without `sessionId` as global events. This is correct for `task_state`, `inbox_updated`, etc. However, if the server ever sends a session-scoped event without tagging `sessionId` (e.g., due to a race before `resolvedSessionId` is set), it will bypass the filter and leak into the wrong session. This is a server-side responsibility but worth noting — consider adding a comment documenting which event types are expected to be global.

packages/client/src/protocol-parser.ts
```diff
@@ -97,6 +103,25 @@ export function parseServerMessage(
       result.connectionUpdate = { status: 'disconnected' };
       break;
 
+    // ── v2 handshake events ────────────────────────────────────────────────
+
+    case 'reconnected':
```
dimakis, line 108: 🟡 **missing_tests**: New protocol-parser cases (`reconnected`, `session_switched`, `session_cleared`) and new callbacks (`onSendQueued`, `onTokensHydrated`) have no unit tests in `protocol-parser.test.ts`. The store-level integration tests cover some paths, but the parser's own test file should cover these branches directly. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -0,0 +1,184 @@
...
+
+      // Track seq for reconnect replay
+      if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
+        this.seqBySession.set(msg.sessionId as string, msg.seq as number);
+      }
+
+      this.listener?.(msg);
+    };
+
+    ws.onclose = () => {
```
dimakis, line 158: 🔵 **bugs**: `onclose` sets `this.ws = null` then schedules reconnect. If `disconnect()` is called between `onclose` firing and the timer callback, `disconnect()` sees `this.ws` as null and skips `ws.close()`, but the timer is cleared correctly. This is fine. However, `onerror` is empty and the WS reference in `onclose` still triggers reconnect after an explicit `disconnect()` followed by a delayed close event. Consider guarding reconnect with a `_disposed` flag if `disconnect()` is ever called while a close is in-flight. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.doConnect();
+  }
+
+  disconnect(): void {
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer);
+      this.reconnectTimer = null;
+    }
+    if (this.ws) {
+      this.ws.onclose = null;
```
dimakis, line 50: 🔵 **style**: `disconnect()` nullifies `ws.onclose` to prevent reconnect but leaves `ws.onmessage`, `ws.onopen`, and `ws.onerror` intact. While functionally harmless (the socket is closing), nullifying all handlers would be more defensive and consistent. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid migration from per-session WsPool to multiplexed MitzoConnection. One real bug: `switchSession` doesn't clear `pendingSend`, risking stale message delivery to the wrong session. Protocol parser additions lack direct unit tests._

- 🟡 **bugs** (L148): `switchSession()` does not clear `parserState.pendingSend`. If a user queues a follow-up message in session A then switches to session B, the stale message will fire on session B's first `session_end`. `newSession()` correctly clears it (line 175) — `switchSession` should do the same. `[fixable]`
- 🔵 **unsafe_assumptions** (L416): The session filter accepts events without `sessionId` as global events. This is correct for `task_state`, `inbox_updated`, etc. However, if the server ever sends a session-scoped event without tagging `sessionId` (e.g., due to a race before `resolvedSessionId` is set), it will bypass the filter and leak into the wrong session. This is a server-side responsibility but worth noting — consider adding a comment documenting which event types are expected to be global.


_Solid migration from per-session WsPool to multiplexed MitzoConnection. One real bug: `switchSession` doesn't clear `pendingSend`, risking stale message delivery to the wrong session. Protocol parser additions lack direct unit tests._

- 🟡 **missing_tests** (L108): New protocol-parser cases (`reconnected`, `session_switched`, `session_cleared`) and new callbacks (`onSendQueued`, `onTokensHydrated`) have no unit tests in `protocol-parser.test.ts`. The store-level integration tests cover some paths, but the parser's own test file should cover these branches directly. `[fixable]`


_Solid migration from per-session WsPool to multiplexed MitzoConnection. One real bug: `switchSession` doesn't clear `pendingSend`, risking stale message delivery to the wrong session. Protocol parser additions lack direct unit tests._

- 🔵 **bugs** (L158): `onclose` sets `this.ws = null` then schedules reconnect. If `disconnect()` is called between `onclose` firing and the timer callback, `disconnect()` sees `this.ws` as null and skips `ws.close()`, but the timer is cleared correctly. This is fine. However, `onerror` is empty and the WS reference in `onclose` still triggers reconnect after an explicit `disconnect()` followed by a delayed close event. Consider guarding reconnect with a `_disposed` flag if `disconnect()` is ever called while a close is in-flight. `[fixable]`
- 🔵 **style** (L50): `disconnect()` nullifies `ws.onclose` to prevent reconnect but leaves `ws.onmessage`, `ws.onopen`, and `ws.onerror` intact. While functionally harmless (the socket is closing), nullifying all handlers would be more defensive and consistent. `[fixable]`


_Solid migration from per-session WsPool to multiplexed MitzoConnection. One real bug: `switchSession` doesn't clear `pendingSend`, risking stale message delivery to the wrong session. Protocol parser additions lack direct unit tests._

- 🔵 **missing_tests**: No test covers `interruptMessage` with the new v2 protocol (sessionId tagging, WS delivery). The old test suite also didn't cover it, but this is a good time to add coverage since the sending mechanism changed. `[fixable]`

PR #226, opened 2026-04-18
feat: end-to-end OpenTelemetry tracing with Jaeger

dimakis:
## Summary
- Adds full-stack distributed tracing: browser spans link to server spans via W3C `traceparent` propagated in WS JSON payloads
- Jaeger all-in-one via docker-compose (`npm run tracing:up/down`, UI at `:16686`)
- 9 backend spans covering WS lifecycle (connect, subscribe, reattach, send, interrupt, stop, disconnect)
- Frontend browser tracer + spans on wsSend, reconnect, subscribe, disconnect
- Both sides opt-in via env var — zero overhead when unconfigured

## Test plan
- [x] `trace-context.test.ts` — W3C traceparent parsing (valid, malformed, missing)
- [x] `tracing.test.ts` (frontend) — noop tracer, injectTraceparent
- [x] Existing `ws-pool.test.ts` still passes (noop tracer doesn't alter message format)
- [x] Full test suite passes (18381 tests)
- [ ] Manual: start Jaeger, use Mitzo, verify linked traces in Jaeger UI


dimakis:
## Centaur Review

### `server/tracing.ts`
_Well-structured opt-in tracing with good noop fallback. Main concerns: SimpleSpanProcessor on the server will add synchronous latency under load (use BatchSpanProcessor), and the fire-and-forget browser init silently swallows errors._

- 🟡 **unsafe_assumptions** (L29): SimpleSpanProcessor exports each span synchronously, blocking the event loop on every span.end(). On a server handling many concurrent WS connections (each creating multiple spans per message), this adds per-request latency. Use BatchSpanProcessor for the server — it buffers and sends in the background. SimpleSpanProcessor is fine for the browser side where span volume is low. `[fixable]`


### `server/index.ts`
_Well-structured opt-in tracing with good noop fallback. Main concerns: SimpleSpanProcessor on the server will add synchronous latency under load (use BatchSpanProcessor), and the fire-and-forget browser init silently swallows errors._

- 🟡 **regressions** (L299): The subscribe handler previously had a multi-line comment explaining *why* a subscriber is promoted to driver (iOS Safari WS drops every 30-90s, pool reconnects with subscribe). That context was deleted and replaced with a span attribute (`ws.subscribe.outcome: promoted_to_reattach`). Span attributes are runtime metadata — they don't explain intent. The rationale comment should be preserved alongside the tracing code. `[fixable]`
- 🔵 **style** (L270): The connSpan is created at the top of handleChatWs and ended inside the ws.on('close') handler ~300 lines later. This long-lived span (potentially hours for a persistent WS) will accumulate in memory until close. Consider whether a connection-scoped span is the right abstraction here — the per-message spans already capture the interesting work. If kept, add the session_id as an attribute once known (it's only available after a send/reattach). `[fixable]`


### `frontend/src/lib/tracing.ts`
_Well-structured opt-in tracing with good noop fallback. Main concerns: SimpleSpanProcessor on the server will add synchronous latency under load (use BatchSpanProcessor), and the fire-and-forget browser init silently swallows errors._

- 🟡 **bugs** (L22): The async provider registration is fire-and-forget (`void (async () => { ... })()`). If it throws (e.g. invalid endpoint URL, network error during OTLPTraceExporter construction), the error is silently swallowed. Add a `.catch()` to log the failure — otherwise tracing appears configured but silently drops all spans with no diagnostic output. `[fixable]`


### `frontend/src/lib/__tests__/ws-pool.test.ts`
_Well-structured opt-in tracing with good noop fallback. Main concerns: SimpleSpanProcessor on the server will add synchronous latency under load (use BatchSpanProcessor), and the fire-and-forget browser init silently swallows errors._

- 🔵 **missing_tests**: ws-pool.ts was modified to inject traceparent into every outbound message via wsSend and to create spans on connect/disconnect/reattach. The existing ws-pool tests aren't updated to verify that (1) outbound messages include traceparent when tracing is active, or (2) the span creation doesn't break existing behavior when the tracer is noop. The noop path is implicitly covered since tests run without VITE_OTEL_ENDPOINT, but an explicit assertion that wsSend still works correctly would be more robust. `[fixable]`


### `server/trace-context.ts`
_Well-structured opt-in tracing with good noop fallback. Main concerns: SimpleSpanProcessor on the server will add synchronous latency under load (use BatchSpanProcessor), and the fire-and-forget browser init silently swallows errors._

- 🔵 **unsafe_assumptions** (L11): The regex only accepts version `00`. The W3C Trace Context spec says receivers SHOULD accept unknown versions and parse the first 55 characters the same way. A traceparent starting with e.g. `01-...` (future version) would be rejected. Low risk today but worth a TODO comment. `[fixable]`

frontend/src/lib/tracing.ts
```diff
@@ -0,0 +1,64 @@
...
+    provider.register();
+  })().catch((err) => {
+    console.warn('[mitzo] browser tracing init failed', err);
+  });
+}
+
+/**
+ * Browser tracer instance. Returns noop spans when OTel is not configured.
+ */
+export const tracer = trace.getTracer('mitzo-frontend', '1.0.0');
```
dimakis, line 45: 🔵 **unsafe_assumptions**: The `tracer` is exported synchronously at module load, but the WebTracerProvider is registered asynchronously (line 22). Any span created before the async import completes will be a noop span. This is by design (documented in the comment), but means early WS operations (initial connect, first subscribe) will always produce noop spans even when tracing is configured. Consider awaiting provider registration before the app starts, or document this known gap. `[fixable]`

server/trace-context.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * WS doesn't have per-message HTTP headers, so we embed a `traceparent`
+ * field in the JSON payload. These helpers parse it into an OTel Context
+ * so backend spans can join the same trace as the frontend.
+ */
+
+import { trace, context, SpanContext, TraceFlags, ROOT_CONTEXT } from '@opentelemetry/api';
+
+// TODO: W3C spec says receivers SHOULD accept unknown versions (not just 00)
+const TRACEPARENT_RE = /^00-([0-9a-f]{32})-([0-9a-f]{16})-([0-9a-f]{2})$/;
```
dimakis, line 12: 🔵 **style**: The TODO comment about accepting unknown versions (not just 00) per W3C spec is reasonable but could lead to silent incompatibility with future traceparent versions. If this is intentional scope limitation, remove the TODO; if it should be fixed, fix it now — the regex change is trivial (replace `00` with `[0-9a-f]{2}`). `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid end-to-end tracing implementation with clean opt-in design; main concern is spans not protected by try/finally in the server WS handler, which will leak on exceptions when tracing is active._

- 🟡 **bugs**: Spans created in each message handler branch (subscribe, reattach, send, permission_response, set_mode, interrupt, stop) are not wrapped in try/finally. If any call between startSpan and span.end() throws (e.g. transport.send, reattachChat, registry.addObserver), the span will never end. The outer try/catch at line 565 does not end these per-message spans. Use try/finally around each span's lifetime to guarantee span.end() is called. `[fixable]`
- 🟡 **regressions**: The comment `// Do NOT reassign clientId — see replayMissedEvents doc.` was removed from the subscribe-promoted-to-reattach path and replaced with a span attribute. This comment documented a subtle behavioral constraint (why clientId is intentionally NOT reassigned in the subscribe path, unlike the reattach path where it IS reassigned). The span attribute doesn't convey this to future developers. The comment should be preserved. `[fixable]`


### `frontend/src/lib/ws-pool.ts`
_Solid end-to-end tracing implementation with clean opt-in design; main concern is spans not protected by try/finally in the server WS handler, which will leak on exceptions when tracing is active._

- 🟡 **bugs**: In wsSend(), `msg` parameter is typed as `unknown` but cast to `Record<string, unknown>` via `as`. If a caller passes a non-object value (primitive, array, null), the spread `{ ...msgObj, ...injectTraceparent(span) }` will silently produce unexpected results — primitives spread to empty objects, losing the original message. The previous code simply called JSON.stringify(msg) which handled any type. Add a typeof guard or narrow the type before spreading. `[fixable]`


### `frontend/src/lib/tracing.ts`
_Solid end-to-end tracing implementation with clean opt-in design; main concern is spans not protected by try/finally in the server WS handler, which will leak on exceptions when tracing is active._

- 🔵 **unsafe_assumptions** (L45): The `tracer` is exported synchronously at module load, but the WebTracerProvider is registered asynchronously (line 22). Any span created before the async import completes will be a noop span. This is by design (documented in the comment), but means early WS operations (initial connect, first subscribe) will always produce noop spans even when tracing is configured. Consider awaiting provider registration before the app starts, or document this known gap. `[fixable]`


### `server/trace-context.ts`
_Solid end-to-end tracing implementation with clean opt-in design; main concern is spans not protected by try/finally in the server WS handler, which will leak on exceptions when tracing is active._

- 🔵 **style** (L12): The TODO comment about accepting unknown versions (not just 00) per W3C spec is reasonable but could lead to silent incompatibility with future traceparent versions. If this is intentional scope limitation, remove the TODO; if it should be fixed, fix it now — the regex change is trivial (replace `00` with `[0-9a-f]{2}`). `[fixable]`


### `frontend/src/lib/__tests__/tracing.test.ts`
_Solid end-to-end tracing implementation with clean opt-in design; main concern is spans not protected by try/finally in the server WS handler, which will leak on exceptions when tracing is active._

- 🔵 **missing_tests**: The frontend tracing tests cover the noop path well but don't test the ws-pool.ts integration — specifically that wsSend injects traceparent into outgoing messages or that connectEntry creates spans for reattach/subscribe. These are the primary user-facing behaviors of the frontend tracing addition. `[fixable]`

PR #227, opened 2026-04-18
fix: reconcile orphaned sessions between SDK and EventStore

dimakis:
## Summary

- **Self-healing EventStore**: `getSessions()` now reconciles SDK-discovered sessions that aren't tracked in the EventStore (orphaned by server restarts mid-query, or created by external agents in worktrees).
- **Resilient `switch_session`**: Falls back to `getSessionInfo()` (targeted SDK lookup) when the EventStore misses, backfills the row, then proceeds. Fixes the missing OTel span status on the not-found path that made failures invisible in Jaeger.
- **Graceful deploy**: `deploy.sh` sends SIGTERM and waits 3s before `kickstart -k`, giving the query loop's `finally` block a chance to persist session metadata. `start.sh` rotates logs before starting so launchd's truncate-on-start doesn't destroy the previous instance's history.
- **Early registration on resume**: `startChat()` pre-registers resumed sessions in the EventStore before the query loop starts, closing the window where a restart mid-resume loses the session.

## Root cause

The EventStore row for a session was only created when `runQueryLoop` received its first `assistant` event from the Claude SDK. A server restart (deploy via `launchctl kickstart -k` which sends SIGKILL) before that point meant the session was lost from the EventStore — but the SDK's `listSessions({ includeWorktrees: true })` still discovered the conversation JSONL file. The mobile UI showed the session in the sidebar, but tapping it triggered "Session not found" because `switch_session` checked the EventStore exclusively.

## Test plan

- [x] All 1727 existing server tests pass
- [x] New test: `handleSwitchSession` falls back to SDK discovery when EventStore misses
- [x] New test: error still sent when both EventStore and SDK discovery fail
- [x] Existing `switch_session` tests updated for async handler
- [x] TypeScript type-check clean (`tsc --noEmit`)
- [x] Full server build clean (`npm run build:server`)

### DIFF
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -261,13 +261,14 @@ export class EventStore {
       );
     } else {
       this.db!.prepare(
-        'INSERT INTO sessions (session_id, summary, branch, cwd, mode, initial_prompt, wt_id, goal_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?)',
+        'INSERT INTO sessions (session_id, summary, branch, cwd, mode, is_active, initial_prompt, wt_id, goal_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)',
       ).run(
         meta.sessionId,
         meta.summary ?? null,
         meta.branch ?? null,
         meta.cwd ?? null,
         meta.mode ?? 'agent',
+        meta.isActive === false ? 0 : 1,
         meta.initialPrompt ?? null,
         meta.wtId ?? null,
         meta.goalId ?? null,
@@ -282,15 +283,21 @@ export class EventStore {
 
   /**
    * Return the subset of `sessionIds` that already exist in the sessions table.
-   * Single query regardless of input size — used for batch reconciliation.
+   * Batches into chunks of 500 to stay within SQLite's SQLITE_MAX_VARIABLE_NUMBER.
    */
   getKnownSessionIds(sessionIds: string[]): Set<string> {
     if (sessionIds.length === 0) return new Set();
-    const placeholders = sessionIds.map(() => '?').join(',');
-    const rows = this.db!.prepare(
-      `SELECT session_id FROM sessions WHERE session_id IN (${placeholders})`,
-    ).all(...sessionIds) as Array<{ session_id: string }>;
-    return new Set(rows.map((r) => r.session_id));
+    const CHUNK = 500;
+    const result = new Set<string>();
+    for (let i = 0; i < sessionIds.length; i += CHUNK) {
+      const chunk = sessionIds.slice(i, i + CHUNK);
+      const placeholders = chunk.map(() => '?').join(',');
+      const rows = this.db!.prepare(
+        `SELECT session_id FROM sessions WHERE session_id IN (${placeholders})`,
+      ).all(...chunk) as Array<{ session_id: string }>;
+      for (const r of rows) result.add(r.session_id);
+    }
+    return result;
   }
 
   listSessions(limit?: number): SessionMeta[] {
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -280,6 +280,19 @@ export class EventStore {
     return row ? rowToSession(row) : null;
   }
 
+  /**
+   * Return the subset of `sessionIds` that already exist in the sessions table.
+   * Single query regardless of input size — used for batch reconciliation.
+   */
+  getKnownSessionIds(sessionIds: string[]): Set<string> {
+    if (sessionIds.length === 0) return new Set();
+    const placeholders = sessionIds.map(() => '?').join(',');
+    const rows = this.db!.prepare(
+      `SELECT session_id FROM sessions WHERE session_id IN (${placeholders})`,
+    ).all(...sessionIds) as Array<{ session_id: string }>;
+    return new Set(rows.map((r) => r.session_id));
+  }
+
   listSessions(limit?: number): SessionMeta[] {
     const rows =
       limit != null ? this.stmts.listSessionsLimited.all(limit) : this.stmts.listSessions.all();
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -8,12 +8,8 @@ npm run build:server
 echo "Building frontend..."
 npm run build
 
-echo "Gracefully stopping service..."
-SERVICE="gui/$(id -u)/com.mitzo.server"
-launchctl kill SIGTERM "$SERVICE" 2>/dev/null || true
-sleep 3
-
 echo "Restarting service..."
-launchctl kickstart -k "$SERVICE"
+# kickstart -k sends SIGTERM and waits for termination before restarting.
+launchctl kickstart -k "gui/$(id -u)/com.mitzo.server"
 
 echo "Deployed and restarted."
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -8,7 +8,12 @@ npm run build:server
 echo "Building frontend..."
 npm run build
 
+echo "Gracefully stopping service..."
+SERVICE="gui/$(id -u)/com.mitzo.server"
+launchctl kill SIGTERM "$SERVICE" 2>/dev/null || true
+sleep 3
+
 echo "Restarting service..."
-launchctl kickstart -k "gui/$(id -u)/com.mitzo.server"
+launchctl kickstart -k "$SERVICE"
 
 echo "Deployed and restarted."
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -13,6 +13,6 @@ for f in "$LOGDIR/server-stdout.log" "$LOGDIR/server-stderr.log"; do
 done
 
 # Prune rotated logs older than 7 days
-find "$LOGDIR" -name 'server-*.2*.log' -mtime +7 -delete 2>/dev/null || true
+find "$LOGDIR" -name 'server-std*.????????-??????.log' -mtime +7 -delete 2>/dev/null || true
 
 exec node dist/index.js
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -1,3 +1,18 @@
 #!/bin/bash
 cd "$(dirname "$0")/.."
+
+LOGDIR="logs"
+mkdir -p "$LOGDIR"
+
+# Rotate previous logs before starting (launchd truncates StandardOutPath
+# on each process start, so we preserve the old log here instead).
+for f in "$LOGDIR/server-stdout.log" "$LOGDIR/server-stderr.log"; do
+  if [ -s "$f" ]; then
+    mv "$f" "${f%.log}.$(date +%Y%m%d-%H%M%S).log"
+  fi
+done
+
+# Prune rotated logs older than 7 days
+find "$LOGDIR" -name 'server-*.2*.log' -mtime +7 -delete 2>/dev/null || true
+
 exec node dist/index.js
--- a/server/__tests__/discover-session.test.ts
+++ b/server/__tests__/discover-session.test.ts
@@ -132,7 +132,7 @@ describe('getSessions reconciliation', () => {
     mockGetKnownSessionIds.mockReturnValue(new Set(['sess-known']));
 
     const { getSessions } = await import('../chat.js');
-    await getSessions();
+    const result = await getSessions();
 
     expect(mockGetKnownSessionIds).toHaveBeenCalled();
     expect(mockUpsertSession).toHaveBeenCalledTimes(1);
@@ -142,7 +142,13 @@ describe('getSessions reconciliation', () => {
         summary: 'Orphan',
         cwd: '/projects/bar',
         branch: 'feat',
+        isActive: false,
       }),
     );
+
+    // Backfilled sessions should appear in the returned list
+    const ids = result.sessions.map((s: { id: string }) => s.id);
+    expect(ids).toContain('sess-orphan');
+    expect(ids).toContain('sess-known');
   });
 });
--- a/server/__tests__/discover-session.test.ts
+++ b/server/__tests__/discover-session.test.ts
@@ -1,0 +1,148 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const mockGetSessionInfo = vi.fn();
+const mockUpsertSession = vi.fn();
+const mockGetSession = vi.fn();
+const mockGetKnownSessionIds = vi.fn();
+
+vi.mock('@anthropic-ai/claude-agent-sdk', () => ({
+  query: vi.fn(),
+  listSessions: vi.fn().mockResolvedValue([]),
+  getSessionInfo: (...args: unknown[]) => mockGetSessionInfo(...args),
+  getSessionMessages: vi.fn().mockResolvedValue([]),
+  renameSession: vi.fn(),
+}));
+
+const mockEventStore = {
+  upsertSession: mockUpsertSession,
+  getSession: mockGetSession,
+  getKnownSessionIds: mockGetKnownSessionIds,
+  listSessions: vi.fn().mockReturnValue([]),
+  getEventsAfter: vi.fn().mockReturnValue([]),
+  getSessionEvents: vi.fn().mockReturnValue([]),
+  markSessionInactive: vi.fn(),
+  hideSession: vi.fn(),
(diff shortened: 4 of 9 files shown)


server/index.ts
```diff
@@ -245,13 +245,11 @@ function handleChatWsV2(ws: WebSocket, connectionId: string) {
   }, HEARTBEAT_INTERVAL_MS);
 
   ws.on('message', (raw) => {
-    try {
-      dispatchV2Message(connectionId, transport, raw.toString(), v2Ctx);
-    } catch (err: unknown) {
+    dispatchV2Message(connectionId, transport, raw.toString(), v2Ctx).catch((err: unknown) => {
```
dimakis: 🔴 **regressions**: Making `dispatchV2Message` async and calling it without `await` in the WS `message` handler introduces message reordering. If a `switch_session` (which now awaits `discoverSession`) is in flight when a subsequent `send` message arrives, the `send` may be dispatched before `switch_session` completes, executing against the wrong (or no) active session. The previous synchronous dispatch guaranteed FIFO ordering. Consider serializing dispatch per connection (e.g., a per-connection promise chain). `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,66 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
       }
     } catch {
       // Expected when session dir doesn't exist
     }
   }
+
+  // Reconcile: backfill EventStore for SDK-discovered sessions that Mitzo
+  // doesn't track (e.g. sessions orphaned by a server restart mid-query,
+  // or created by external agents in a worktree).
+  for (const [sessionId, entry] of seen) {
```
dimakis, line 873: 🟡 **unsafe_assumptions**: The reconciliation loop in `getSessions` calls `eventStore.getSession()` for every SDK-discovered session on every page fetch. With hundreds of sessions across multiple dirs, this is O(n) synchronous SQLite reads on a hot path (session list is polled by the frontend). Consider batching: collect all session IDs, do a single query for known IDs, then upsert only the missing ones. `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,66 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
   return { sessions: page, hasMore };
 }
 
+/**
+ * Look up a single session by ID via the Claude SDK.
+ * Used as a fallback when the EventStore doesn't have the session yet
+ * (e.g. orphaned by a restart or created externally). If found, backfills
+ * the EventStore so subsequent lookups are fast.
+ */
+export async function discoverSession(
```
dimakis, line 902: 🟡 **missing_tests**: `discoverSession` has no dedicated unit tests. The WS handler test covers it indirectly via the mock, but the actual function logic (SDK call, EventStore backfill, error handling) is untested. A unit test for the success path, SDK-returns-null path, and SDK-throws path would be valuable. `[fixable]`

server/chat.ts
dimakis, line 873: 🔵 **missing_tests**: The `getSessions` reconciliation loop (backfilling EventStore for orphaned sessions) has no test coverage. A test verifying that sessions present in the SDK but absent from EventStore get upserted would prevent regressions. `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,66 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
             summary: s.summary,
             lastModified: s.lastModified,
             branch: s.gitBranch,
+            cwd: s.cwd ?? dir,
```
dimakis: 🔵 **bugs**: The fallback `cwd: s.cwd ?? dir` uses the session directory path as a CWD fallback when the SDK doesn't provide one. This session directory (e.g. `~/.claude/projects/...`) is not a valid working directory for the project — it's the SDK's internal storage path. If `s.cwd` is null, it would be more accurate to use `BASE_REPO` or omit `cwd` entirely. `[fixable]`

scripts/deploy.sh
```diff
@@ -8,7 +8,12 @@ npm run build:server
 echo "Building frontend..."
 npm run build
 
+echo "Gracefully stopping service..."
```
dimakis: 🔵 **style**: The `launchctl kill SIGTERM` + `sleep 3` before `kickstart -k` is redundant — `kickstart -k` already sends SIGTERM and waits for termination before restarting. The extra kill + sleep adds 3 seconds to every deploy with no benefit. If graceful drain is needed, `kickstart -k` handles it. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_The async dispatch change introduces a message reordering risk that could cause race conditions between switch_session and subsequent messages on the same connection. The reconciliation logic is sound in concept but needs serialization or per-connection queuing to be safe._

- 🔴 **regressions** (L248): Making `dispatchV2Message` async and calling it without `await` in the WS `message` handler introduces message reordering. If a `switch_session` (which now awaits `discoverSession`) is in flight when a subsequent `send` message arrives, the `send` may be dispatched before `switch_session` completes, executing against the wrong (or no) active session. The previous synchronous dispatch guaranteed FIFO ordering. Consider serializing dispatch per connection (e.g., a per-connection promise chain). `[fixable]`


### `server/chat.ts`
_The async dispatch change introduces a message reordering risk that could cause race conditions between switch_session and subsequent messages on the same connection. The reconciliation logic is sound in concept but needs serialization or per-connection queuing to be safe._

- 🟡 **unsafe_assumptions** (L871): The reconciliation loop in `getSessions` calls `eventStore.getSession()` for every SDK-discovered session on every page fetch. With hundreds of sessions across multiple dirs, this is O(n) synchronous SQLite reads on a hot path (session list is polled by the frontend). Consider batching: collect all session IDs, do a single query for known IDs, then upsert only the missing ones. `[fixable]`
- 🟡 **missing_tests** (L896): `discoverSession` has no dedicated unit tests. The WS handler test covers it indirectly via the mock, but the actual function logic (SDK call, EventStore backfill, error handling) is untested. A unit test for the success path, SDK-returns-null path, and SDK-throws path would be valuable. `[fixable]`
- 🔵 **missing_tests** (L871): The `getSessions` reconciliation loop (backfilling EventStore for orphaned sessions) has no test coverage. A test verifying that sessions present in the SDK but absent from EventStore get upserted would prevent regressions. `[fixable]`
- 🔵 **bugs** (L859): The fallback `cwd: s.cwd ?? dir` uses the session directory path as a CWD fallback when the SDK doesn't provide one. This session directory (e.g. `~/.claude/projects/...`) is not a valid working directory for the project — it's the SDK's internal storage path. If `s.cwd` is null, it would be more accurate to use `BASE_REPO` or omit `cwd` entirely. `[fixable]`


### `scripts/deploy.sh`
_The async dispatch change introduces a message reordering risk that could cause race conditions between switch_session and subsequent messages on the same connection. The reconciliation logic is sound in concept but needs serialization or per-connection queuing to be safe._

- 🔵 **style** (L11): The `launchctl kill SIGTERM` + `sleep 3` before `kickstart -k` is redundant — `kickstart -k` already sends SIGTERM and waits for termination before restarting. The extra kill + sleep adds 3 seconds to every deploy with no benefit. If graceful drain is needed, `kickstart -k` handles it. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -280,6 +280,19 @@ export class EventStore {
     return row ? rowToSession(row) : null;
   }
 
+  /**
+   * Return the subset of `sessionIds` that already exist in the sessions table.
+   * Single query regardless of input size — used for batch reconciliation.
+   */
+  getKnownSessionIds(sessionIds: string[]): Set<string> {
+    if (sessionIds.length === 0) return new Set();
+    const placeholders = sessionIds.map(() => '?').join(',');
```
dimakis: 🟡 **missing_tests**: New `getKnownSessionIds` method has no unit test in the protocol package. The discover-session test mocks it rather than exercising the real SQL. Add a test in the protocol package that verifies correct set-membership semantics (empty input, partial match, full match, unknown IDs). `[fixable]`

dimakis: 🟡 **unsafe_assumptions**: SQLite has a default `SQLITE_MAX_VARIABLE_NUMBER` limit (typically 999 or 32766 depending on version). If `sessionIds` exceeds this, the query will throw. The SDK's `listSessions` could return a large number of sessions. Consider chunking into batches (e.g. 500) or documenting the assumption. `[fixable]`

server/__tests__/discover-session.test.ts
```diff
@@ -0,0 +1,148 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
dimakis, line 1: 🔵 **missing_tests**: The `getSessions reconciliation` test verifies backfill is called but doesn't assert that the returned session list actually includes the backfilled sessions (i.e., that the reconciliation is visible to the caller). Consider asserting on the return value of `getSessions()`. `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,67 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
           });
         }
       }
     } catch {
       // Expected when session dir doesn't exist
     }
   }
+
+  // Reconcile: backfill EventStore for SDK-discovered sessions that Mitzo
+  // doesn't track (e.g. sessions orphaned by a server restart mid-query,
```
dimakis, line 869: 🔵 **style**: The reconciliation loop logs `log.info` per orphaned session. In workspaces with many worktree sessions this could be noisy on every `getSessions` call. Consider `log.debug` or batching into a single log line with the count. `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,67 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
+ * (e.g. orphaned by a restart or created externally). If found, backfills
+ * the EventStore so subsequent lookups are fast.
+ */
+export async function discoverSession(
+  sessionId: string,
+): Promise<import('./event-store.js').SessionMeta | null> {
+  try {
+    const info = await getSessionInfo(sessionId);
+    if (!info) return null;
+    eventStore.upsertSession({
```
dimakis, line 908: 🔵 **bugs**: `discoverSession` calls `eventStore.upsertSession` then immediately returns `eventStore.getSession(sessionId)`. If `upsertSession` does not set all `SessionMeta` fields (e.g. `mode` defaults to something unexpected), the returned meta may have surprising defaults. The upsert only provides `sessionId`, `summary`, `cwd`, `branch` — but `SessionMeta` includes `mode`, `isActive`, `promptCount`, etc. Verify the DB defaults align with caller expectations in `handleSwitchSession`.

scripts/start.sh
```diff
@@ -1,3 +1,18 @@
...
+
+LOGDIR="logs"
+mkdir -p "$LOGDIR"
+
+# Rotate previous logs before starting (launchd truncates StandardOutPath
+# on each process start, so we preserve the old log here instead).
+for f in "$LOGDIR/server-stdout.log" "$LOGDIR/server-stderr.log"; do
+  if [ -s "$f" ]; then
+    mv "$f" "${f%.log}.$(date +%Y%m%d-%H%M%S).log"
+  fi
```
dimakis, line 12: 🔵 **style**: The find pattern `'server-*.2*.log'` relies on rotated filenames containing a `2` (from year 2xxx). This is fragile — a more explicit pattern like `'server-std*.*.log'` or matching the full date format would be more robust. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid reconciliation strategy with good error handling and test coverage for the core flows. Main gaps are missing unit tests for the new EventStore method and a potential SQLite variable limit issue with large session lists._

- 🟡 **missing_tests** (L289): New `getKnownSessionIds` method has no unit test in the protocol package. The discover-session test mocks it rather than exercising the real SQL. Add a test in the protocol package that verifies correct set-membership semantics (empty input, partial match, full match, unknown IDs). `[fixable]`
- 🟡 **unsafe_assumptions** (L289): SQLite has a default `SQLITE_MAX_VARIABLE_NUMBER` limit (typically 999 or 32766 depending on version). If `sessionIds` exceeds this, the query will throw. The SDK's `listSessions` could return a large number of sessions. Consider chunking into batches (e.g. 500) or documenting the assumption. `[fixable]`


### `server/__tests__/discover-session.test.ts`
_Solid reconciliation strategy with good error handling and test coverage for the core flows. Main gaps are missing unit tests for the new EventStore method and a potential SQLite variable limit issue with large session lists._

- 🔵 **missing_tests** (L1): The `getSessions reconciliation` test verifies backfill is called but doesn't assert that the returned session list actually includes the backfilled sessions (i.e., that the reconciliation is visible to the caller). Consider asserting on the return value of `getSessions()`. `[fixable]`


### `server/chat.ts`
_Solid reconciliation strategy with good error handling and test coverage for the core flows. Main gaps are missing unit tests for the new EventStore method and a potential SQLite variable limit issue with large session lists._

- 🔵 **style** (L869): The reconciliation loop logs `log.info` per orphaned session. In workspaces with many worktree sessions this could be noisy on every `getSessions` call. Consider `log.debug` or batching into a single log line with the count. `[fixable]`
- 🔵 **bugs** (L903): `discoverSession` calls `eventStore.upsertSession` then immediately returns `eventStore.getSession(sessionId)`. If `upsertSession` does not set all `SessionMeta` fields (e.g. `mode` defaults to something unexpected), the returned meta may have surprising defaults. The upsert only provides `sessionId`, `summary`, `cwd`, `branch` — but `SessionMeta` includes `mode`, `isActive`, `promptCount`, etc. Verify the DB defaults align with caller expectations in `handleSwitchSession`.


### `scripts/start.sh`
_Solid reconciliation strategy with good error handling and test coverage for the core flows. Main gaps are missing unit tests for the new EventStore method and a potential SQLite variable limit issue with large session lists._

- 🔵 **style** (L12): The find pattern `'server-*.2*.log'` relies on rotated filenames containing a `2` (from year 2xxx). This is fragile — a more explicit pattern like `'server-std*.*.log'` or matching the full date format would be more robust. `[fixable]`

server/chat.ts
```diff
@@ -843,20 +856,67 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
+  // or created by external agents in a worktree).
+  const knownIds = eventStore.getKnownSessionIds(Array.from(seen.keys()));
+  for (const [sessionId, entry] of seen) {
+    if (!knownIds.has(sessionId)) {
+      eventStore.upsertSession({
+        sessionId,
+        summary: entry.summary || null,
+        cwd: entry.cwd ?? (BASE_REPO || null),
+        branch: entry.branch ?? null,
+      });
```
dimakis, line 881: 🟡 **bugs**: `getSessions()` now backfills orphaned SDK sessions into the EventStore with only `summary/cwd/branch`, but it never marks them inactive/hidden. Because `listSessions()` reads from the EventStore and `markSessionInactive()` is only called on normal teardown, externally-created or already-finished SDK sessions can become permanently listed as active Mitzo sessions even though no live session exists. Reconcile should preserve the existing inactive semantics or explicitly set a safe default state for backfilled rows. `[fixable]`

dimakis (commented):
## Codex Review

### `server/chat.ts`
_PR mostly addresses session reconciliation correctly, but the new EventStore backfill path appears to misclassify orphaned SDK sessions as active._

- 🟡 **bugs** (L879): `getSessions()` now backfills orphaned SDK sessions into the EventStore with only `summary/cwd/branch`, but it never marks them inactive/hidden. Because `listSessions()` reads from the EventStore and `markSessionInactive()` is only called on normal teardown, externally-created or already-finished SDK sessions can become permanently listed as active Mitzo sessions even though no live session exists. Reconcile should preserve the existing inactive semantics or explicitly set a safe default state for backfilled rows. `[fixable]`

dimakis (commented):
## Codex Review

### `server/chat.ts`
_Good reconciliation fix overall, but SDK-backed session discovery currently mislabels recovered active sessions as inactive._

- 🟡 **bugs**: `discoverSession()` backfills the EventStore with `isActive: false` unconditionally. If the requested session is currently running in this process but was missing from the EventStore (the exact orphaned-session case this PR targets), switching to it will still report it as inactive in session metadata and session lists, regressing active-state accuracy until some later event updates the row. `[fixable]`

dimakis:
Superseded by new PR targeting main (original was against the now-merged refactor/ws-v2-phase1 branch).

PR #229, opened 2026-04-18
fix: v2 review round 2 — iOS reconnect, interrupt routing, skill policy

dimakis:
## Summary

- **iOS reconnect resilience**: heartbeat timer detects silently-dead sockets, `visibilitychange` and `pageshow` listeners force reconnect on foreground return
- **handleInterruptV2**: now calls `watch`/`setActive` so the interrupting connection receives session events
- **Skill policy timing**: `applySkillPolicy` helper ensures `setSkillPolicy` is called with the correct `clientId` after session registration
- **InboxView cleanup**: removed redundant raw `fetch` — now driven entirely by store's `loadInbox` + WS `inbox_updated` events, with optimistic removal via `pendingRemovals`

## Test plan

- [x] 4 new tests for iOS heartbeat reconnect and browser listener handling (`connection.test.ts`)
- [x] 2 new tests for `handleInterruptV2` watch/activate and no-op paths (`ws-handler-v2.test.ts`)
- [x] All 1739 tests pass

### DIFF
--- a/packages/client/__tests__/connection.test.ts
+++ b/packages/client/__tests__/connection.test.ts
@@ -295,4 +295,36 @@ describe('MitzoConnection', () => {
       vi.useRealTimers();
     });
   });
+
+  describe('iOS reconnect resilience', () => {
+    it('reconnects when heartbeat detects dead socket', () => {
+      vi.useFakeTimers();
+      const conn = createConnection();
+      conn.onMessage(() => {});
+      const ws = openWithHandshake(conn);
+
+      // Simulate iOS silent death — readyState changes without onclose firing
+      ws.readyState = 3; // CLOSED
+
+      // Heartbeat fires every 5s
+      vi.advanceTimersByTime(5_000);
+
+      // A new WS should have been created
+      expect(lastWs).not.toBe(ws);
+
+      vi.useRealTimers();
+    });
+
+    // Browser-specific tests (visibilitychange, pageshow) require a DOM
+    // environment and are covered by the frontend test suite. Here we verify
+    // the heartbeat path which works in Node.
+
+    it('skips browser listeners in non-browser environment', () => {
+      // In Node, document is undefined — connect should not throw
+      const conn = createConnection();
+      conn.onMessage(() => {});
+      openWithHandshake(conn);
+      conn.disconnect();
+    });
+  });
 });
--- a/packages/client/__tests__/connection.test.ts
+++ b/packages/client/__tests__/connection.test.ts
@@ -319,6 +319,21 @@ describe('MitzoConnection', () => {
     // environment and are covered by the frontend test suite. Here we verify
     // the heartbeat path which works in Node.
 
+    it('emits _close when heartbeat detects dead socket', () => {
+      vi.useFakeTimers();
+      const conn = createConnection();
+      const received: Record<string, unknown>[] = [];
+      conn.onMessage((msg) => received.push(msg));
+      openWithHandshake(conn);
+
+      // Kill socket silently
+      lastWs!.readyState = 3;
+      vi.advanceTimersByTime(5_000);
+
+      expect(received.some((m) => m.type === '_close')).toBe(true);
+      vi.useRealTimers();
+    });
+
     it('skips browser listeners in non-browser environment', () => {
       // In Node, document is undefined — connect should not throw
       const conn = createConnection();
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -18,6 +18,7 @@ export interface MitzoConnectionConfig {
 export type ConnectionListener = (msg: Record<string, unknown>) => void;
 
 const MAX_PENDING_SENDS = 100;
+const HEARTBEAT_INTERVAL_MS = 5_000;
 
 export class MitzoConnection {
   private ws: WebSocketLike | null = null;
@@ -28,6 +29,9 @@ export class MitzoConnection {
   private seqBySession = new Map<string, number>();
   private pendingSends: string[] = [];
   private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
+  private heartbeatTimer: ReturnType<typeof setInterval> | null = null;
+  private boundOnVisibility: (() => void) | null = null;
+  private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
   private config: Required<MitzoConnectionConfig>;
 
   constructor(config: MitzoConnectionConfig) {
@@ -39,9 +43,13 @@ export class MitzoConnection {
 
   connect(): void {
     this.doConnect();
+    this.startHeartbeat();
+    this.addBrowserListeners();
   }
 
   disconnect(): void {
+    this.stopHeartbeat();
+    this.removeBrowserListeners();
     if (this.reconnectTimer) {
       clearTimeout(this.reconnectTimer);
       this.reconnectTimer = null;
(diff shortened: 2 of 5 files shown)


dimakis:
## Centaur Review

### `frontend/src/pages/InboxView.tsx`
_Solid fixes for iOS reconnect, interrupt routing, and skill policy timing. Two behavioral issues: InboxView loading state regresses (empty flash instead of spinner), and checkAndReconnect omits the _close event that the heartbeat path emits._

- 🟡 **regressions** (L193): Loading state regression: the sync effect (line 188–194) fires on mount with the store's initial empty array, immediately setting `loading = false` before `loadInbox()` completes. Previously, `loadItems` kept `loading = true` until `.finally()`. Users will now see the empty state flash instead of a loading spinner on first load. `[fixable]`


_Solid fixes for iOS reconnect, interrupt routing, and skill policy timing. Two behavioral issues: InboxView loading state regresses (empty flash instead of spinner), and checkAndReconnect omits the _close event that the heartbeat path emits._

- 🟡 **bugs** (L246): `checkAndReconnect()` (called from visibilitychange/pageshow) does not emit `{ type: '_close' }` to the listener before reconnecting, but the heartbeat path at line 205 does. This inconsistency means the app layer won't be notified of a connection loss when recovery is triggered by a browser visibility event. `[fixable]`


_Solid fixes for iOS reconnect, interrupt routing, and skill policy timing. Two behavioral issues: InboxView loading state regresses (empty flash instead of spinner), and checkAndReconnect omits the _close event that the heartbeat path emits._

- 🔵 **missing_tests** (L426): The handleInterruptV2 test verifies watch/setActive state but does not assert that `interruptChat` was called with the correct arguments (`found.clientId`, `msg.prompt`, images, contextBlocks, clientMsgId`). Since this handler's primary purpose is routing the interrupt, asserting the call would strengthen coverage. `[fixable]`
- 🔵 **missing_tests**: No test covers the skill policy timing fix in `handleSendV2`. The key change is that `applySkillPolicy` now targets the correct `clientId` depending on the path (resume vs. new session). A test that verifies `setSkillPolicy` is called with `found.clientId` on the resume path (not `connectionId`) would prevent regression of the fix. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -181,4 +189,67 @@ export class MitzoConnection {
...
+  // ─── iOS reconnect resilience ─────────────────────────────────────────────
+  // iOS Safari silently kills WebSocket connections when the app is backgrounded
+  // or the screen is locked. The onclose event may never fire. These listeners
+  // detect when the app returns to the foreground and force a reconnect if the
+  // socket is dead.
+
+  private startHeartbeat(): void {
+    if (typeof globalThis.setInterval === 'undefined') return;
+    this.heartbeatTimer = setInterval(() => {
+      if (this.ws && this.ws.readyState !== WS_READY_STATE.OPEN && !this.reconnectTimer) {
```
dimakis, line 202: 🔴 **bugs**: Heartbeat treats CONNECTING (readyState=0) the same as CLOSED/CLOSING. After the heartbeat tears down a dead socket and calls doConnect(), the new WS starts in CONNECTING state. 5s later the next heartbeat tick sees `readyState !== OPEN` → tears it down again → reconnect loop. Fix: check `readyState > WS_READY_STATE.OPEN` (i.e. CLOSING or CLOSED) instead of `!== OPEN`. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -181,4 +189,67 @@ export class MitzoConnection {
...
+      this.boundOnVisibility = null;
+    }
+    if (this.boundOnPageShow) {
+      globalThis.removeEventListener('pageshow', this.boundOnPageShow);
+      this.boundOnPageShow = null;
+    }
+  }
+
+  private checkAndReconnect(): void {
+    if (!this.ws || this.ws.readyState !== WS_READY_STATE.OPEN) {
```
dimakis, line 247: 🟡 **bugs**: Same CONNECTING-state issue in checkAndReconnect(): `readyState !== OPEN` will tear down a socket that's still connecting (e.g., user switches back to app while a reconnect is mid-handshake). Should also exclude CONNECTING state. `[fixable]`

packages/client/__tests__/connection.test.ts
```diff
@@ -295,4 +295,51 @@ describe('MitzoConnection', () => {
...
+
+      // A new WS should have been created
+      expect(lastWs).not.toBe(ws);
+
+      vi.useRealTimers();
+    });
+
+    // Browser-specific tests (visibilitychange, pageshow) require a DOM
+    // environment and are covered by the frontend test suite. Here we verify
+    // the heartbeat path which works in Node.
```
dimakis, line 320: 🟡 **missing_tests**: No test verifying the heartbeat leaves a CONNECTING socket alone. Given the bug above, a test like `ws.readyState = 0; advanceTimersByTime(5000); expect(lastWs).toBe(ws)` would catch the regression. `[fixable]`

server/__tests__/ws-handler-v2.test.ts
```diff
@@ -415,6 +418,103 @@ describe('handleSendV2 auto-watch', () => {
   });
```
dimakis, line 418: 🔵 **missing_tests**: Skill policy test covers the active-resume path but not the resume-inactive or new-session paths where `applySkillPolicy(connectionId)` is called. A test for the resume path with a skill would increase confidence that the policy is applied after startChat. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -364,6 +373,8 @@ export function handleInterruptV2(
 ): void {
```
dimakis, line 373: 🔵 **regressions**: handleInterruptV2 doesn't resolve slash commands or apply skill policy. If a user sends `/simplify` as an interrupt, it will be sent as raw text rather than resolved. This matches the pre-PR behavior (interrupts were always raw) so not a regression, but worth noting as a gap.

dimakis (commented):
## Centaur Review

_Heartbeat and checkAndReconnect treat CONNECTING sockets as dead, causing a 5-second reconnect loop after any iOS recovery; skill policy and interrupt routing fixes look correct._

- 🔴 **bugs** (L202): Heartbeat treats CONNECTING (readyState=0) the same as CLOSED/CLOSING. After the heartbeat tears down a dead socket and calls doConnect(), the new WS starts in CONNECTING state. 5s later the next heartbeat tick sees `readyState !== OPEN` → tears it down again → reconnect loop. Fix: check `readyState > WS_READY_STATE.OPEN` (i.e. CLOSING or CLOSED) instead of `!== OPEN`. `[fixable]`
- 🟡 **bugs** (L247): Same CONNECTING-state issue in checkAndReconnect(): `readyState !== OPEN` will tear down a socket that's still connecting (e.g., user switches back to app while a reconnect is mid-handshake). Should also exclude CONNECTING state. `[fixable]`


_Heartbeat and checkAndReconnect treat CONNECTING sockets as dead, causing a 5-second reconnect loop after any iOS recovery; skill policy and interrupt routing fixes look correct._

- 🟡 **missing_tests** (L320): No test verifying the heartbeat leaves a CONNECTING socket alone. Given the bug above, a test like `ws.readyState = 0; advanceTimersByTime(5000); expect(lastWs).toBe(ws)` would catch the regression. `[fixable]`


_Heartbeat and checkAndReconnect treat CONNECTING sockets as dead, causing a 5-second reconnect loop after any iOS recovery; skill policy and interrupt routing fixes look correct._

- 🔵 **missing_tests** (L418): Skill policy test covers the active-resume path but not the resume-inactive or new-session paths where `applySkillPolicy(connectionId)` is called. A test for the resume path with a skill would increase confidence that the policy is applied after startChat. `[fixable]`


### `server/ws-handler-v2.ts`
_Heartbeat and checkAndReconnect treat CONNECTING sockets as dead, causing a 5-second reconnect loop after any iOS recovery; skill policy and interrupt routing fixes look correct._

- 🔵 **regressions** (L373): handleInterruptV2 doesn't resolve slash commands or apply skill policy. If a user sends `/simplify` as an interrupt, it will be sent as raw text rather than resolved. This matches the pre-PR behavior (interrupts were always raw) so not a regression, but worth noting as a gap.

PR #222, opened 2026-04-17
feat: v2 WS protocol phase 1a+1b — ConnectionRegistry + sessionId tagging

dimakis:
## Summary

Phase 1a and 1b of the single-WS architecture migration. Additive — no breaking changes, v1 protocol untouched.

### Phase 1a — ConnectionRegistry

New `packages/harness/src/connection-registry.ts`: tracks WS connections and their relationship to sessions, separate from SessionRegistry (which tracks SDK sessions).

- Each browser tab has one connection (connectionId)
- A connection can watch multiple sessions (receiving events for each)
- Each connection has one active session (foreground for that tab)
- API: register, get, remove, watch, unwatch, setActive, getConnectionsWatching, broadcast
- 15 tests

### Phase 1b — sessionId tagging on outbound events

- `sendOrBuffer` in `server/query-loop.ts` now adds `sessionId` to every v2 payload before delivery and persistence
- `worktree_opened` events from the lazy worktree HTTP endpoint also include `sessionId`
- v1 clients ignore the extra field; v2 clients will use it for session demuxing on a single WS

## Test plan

- [x] 15 ConnectionRegistry tests pass
- [x] 382 query-loop tests pass (sessionId tagging is additive)
- [x] 213 harness tests pass
- [ ] CI green

## What's next

Phase 1c: v2 message handler in server/index.ts (reconnect, watch/unwatch, switch_session, send with explicit sessionId, OTel spans).

Made with [Cursor](https://cursor.com)

PR #231, opened 2026-04-18
fix: reconcile orphaned sessions between SDK and EventStore

dimakis:
## Summary

Fixes the "Session expired" error on mobile when tapping sessions that were orphaned by a server restart mid-query or created by external agents in worktrees.

- **Self-healing EventStore**: `getSessions()` reconciles SDK-discovered sessions into the EventStore using a single batch query (`getKnownSessionIds`), chunked to stay within SQLite variable limits. Backfilled sessions are marked inactive.
- **Resilient `switch_session`**: Falls back to `getSessionInfo()` (targeted SDK lookup) when the EventStore misses, backfills the row, then proceeds. Fixes the missing OTel span status on the not-found path.
- **FIFO dispatch**: Per-connection promise chain in `handleChatWsV2` serializes async dispatch, preventing message reordering when `switch_session` awaits SDK discovery.
- **Early registration on resume**: `startChat()` pre-registers resumed sessions in the EventStore before the query loop starts.
- **Log rotation**: `start.sh` rotates logs before starting so launchd's truncate-on-start doesn't destroy the previous instance's history.

Supersedes #227 (rebased onto main after #222 merged).

## Local CI results

- [x] Build workspace packages — pass
- [x] Lint — pass (1 pre-existing warning in untracked local file)
- [x] Format check — pass
- [x] Type check (server + frontend) — pass
- [x] Tests — 1734 passed (guard: ≥250)
- [x] Full build — pass

## Test plan

- [x] `getKnownSessionIds`: empty, partial, full match, unknown IDs, large batch (600 IDs spanning chunk boundary)
- [x] `discoverSession`: success+backfill, SDK returns null, SDK throws
- [x] `getSessions` reconciliation: verifies orphaned sessions get backfilled with `isActive: false`
- [x] `handleSwitchSession`: SDK fallback succeeds, error when both fail, existing tests updated for async
- [x] All 1734 existing tests pass

### DIFF
--- a/packages/protocol/__tests__/event-store.test.ts
+++ b/packages/protocol/__tests__/event-store.test.ts
@@ -318,6 +318,46 @@ describe('EventStore', () => {
     });
   });
 
+  describe('getKnownSessionIds', () => {
+    it('returns empty set for empty input', () => {
+      expect(store.getKnownSessionIds([])).toEqual(new Set());
+    });
+
+    it('returns empty set when none of the IDs exist', () => {
+      expect(store.getKnownSessionIds(['a', 'b', 'c'])).toEqual(new Set());
+    });
+
+    it('returns only IDs that exist in the store', () => {
+      store.upsertSession({ sessionId: 'sess-1' });
+      store.upsertSession({ sessionId: 'sess-3' });
+
+      const result = store.getKnownSessionIds(['sess-1', 'sess-2', 'sess-3', 'sess-4']);
+      expect(result).toEqual(new Set(['sess-1', 'sess-3']));
+    });
+
+    it('returns all IDs when all exist', () => {
+      store.upsertSession({ sessionId: 'a' });
+      store.upsertSession({ sessionId: 'b' });
+
+      expect(store.getKnownSessionIds(['a', 'b'])).toEqual(new Set(['a', 'b']));
+    });
+
+    it('handles large batches via chunking', () => {
+      const ids: string[] = [];
+      for (let i = 0; i < 600; i++) {
+        const id = `sess-${i}`;
+        ids.push(id);
+        if (i % 2 === 0) store.upsertSession({ sessionId: id });
+      }
+      const result = store.getKnownSessionIds(ids);
+      expect(result.size).toBe(300);
+      expect(result.has('sess-0')).toBe(true);
+      expect(result.has('sess-1')).toBe(false);
+      expect(result.has('sess-598')).toBe(true);
+      expect(result.has('sess-599')).toBe(false);
+    });
+  });
+
   describe('close', () => {
     it('is safe to call multiple times', () => {
       store.close();
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -261,13 +261,14 @@ export class EventStore {
       );
     } else {
       this.db!.prepare(
-        'INSERT INTO sessions (session_id, summary, branch, cwd, mode, initial_prompt, wt_id, goal_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?)',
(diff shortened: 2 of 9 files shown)


dimakis:
## Centaur Review

### `server/chat.ts`
_Solid reconciliation mechanism with good test coverage. Main concern is the global dispatch chain serializing all WS message types, which could delay control signals (stop/interrupt) behind slow SDK discovery calls._

- 🟡 **bugs** (L878): `entry.cwd ?? (BASE_REPO || null)` — if `BASE_REPO` is the empty string `""`, `||` coerces it to `null`. This is probably intentional (empty string is not a valid cwd), but `??` is used for `entry.cwd` while `||` is used for `BASE_REPO`, creating an inconsistency. If `entry.cwd` is `""` it will be kept as-is, but `BASE_REPO=""` becomes `null`. Minor, but worth aligning. `[fixable]`
- 🔵 **style** (L904): `discoverSession` uses an inline `import('./event-store.js').SessionMeta` type annotation instead of importing `SessionMeta` from the existing import at line 39. This is inconsistent with the rest of the file — prefer a direct import for readability. `[fixable]`


### `server/index.ts`
_Solid reconciliation mechanism with good test coverage. Main concern is the global dispatch chain serializing all WS message types, which could delay control signals (stop/interrupt) behind slow SDK discovery calls._

- 🟡 **unsafe_assumptions** (L250): The `dispatchChain` promise chain serializes ALL message types for a connection, not just `switch_session`. This means a slow SDK `getSessionInfo` call during `switch_session` blocks processing of subsequent `send`, `stop`, and `interrupt` messages on that connection. In particular, `stop`/`interrupt` (which should be low-latency control signals) could be delayed. Consider only serializing `switch_session` or using a lock per-session-id rather than a global chain. `[fixable]`


### `server/__tests__/discover-session.test.ts`
_Solid reconciliation mechanism with good test coverage. Main concern is the global dispatch chain serializing all WS message types, which could delay control signals (stop/interrupt) behind slow SDK discovery calls._

- 🔵 **missing_tests**: The `getSessions` reconciliation test doesn't verify behavior when `getKnownSessionIds` returns all IDs (i.e. no orphans exist). A no-op reconciliation test would confirm `upsertSession` is NOT called when everything is already known. `[fixable]`
- 🔵 **missing_tests**: No test covers the `startChat` resume pre-registration path (lines 481-487 in chat.ts). This is a new code path that ensures resumed sessions are discoverable even if the query loop dies early. `[fixable]`


### `server/ws-handler-v2.ts`
_Solid reconciliation mechanism with good test coverage. Main concern is the global dispatch chain serializing all WS message types, which could delay control signals (stop/interrupt) behind slow SDK discovery calls._

- 🟡 **regressions** (L233): `handleSwitchSession` now throws from line 239 (`throw err`) after recording the span error. Since `dispatchV2Message` doesn't catch errors from `await handleSwitchSession(...)`, the throw propagates to the caller. In `index.ts` the `.catch()` on the promise chain handles it, but the old code never threw — the try/catch in the previous sync `dispatchV2Message` would have caught it. Verify that unexpected SDK errors (e.g. network timeout in `discoverSession`) don't crash the dispatch chain or leave it in a broken state. The chain's `.catch()` does NOT re-resolve cleanly — it returns `undefined`, so the chain continues, but an unhandled throw from `handleSwitchSession` could leave the span in a bad state if `finally` doesn't run (it does here, so this is OK, but fragile).


### `scripts/start.sh`
_Solid reconciliation mechanism with good test coverage. Main concern is the global dispatch chain serializing all WS message types, which could delay control signals (stop/interrupt) behind slow SDK discovery calls._

- 🔵 **style** (L13): The `find` command for log pruning uses `-name 'server-std*.????????-??????.log'` which is a fairly loose glob. Consider a tighter pattern like `'server-std*.[0-9]*-[0-9]*.log'` to avoid accidentally matching unrelated files, though the risk is low given the dedicated `logs/` directory. `[fixable]`

PR #232, opened 2026-04-18
fix(harness): fire abort signal before clearing closingOut set

dimakis:
## Summary

- abort() deleted from closingOut before calling abortController.abort(), so abort signal listeners could never observe isClosingOut()=true
- This meant closeoutSession always reported status 'closed', never 'abandoned'
- Reorder: fire the signal first, then clean up

## Test plan

- [x] Added test verifying isClosingOut()=true during abort signal

### DIFF
--- a/packages/harness/src/session-registry.ts
+++ b/packages/harness/src/session-registry.ts
@@ -293,8 +293,10 @@ export class SessionRegistry {
     const session = this.sessions.get(clientId);
     if (session) session.observers.clear();
     this.clearDetachTimer(clientId);
+    this.clearCloseoutTimer(clientId);
     this.sessions.delete(clientId);
     this.attached.delete(clientId);
+    this.closingOut.delete(clientId);
   }
 
   /**
--- a/packages/harness/src/session-registry.ts
+++ b/packages/harness/src/session-registry.ts
@@ -142,7 +142,8 @@ export class SessionRegistry {
         // Phase 2: hard abort after CLOSEOUT_TIMEOUT_MS
         const abortTimer = setTimeout(() => {
           this.closeoutTimers.delete(clientId);
-          this.closingOut.delete(clientId);
+          // Don't delete closingOut here — abort() will do it AFTER
+          // firing the abort signal, so listeners can check isClosingOut().
           if (this.sessions.has(clientId) && !this.attached.has(clientId)) {
             log.info(`closeout timeout for ${clientId}, aborting`);
             this.abort(clientId);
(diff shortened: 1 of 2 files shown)


packages/harness/src/session-registry.ts
```diff
@@ -275,8 +276,10 @@ export class SessionRegistry {
 
     this.clearDetachTimer(clientId);
     this.clearCloseoutTimer(clientId);
-    this.closingOut.delete(clientId);
+    // Fire abort signal BEFORE clearing closingOut — abort listeners
+    // check isClosingOut() to distinguish 'abandoned' vs 'closed' status.
     session.abortController.abort();
+    this.closingOut.delete(clientId);
```
dimakis: 🔵 **regressions**: After `abort()`, `closingOut.delete` runs on line 282 but `sessions.delete` runs on line 284. Between those two lines, `isClosingOut('client-1')` returns false while `isActive('client-1')` still returns true — a brief inconsistent state. This is fine in practice (single-threaded, no yielding between), but worth noting.

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `packages/harness/__tests__/session-registry.test.ts`
_Clean, well-targeted fix — reorders abort signal vs closingOut cleanup so chat.ts abort listeners can correctly distinguish 'abandoned' from 'closed'. Test directly validates the invariant._

- 🔵 **missing_tests**: The test covers the closeout-timeout path but not a direct `abort()` call while closing out (e.g., user explicitly stops a session mid-closeout). Consider a test that calls `registry.abort('client-1')` directly during closeout and verifies the abort listener still sees `isClosingOut=true`. `[fixable]`


_Clean, well-targeted fix — reorders abort signal vs closingOut cleanup so chat.ts abort listeners can correctly distinguish 'abandoned' from 'closed'. Test directly validates the invariant._

- 🔵 **regressions** (L282): After `abort()`, `closingOut.delete` runs on line 282 but `sessions.delete` runs on line 284. Between those two lines, `isClosingOut('client-1')` returns false while `isActive('client-1')` still returns true — a brief inconsistent state. This is fine in practice (single-threaded, no yielding between), but worth noting.

dimakis:
## Centaur Review

_Correct fix — moving `closingOut.delete` after `abort()` ensures listeners can distinguish abandoned sessions. Minor gap: `remove()` doesn't clean up `closingOut` either (pre-existing)._

- 🟡 **regressions** (L292): `remove()` never clears `closingOut`. If a session completes naturally while in closeout phase (e.g., the closeout handler finishes the SDK query gracefully), `closingOut` will leak the clientId. This was also true before this PR, but the PR would be a good opportunity to fix it — add `this.closingOut.delete(clientId)` in `remove()`. `[fixable]`


### `packages/harness/__tests__/session-registry.test.ts`
_Correct fix — moving `closingOut.delete` after `abort()` ensures listeners can distinguish abandoned sessions. Minor gap: `remove()` doesn't clean up `closingOut` either (pre-existing)._

- 🔵 **missing_tests**: No test covers calling `abort()` directly (not via closeout timeout) on a session that is currently in closingOut state. The new ordering guarantee matters for all `abort()` callers (e.g., `dispose()`, explicit user abort during closeout), not just the timeout path. A test like: enter closeout → call `abort()` directly → verify listener sees `isClosingOut=true` would strengthen coverage. `[fixable]`

PR #230, opened 2026-04-18
feat: replace console.log logger with Pino structured logging

dimakis:
## Summary

- Replace the custom `console.log`-based logger (`server/logger.ts`) with **Pino** for structured JSON output, daily file rotation via `pino-roll`, and OTel trace context correlation (`trace_id`/`span_id` injected via mixin)
- Replace the harness logger (`packages/harness/src/logger.ts`) with Pino (stdout-only, no file transport or OTel — appropriate for a library)
- Swap `SimpleSpanProcessor` → `BatchSpanProcessor` in `server/tracing.ts` to reduce overhead ahead of deep span instrumentation

The `createLogger(module)` API is unchanged — zero call-site modifications across 17+ server files and 5 harness files.

## What changed

| File | Change |
|------|--------|
| `server/logger.ts` | Pino with dual transport (pino-roll file + stdout), OTel mixin, error serialization |
| `packages/harness/src/logger.ts` | Pino with stdout only |
| `server/tracing.ts` | `SimpleSpanProcessor` → `BatchSpanProcessor` |
| `server/__tests__/logger.test.ts` | 8 new tests (API, levels, context, OTel injection, error serialization) |
| `package.json` | Added `pino`, `pino-roll`, `pino-pretty` (dev) |
| `packages/harness/package.json` | Added `pino` |

## Log output format

```json
{
  "level": 30,
  "time": 1776518761968,
  "pid": 99974,
  "hostname": "Dimitris-MacBook-Pro",
  "module": "server",
  "msg": "Chat Agent running on https://localhost:3100 (TLS)",
  "trace_id": "abc123...",
  "span_id": "def456..."
}
```

`trace_id`/`span_id` appear when an OTel span is active (will light up with deep instrumentation in workstream 1).

## Test plan

- [x] 8 new logger unit tests passing
- [x] 129 test files / 1734 tests passing (zero regressions)
- [x] Manual smoke test: server starts, structured JSON in `logs/server-stdout.log` and `logs/server.2026-04-18.*.log` (pino-roll rotation)
- [x] Existing logger mocks (`vi.mock('../logger.js', ...)`) unaffected — they replace the whole module

Made with [Cursor](https://cursor.com)

PR #233, opened 2026-04-18
feat: replace console.log logger with Pino structured logging

dimakis:
## Summary

- Rewrite server/logger.ts with Pino: structured JSON, OTel trace context mixin, pino-roll daily rotation, pino-pretty for dev
- Rewrite packages/harness/src/logger.ts with Pino (stdout only, no file transport)
- Swap SimpleSpanProcessor to BatchSpanProcessor for lower overhead
- Add 8 logger tests

## Review fixes (Centaur)

- Lazy rootLogger init — deferred to first createLogger() call to avoid import-time side effects
- ESM-safe pino-pretty guard via createRequire instead of bare require.resolve
- Save/restore process.env.LOG_LEVEL in test beforeEach/afterEach

Replaces #230.


packages/harness/src/logger.ts
```diff
@@ -1,47 +1,27 @@
...
-
-  if (level === 'error') {
-    console.error(prefix, message, ...(hasContext ? [context] : []));
-  } else if (level === 'warn') {
-    console.warn(prefix, message, ...(hasContext ? [context] : []));
-  } else {
-    console.log(prefix, message, ...(hasContext ? [context] : []));
-  }
-}
+const rootLogger = pino({
```
dimakis, line 12: 🟡 **bugs**: Harness logger eagerly initializes `rootLogger` at module load time, so `LOG_LEVEL` is captured once at import. This is fine for production but differs from the server logger's lazy `getRootLogger()` pattern. If any consumer changes `LOG_LEVEL` after import (e.g. tests), the harness logger won't pick it up. The old implementation had the same behavior, so not a regression — but a missed opportunity to align the two loggers. `[fixable]`

server/logger.ts
```diff
@@ -1,47 +1,109 @@
...
+    return true;
+  } catch {
+    return false;
+  }
 }
 
-function emit(entry: LogEntry): void {
-  if (LEVEL_ORDER[entry.level] < LEVEL_ORDER[configuredLevel]) return;
+function buildProductionDestination(level: LogLevel): DestinationStream {
+  const logDir = join(process.cwd(), 'logs');
```
dimakis, line 35: 🟡 **unsafe_assumptions**: `buildProductionDestination` defaults `logFile` to `join(process.cwd(), 'logs', 'server.log')`. When the server is started from a different working directory (e.g. via launchd, systemd, or a test runner), `process.cwd()` won't be the repo root and logs will land in an unexpected location. Consider using `__dirname` or an env-var-based root instead. `[fixable]`

server/logger.ts
```diff
@@ -1,47 +1,109 @@
...
+function otelMixin(): Record<string, unknown> {
+  const span = trace.getSpan(context.active());
+  if (!span) return {};
+  const ctx = span.spanContext();
+  return { trace_id: ctx.traceId, span_id: ctx.spanId };
+}
 
-const configuredLevel: LogLevel =
-  (process.env.LOG_LEVEL as LogLevel) in LEVEL_ORDER ? (process.env.LOG_LEVEL as LogLevel) : 'info';
+import { createRequire } from 'module';
```
dimakis, line 22: 🔵 **style**: The `import { createRequire } from 'module'` statement is placed after the `otelMixin` function definition rather than with the other imports at the top of the file. Moving it to the top-level import block would improve readability. `[fixable]`

server/tracing.ts
```diff
@@ -7,7 +7,7 @@
  * code has zero overhead.
  */
 
-import { NodeTracerProvider, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-node';
+import { NodeTracerProvider, BatchSpanProcessor } from '@opentelemetry/sdk-trace-node';
```
dimakis, line 10: 🔵 **regressions**: Switching from `SimpleSpanProcessor` to `BatchSpanProcessor` is a good production change (reduces overhead), but note that `BatchSpanProcessor` buffers spans and flushes asynchronously. The existing `shutdown` handler calls `provider.shutdown()` which should flush, but very short-lived processes may lose the final batch. Low risk for a long-running server, but worth noting.

dimakis (commented):
## Centaur Review

### `packages/harness/src/logger.ts`
_Solid migration from console to Pino with good test coverage for the server logger, but the harness logger has no tests, and the production transport paths (pino-roll, LOGGER_SYNC) are untested._

- 🟡 **bugs** (L12): Harness logger eagerly initializes `rootLogger` at module load time, so `LOG_LEVEL` is captured once at import. This is fine for production but differs from the server logger's lazy `getRootLogger()` pattern. If any consumer changes `LOG_LEVEL` after import (e.g. tests), the harness logger won't pick it up. The old implementation had the same behavior, so not a regression — but a missed opportunity to align the two loggers. `[fixable]`
- 🟡 **missing_tests**: No tests exist for the harness logger (`packages/harness/src/logger.ts`). The server logger has thorough tests, but the harness package — which also changed from console to Pino — has zero test coverage for the new implementation. `[fixable]`


### `server/logger.ts`
_Solid migration from console to Pino with good test coverage for the server logger, but the harness logger has no tests, and the production transport paths (pino-roll, LOGGER_SYNC) are untested._

- 🟡 **unsafe_assumptions** (L35): `buildProductionDestination` defaults `logFile` to `join(process.cwd(), 'logs', 'server.log')`. When the server is started from a different working directory (e.g. via launchd, systemd, or a test runner), `process.cwd()` won't be the repo root and logs will land in an unexpected location. Consider using `__dirname` or an env-var-based root instead. `[fixable]`
- 🔵 **style** (L22): The `import { createRequire } from 'module'` statement is placed after the `otelMixin` function definition rather than with the other imports at the top of the file. Moving it to the top-level import block would improve readability. `[fixable]`


### `server/__tests__/logger.test.ts`
_Solid migration from console to Pino with good test coverage for the server logger, but the harness logger has no tests, and the production transport paths (pino-roll, LOGGER_SYNC) are untested._

- 🔵 **missing_tests**: No tests cover the `buildProductionDestination` path (pino-roll multi-transport), the `LOGGER_SYNC` env var branch, or the `isPinoPrettyAvailable` fallback. These are the most complex new code paths in the server logger. The existing tests all bypass them by passing a custom `dest`. `[fixable]`


### `server/tracing.ts`
_Solid migration from console to Pino with good test coverage for the server logger, but the harness logger has no tests, and the production transport paths (pino-roll, LOGGER_SYNC) are untested._

- 🔵 **regressions** (L10): Switching from `SimpleSpanProcessor` to `BatchSpanProcessor` is a good production change (reduces overhead), but note that `BatchSpanProcessor` buffers spans and flushes asynchronously. The existing `shutdown` handler calls `provider.shutdown()` which should flush, but very short-lived processes may lose the final batch. Low risk for a long-running server, but worth noting.

PR #234, opened 2026-04-18
fix(client): guard stopGeneration and respondToPermission against null sessionId

dimakis:
## Summary

- `stopGeneration` sent `sessionId: null` which `V2StopMessage` rejects (requires non-empty string). Guard with early return — can't stop what we can't find.
- `respondToPermission` had the same issue but server resolves by `permId` not `sessionId`. Made `sessionId` optional in `V2PermissionResponseMessage` schema so first-turn permission responses work.

Found by Centaur review on #230.

## Test plan

- [x] Updated test: stopGeneration before session_id is a no-op (not null send)
- [x] Updated test: respondToPermission omits sessionId field instead of sending null
- [x] All 339 schema/handler tests pass
- [x] All 118 store tests pass

### DIFF
--- a/packages/client/__tests__/store.test.ts
+++ b/packages/client/__tests__/store.test.ts
@@ -477,7 +477,7 @@ describe('stopGeneration', () => {
     expect(stop!.sessionId).toBe('test-session');
   });
 
-  it('sends stop with null sessionId during first turn before session_id arrives', () => {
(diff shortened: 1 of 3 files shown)


dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #235, opened 2026-04-18
feat: deterministic session isolation — worktree guard, POST /api/sessions, system prompt fixes

dimakis:
## Summary

Mitzo-side foundation for deterministic session isolation. Every session gets isolated worktrees for all configured repos. Mitzo enforces isolation via a tool-level path guard.

- **POST /api/sessions** — creates worktrees for primary + all `.mitzo.json` repos, returns manifest
- **Worktree path guard** in `canUseTool` — runs before `shouldAutoAllow`, inspects Write/Edit/Bash tool inputs, denies out-of-worktree paths with a redirect message the agent self-corrects from. Read operations unrestricted.
- **`buildWorktreeSystemPrompt`** — now lists ALL repos including primary (critical for the "cd back to mgmt" navigation problem)
- **Env loop fix** — `MITZO_REPO_PRIMARY` now set alongside secondary repo env vars
- **`createWorktree` refactor** — optional dir/branch/prefix overrides, backward-compatible
- **`createSessionWorktrees`** — bulk helper for endpoint and chat.ts
- **Cleanup unification** — scans `.cursor/worktrees/` alongside `.claude/worktrees/`
- **Internal token persisted** to `~/.mitzo/internal-token` so session hooks can authenticate

## Test plan

- [x] 13 worktree guard tests (path matching, Bash parsing, read pass-through, suggestions)
- [x] 467 harness tests pass
- [x] Full build compiles clean
- [x] All real test failures are from stale worktree duplicates (pre-existing, not this PR)

### DIFF
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -1,0 +1,98 @@
+import type { ManagedSession } from './session-registry.js';
+
+const WRITE_TOOLS = new Set(['Write', 'Edit', 'StrReplace', 'EditNotebook', 'MultiEdit']);
+const SHELL_TOOLS = new Set(['Bash', 'Shell']);
+const PATH_FIELDS = ['file_path', 'path', 'target_notebook'];
+
+/**
+ * Extract absolute paths from a shell command string. Heuristic — catches
+ * explicit absolute paths but can't catch every indirect construction.
+ */
+function extractAbsolutePaths(command: string): string[] {
+  const matches = command.match(/(?:^|\s|=|")(\/[\w/.@-]+)/g);
+  if (!matches) return [];
+  return matches.map((m) => m.trim().replace(/^["=]/, ''));
+}
+
+function findAllowedWorktree(
+  absolutePath: string,
+  worktreePaths: Map<string, { path: string; wtId: string }>,
+): { repoName: string; worktreePath: string } | null {
+  for (const [name, { path }] of worktreePaths) {
+    if (absolutePath.startsWith(path + '/') || absolutePath === path) {
+      return { repoName: name, worktreePath: path };
+    }
+  }
+  return null;
+}
+
+/**
+ * Find the worktree path that should be used for a given out-of-bounds path.
+ * Maps /repo-root/some/file to /repo-root/.claude/worktrees/ID/some/file.
+ */
+function suggestWorktreePath(
+  absolutePath: string,
+  worktreePaths: Map<string, { path: string; wtId: string }>,
+): string | null {
+  for (const [, { path: wtPath }] of worktreePaths) {
+    // worktree path format: /repo-root/.claude/worktrees/<id>
+    // extract repo root: everything before /.claude/worktrees/ or /.cursor/worktrees/
+    const worktreeMarker = wtPath.match(/^(.+?)\/(\.claude|\.cursor)\/worktrees\/.+$/);
+    if (!worktreeMarker) continue;
+
+    const repoRoot = worktreeMarker[1];
+    if (absolutePath.startsWith(repoRoot + '/') || absolutePath === repoRoot) {
+      const relativePath = absolutePath.slice(repoRoot.length);
+      return wtPath + relativePath;
+    }
+  }
+  return null;
+}
+
+/**
+ * Check if a tool invocation violates worktree isolation.
+ * Returns null if allowed, or a deny message with the correct path.
+ */
+export function checkWorktreePolicy(
+  session: ManagedSession,
+  toolName: string,
+  toolInput: Record<string, unknown>,
+): string | null {
+  if (session.worktreePaths.size === 0) return null;
+
+  if (WRITE_TOOLS.has(toolName)) {
+    for (const field of PATH_FIELDS) {
+      const filePath = toolInput[field];
+      if (typeof filePath !== 'string' || !filePath.startsWith('/')) continue;
+
+      const allowed = findAllowedWorktree(filePath, session.worktreePaths);
+      if (allowed) continue;
+
+      const suggestion = suggestWorktreePath(filePath, session.worktreePaths);
+      if (suggestion) {
+        return `Path ${filePath} is outside session worktrees. ` + `Use ${suggestion} instead.`;
+      }
+      return `Path ${filePath} is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+    }
+  }
+
+  if (SHELL_TOOLS.has(toolName)) {
+    const command = toolInput.command;
+    if (typeof command !== 'string') return null;
+
+    const paths = extractAbsolutePaths(command);
+    for (const p of paths) {
+      if (findAllowedWorktree(p, session.worktreePaths)) continue;
+
+      const suggestion = suggestWorktreePath(p, session.worktreePaths);
+      if (suggestion) {
+        return (
+          `Shell command references ${p} which is outside session worktrees. ` +
+          `Use ${suggestion} instead.`
+        );
+      }
+    }
+  }
+
+  return null;
+}
--- a/server/app.ts
+++ b/server/app.ts
@@ -22,8 +22,13 @@ import {
   getRepoConfig,
   eventStore,
   registry,
+  generateWtId,
 } from './chat.js';
-import { createWorktree, listWorktrees } from './worktree.js';
+import {
+  createWorktree,
+  createSessionWorktrees as createAllWorktrees,
+  listWorktrees,
+} from './worktree.js';
 import { GIT_BRANCH_TIMEOUT_MS } from './constants.js';
 import { INTERNAL_TOKEN } from './internal-token.js';
 import { getLocalCommit, isUpdateAvailable } from './git-version.js';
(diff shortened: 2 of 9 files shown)


packages/harness/src/worktree-guard.ts
```diff
@@ -0,0 +1,98 @@
...
+      }
+      return `Path ${filePath} is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+    }
+  }
+
+  if (SHELL_TOOLS.has(toolName)) {
+    const command = toolInput.command;
+    if (typeof command !== 'string') return null;
+
+    const paths = extractAbsolutePaths(command);
```
dimakis, line 83: 🟡 **bugs**: Shell guard silently allows paths outside ALL known repos. When `suggestWorktreePath` returns null (path not under any configured repo root), the loop continues without denying. Compare to the Write tool handling (line 75) which has a fallback deny: `return 'Path ... is outside session worktrees. Check $MITZO_REPO_* env vars...'`. A Bash command like `echo secret > /unknown/repo/file` passes the guard, while a Write to the same path is blocked. If the asymmetry is intentional (shell needs system paths like /usr/bin), add a comment explaining the design choice. `[fixable]`

packages/harness/src/worktree-guard.ts
```diff
@@ -0,0 +1,98 @@
...
+const WRITE_TOOLS = new Set(['Write', 'Edit', 'StrReplace', 'EditNotebook', 'MultiEdit']);
+const SHELL_TOOLS = new Set(['Bash', 'Shell']);
+const PATH_FIELDS = ['file_path', 'path', 'target_notebook'];
+
+/**
+ * Extract absolute paths from a shell command string. Heuristic — catches
+ * explicit absolute paths but can't catch every indirect construction.
+ */
+function extractAbsolutePaths(command: string): string[] {
+  const matches = command.match(/(?:^|\s|=|")(\/[\w/.@-]+)/g);
```
dimakis, line 12: 🟡 **unsafe_assumptions**: The `extractAbsolutePaths` regex `(?:^|\s|=|")(\/)` does not match paths in single quotes (e.g., `cat '/repo/root/secret'`), paths after no-space redirections (`>/repo/file`), or paths containing `+`, `~`, `#`. Since Claude regularly emits single-quoted paths in Bash commands, this is a realistic bypass vector for the worktree guard. Consider adding `'` to the prefix group and expanding the character class to include at least `+~`. `[fixable]`

server/worktree.ts
```diff
@@ -47,6 +57,39 @@ export function createWorktree(sessionId: string, baseRepo: string): string {
   return worktreePath;
 }
 
+/**
```
dimakis, line 60: 🟡 **missing_tests**: The new `createSessionWorktrees` function has no unit tests. Key behaviors to cover: primary + secondary repo creation, shared branch naming, error resilience when a secondary repo fails (should log warning but not throw), and return value shape. The existing `worktree.test.ts` only tests `cleanupStaleWorktrees` and the base `createWorktree`. `[fixable]`

server/chat.ts
```diff
@@ -270,26 +270,30 @@ function buildTaskPromptForSession(clientId: string): string {
...
  * Build system prompt section listing all session worktrees.
- * Tells the agent exactly where each repo's worktree lives.
+ * Lists ALL repos including primary so the agent has a complete lookup table
+ * for navigating between repos (the "cd back" problem).
  */
 export function buildWorktreeSystemPrompt(
   repoWorktrees: Map<string, { path: string; wtId: string }>,
 ): string {
-  if (repoWorktrees.size <= 1) return ''; // Only primary or none
+  if (repoWorktrees.size === 0) return '';
```
dimakis, line 279: 🔵 **missing_tests**: `buildWorktreeSystemPrompt` changed from `size <= 1` guard (skip for single repo) to `size === 0` (show for any worktree), and now includes primary in the listing. No tests exist for this function. The behavioral change — single-repo sessions now get the worktree system prompt — deserves at least a basic assertion. `[fixable]`

server/app.ts
```diff
@@ -331,6 +336,40 @@ app.post('/api/repos/open', (req, res) => {
...
 });
 
+app.post('/api/sessions', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+
+  const { source } = req.body || {};
+  if (!source || typeof source !== 'string') {
```
dimakis, line 370: 🔵 **unsafe_assumptions**: `source` is validated as a non-empty string but not checked against an allowlist. The error message says `cursor | claude | mitzo` but any string is accepted — values besides `cursor` silently map to `.claude` prefix. Consider validating with an explicit set or Zod enum to catch misuse early. `[fixable]`

server/internal-token.ts
```diff
@@ -1,4 +1,16 @@
...
 
 /** Shared token for MCP server → Mitzo API calls (generated once at startup). */
 export const INTERNAL_TOKEN = randomBytes(32).toString('hex');
+
+// Persist to ~/.mitzo/internal-token so session hooks can authenticate
+try {
+  const dir = join(homedir(), '.mitzo');
+  mkdirSync(dir, { recursive: true, mode: 0o700 });
+  writeFileSync(join(dir, 'internal-token'), INTERNAL_TOKEN, { mode: 0o600 });
+} catch {
```
dimakis, line 14: 🔵 **style**: Empty `catch {}` block silently swallows file-write errors. If the token file can't be written (e.g., permissions, disk full), all external hooks will silently fail to authenticate with no diagnostic trail. Consider logging a warning via the project's logger rather than swallowing completely. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🟡 **bugs** (L83): Shell guard silently allows paths outside ALL known repos. When `suggestWorktreePath` returns null (path not under any configured repo root), the loop continues without denying. Compare to the Write tool handling (line 75) which has a fallback deny: `return 'Path ... is outside session worktrees. Check $MITZO_REPO_* env vars...'`. A Bash command like `echo secret > /unknown/repo/file` passes the guard, while a Write to the same path is blocked. If the asymmetry is intentional (shell needs system paths like /usr/bin), add a comment explaining the design choice. `[fixable]`
- 🟡 **unsafe_assumptions** (L12): The `extractAbsolutePaths` regex `(?:^|\s|=|")(\/)` does not match paths in single quotes (e.g., `cat '/repo/root/secret'`), paths after no-space redirections (`>/repo/file`), or paths containing `+`, `~`, `#`. Since Claude regularly emits single-quoted paths in Bash commands, this is a realistic bypass vector for the worktree guard. Consider adding `'` to the prefix group and expanding the character class to include at least `+~`. `[fixable]`


### `packages/harness/__tests__/worktree-guard.test.ts`
_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🟡 **missing_tests**: Shell/Bash test cases are thin — no coverage for: redirection operators (`cmd > /path`), single-quoted paths (`cat '/path'`), paths with special characters (`+`, `~`), or the case where a shell command references a path outside ALL known repos (the silent-allow behavior at line 94). These are the boundary conditions where the heuristic regex could fail. `[fixable]`


### `server/worktree.ts`
_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🟡 **missing_tests** (L60): The new `createSessionWorktrees` function has no unit tests. Key behaviors to cover: primary + secondary repo creation, shared branch naming, error resilience when a secondary repo fails (should log warning but not throw), and return value shape. The existing `worktree.test.ts` only tests `cleanupStaleWorktrees` and the base `createWorktree`. `[fixable]`


### `server/chat.ts`
_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🔵 **missing_tests** (L279): `buildWorktreeSystemPrompt` changed from `size <= 1` guard (skip for single repo) to `size === 0` (show for any worktree), and now includes primary in the listing. No tests exist for this function. The behavioral change — single-repo sessions now get the worktree system prompt — deserves at least a basic assertion. `[fixable]`


### `server/app.ts`
_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🔵 **unsafe_assumptions** (L346): `source` is validated as a non-empty string but not checked against an allowlist. The error message says `cursor | claude | mitzo` but any string is accepted — values besides `cursor` silently map to `.claude` prefix. Consider validating with an explicit set or Zod enum to catch misuse early. `[fixable]`


### `server/internal-token.ts`
_Solid worktree guard design with good test coverage for the happy path, but the shell-command path extraction regex has realistic bypass vectors (single-quoted paths) and the shell guard silently allows paths outside all known repos — an asymmetry with the stricter write-tool handling._

- 🔵 **style** (L14): Empty `catch {}` block silently swallows file-write errors. If the token file can't be written (e.g., permissions, disk full), all external hooks will silently fail to authenticate with no diagnostic trail. Consider logging a warning via the project's logger rather than swallowing completely. `[fixable]`

PR #236, opened 2026-04-18
feat: Capacitor iOS app shell with transport abstraction

dimakis:
## Summary

- Add `apiFetch()` wrapper with configurable `VITE_API_BASE_URL` — centralises all 23+ API fetch calls behind a single function that supports both browser same-origin and Capacitor cross-origin
- Server-side: Bearer token auth fallback, CORS middleware (`CORS_ALLOWED_ORIGINS` env var), WebSocket query-param auth, JWT in login response body
- Capacitor iOS project initialized with lifecycle hooks wiring `appStateChange` to `MitzoConnection.checkAndReconnect()`
- `@mitzo/client`: expose `checkAndReconnect()` as public for native lifecycle integration

## Motivation

iOS Safari silently kills WebSocket connections when backgrounded. Despite extensive v2 protocol mitigations (snapshots, heartbeat, visibility listeners), the browser remains the bottleneck. Capacitor wraps the existing React frontend in a native iOS shell — reliable WS lifecycle, no gesture gates for audio, native push later.

The transport abstraction (`apiFetch`, `getWsBaseUrl`, `getWsChatUrl`) also benefits Mitzo-Theia by decoupling server URL construction from `window.location`.

## Test plan

- [ ] `npm test` — 343 tests pass (lib + client package)
- [ ] `npm run dev` — browser version works unchanged (VITE_API_BASE_URL defaults to empty)
- [ ] Set `CORS_ALLOWED_ORIGINS=capacitor://localhost` in server `.env`
- [ ] `npm run build:ios` in `frontend/` — builds and syncs to Xcode project
- [ ] `npm run open:ios` — opens Xcode, build to simulator
- [ ] Login flow: passphrase -> JWT stored in localStorage
- [ ] Chat streaming works over Tailscale
- [ ] Background app 30s -> return -> WS reconnects


server/app.ts
```diff
@@ -587,7 +607,7 @@ app.post('/api/auth/login', loginLimiter, async (req, res) => {
     sameSite: 'strict',
     maxAge: MAX_AGE_HOURS * 60 * 60 * 1000,
   });
-  res.json({ ok: true });
+  res.json({ ok: true, token });
```
dimakis, line 612: 🔴 **regressions**: Login response changed from `{ ok: true }` to `{ ok: true, token }`. The existing test at `server/__tests__/routes.test.ts:167` uses `expect(res.body).toEqual({ ok: true })` which performs deep equality — this will fail because the response now includes the extra `token` field. `[fixable]`

dimakis (commented):
## Centaur Review

Found **10** issue(s) (1 critical) (6 warning).


### `server/app.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🔴 **regressions** (L610): Login response changed from `{ ok: true }` to `{ ok: true, token }`. The existing test at `server/__tests__/routes.test.ts:167` uses `expect(res.body).toEqual({ ok: true })` which performs deep equality — this will fail because the response now includes the extra `token` field. `[fixable]`
- 🔵 **missing_tests** (L140): The CORS middleware has no integration tests. Should verify: (1) allowed origins get correct `Access-Control-*` headers, (2) disallowed origins don't get CORS headers, (3) preflight `OPTIONS` returns 204. `[fixable]`


### `frontend/src/lib/api-fetch.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **bugs** (L23): Dead code in `getWsBaseUrl()`: the token is retrieved from localStorage but never used. The ternary `return token ? \`${wsBase}\` : wsBase` returns the identical string in both branches. This looks like an incomplete refactor — the token should either be appended as a query param (as done in `getWsChatUrl()`) or the retrieval should be removed. `[fixable]`


### `frontend/src/lib/capacitor.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **bugs** (L22): `registerCapacitorLifecycle()` is exported but never imported or called anywhere in the app. The iOS app will not reconnect WebSockets when returning from background — the core feature this file implements is dead code. It needs to be wired up in `main.tsx` or `client-store.ts` after store creation. `[fixable]`


### `frontend/package.json`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **unsafe_assumptions** (L9): The `build:ios` script hardcodes a Tailscale IP (`https://100.91.50.57:3100`). This is machine-specific and violates the project convention (CLAUDE.md: "No hardcoded machine-specific paths in source code"). Should be read from an env var or `.env` file, e.g. `VITE_API_BASE_URL=$MITZO_SERVER_URL`. `[fixable]`
- 🔵 **style**: `@capacitor/status-bar` is listed as a dependency but never imported anywhere in the frontend source code. Remove it to avoid unnecessary bundle/dependency bloat. `[fixable]`


### `frontend/capacitor.config.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **unsafe_assumptions** (L12): `allowNavigation: ['*.ts.net', '100.*']` — the `100.*` pattern matches the entire 100.0.0.0/8 range (~16M IPs). While Tailscale uses 100.64-127.x.x, this allows navigation to any 100.x.x.x host. Consider narrowing to Tailscale's CGNAT range (e.g. `100.64.*` through `100.127.*`) or documenting the security trade-off. `[fixable]`


### `server/auth.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **missing_tests** (L54): The new Bearer token authentication path (`extractBearerToken` + fallback in `authMiddleware`) has no test coverage. Need tests verifying: (1) requests with valid `Authorization: Bearer <token>` header succeed, (2) requests with invalid/expired Bearer tokens are rejected, (3) Bearer auth works as fallback when no cookie is present. `[fixable]`


### `server/index.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🟡 **missing_tests** (L176): WebSocket query-param token auth (`?token=` in upgrade handler) has no test coverage. This is a new auth vector that should be tested for both valid and invalid tokens. `[fixable]`


### `frontend/src/lib/constants.ts`
_Transport abstraction is well-structured, but the login response change will break an existing test, `registerCapacitorLifecycle` is dead code (never called), `getWsBaseUrl` has a no-op token branch, and the `build:ios` script hardcodes a machine-specific IP._

- 🔵 **style** (L24): Top-level `import` statement at line 24 is placed between constant declarations rather than at the top of the file. Move it to the top with the other imports for consistency. `[fixable]`

server/app.ts
```diff
@@ -137,6 +137,26 @@ export const yapperWsProxy = createProxyMiddleware({
 app.use('/api/yapper', yapperHttpProxy);
 app.use('/api/yapper-ws', yapperWsProxy);
 
+// --- CORS for non-same-origin clients (Capacitor iOS, etc.) ---
```
dimakis, line 120: 🔵 **missing_tests**: The CORS middleware has no integration tests. Should verify: (1) allowed origins get correct `Access-Control-*` headers, (2) disallowed origins don't get CORS headers, (3) preflight `OPTIONS` returns 204. `[fixable]`

frontend/src/lib/api-fetch.ts
```diff
@@ -0,0 +1,44 @@
...
+  return import.meta.env.VITE_API_BASE_URL || '';
+}
+
+export function getWsBaseUrl(): string {
+  const base = getApiBaseUrl();
+  if (base) {
+    const wsBase = base.replace(/^http/, 'ws');
+    const token =
+      typeof localStorage !== 'undefined' ? localStorage.getItem(AUTH_TOKEN_KEY) : null;
+    return token ? `${wsBase}` : wsBase;
```
dimakis: 🟡 **bugs**: Dead code in `getWsBaseUrl()`: the token is retrieved from localStorage but never used. The ternary `return token ? \`${wsBase}\` : wsBase` returns the identical string in both branches. This looks like an incomplete refactor — the token should either be appended as a query param (as done in `getWsChatUrl()`) or the retrieval should be removed. `[fixable]`

frontend/src/lib/capacitor.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export function isCapacitor(): boolean {
+  return Capacitor.isNativePlatform();
+}
+
+/**
+ * Wire Capacitor app lifecycle events to MitzoConnection reconnect.
+ * Call once after creating the store. No-op in browser environments.
+ */
+export function registerCapacitorLifecycle(connection: MitzoConnection): void {
```
dimakis: 🟡 **bugs**: `registerCapacitorLifecycle()` is exported but never imported or called anywhere in the app. The iOS app will not reconnect WebSockets when returning from background — the core feature this file implements is dead code. It needs to be wired up in `main.tsx` or `client-store.ts` after store creation. `[fixable]`

frontend/package.json
```diff
@@ -5,9 +5,16 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "build:ios": "VITE_API_BASE_URL=https://100.91.50.57:3100 tsc -b && VITE_API_BASE_URL=https://100.91.50.57:3100 vite build && npx cap sync ios",
```
dimakis: 🟡 **unsafe_assumptions**: The `build:ios` script hardcodes a Tailscale IP (`https://100.91.50.57:3100`). This is machine-specific and violates the project convention (CLAUDE.md: "No hardcoded machine-specific paths in source code"). Should be read from an env var or `.env` file, e.g. `VITE_API_BASE_URL=$MITZO_SERVER_URL`. `[fixable]`

frontend/capacitor.config.ts
```diff
@@ -0,0 +1,16 @@
...
+const config: CapacitorConfig = {
+  appId: 'com.mitzo.app',
+  appName: 'Mitzo',
+  webDir: 'dist',
+  ios: {
+    contentInset: 'automatic',
+  },
+  server: {
+    // Allow mixed content for development (Tailscale HTTP)
+    allowNavigation: ['*.ts.net', '100.*'],
```
dimakis: 🟡 **unsafe_assumptions**: `allowNavigation: ['*.ts.net', '100.*']` — the `100.*` pattern matches the entire 100.0.0.0/8 range (~16M IPs). While Tailscale uses 100.64-127.x.x, this allows navigation to any 100.x.x.x host. Consider narrowing to Tailscale's CGNAT range (e.g. `100.64.*` through `100.127.*`) or documenting the security trade-off. `[fixable]`

server/auth.ts
```diff
@@ -51,10 +51,16 @@ export async function verifyToken(token: string): Promise<boolean> {
   }
 }
 
+function extractBearerToken(req: Request): string | undefined {
```
dimakis, line 54: 🟡 **missing_tests**: The new Bearer token authentication path (`extractBearerToken` + fallback in `authMiddleware`) has no test coverage. Need tests verifying: (1) requests with valid `Authorization: Bearer <token>` header succeed, (2) requests with invalid/expired Bearer tokens are rejected, (3) Bearer auth works as fallback when no cookie is present. `[fixable]`

server/index.ts
```diff
@@ -173,7 +173,9 @@ server.on('upgrade', async (req, socket, head) => {
     return;
   }
 
-  const authed = await verifyWsAuth(req.headers.cookie);
+  const authed =
```
dimakis, line 176: 🟡 **missing_tests**: WebSocket query-param token auth (`?token=` in upgrade handler) has no test coverage. This is a new auth vector that should be tested for both valid and invalid tokens. `[fixable]`

frontend/src/lib/constants.ts
```diff
@@ -21,9 +21,10 @@ export const TOOL_GROUP_THRESHOLD = 3;
 export const MAX_IMAGE_ATTACHMENTS = 4;
 
 // --- Voice / Yapper ---
+import { getApiBaseUrl } from './api-fetch';
```
dimakis: 🔵 **style**: Top-level `import` statement at line 24 is placed between constant declarations rather than at the top of the file. Move it to the top with the other imports for consistency. `[fixable]`

frontend/src/lib/capacitor.ts
```diff
@@ -0,0 +1,32 @@
...
+ * existing browser listeners trigger reconnection.
+ *
+ * Call once at app startup. No-op in browser environments.
+ */
+export function registerCapacitorLifecycle(): void {
+  if (!isCapacitor()) return;
+
+  App.addListener('appStateChange', ({ isActive }) => {
+    if (isActive && typeof document !== 'undefined') {
+      document.dispatchEvent(new Event('visibilitychange'));
```
dimakis: 🔴 **bugs**: Synthetic `visibilitychange` event won't trigger reconnection. `MitzoConnection.addBrowserListeners()` (connection.ts:222) checks `document.visibilityState === 'visible'` before calling `checkAndReconnect()`. Dispatching a synthetic Event does NOT change the read-only `document.visibilityState` property — in a Capacitor app returning from background, it may still be `'hidden'`, so the guard fails silently. The method was made public (connection.ts:246) precisely for this case — call `clientStore.connection.checkAndReconnect()` directly instead of dispatching a synthetic DOM event. `[fixable]`

server/app.ts
```diff
@@ -137,6 +137,26 @@ export const yapperWsProxy = createProxyMiddleware({
...
+const CORS_ALLOWED_ORIGINS = process.env.CORS_ALLOWED_ORIGINS?.split(',').map((s) => s.trim()) ?? [];
+
+if (CORS_ALLOWED_ORIGINS.length > 0) {
```
dimakis, line 125: 🟡 **bugs**: CORS middleware is mounted AFTER the yapper proxy (lines 137-138). Requests to `/api/yapper` from Capacitor (origin `capacitor://localhost`) are handled by the proxy before CORS headers are set, so cross-origin voice requests will fail. Move the CORS middleware before the proxy, or apply CORS headers inside the proxy middleware. `[fixable]`

frontend/src/lib/constants.ts
```diff
@@ -22,8 +24,7 @@ export const MAX_IMAGE_ATTACHMENTS = 4;
 
 // --- Voice / Yapper ---
 export const YAPPER_URL =
-  import.meta.env.VITE_YAPPER_URL ||
-  `${typeof window !== 'undefined' ? `${window.location.protocol}//${window.location.host}` : 'http://localhost'}/api/yapper`;
+  import.meta.env.VITE_YAPPER_URL || `${getApiBaseUrl()}/api/yapper`;
```
dimakis: 🟡 **unsafe_assumptions**: When `VITE_API_BASE_URL` is unset, `YAPPER_URL` becomes the relative path `/api/yapper`. The old code built an absolute URL from `window.location`. If `YAPPER_URL` is later used in a `WebSocket` constructor (which requires an absolute URL), this will throw. Ensure all consumers use `apiFetch()` or prepend the base URL. `[fixable]`

frontend/package.json
```diff
@@ -5,9 +5,15 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "build:ios": "tsc -b && vite build && npx cap sync ios",
+    "open:ios": "npx cap open ios"
```
dimakis, line 10: 🔵 **style**: `@capacitor/cli` and `@capacitor/ios` are build/CLI tools and should be in `devDependencies`, not `dependencies`. Only `@capacitor/core` and `@capacitor/app` are runtime dependencies. `[fixable]`

server/index.ts
dimakis, line 176: 🟡 **unsafe_assumptions**: WebSocket auth token in query param (`?token=<jwt>`) can be logged in server access logs, reverse proxy logs, and browser history. This is a common WS auth trade-off but worth documenting. Consider adding a log-scrubbing note or rotating these tokens with a shorter TTL than cookie JWTs.

server/__tests__/cors-ws-auth.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(res.status).toBe(204);
+    expect(res.headers['access-control-allow-origin']).toBeUndefined();
+  });
+});
+
+describe('WS query-param auth (verifyToken)', () => {
+  it('verifyToken accepts valid JWT', async () => {
+    const { login, verifyToken } = await import('../auth.js');
+    const token = await login(process.env.AUTH_PASSPHRASE!);
+    expect(token).toBeTruthy();
```
dimakis, line 112: 🟡 **regressions**: Test name says 'does not return 204 for OPTIONS from disallowed origin' but the assertion on line 124 checks `expect(res.status).toBe(204)` — the test passes because OPTIONS always returns 204. The test name is misleading and contradicts the assertion. Rename to 'returns 204 for OPTIONS from disallowed origin but without CORS headers'. `[fixable]`

dimakis (commented):
## Centaur Review

Found **7** issue(s) (1 critical) (5 warning).


### `frontend/src/lib/capacitor.ts`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🔴 **bugs** (L29): Synthetic `visibilitychange` event won't trigger reconnection. `MitzoConnection.addBrowserListeners()` (connection.ts:222) checks `document.visibilityState === 'visible'` before calling `checkAndReconnect()`. Dispatching a synthetic Event does NOT change the read-only `document.visibilityState` property — in a Capacitor app returning from background, it may still be `'hidden'`, so the guard fails silently. The method was made public (connection.ts:246) precisely for this case — call `clientStore.connection.checkAndReconnect()` directly instead of dispatching a synthetic DOM event. `[fixable]`
- 🟡 **missing_tests**: No tests for `registerCapacitorLifecycle()` or `isCapacitor()`. The project follows TDD — this module has testable logic (native platform detection, event dispatch) that should be covered, especially given the visibilitychange bug. `[fixable]`


### `server/app.ts`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🟡 **bugs** (L143): CORS middleware is mounted AFTER the yapper proxy (lines 137-138). Requests to `/api/yapper` from Capacitor (origin `capacitor://localhost`) are handled by the proxy before CORS headers are set, so cross-origin voice requests will fail. Move the CORS middleware before the proxy, or apply CORS headers inside the proxy middleware. `[fixable]`


### `frontend/src/lib/constants.ts`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🟡 **unsafe_assumptions** (L27): When `VITE_API_BASE_URL` is unset, `YAPPER_URL` becomes the relative path `/api/yapper`. The old code built an absolute URL from `window.location`. If `YAPPER_URL` is later used in a `WebSocket` constructor (which requires an absolute URL), this will throw. Ensure all consumers use `apiFetch()` or prepend the base URL. `[fixable]`


### `frontend/package.json`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🔵 **style** (L10): `@capacitor/cli` and `@capacitor/ios` are build/CLI tools and should be in `devDependencies`, not `dependencies`. Only `@capacitor/core` and `@capacitor/app` are runtime dependencies. `[fixable]`


### `server/index.ts`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🟡 **unsafe_assumptions** (L176): WebSocket auth token in query param (`?token=<jwt>`) can be logged in server access logs, reverse proxy logs, and browser history. This is a common WS auth trade-off but worth documenting. Consider adding a log-scrubbing note or rotating these tokens with a shorter TTL than cookie JWTs.


### `server/__tests__/cors-ws-auth.test.ts`
_The Capacitor reconnection logic has a critical bug — synthetic visibilitychange events won't trigger MitzoConnection's reconnect because document.visibilityState remains unchanged. CORS middleware ordering also blocks voice API calls from native clients._

- 🟡 **regressions** (L121): Test name says 'does not return 204 for OPTIONS from disallowed origin' but the assertion on line 124 checks `expect(res.status).toBe(204)` — the test passes because OPTIONS always returns 204. The test name is misleading and contradicts the assertion. Rename to 'returns 204 for OPTIONS from disallowed origin but without CORS headers'. `[fixable]`

PR #237, opened 2026-04-18
fix(server): per-session clientId in v2 to fix active dots

dimakis:
## Summary

- `handleSendV2` was passing the shared `connectionId` as `clientId` to `startChat`, causing `registry.register()` to overwrite the previous session when the user switched between sessions on the same v2 connection
- `getActiveSessions()` returned at most one session per connection, so the home screen active dots were broken (no dots showed)
- Generate unique per-session keys (`connectionId:sessionId` for resume, `connectionId:new-<ts>` for create) so each session gets its own registry entry
- All downstream lookups use `findBySessionId()` which searches by sessionId, not clientId, so they remain correct

## Test plan

- [x] All 121 ws-handler-v2 tests pass
- [x] All 1700+ relevant server tests pass
- [ ] Deploy and verify active dots appear on mobile home screen for multiple concurrent sessions
- [ ] Verify dots show green (attached) during active query and orange (detached) after WS reconnect

### DIFF
--- a/server/ws-handler-v2.ts
+++ b/server/ws-handler-v2.ts
@@ -347,11 +347,14 @@ export function handleSendV2(
         return;
       }
 
-      // Session exists in store but no active driver — start with resume
+      // Session exists in store but no active driver — start with resume.
+      // Use a per-session clientId so multiple sessions from the same v2
+      // connection each get their own registry entry (prevents overwrites).
+      const sessionClientId = `${connectionId}:${sessionId}`;
       ctx.connRegistry.watch(connectionId, sessionId);
       ctx.connRegistry.setActive(connectionId, sessionId);
       span.setAttribute('routing.decision', 'resume');
-      startChat(transport, connectionId, prompt, {
+      startChat(transport, sessionClientId, prompt, {
         resume: sessionId,
         cwd: msg.cwd,
         model: msg.model,
@@ -361,17 +364,17 @@ export function handleSendV2(
         contextBlocks: msg.contextBlocks,
         clientMsgId: msg.clientMsgId,
       });
-      // startChat registers under connectionId — safe to apply now
-      applySkillPolicy(connectionId);
+      applySkillPolicy(sessionClientId);
     } else {
-      // null sessionId — new session. Supply onSessionResolved so the
-      // connection auto-watches once the SDK resolves the session ID.
+      // null sessionId — new session. Use a unique clientId so concurrent
+      // new-session starts don't collide in the registry.
+      const newClientId = `${connectionId}:new-${Date.now()}`;
       span.setAttribute('routing.decision', 'create');
       const onSessionResolved = (resolvedId: string) => {
         ctx.connRegistry.watch(connectionId, resolvedId);
         ctx.connRegistry.setActive(connectionId, resolvedId);
       };
-      startChat(transport, connectionId, prompt, {
+      startChat(transport, newClientId, prompt, {
         cwd: msg.cwd,
         model: msg.model,
         extraTools: msg.extraTools,
@@ -381,8 +384,7 @@ export function handleSendV2(
         clientMsgId: msg.clientMsgId,
         onSessionResolved,
       });
-      // startChat registers under connectionId — safe to apply now
-      applySkillPolicy(connectionId);
+      applySkillPolicy(newClientId);
     }
 
     span.setStatus({ code: SpanStatusCode.OK });
--- a/server/ws-handler-v2.ts
+++ b/server/ws-handler-v2.ts
@@ -32,6 +32,7 @@ type StopMsg = z.infer<typeof V2StopMessage>;
 type InterruptMsg = z.infer<typeof V2InterruptMessage>;
 type PermissionMsg = z.infer<typeof V2PermissionResponseMessage>;
 type SetModeMsg = z.infer<typeof V2SetModeMessage>;
+import { randomUUID } from 'crypto';
 import { tracer } from './tracing.js';
 import { SpanStatusCode } from '@opentelemetry/api';
 import { resolvePending } from './permissions.js';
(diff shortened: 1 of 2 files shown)


dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_The core approach is sound — per-session clientIds fix the registry overwrite problem. Two concerns: Date.now()-based keys have millisecond collision risk, and the tests were not updated to assert the new composite key format._

- 🟡 **unsafe_assumptions** (L345): `Date.now()` is not collision-safe for the new-session clientId. If two new-session requests arrive in the same millisecond on the same connection, they'll get the same `newClientId`. Consider using a counter or crypto.randomUUID() instead. `[fixable]`
- 🔴 **regressions** (L339): On the resume path, `connRegistry.watch()` and `connRegistry.setActive()` are called with the raw `connectionId` (not the composite `sessionClientId`). But `startChat` and `applySkillPolicy` now use the composite key. This is actually correct — the connRegistry tracks connections, not sessions. However, when the connection disconnects (index.ts:280), `findBySessionId` returns the composite `clientId`, which is then used for `registry.get()` and `detachChat()`. This works because the SessionRegistry maps sessionId→clientId correctly. No regression here on closer inspection.
- 🔵 **style** (L345): The variable name `newClientId` is inconsistent with `sessionClientId` used in the resume path. Both serve the same purpose (a per-session-scoped clientId). Consider a unified naming convention, e.g., both could be `sessionClientId` with different construction logic. `[fixable]`


_The core approach is sound — per-session clientIds fix the registry overwrite problem. Two concerns: Date.now()-based keys have millisecond collision risk, and the tests were not updated to assert the new composite key format._

- 🟡 **missing_tests**: The PR changes only the implementation file but not the tests. The existing 'auto-watch' test for the resume path (line 396) verifies `startChat` is called with connectionId='c1', but the PR now passes `c1:sess-resume`. The test should assert the new composite `clientId` format. Similarly, the create-path test (line 429) should verify the `newClientId` format, and the skill-policy tests should verify `setSkillPolicy` is called with the composite key on the resume and create paths (only the active-resume path is currently tested). `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -335,17 +338,17 @@ export function handleSendV2(
...
       });
-      // startChat registers under connectionId — safe to apply now
-      applySkillPolicy(connectionId);
+      applySkillPolicy(sessionClientId);
     } else {
-      // null sessionId — new session. Supply onSessionResolved so the
-      // connection auto-watches once the SDK resolves the session ID.
+      // null sessionId — new session. Use a unique clientId so concurrent
+      // new-session starts don't collide in the registry.
+      const newClientId = `${connectionId}:new-${Date.now()}`;
```
dimakis: 🟡 **unsafe_assumptions**: `Date.now()` is not collision-safe for the new-session clientId. If two new-session requests arrive in the same millisecond on the same connection, they'll get the same `newClientId`. Consider using a counter or crypto.randomUUID() instead. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -335,17 +338,17 @@ export function handleSendV2(
         contextBlocks: msg.contextBlocks,
         clientMsgId: msg.clientMsgId,
```
dimakis, line 366: 🔴 **regressions**: On the resume path, `connRegistry.watch()` and `connRegistry.setActive()` are called with the raw `connectionId` (not the composite `sessionClientId`). But `startChat` and `applySkillPolicy` now use the composite key. This is actually correct — the connRegistry tracks connections, not sessions. However, when the connection disconnects (index.ts:280), `findBySessionId` returns the composite `clientId`, which is then used for `registry.get()` and `detachChat()`. This works because the SessionRegistry maps sessionId→clientId correctly. No regression here on closer inspection.

server/ws-handler-v2.ts
dimakis: 🔵 **style**: The variable name `newClientId` is inconsistent with `sessionClientId` used in the resume path. Both serve the same purpose (a per-session-scoped clientId). Consider a unified naming convention, e.g., both could be `sessionClientId` with different construction logic. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `server/ws-handler-v2.ts`
_Clean, correct fix — clientId is treated as an opaque key throughout the codebase so the composite format change is safe. Two minor suggestions: move the misplaced import and add a test for clearSkillPolicy with composite clientId._

- 🔵 **style** (L35): The `import { randomUUID } from 'crypto'` is placed between the type aliases and other imports (line 35), breaking the import grouping convention used in this file. Node built-in imports should go with the other imports at the top (lines 11-24), not after the type alias block. `[fixable]`


_Clean, correct fix — clientId is treated as an opaque key throughout the codebase so the composite format change is safe. Two minor suggestions: move the misplaced import and add a test for clearSkillPolicy with composite clientId._

- 🔵 **missing_tests**: No test verifies that `clearSkillPolicy` receives the composite clientId on the resume/create paths when the prompt is NOT a slash command. The existing test only covers `setSkillPolicy` on the resume path with a skill. A test for the `else` branch (plain text prompt → `clearSkillPolicy(sessionClientId)`) would complete coverage of the `applySkillPolicy` helper. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -32,6 +32,7 @@ type StopMsg = z.infer<typeof V2StopMessage>;
 type InterruptMsg = z.infer<typeof V2InterruptMessage>;
 type PermissionMsg = z.infer<typeof V2PermissionResponseMessage>;
 type SetModeMsg = z.infer<typeof V2SetModeMessage>;
+import { randomUUID } from 'crypto';
```
dimakis, line 35: 🔵 **style**: The `import { randomUUID } from 'crypto'` is placed between the type aliases and other imports (line 35), breaking the import grouping convention used in this file. Node built-in imports should go with the other imports at the top (lines 11-24), not after the type alias block. `[fixable]`

PR #238, opened 2026-04-18
feat: capacitor phase 4 — status bar, safe area, push notifications

dimakis:
## Summary

- **Status bar styling**: `@capacitor/status-bar` integration — dark style (light text) with app background color, configured at startup via `configureStatusBar()`. No-op in browser.
- **Safe area insets**: `viewport-fit=cover` in index.html + `env(safe-area-inset-*)` on body for notch/Dynamic Island clearance. Bottom insets already handled by tab bar.
- **APNs push notifications**: Server-side `apns.ts` module with device token management (JSON file persistence), APNs provider via `@parse/node-apn`, and `sendTurnCompleteNotification` wired into query-loop alongside ntfy/Pushover. Fires when agent finishes and client is detached (backgrounded).
- **Push registration routes**: `POST/DELETE /api/push/register` for device token CRUD from native clients.
- **Frontend push integration**: `@capacitor/push-notifications` — permission request, token registration, foreground handling, tap-to-navigate. Wired into `client-store.ts`.
- **Test fixes**: Updated `useCalendarData`, `useTodoData`, `useSessionList` tests to mock `apiFetch` instead of `fetch` (broken by capacitor PR migration).

## Setup required for push

1. Generate APNs key (.p8) in App Store Connect > Keys
2. Set env vars: `APNS_KEY_PATH`, `APNS_KEY_ID`, `APNS_TEAM_ID`
3. Enable Push Notifications capability in Xcode

## Test plan

- [ ] `npm test` passes (new tests: capacitor, push, apns, push-routes + fixed hook tests)
- [ ] Browser regression: status bar and push init are no-ops, no behavior change
- [ ] Capacitor build: `npm run build:ios && npx cap sync ios` succeeds with 3 plugins
- [ ] Status bar renders with light text on dark background in simulator
- [ ] Safe area clearance visible on notched devices (Dynamic Island, notch)
- [ ] Push permission prompt appears on first launch
- [ ] Device token registered with server on permission grant

### DIFF
--- a/frontend/src/lib/__tests__/push.test.ts
+++ b/frontend/src/lib/__tests__/push.test.ts
@@ -29,20 +29,22 @@ vi.mock('../api-fetch', () => ({
 import { Capacitor } from '@capacitor/core';
 import { PushNotifications } from '@capacitor/push-notifications';
 import { apiFetch } from '../api-fetch';
-import { initPushNotifications } from '../push';
+import { initPushNotifications, _resetForTest } from '../push';
 
 beforeEach(() => {
   vi.clearAllMocks();
+  _resetForTest();
   Object.keys(pushListeners).forEach((k) => delete pushListeners[k]);
   // Restore default mock returns after clearAllMocks resets them
   vi.mocked(PushNotifications.requestPermissions).mockResolvedValue({ receive: 'granted' });
   vi.mocked(PushNotifications.register).mockResolvedValue(undefined);
-  vi.mocked(PushNotifications.addListener).mockImplementation(
-    (event: string, cb: (data: unknown) => void) => {
-      pushListeners[event] = cb;
-      return Promise.resolve() as ReturnType<typeof PushNotifications.addListener>;
-    },
-  );
+  vi.mocked(PushNotifications.addListener).mockImplementation(((
+    event: string,
+    cb: (data: unknown) => void,
+  ) => {
+    pushListeners[event] = cb;
+    return Promise.resolve({ remove: vi.fn() });
+  }) as typeof PushNotifications.addListener);
 });
 
 describe('initPushNotifications', () => {
--- a/frontend/src/lib/__tests__/push.test.ts
+++ b/frontend/src/lib/__tests__/push.test.ts
@@ -1,0 +1,111 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// Mock @capacitor/core
+vi.mock('@capacitor/core', () => ({
+  Capacitor: {
+    isNativePlatform: vi.fn(() => false),
+  },
+}));
+
+// Mock @capacitor/push-notifications — capture listeners
+const pushListeners: Record<string, (data: unknown) => void> = {};
+vi.mock('@capacitor/push-notifications', () => ({
+  PushNotifications: {
+    requestPermissions: vi.fn().mockResolvedValue({ receive: 'granted' }),
+    register: vi.fn().mockResolvedValue(undefined),
+    addListener: vi.fn((event: string, cb: (data: unknown) => void) => {
+      pushListeners[event] = cb;
+      return Promise.resolve();
+    }),
+  },
+}));
+
+// Mock api-fetch
+vi.mock('../api-fetch', () => ({
+  apiFetch: vi.fn().mockResolvedValue({ ok: true }),
+}));
+
+import { Capacitor } from '@capacitor/core';
+import { PushNotifications } from '@capacitor/push-notifications';
+import { apiFetch } from '../api-fetch';
+import { initPushNotifications } from '../push';
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  Object.keys(pushListeners).forEach((k) => delete pushListeners[k]);
+  // Restore default mock returns after clearAllMocks resets them
+  vi.mocked(PushNotifications.requestPermissions).mockResolvedValue({ receive: 'granted' });
+  vi.mocked(PushNotifications.register).mockResolvedValue(undefined);
+  vi.mocked(PushNotifications.addListener).mockImplementation(
+    (event: string, cb: (data: unknown) => void) => {
+      pushListeners[event] = cb;
+      return Promise.resolve() as ReturnType<typeof PushNotifications.addListener>;
+    },
+  );
+});
+
+describe('initPushNotifications', () => {
+  it('no-ops in browser environment', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(false);
+    await initPushNotifications();
+    expect(PushNotifications.requestPermissions).not.toHaveBeenCalled();
+  });
+
+  it('requests permissions on native platform', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(true);
+    await initPushNotifications();
+    expect(PushNotifications.requestPermissions).toHaveBeenCalled();
+  });
+
+  it('calls register after permission granted', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(true);
+    await initPushNotifications();
+    expect(PushNotifications.register).toHaveBeenCalled();
+  });
+
+  it('does not register when permission denied', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(true);
+    vi.mocked(PushNotifications.requestPermissions).mockResolvedValue({
+      receive: 'denied',
+    });
+    await initPushNotifications();
+    expect(PushNotifications.register).not.toHaveBeenCalled();
+  });
+
+  it('registers device token with server on registration event', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(true);
+    await initPushNotifications();
+
+    // Simulate registration success
+    pushListeners['registration']({ value: 'device-token-xyz' });
+
+    expect(apiFetch).toHaveBeenCalledWith('/api/push/register', {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ token: 'device-token-xyz' }),
+    });
+  });
+
+  it('registers listeners for push events', async () => {
+    vi.mocked(Capacitor.isNativePlatform).mockReturnValue(true);
+    await initPushNotifications();
+
+    expect(PushNotifications.addListener).toHaveBeenCalledWith(
+      'registration',
+      expect.any(Function),
+    );
+    expect(PushNotifications.addListener).toHaveBeenCalledWith(
+      'registrationError',
+      expect.any(Function),
+    );
+    expect(PushNotifications.addListener).toHaveBeenCalledWith(
+      'pushNotificationReceived',
+      expect.any(Function),
+    );
+    expect(PushNotifications.addListener).toHaveBeenCalledWith(
+      'pushNotificationActionPerformed',
+      expect.any(Function),
+    );
+  });
+});
--- a/frontend/src/lib/capacitor.ts
+++ b/frontend/src/lib/capacitor.ts
@@ -1,11 +1,4 @@
-/**
- * Capacitor-specific lifecycle integration.
- *
- * When running as a native iOS app, registers appStateChange listener
- * to force a WebSocket reconnect check on resume. The callback directly
- * invokes MitzoConnection.checkAndReconnect() via the store rather than
- * relying on synthetic DOM events.
- */
+// Capacitor-specific lifecycle integration for native iOS.
 
 import { Capacitor } from '@capacitor/core';
 import { App } from '@capacitor/app';
@@ -15,15 +8,7 @@ export function isCapacitor(): boolean {
   return Capacitor.isNativePlatform();
 }
 
-/**
- * Register Capacitor app lifecycle events. When the app returns to the
- * foreground, calls the provided callback to force a WebSocket reconnect
- * check. This directly invokes MitzoConnection.checkAndReconnect() rather
- * than relying on synthetic visibilitychange events (which won't update
- * the read-only document.visibilityState property).
- *
- * Call once at app startup. No-op in browser environments.
- */
+/** Register app lifecycle events. Calls onResume on foreground return. No-op in browser. */
 export function registerCapacitorLifecycle(onResume: () => void): void {
   if (!isCapacitor()) return;
 
@@ -32,12 +17,7 @@ export function registerCapacitorLifecycle(onResume: () => void): void {
   });
 }
 
-/**
- * Configure the native status bar appearance. Sets light text on the dark
- * app background (#0f0f1a) so the status bar blends with the UI.
- *
- * No-op in browser environments.
- */
+/** Configure native status bar — light text on dark background. No-op in browser. */
 export async function configureStatusBar(): Promise<void> {
   if (!isCapacitor()) return;
 
--- a/frontend/src/lib/capacitor.ts
+++ b/frontend/src/lib/capacitor.ts
@@ -9,6 +9,7 @@
 
 import { Capacitor } from '@capacitor/core';
 import { App } from '@capacitor/app';
+import { StatusBar, Style } from '@capacitor/status-bar';
 
 export function isCapacitor(): boolean {
   return Capacitor.isNativePlatform();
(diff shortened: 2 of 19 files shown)


dimakis (commented):
## Centaur Review

Found **8** issue(s) (2 warning).


### `frontend/src/lib/push.ts`
_Solid push notification implementation with good test coverage for the backend. Key concerns: bare `require()` in ESM context may break in compiled output, notification tap handler bypasses React Router, and `sendPermissionNotification` is dead code._

- 🟡 **bugs** (L41): Navigation via `window.location.pathname` bypasses React Router, causing a full page reload instead of a client-side transition. The rest of the codebase uses `useNavigate()` from react-router-dom. Since this runs inside a Capacitor listener callback (not a React component), consider using `window.location.href = '/chat/...'` at minimum, or importing the router instance directly. The `window.location.hash = ''` on line 40 is also a no-op since the app uses BrowserRouter, not HashRouter. `[fixable]`
- 🔵 **style** (L1): 8-line module-level docstring violates the project's 'no comments / one short line max' convention. `[fixable]`
- 🔵 **unsafe_assumptions** (L21): No guard against calling `initPushNotifications()` multiple times. Each call registers new listeners via `addListener`, which could result in duplicate token registration calls and duplicate navigation handlers. Consider a module-level `initialized` flag. `[fixable]`


### `server/apns.ts`
_Solid push notification implementation with good test coverage for the backend. Key concerns: bare `require()` in ESM context may break in compiled output, notification tap handler bypasses React Router, and `sendPermissionNotification` is dead code._

- 🟡 **bugs** (L83): Using bare `require('@parse/node-apn')` in an ESM project (`"type": "module"` in package.json). The project convention for CommonJS interop is `createRequire` (see `server/logger.ts`). Bare `require()` is not defined in ESM modules — this works in tsx/ts-node but may break in the compiled `dist/` output from `tsc`. Use `import { createRequire } from 'module'` or dynamic `await import()`. `[fixable]`
- 🔵 **regressions** (L146): `sendPermissionNotification()` is exported but never called anywhere. The permission handler (`packages/harness/src/permission-handler.ts`) calls the ntfy and Pushover equivalents but not the APNs one. This is dead code — either wire it up in the permission handler or remove it to avoid confusion. `[fixable]`
- 🔵 **style** (L1): Multi-paragraph module-level docstring (14 lines of comments). Project convention per CLAUDE.md is 'default to writing no comments' with a 'one short line max'. The function-level behavior is clear from the code and types. `[fixable]`


### `frontend/src/lib/__tests__/push.test.ts`
_Solid push notification implementation with good test coverage for the backend. Key concerns: bare `require()` in ESM context may break in compiled output, notification tap handler bypasses React Router, and `sendPermissionNotification` is dead code._

- 🔵 **missing_tests**: No test for the `pushNotificationActionPerformed` navigation behavior (the handler that sets `window.location.pathname`). Also no test for calling `initPushNotifications()` multiple times — there's no guard against double-registration of listeners. `[fixable]`


### `frontend/src/lib/capacitor.ts`
_Solid push notification implementation with good test coverage for the backend. Key concerns: bare `require()` in ESM context may break in compiled output, notification tap handler bypasses React Router, and `sendPermissionNotification` is dead code._

- 🔵 **style** (L34): 5-line docstring on `configureStatusBar` — exceeds the project's 'one short line max' comment convention. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  await PushNotifications.addListener('pushNotificationReceived', (_notification) => {
+    // Foreground notification — the app is already open, so no action needed.
+    // The WS connection handles live updates.
+  });
+
+  await PushNotifications.addListener('pushNotificationActionPerformed', (action) => {
+    // User tapped a notification — navigate to the relevant session
+    const data = (action as { notification: { data: Record<string, string> } }).notification.data;
+    if (data?.sessionId) {
```
dimakis, line 39: 🟡 **bugs**: Navigation via `window.location.pathname` bypasses React Router, causing a full page reload instead of a client-side transition. The rest of the codebase uses `useNavigate()` from react-router-dom. Since this runs inside a Capacitor listener callback (not a React component), consider using `window.location.href = '/chat/...'` at minimum, or importing the router instance directly. The `window.location.hash = ''` on line 40 is also a no-op since the app uses BrowserRouter, not HashRouter. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -0,0 +1,48 @@
+/**
```
dimakis: 🔵 **style**: 8-line module-level docstring violates the project's 'no comments / one short line max' convention. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -0,0 +1,48 @@
...
+import { apiFetch } from './api-fetch';
+
+export async function initPushNotifications(): Promise<void> {
+  if (!Capacitor.isNativePlatform()) return;
+
+  const permission = await PushNotifications.requestPermissions();
+  if (permission.receive !== 'granted') return;
+
+  // Register listeners before calling register()
+  await PushNotifications.addListener('registration', (token) => {
```
dimakis, line 21: 🔵 **unsafe_assumptions**: No guard against calling `initPushNotifications()` multiple times. Each call registers new listeners via `addListener`, which could result in duplicate token registration calls and duplicate navigation handlers. Consider a module-level `initialized` flag. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,159 @@
...
+let apnProvider: import('@parse/node-apn').Provider | null = null;
+
+function getProvider(): import('@parse/node-apn').Provider | null {
+  if (apnProvider) return apnProvider;
+  if (!isConfigured()) return null;
+
+  try {
+    // Dynamic import to avoid requiring the module when not configured
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const apn = require('@parse/node-apn');
```
dimakis, line 72: 🟡 **bugs**: Using bare `require('@parse/node-apn')` in an ESM project (`"type": "module"` in package.json). The project convention for CommonJS interop is `createRequire` (see `server/logger.ts`). Bare `require()` is not defined in ESM modules — this works in tsx/ts-node but may break in the compiled `dist/` output from `tsc`. Use `import { createRequire } from 'module'` or dynamic `await import()`. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,159 @@
...
+  sessionId?: string,
+  snippet?: string,
+): Promise<void> {
+  await sendPush('Mitzo: Agent replied', snippet || 'The agent has finished its turn.', {
+    type: 'turn_complete',
+    sessionId,
+  });
+}
+
+export async function sendPermissionNotification(
```
dimakis: 🔵 **regressions**: `sendPermissionNotification()` is exported but never called anywhere. The permission handler (`packages/harness/src/permission-handler.ts`) calls the ntfy and Pushover equivalents but not the APNs one. This is dead code — either wire it up in the permission handler or remove it to avoid confusion. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,159 @@
+/**
```
dimakis: 🔵 **style**: Multi-paragraph module-level docstring (14 lines of comments). Project convention per CLAUDE.md is 'default to writing no comments' with a 'one short line max'. The function-level behavior is clear from the code and types. `[fixable]`

frontend/src/lib/capacitor.ts
```diff
@@ -30,3 +31,16 @@ export function registerCapacitorLifecycle(onResume: () => void): void {
     if (isActive) onResume();
   });
 }
+
```
dimakis, line 19: 🔵 **style**: 5-line docstring on `configureStatusBar` — exceeds the project's 'one short line max' comment convention. `[fixable]`

frontend/src/lib/__tests__/push.test.ts
```diff
@@ -0,0 +1,111 @@
...
+vi.mock('../api-fetch', () => ({
+  apiFetch: vi.fn().mockResolvedValue({ ok: true }),
+}));
+
+import { Capacitor } from '@capacitor/core';
+import { PushNotifications } from '@capacitor/push-notifications';
+import { apiFetch } from '../api-fetch';
+import { initPushNotifications } from '../push';
+
+beforeEach(() => {
```
dimakis, line 34: 🔴 **bugs**: The `initialized` module-level flag in `push.ts` is never reset between tests. `vi.clearAllMocks()` in `beforeEach` does not reset module-level state. After the first test that calls `initPushNotifications()` on a native platform, the `initialized` flag stays `true`, causing all subsequent tests to silently no-op. Only the first test to set `isNativePlatform=true` actually exercises the code path. Tests appear to pass because `vi.mock` hoists and the mock module is shared, but the `initialized` guard means only one test truly runs through the logic. Use `vi.resetModules()` + dynamic `import()` in `beforeEach` (like `apns.test.ts` does) to get fresh module state per test. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  await PushNotifications.addListener('pushNotificationReceived', (_notification) => {
+    // Foreground — WS handles live updates, no action needed
+  });
+
+  await PushNotifications.addListener('pushNotificationActionPerformed', (action) => {
+    const data = (action as { notification: { data: Record<string, string> } }).notification.data;
+    if (data?.sessionId) {
+      window.location.href = `/chat/${data.sessionId}`;
```
dimakis, line 40: 🟡 **unsafe_assumptions**: The `sessionId` from a push notification payload is interpolated directly into `window.location.href` without validation. While `window.location.href` assignment is generally safe from XSS (browsers won't execute `javascript:` URIs set this way), the sessionId should still be validated to prevent open redirect or path traversal (e.g., `../../some/other/path`). Consider validating it matches an expected format like `/^[a-zA-Z0-9_-]+$/` before navigation. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function initPushNotifications(): Promise<void> {
+  if (!Capacitor.isNativePlatform() || initialized) return;
+  initialized = true;
+
+  const permission = await PushNotifications.requestPermissions();
+  if (permission.receive !== 'granted') return;
+
+  await PushNotifications.addListener('registration', (token) => {
+    apiFetch('/api/push/register', {
```
dimakis, line 22: 🔵 **style**: The `apiFetch` call in the `registration` listener fires without error handling. If the server is unreachable, the token registration silently fails and the device won't receive push notifications. Consider adding a `.catch()` with a retry or at least a `console.warn`. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,132 @@
...
+  try {
+    const apn = require('@parse/node-apn');
+    const notification = new apn.Notification();
+    notification.alert = { title, body };
+    notification.topic = APNS_BUNDLE_ID;
+    notification.sound = 'default';
+    notification.badge = 1;
+    if (data) notification.payload = data;
+
+    const result = await provider.send(notification, tokens);
```
dimakis, line 108: 🟡 **unsafe_assumptions**: `sendPush` passes the `tokens` array reference directly to `provider.send()`, then mutates the same array via `removeToken()` in the failure-handling loop. While Node.js is single-threaded and `provider.send()` returns a result (not an iterator), if the APNs library retains or iterates the array reference internally after returning, the mutation could cause subtle issues. Safer to pass `[...tokens]` (a copy) to `provider.send()`. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,132 @@
...
+  try {
+    writeFileSync(tokenStorePath, JSON.stringify(tokens));
+  } catch (err: unknown) {
+    log.error('failed to persist device tokens', {
+      error: err instanceof Error ? err.message : 'unknown',
+    });
+  }
+}
+
+export function registerToken(token: string): void {
```
dimakis, line 46: 🟡 **unsafe_assumptions**: No upper bound on the number of device tokens that can be registered. Each authenticated call to `POST /api/push/register` appends to an unbounded array and persists to disk. While auth-gated, a compromised session or misbehaving client could accumulate thousands of tokens, growing the JSON file and the array passed to `provider.send()`. Consider capping the array size (e.g., 50 tokens) and evicting the oldest on overflow. `[fixable]`

server/apns.ts
```diff
@@ -0,0 +1,132 @@
...
+
+import { readFileSync, writeFileSync, existsSync } from 'fs';
+import { createRequire } from 'module';
+import { createLogger } from '@mitzo/harness';
+
+const require = createRequire(import.meta.url);
+
+const log = createLogger('apns');
+
+const APNS_KEY_PATH = process.env.APNS_KEY_PATH;
```
dimakis, line 11: 🔵 **style**: APNs configuration constants (`APNS_KEY_PATH`, etc.) are captured once at module load time. This is fine for production but makes testing awkward — the `isConfigured` tests must use `vi.resetModules()` to get fresh captures. Reading directly from `process.env` inside `isConfigured()` and `getProvider()` would be simpler and more testable, consistent with how other config is typically handled. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/lib/__tests__/push.test.ts`
_Solid feature PR with good test coverage for the token store and routes, but the frontend push test suite has a critical `initialized` flag leak across tests that means only one test truly exercises the native code path — fix with `vi.resetModules()` per test._

- 🔴 **bugs** (L34): The `initialized` module-level flag in `push.ts` is never reset between tests. `vi.clearAllMocks()` in `beforeEach` does not reset module-level state. After the first test that calls `initPushNotifications()` on a native platform, the `initialized` flag stays `true`, causing all subsequent tests to silently no-op. Only the first test to set `isNativePlatform=true` actually exercises the code path. Tests appear to pass because `vi.mock` hoists and the mock module is shared, but the `initialized` guard means only one test truly runs through the logic. Use `vi.resetModules()` + dynamic `import()` in `beforeEach` (like `apns.test.ts` does) to get fresh module state per test. `[fixable]`
- 🟡 **missing_tests**: No test covers the `pushNotificationActionPerformed` handler that navigates to `/chat/{sessionId}`. This is the most complex listener (data extraction + navigation) and the one most likely to break. Add a test that fires the listener and asserts `window.location.href` is set correctly. `[fixable]`


### `frontend/src/lib/push.ts`
_Solid feature PR with good test coverage for the token store and routes, but the frontend push test suite has a critical `initialized` flag leak across tests that means only one test truly exercises the native code path — fix with `vi.resetModules()` per test._

- 🟡 **unsafe_assumptions** (L35): The `sessionId` from a push notification payload is interpolated directly into `window.location.href` without validation. While `window.location.href` assignment is generally safe from XSS (browsers won't execute `javascript:` URIs set this way), the sessionId should still be validated to prevent open redirect or path traversal (e.g., `../../some/other/path`). Consider validating it matches an expected format like `/^[a-zA-Z0-9_-]+$/` before navigation. `[fixable]`
- 🔵 **style** (L17): The `apiFetch` call in the `registration` listener fires without error handling. If the server is unreachable, the token registration silently fails and the device won't receive push notifications. Consider adding a `.catch()` with a retry or at least a `console.warn`. `[fixable]`


### `server/apns.ts`
_Solid feature PR with good test coverage for the token store and routes, but the frontend push test suite has a critical `initialized` flag leak across tests that means only one test truly exercises the native code path — fix with `vi.resetModules()` per test._

- 🟡 **unsafe_assumptions** (L108): `sendPush` passes the `tokens` array reference directly to `provider.send()`, then mutates the same array via `removeToken()` in the failure-handling loop. While Node.js is single-threaded and `provider.send()` returns a result (not an iterator), if the APNs library retains or iterates the array reference internally after returning, the mutation could cause subtle issues. Safer to pass `[...tokens]` (a copy) to `provider.send()`. `[fixable]`
- 🟡 **unsafe_assumptions** (L46): No upper bound on the number of device tokens that can be registered. Each authenticated call to `POST /api/push/register` appends to an unbounded array and persists to disk. While auth-gated, a compromised session or misbehaving client could accumulate thousands of tokens, growing the JSON file and the array passed to `provider.send()`. Consider capping the array size (e.g., 50 tokens) and evicting the oldest on overflow. `[fixable]`
- 🔵 **style** (L11): APNs configuration constants (`APNS_KEY_PATH`, etc.) are captured once at module load time. This is fine for production but makes testing awkward — the `isConfigured` tests must use `vi.resetModules()` to get fresh captures. Reading directly from `process.env` inside `isConfigured()` and `getProvider()` would be simpler and more testable, consistent with how other config is typically handled. `[fixable]`


### `server/__tests__/apns.test.ts`
_Solid feature PR with good test coverage for the token store and routes, but the frontend push test suite has a critical `initialized` flag leak across tests that means only one test truly exercises the native code path — fix with `vi.resetModules()` per test._

- 🔵 **missing_tests**: The `sendPush` test only covers the 'not configured' path. The happy path (configured provider, token cleanup on 410) and error handling path (provider.send throws) are untested. These are the most complex code paths in the module. `[fixable]`

PR #239, opened 2026-04-18
fix: mobile WS reconnect — prevent event loss on iOS background/resume

dimakis:
## Summary

- **Server**: `sendOrBuffer` now gates v2 broadcast on `hasOpenWatchers(sessionId)` instead of checking the originating `clientId` — after a WS reconnect the old connection is gone but new ones may be watching. Also reattaches detached sessions on reconnect and on send.
- **Client**: `defuseOldWs()` nulls all handlers on the old WebSocket before creating a new one, so the stale `onclose` can't overwrite `this.ws` / `this._connected` and cause cascading reconnect cycles.
- **Client**: Foreground recovery — on `visibilitychange → visible`, if the store has an active session but no messages (iOS evicted the page), re-fetches from the REST API.

## Root cause

iOS Safari kills WebSocket connections when the app is backgrounded. Three things then go wrong:

1. The query loop's closed-over `clientId` no longer exists in `ConnectionRegistry` → events silently dropped
2. `checkAndReconnect()` replaces the WS but the old socket's `onclose` fires later, setting `this.ws = null` and `_connected = false` on the **new** connection — causing duplicate reconnect cycles
3. If iOS fully evicts the page from memory, the in-memory Zustand store and `seqBySession` map are lost — no reconnect replay happens and messages vanish

## Test plan

- [x] Send a message on mobile, switch away, come back — response visible without pull-to-refresh
- [x] Second prompt in same session displays correctly
- [ ] Desktop/Safari — no regression on normal WS lifecycle
- [ ] `npm test` passes

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -108,6 +108,19 @@ export class MitzoConnection {
 
   // ─── Internal ──────────────────────────────────────────────────────────────
 
+  /**
+   * Detach handlers from the old WS so its delayed onclose can't overwrite
+   * `this.ws` / `this._connected` after a new WS has been created.
+   */
+  private defuseOldWs(): void {
+    if (this.ws) {
+      this.ws.onclose = null;
+      this.ws.onmessage = null;
+      this.ws.onerror = null;
+      this.ws = null;
+    }
+  }
+
   private doConnect(): void {
     if (
       this.ws?.readyState === WS_READY_STATE.OPEN ||
@@ -200,7 +213,7 @@ export class MitzoConnection {
     if (typeof globalThis.setInterval === 'undefined') return;
     this.heartbeatTimer = setInterval(() => {
       if (this.ws && this.ws.readyState !== WS_READY_STATE.OPEN && !this.reconnectTimer) {
-        this.ws = null;
+        this.defuseOldWs();
         this._connected = false;
         this.listener?.({ type: '_close' });
         this.doConnect();
@@ -221,6 +234,9 @@ export class MitzoConnection {
     this.boundOnVisibility = () => {
       if (document.visibilityState === 'visible') {
         this.checkAndReconnect();
+        // Notify the store so it can re-hydrate messages if the page was
+        // evicted from memory (iOS bfcache discard) and state was lost.
+        this.listener?.({ type: '_foreground' });
       }
     };
 
@@ -247,7 +263,7 @@ export class MitzoConnection {
   checkAndReconnect(): void {
     if (!this.ws || this.ws.readyState !== WS_READY_STATE.OPEN) {
       if (this.reconnectTimer) return;
-      this.ws = null;
+      this.defuseOldWs();
       this._connected = false;
       this.listener?.({ type: '_close' });
       this.doConnect();
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -571,7 +571,11 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
               }
             }
           })
-          .catch(() => {});
+          .catch((err) => {
+            if (typeof console !== 'undefined') {
+              console.warn('[mitzo] foreground recovery fetch failed', err);
+            }
+          });
       }
       return;
     }
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -552,6 +552,30 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
   };
 
   function wsListener(msg: Record<string, unknown>) {
+    // Foreground recovery: when the page becomes visible again (iOS may have
+    // evicted it from memory, losing in-memory state), re-fetch messages from
+    // the REST API if we have an active session but no messages in the store.
+    if (msg.type === '_foreground') {
+      const { sessions, messages: msgs } = store.getState();
+      if (sessions.active && msgs.messages.length === 0 && !msgs.current) {
+        api
+          .getSessionMessages(sessions.active)
+          .then((restored) => {
+            if (Array.isArray(restored) && restored.length > 0) {
+              // Only restore if store is still empty (avoid clobbering live data)
+              const current = store.getState().messages;
+              if (current.messages.length === 0 && !current.current) {
+                store.setState((s) => ({
+                  messages: messagesReducer(s.messages, { type: 'RESTORE', messages: restored }),
+                }));
+              }
+            }
+          })
+          .catch(() => {});
+      }
+      return;
+    }
+
     const eventSessionId = msg.sessionId as string | undefined;
 
     // Session-scoped event filtering for multiplexed v2 connections:
--- a/packages/harness/src/connection-registry.ts
+++ b/packages/harness/src/connection-registry.ts
@@ -84,6 +84,17 @@ export class ConnectionRegistry {
   }
 
   /**
+   * Check if at least one open connection is watching the given session.
+   * Short-circuits on the first match — no array allocation.
+   */
+  hasOpenWatchers(sessionId: string): boolean {
+    for (const conn of this.connections.values()) {
+      if (conn.watchedSessions.has(sessionId) && conn.transport.isOpen()) return true;
+    }
+    return false;
+  }
+
+  /**
    * Send a message to all open connections watching a session.
    * Catches send errors to prevent one failing transport from
    * aborting the broadcast loop.
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -81,9 +81,11 @@ function sendOrBuffer(
     }
   }
 
-  // v2 path: deliver via ConnectionRegistry when clientId is a registered v2
-  // connection AND sessionId is resolved (broadcast needs a session target).
-  if (connRegistry?.get(clientId) && sessionId) {
+  // v2 path: deliver via ConnectionRegistry when there are open connections
+  // watching this session. Uses sessionId-based fan-out instead of checking
+  // whether the originating clientId is still registered — after a WS
+  // reconnect the original connection is gone but new ones may be watching.
+  if (sessionId && connRegistry?.hasOpenWatchers(sessionId)) {
     connRegistry.broadcast(sessionId, enriched);
     return;
   }
@@ -399,8 +401,14 @@ export async function runQueryLoop(
         }
 
         emit(v2('session_end', { sessionId: msg.session_id, usage: usageData }));
-        if (connRegistry?.get(clientId)) {
-          connRegistry.setActive(clientId, null);
+        const resultSid = (msg.session_id as string) || currentSession.sessionId;
+        if (resultSid && connRegistry?.hasOpenWatchers(resultSid)) {
+          for (const { connectionId: cid } of connRegistry.getConnectionsWatching(
+            resultSid,
+            true,
+          )) {
+            connRegistry.setActive(cid, null);
+          }
         }
         if (!registry.isAttached(clientId)) {
           const snippet = extractSnippet(snapshotBlocks, NOTIFY_SNIPPET_MAX_CHARS);
@@ -669,14 +677,18 @@ export async function runQueryLoop(
       finalSession.currentSnapshot = null;
       if (!doneSent) {
         const endMsg = v2('session_end', { sessionId: finalSession.sessionId });
-        if (connRegistry?.get(clientId) && finalSession.sessionId) {
-          connRegistry.broadcast(finalSession.sessionId, endMsg);
+        const sid = finalSession.sessionId;
+        if (sid && connRegistry?.hasOpenWatchers(sid)) {
+          connRegistry.broadcast(sid, endMsg);
         } else {
           send(finalSession.transport, endMsg);
           broadcastToObservers(finalSession.observers, endMsg);
         }
-        if (connRegistry?.get(clientId)) {
-          connRegistry.setActive(clientId, null);
+        if (sid && connRegistry?.hasOpenWatchers(sid)) {
+          // Clear active session on all watching connections
+          for (const { connectionId: cid } of connRegistry.getConnectionsWatching(sid, true)) {
+            connRegistry.setActive(cid, null);
+          }
         }
       }
       registry.remove(clientId);
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -407,7 +407,10 @@ export async function runQueryLoop(
             resultSid,
             true,
           )) {
-            connRegistry.setActive(cid, null);
+            const conn = connRegistry.get(cid);
+            if (conn?.activeSession === resultSid) {
+              connRegistry.setActive(cid, null);
+            }
           }
         }
         if (!registry.isAttached(clientId)) {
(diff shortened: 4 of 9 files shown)


server/ws-handler-v2.ts
```diff
@@ -118,9 +119,23 @@ export function handleReconnect(
...
       }
 
-      // Check if session has an active driver in the SessionRegistry
+      // Check if session has an active driver in the SessionRegistry.
+      // If the driver is still running but was detached by the old WS close,
+      // reattach it with the new connection's transport so future events
+      // from the query loop can be delivered via the v1 fallback path.
       const found = ctx.sessionRegistry.findBySessionId(entry.sessionId);
       const running = found ? ctx.sessionRegistry.isActive(found.clientId) : false;
+      if (found && running && !ctx.sessionRegistry.isAttached(found.clientId)) {
```
dimakis, line 128: 🟡 **missing_tests**: The reattach-on-reconnect path (lines 128-137) and reattach-on-send path (lines 327-336) have no test coverage. In ws-handler-v2.test.ts, `isAttached` is mocked to return `true`, so the `!isAttached` branch is never exercised. Add tests where `isAttached` returns `false` and verify `reattach`/`reattachChat` are called with the correct arguments. `[fixable]`

dimakis (commented):
## Centaur Review

Found **6** issue(s) (4 warning).


### `server/ws-handler-v2.ts`
_Core fix is sound — session-based fan-out in sendOrBuffer correctly solves the reconnect event-loss bug and the new query-loop test validates the key scenario. Main gaps are untested reattach paths in ws-handler-v2, untested client-side changes, and an overbroad setActive clearing that could affect connections watching multiple sessions._

- 🟡 **missing_tests** (L128): The reattach-on-reconnect path (lines 128-137) and reattach-on-send path (lines 327-336) have no test coverage. In ws-handler-v2.test.ts, `isAttached` is mocked to return `true`, so the `!isAttached` branch is never exercised. Add tests where `isAttached` returns `false` and verify `reattach`/`reattachChat` are called with the correct arguments. `[fixable]`
- 🔵 **style** (L131): Inconsistent reattach API: `handleReconnect` calls `ctx.sessionRegistry.reattach()` directly (line 131) while `handleSendV2` uses the `reattachChat()` wrapper (line 334). If `reattachChat` gains additional logic (e.g., logging, metrics), only one path benefits. Use the same entry point in both places. `[fixable]`


_Core fix is sound — session-based fan-out in sendOrBuffer correctly solves the reconnect event-loss bug and the new query-loop test validates the key scenario. Main gaps are untested reattach paths in ws-handler-v2, untested client-side changes, and an overbroad setActive clearing that could affect connections watching multiple sessions._

- 🟡 **missing_tests** (L115): No tests for `defuseOldWs()`. The existing heartbeat test (connection.test.ts:300) verifies a new WS is created but doesn't assert that the old WS's `onclose`/`onmessage`/`onerror` handlers are nullified — which is the core bug this method fixes. A test should verify that a delayed `onclose` on the old WS does not reset `this.ws` or `this._connected` after a new WS is created. `[fixable]`


_Core fix is sound — session-based fan-out in sendOrBuffer correctly solves the reconnect event-loss bug and the new query-loop test validates the key scenario. Main gaps are untested reattach paths in ws-handler-v2, untested client-side changes, and an overbroad setActive clearing that could affect connections watching multiple sessions._

- 🟡 **missing_tests** (L555): No tests for the `_foreground` recovery path. Should verify: (1) REST re-fetch fires when `sessions.active` is set but messages are empty, (2) re-fetch is skipped when messages already exist, (3) the double-check guard prevents clobbering data that arrived between the fetch start and resolve. `[fixable]`
- 🔵 **style** (L575): `.catch(() => {})` silently swallows API errors during foreground recovery. A `console.warn` or debug-level log would aid diagnosis on user devices where this recovery path fires unexpectedly. `[fixable]`


### `server/query-loop.ts`
_Core fix is sound — session-based fan-out in sendOrBuffer correctly solves the reconnect event-loss bug and the new query-loop test validates the key scenario. Main gaps are untested reattach paths in ws-handler-v2, untested client-side changes, and an overbroad setActive clearing that could affect connections watching multiple sessions._

- 🟡 **regressions** (L404): The `setActive(cid, null)` loop clears the active session on ALL open connections watching the ending session. If a connection watches session X (from a prior reconnect) but currently has session Z as its active session, this would incorrectly clear Z. The old code only cleared the originating `clientId`. Consider guarding: `if (conn.activeSession === resultSid) connRegistry.setActive(cid, null)`. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -118,9 +119,23 @@ export function handleReconnect(
...
+        const conn = ctx.connRegistry.get(connectionId);
+        if (conn) {
+          ctx.sessionRegistry.reattach(found.clientId, conn.transport);
```
dimakis: 🔵 **style**: Inconsistent reattach API: `handleReconnect` calls `ctx.sessionRegistry.reattach()` directly (line 131) while `handleSendV2` uses the `reattachChat()` wrapper (line 334). If `reattachChat` gains additional logic (e.g., logging, metrics), only one path benefits. Use the same entry point in both places. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -108,6 +108,19 @@ export class MitzoConnection {
 
   // ─── Internal ──────────────────────────────────────────────────────────────
 
+  /**
+   * Detach handlers from the old WS so its delayed onclose can't overwrite
+   * `this.ws` / `this._connected` after a new WS has been created.
+   */
+  private defuseOldWs(): void {
```
dimakis, line 115: 🟡 **missing_tests**: No tests for `defuseOldWs()`. The existing heartbeat test (connection.test.ts:300) verifies a new WS is created but doesn't assert that the old WS's `onclose`/`onmessage`/`onerror` handlers are nullified — which is the core bug this method fixes. A test should verify that a delayed `onclose` on the old WS does not reset `this.ws` or `this._connected` after a new WS is created. `[fixable]`

packages/client/src/store.ts
```diff
@@ -552,6 +552,30 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
   };
 
   function wsListener(msg: Record<string, unknown>) {
+    // Foreground recovery: when the page becomes visible again (iOS may have
```
dimakis, line 555: 🟡 **missing_tests**: No tests for the `_foreground` recovery path. Should verify: (1) REST re-fetch fires when `sessions.active` is set but messages are empty, (2) re-fetch is skipped when messages already exist, (3) the double-check guard prevents clobbering data that arrived between the fetch start and resolve. `[fixable]`

packages/client/src/store.ts
```diff
@@ -552,6 +552,30 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+              const current = store.getState().messages;
+              if (current.messages.length === 0 && !current.current) {
+                store.setState((s) => ({
+                  messages: messagesReducer(s.messages, { type: 'RESTORE', messages: restored }),
+                }));
+              }
+            }
+          })
+          .catch(() => {});
+      }
```
dimakis, line 579: 🔵 **style**: `.catch(() => {})` silently swallows API errors during foreground recovery. A `console.warn` or debug-level log would aid diagnosis on user devices where this recovery path fires unexpectedly. `[fixable]`

server/query-loop.ts
```diff
@@ -399,8 +401,14 @@ export async function runQueryLoop(
         }
 
         emit(v2('session_end', { sessionId: msg.session_id, usage: usageData }));
-        if (connRegistry?.get(clientId)) {
-          connRegistry.setActive(clientId, null);
+        const resultSid = (msg.session_id as string) || currentSession.sessionId;
```
dimakis, line 404: 🟡 **regressions**: The `setActive(cid, null)` loop clears the active session on ALL open connections watching the ending session. If a connection watches session X (from a prior reconnect) but currently has session Z as its active session, this would incorrectly clear Z. The old code only cleared the originating `clientId`. Consider guarding: `if (conn.activeSession === resultSid) connRegistry.setActive(cid, null)`. `[fixable]`

packages/harness/src/connection-registry.ts
```diff
@@ -83,6 +83,17 @@ export class ConnectionRegistry {
     return result;
   }
 
+  /**
+   * Check if at least one open connection is watching the given session.
+   * Short-circuits on the first match — no array allocation.
+   */
+  hasOpenWatchers(sessionId: string): boolean {
```
dimakis, line 90: 🟡 **missing_tests**: `hasOpenWatchers` is a new public method used in 3 critical delivery paths in `query-loop.ts`, but it has no unit tests in `packages/harness/__tests__/connection-registry.test.ts`. Sibling methods like `getConnectionsWatching` and `broadcast` are tested. Should cover: returns true with an open watcher, false with no watchers, and false when all watchers are closed. `[fixable]`

server/query-loop.ts
```diff
@@ -81,9 +81,11 @@ function sendOrBuffer(
...
   }
 
-  // v2 path: deliver via ConnectionRegistry when clientId is a registered v2
-  // connection AND sessionId is resolved (broadcast needs a session target).
-  if (connRegistry?.get(clientId) && sessionId) {
+  // v2 path: deliver via ConnectionRegistry when there are open connections
+  // watching this session. Uses sessionId-based fan-out instead of checking
+  // whether the originating clientId is still registered — after a WS
+  // reconnect the original connection is gone but new ones may be watching.
+  if (sessionId && connRegistry?.hasOpenWatchers(sessionId)) {
```
dimakis, line 88: 🟡 **bugs**: Before `sessionId` is resolved (early in the query loop), `sendOrBuffer` is called with `sessionId = undefined`. The old code's v2 guard (`connRegistry?.get(clientId)`) would deliver early events (like `session_id`) via the v2 path if the clientId was registered. The new guard (`sessionId && connRegistry?.hasOpenWatchers(sessionId)`) correctly skips this when sessionId is undefined and falls through to the v1 path. However, after session resolution, if the WS has already reconnected and the old clientId is gone from both the connRegistry AND the session is detached (`isAttached` returns false), early events emitted before reattach would be silently dropped (v2 guard fails, v1 guard fails). The event store persistence at line 76-81 still fires though, so replay on reconnect would recover them. Low risk but worth verifying that `handleReconnect` reattach happens before events are emitted in this window.

server/query-loop.ts
```diff
@@ -399,8 +401,17 @@ export async function runQueryLoop(
         }
 
         emit(v2('session_end', { sessionId: msg.session_id, usage: usageData }));
-        if (connRegistry?.get(clientId)) {
-          connRegistry.setActive(clientId, null);
+        const resultSid = (msg.session_id as string) || currentSession.sessionId;
+        if (resultSid && connRegistry?.hasOpenWatchers(resultSid)) {
```
dimakis, line 405: 🔵 **style**: The 7-line active-session cleanup loop (iterate watchers, check `activeSession`, call `setActive(cid, null)`) is duplicated verbatim at lines 405-414 and 688-697. Consider extracting a `clearActiveWatchers(connRegistry, sessionId)` helper to reduce duplication. `[fixable]`

packages/client/src/store.ts
```diff
@@ -552,6 +552,34 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
   };
 
   function wsListener(msg: Record<string, unknown>) {
+    // Foreground recovery: when the page becomes visible again (iOS may have
+    // evicted it from memory, losing in-memory state), re-fetch messages from
+    // the REST API if we have an active session but no messages in the store.
+    if (msg.type === '_foreground') {
```
dimakis, line 558: 🔵 **unsafe_assumptions**: The `_foreground` recovery fetch races with WS reconnect replay: `checkAndReconnect()` and `_foreground` fire back-to-back in the visibility handler. If replay events from `handleReconnect` haven't arrived yet when the REST fetch resolves, RESTORE overwrites the store; subsequent replay events could then create duplicate messages. The double-check (`current.messages.length === 0`) mitigates this well in practice, but the window exists if the REST API responds faster than the WS handshake + replay. The tests cover the reverse race (live data before fetch resolve) but not this direction.

packages/client/src/connection.ts
```diff
@@ -221,6 +234,9 @@ export class MitzoConnection {
     this.boundOnVisibility = () => {
       if (document.visibilityState === 'visible') {
         this.checkAndReconnect();
+        // Notify the store so it can re-hydrate messages if the page was
```
dimakis, line 237: 🔵 **style**: `_foreground` is emitted on every `visibilitychange` to 'visible', even when the WS was never disconnected. This is correctly handled by the store's emptiness check, but a brief comment noting this intentional design (iOS bfcache eviction may preserve connection but lose JS state) would clarify why it's unconditional. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid fix for a real iOS reconnect problem. The `defuseOldWs` approach and sessionId-based fan-out are well-designed. Main gap is missing unit tests for the new `hasOpenWatchers` method, and a theoretical race between REST recovery and WS replay that's mitigated but not fully covered by tests._

- 🟡 **missing_tests** (L90): `hasOpenWatchers` is a new public method used in 3 critical delivery paths in `query-loop.ts`, but it has no unit tests in `packages/harness/__tests__/connection-registry.test.ts`. Sibling methods like `getConnectionsWatching` and `broadcast` are tested. Should cover: returns true with an open watcher, false with no watchers, and false when all watchers are closed. `[fixable]`


### `server/query-loop.ts`
_Solid fix for a real iOS reconnect problem. The `defuseOldWs` approach and sessionId-based fan-out are well-designed. Main gap is missing unit tests for the new `hasOpenWatchers` method, and a theoretical race between REST recovery and WS replay that's mitigated but not fully covered by tests._

- 🟡 **bugs** (L88): Before `sessionId` is resolved (early in the query loop), `sendOrBuffer` is called with `sessionId = undefined`. The old code's v2 guard (`connRegistry?.get(clientId)`) would deliver early events (like `session_id`) via the v2 path if the clientId was registered. The new guard (`sessionId && connRegistry?.hasOpenWatchers(sessionId)`) correctly skips this when sessionId is undefined and falls through to the v1 path. However, after session resolution, if the WS has already reconnected and the old clientId is gone from both the connRegistry AND the session is detached (`isAttached` returns false), early events emitted before reattach would be silently dropped (v2 guard fails, v1 guard fails). The event store persistence at line 76-81 still fires though, so replay on reconnect would recover them. Low risk but worth verifying that `handleReconnect` reattach happens before events are emitted in this window.
- 🔵 **style** (L405): The 7-line active-session cleanup loop (iterate watchers, check `activeSession`, call `setActive(cid, null)`) is duplicated verbatim at lines 405-414 and 688-697. Consider extracting a `clearActiveWatchers(connRegistry, sessionId)` helper to reduce duplication. `[fixable]`


_Solid fix for a real iOS reconnect problem. The `defuseOldWs` approach and sessionId-based fan-out are well-designed. Main gap is missing unit tests for the new `hasOpenWatchers` method, and a theoretical race between REST recovery and WS replay that's mitigated but not fully covered by tests._

- 🔵 **unsafe_assumptions** (L558): The `_foreground` recovery fetch races with WS reconnect replay: `checkAndReconnect()` and `_foreground` fire back-to-back in the visibility handler. If replay events from `handleReconnect` haven't arrived yet when the REST fetch resolves, RESTORE overwrites the store; subsequent replay events could then create duplicate messages. The double-check (`current.messages.length === 0`) mitigates this well in practice, but the window exists if the REST API responds faster than the WS handshake + replay. The tests cover the reverse race (live data before fetch resolve) but not this direction.


_Solid fix for a real iOS reconnect problem. The `defuseOldWs` approach and sessionId-based fan-out are well-designed. Main gap is missing unit tests for the new `hasOpenWatchers` method, and a theoretical race between REST recovery and WS replay that's mitigated but not fully covered by tests._

- 🔵 **style** (L237): `_foreground` is emitted on every `visibilitychange` to 'visible', even when the WS was never disconnected. This is correctly handled by the store's emptiness check, but a brief comment noting this intentional design (iOS bfcache eviction may preserve connection but lose JS state) would clarify why it's unconditional. `[fixable]`

PR #243, opened 2026-04-19
feat: draft prompt persistence + fast session loading

dimakis:
## Summary
- **Draft prompt persistence** — ChatInput drafts are saved to localStorage per session (debounced 400ms). Navigate away, come back, draft is still there. Cleared on send/interrupt.
- **Fast session loading** — Home Screen now serves sessions from EventStore (SQLite, <1ms) instead of filesystem scanning on every load. Background reconciliation keeps the cache fresh. `?full=1` falls back to filesystem scan.
- **Timestamp reconciliation** — EventStore upsertSession now preserves original timestamps. Startup reconciliation syncs filesystem timestamps eagerly so first page load is accurate.
- **Filter non-Mitzo sessions** — Sessions with 0 turns/prompts (e.g. automated code review sessions) are hidden from the session list.

## Test plan
- [x] `useDraft` hook: 8 tests covering restore, save, clear, migration, undefined sessionId
- [x] Existing ChatInput tests: 25 tests pass (no regressions)
- [x] Route tests: 66 tests pass (updated for cached path + new `?full=1` test)
- [x] Full suite: 1945/1947 pass (2 pre-existing flaky timeouts)
- [x] Manual test on mobile via Tailscale — sessions load with correct timestamps, drafts persist across navigation

### DIFF
--- a/frontend/src/hooks/useDraft.ts
+++ b/frontend/src/hooks/useDraft.ts
@@ -1,4 +1,11 @@
-import { type Dispatch, type SetStateAction, useState, useEffect, useRef, useCallback } from 'react';
+import {
+  type Dispatch,
+  type SetStateAction,
+  useState,
+  useEffect,
+  useRef,
+  useCallback,
+} from 'react';
 
 const KEY_PREFIX = 'mitzo-draft-';
 const DEBOUNCE_MS = 400;
--- a/frontend/src/hooks/useDraft.ts
+++ b/frontend/src/hooks/useDraft.ts
@@ -14,10 +14,7 @@ function draftKey(sessionId: string | undefined): string {
   return `${KEY_PREFIX}${sessionId ?? 'new'}`;
 }
 
-/**
- * Persists draft prompt text to localStorage per session.
- * Returns [text, setText, clearDraft].
- */
+/** Persists draft prompt text to localStorage per session. */
 export function useDraft(
   sessionId: string | undefined,
   initialText?: string,
@@ -33,6 +30,7 @@ export function useDraft(
 
   const timerRef = useRef<ReturnType<typeof setTimeout>>(undefined);
   const sessionRef = useRef(sessionId);
+  const mountedRef = useRef(false);
 
   // When sessionId changes (e.g. new session gets assigned an ID),
   // migrate draft from old key and load any existing draft for new key.
@@ -63,8 +61,12 @@ export function useDraft(
     }
   }, [sessionId]);
 
-  // Debounced save to localStorage on text change
+  // Debounced save to localStorage on text change (skip initial render)
   useEffect(() => {
+    if (!mountedRef.current) {
+      mountedRef.current = true;
+      return;
+    }
     clearTimeout(timerRef.current);
     timerRef.current = setTimeout(() => {
       try {
--- a/frontend/src/hooks/useDraft.ts
+++ b/frontend/src/hooks/useDraft.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef, useCallback } from 'react';
+import { type Dispatch, type SetStateAction, useState, useEffect, useRef, useCallback } from 'react';
 
 const KEY_PREFIX = 'mitzo-draft-';
 const DEBOUNCE_MS = 400;
@@ -14,7 +14,7 @@ function draftKey(sessionId: string | undefined): string {
 export function useDraft(
   sessionId: string | undefined,
   initialText?: string,
-): [string, (val: string) => void, () => void] {
+): [string, Dispatch<SetStateAction<string>>, () => void] {
   const [text, setTextRaw] = useState(() => {
     if (initialText) return initialText;
     try {
@@ -24,7 +24,7 @@ export function useDraft(
     }
   });
 
-  const timerRef = useRef<ReturnType<typeof setTimeout>>();
+  const timerRef = useRef<ReturnType<typeof setTimeout>>(undefined);
   const sessionRef = useRef(sessionId);
 
   // When sessionId changes (e.g. new session gets assigned an ID),
--- a/frontend/src/hooks/useDraft.ts
+++ b/frontend/src/hooks/useDraft.ts
@@ -1,0 +1,88 @@
+import { useState, useEffect, useRef, useCallback } from 'react';
+
+const KEY_PREFIX = 'mitzo-draft-';
+const DEBOUNCE_MS = 400;
+
+function draftKey(sessionId: string | undefined): string {
+  return `${KEY_PREFIX}${sessionId ?? 'new'}`;
+}
+
+/**
+ * Persists draft prompt text to localStorage per session.
+ * Returns [text, setText, clearDraft].
+ */
+export function useDraft(
+  sessionId: string | undefined,
+  initialText?: string,
+): [string, (val: string) => void, () => void] {
+  const [text, setTextRaw] = useState(() => {
+    if (initialText) return initialText;
+    try {
+      return localStorage.getItem(draftKey(sessionId)) ?? '';
+    } catch {
+      return '';
+    }
+  });
+
+  const timerRef = useRef<ReturnType<typeof setTimeout>>();
+  const sessionRef = useRef(sessionId);
+
+  // When sessionId changes (e.g. new session gets assigned an ID),
+  // migrate draft from old key and load any existing draft for new key.
+  useEffect(() => {
+    const prev = sessionRef.current;
+    sessionRef.current = sessionId;
+    if (prev === sessionId) return;
+
+    // If we had a draft under the old key, migrate it
+    const oldKey = draftKey(prev);
+    const newKey = draftKey(sessionId);
+    try {
+      const existing = localStorage.getItem(newKey);
+      if (existing) {
+        // New session already has a draft — use it
+        setTextRaw(existing);
+      } else {
+        // Migrate from old key
+        const old = localStorage.getItem(oldKey);
+        if (old) {
+          localStorage.setItem(newKey, old);
+          // Don't change text state — it's the same draft, just moved
+        }
+      }
+      localStorage.removeItem(oldKey);
+    } catch {
+      // localStorage unavailable — ignore
+    }
+  }, [sessionId]);
+
+  // Debounced save to localStorage on text change
+  useEffect(() => {
+    clearTimeout(timerRef.current);
+    timerRef.current = setTimeout(() => {
+      try {
+        const key = draftKey(sessionRef.current);
+        if (text) {
+          localStorage.setItem(key, text);
+        } else {
+          localStorage.removeItem(key);
+        }
+      } catch {
+        // localStorage full or unavailable — ignore
+      }
+    }, DEBOUNCE_MS);
+
+    return () => clearTimeout(timerRef.current);
+  }, [text]);
+
+  const clearDraft = useCallback(() => {
+    setTextRaw('');
+    try {
+      localStorage.removeItem(draftKey(sessionRef.current));
+    } catch {
+      // ignore
+    }
+  }, []);
+
+  return [text, setTextRaw, clearDraft];
+}
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -265,7 +265,17 @@ export class EventStore {
         ...values,
       );
     } else {
-      const cols = ['session_id', 'summary', 'branch', 'cwd', 'mode', 'is_active', 'initial_prompt', 'wt_id', 'goal_id'];
+      const cols = [
+        'session_id',
+        'summary',
+        'branch',
+        'cwd',
+        'mode',
+        'is_active',
+        'initial_prompt',
+        'wt_id',
+        'goal_id',
+      ];
       const vals: unknown[] = [
         meta.sessionId,
         meta.summary ?? null,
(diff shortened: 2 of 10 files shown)


server/app.ts
```diff
@@ -703,28 +705,59 @@ app.get('/api/sessions/active', (_req, res) => {
...
+    });
+    res.json({ sessions: annotated, hasMore });
+
+    // Background: reconcile with filesystem so EventStore stays fresh
+    if (offset === 0) reconcileSessionsBackground();
+    return;
+  }
+
+  // Full path: filesystem scan (original behavior, for ?full=1)
+  getSessions(offset, limit).then(({ sessions, hasMore }) => {
```
dimakis: 🟡 **bugs**: The `getSessions(offset, limit).then(...)` in the `?full=1` path has no `.catch()` handler. If `getSessions` rejects, the error is swallowed and the HTTP response hangs until client timeout. Add `.catch()` to send a 500 response. `[fixable]`

server/app.ts
```diff
@@ -703,28 +705,59 @@ app.get('/api/sessions/active', (_req, res) => {
   res.json(registry.getActiveSessions());
 });
 
-app.get('/api/sessions', async (req, res) => {
+app.get('/api/sessions', (req, res) => {
```
dimakis: 🔵 **style**: The route handler was changed from `async` to sync, but the `?full=1` branch uses `.then()` instead of `await`. This creates two code styles in one handler and the duplicated annotation logic (activeMap + annotated mapping) across both branches. Consider extracting the annotation logic into a shared helper. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -272,6 +286,18 @@ export class EventStore {
         meta.initialPrompt ?? null,
         meta.wtId ?? null,
         meta.goalId ?? null,
+      ];
+      if (meta.updatedAt !== undefined) {
```
dimakis, line 290: 🔵 **missing_tests**: The new explicit `updatedAt`/`createdAt` timestamp paths in `upsertSession` have no dedicated test coverage. Consider adding tests that verify: (1) explicit `updatedAt` on UPDATE is used instead of auto-generated, (2) explicit `createdAt` on INSERT is persisted, (3) omitting them still auto-generates. `[fixable]`

server/chat.ts
```diff
@@ -979,6 +981,110 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
+}
+
+/**
+ * Background reconciliation: scan filesystem for sessions the EventStore
+ * doesn't know about and sync timestamps from the filesystem.
+ * Call fire-and-forget after serving cached results.
+ */
+let _reconciling = false;
+export function reconcileSessionsBackground(): void {
+  if (_reconciling) return;
```
dimakis, line 1023: 🔵 **unsafe_assumptions**: The `_reconciling` module-level flag prevents concurrent reconciliation, but if `syncSessionTimestamps` throws synchronously before entering the async path, `_reconciling` would still be reset via `finally`. However, if the server is under heavy load and the first page load triggers reconciliation, subsequent page-1 loads within the reconciliation window silently skip sync. This is fine for correctness but worth noting — stale data could persist for the duration of a slow filesystem scan.

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Solid feature PR. The missing `.catch()` on the `?full=1` promise path in app.ts is the main actionable issue — the rest is good with minor test coverage gaps._

- 🟡 **bugs** (L741): The `getSessions(offset, limit).then(...)` in the `?full=1` path has no `.catch()` handler. If `getSessions` rejects, the error is swallowed and the HTTP response hangs until client timeout. Add `.catch()` to send a 500 response. `[fixable]`
- 🔵 **style** (L708): The route handler was changed from `async` to sync, but the `?full=1` branch uses `.then()` instead of `await`. This creates two code styles in one handler and the duplicated annotation logic (activeMap + annotated mapping) across both branches. Consider extracting the annotation logic into a shared helper. `[fixable]`


### `frontend/src/hooks/__tests__/useDraft.test.ts`
_Solid feature PR. The missing `.catch()` on the `?full=1` promise path in app.ts is the main actionable issue — the rest is good with minor test coverage gaps._

- 🔵 **missing_tests**: No test for sessionId changing (e.g. from `undefined` to a real ID). This is the draft migration path in useDraft's useEffect — the test should verify that when sessionId changes, the draft migrates from the old localStorage key to the new one and the old key is removed. `[fixable]`


_Solid feature PR. The missing `.catch()` on the `?full=1` promise path in app.ts is the main actionable issue — the rest is good with minor test coverage gaps._

- 🔵 **missing_tests** (L290): The new explicit `updatedAt`/`createdAt` timestamp paths in `upsertSession` have no dedicated test coverage. Consider adding tests that verify: (1) explicit `updatedAt` on UPDATE is used instead of auto-generated, (2) explicit `createdAt` on INSERT is persisted, (3) omitting them still auto-generates. `[fixable]`


### `server/chat.ts`
_Solid feature PR. The missing `.catch()` on the `?full=1` promise path in app.ts is the main actionable issue — the rest is good with minor test coverage gaps._

- 🔵 **unsafe_assumptions** (L1017): The `_reconciling` module-level flag prevents concurrent reconciliation, but if `syncSessionTimestamps` throws synchronously before entering the async path, `_reconciling` would still be reset via `finally`. However, if the server is under heavy load and the first page load triggers reconciliation, subsequent page-1 loads within the reconciliation window silently skip sync. This is fine for correctness but worth noting — stale data could persist for the duration of a slow filesystem scan.

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Well-structured PR with clean separation between fast-path caching and background reconciliation. Main actionable item is `console.error` instead of `log.error` in app.ts; the rest are suggestions for test coverage and minor style._

- 🟡 **style** (L749): Uses `console.error` instead of the project's structured logger (`log.error`). Every other error path in app.ts uses `log.error` via the Pino logger. This will bypass log-level filtering, JSON output, and OTel trace context. `[fixable]`


### `server/chat.ts`
_Well-structured PR with clean separation between fast-path caching and background reconciliation. Main actionable item is `console.error` instead of `log.error` in app.ts; the rest are suggestions for test coverage and minor style._

- 🟡 **bugs** (L993): The `isActive` check in `getSessionsCached` reads from the EventStore's `is_active` column, which can be stale (only updated when a session ends or `markSessionInactive` is called). A brand-new session that was just created but hasn't had its first turn yet (numTurns=0, promptCount=0) could be filtered out if the DB's `is_active` flag hasn't been set yet by the time `listSessions()` runs. In practice this is a narrow race window since the session is created with `is_active=1`, but the comment says 'Active sessions always show' — worth verifying this matches intent.
- 🔵 **unsafe_assumptions** (L989): `getSessionsCached` calls `eventStore.listSessions()` with no limit, loading ALL non-hidden sessions into memory, then filters and slices in JS. For a personal tool this is fine, but if the session count grows large, pushing the filter and pagination into the SQL query would be more efficient. Not urgent — just flagging for future awareness. `[fixable]`
- 🔵 **missing_tests** (L1034): `syncSessionTimestamps` has no unit tests. Key behaviors worth testing: new session insertion, timestamp drift sync (>60s threshold), `manuallyRenamed` summary preservation, and the `hiddenSessionIds` filter. The `reconcileSessionsBackground` concurrency guard (`_reconciling` flag) is also untested. `[fixable]`
- 🔵 **missing_tests** (L988): `getSessionsCached` has no unit test. The filtering logic (numTurns=0 + promptCount=0 + !isActive → hidden) is meaningful business logic that should have a test to prevent regressions. `[fixable]`


### `frontend/src/hooks/useDraft.ts`
_Well-structured PR with clean separation between fast-path caching and background reconciliation. Main actionable item is `console.error` instead of `log.error` in app.ts; the rest are suggestions for test coverage and minor style._

- 🔵 **style** (L21): Multi-line JSDoc comment (`/** ... */`) on the hook — project convention per CLAUDE.md is to default to no comments or one short line max. The behavior is self-documenting from the return type. `[fixable]`
- 🔵 **bugs** (L70): The debounced-save effect fires on every `text` change, including the initial render. When a draft is restored from localStorage, this writes the same value back after the debounce — a harmless no-op but unnecessary I/O. Could skip the first render with an `isMounted` ref. `[fixable]`

server/app.ts
```diff
@@ -706,25 +708,47 @@ app.get('/api/sessions/active', (_req, res) => {
...
+      // Background: reconcile with filesystem so EventStore stays fresh
+      if (offset === 0) reconcileSessionsBackground();
+      return;
+    }
+
+    // Full path: filesystem scan (original behavior, for ?full=1)
+    const { sessions, hasMore } = await getSessions(offset, limit);
+    res.json({ sessions: annotate(sessions), hasMore });
+  } catch (err) {
+    console.error('GET /api/sessions failed:', err);
```
dimakis: 🟡 **style**: Uses `console.error` instead of the project's structured logger (`log.error`). Every other error path in app.ts uses `log.error` via the Pino logger. This will bypass log-level filtering, JSON output, and OTel trace context. `[fixable]`

server/chat.ts
```diff
@@ -979,6 +981,114 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
+/**
+ * Fast session listing from EventStore (SQLite) — no filesystem scan.
+ * Returns the same shape as getSessions() for API compatibility.
+ */
+export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
+  const all = eventStore.listSessions().filter((m) => {
+    // Hide sessions that were never used through Mitzo (e.g. automated
+    // code review sessions discovered from filesystem).  Active sessions
+    // always show regardless of turn count.
+    if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) return false;
```
dimakis, line 995: 🟡 **bugs**: The `isActive` check in `getSessionsCached` reads from the EventStore's `is_active` column, which can be stale (only updated when a session ends or `markSessionInactive` is called). A brand-new session that was just created but hasn't had its first turn yet (numTurns=0, promptCount=0) could be filtered out if the DB's `is_active` flag hasn't been set yet by the time `listSessions()` runs. In practice this is a narrow race window since the session is created with `is_active=1`, but the comment says 'Active sessions always show' — worth verifying this matches intent.

server/chat.ts
```diff
@@ -979,6 +981,114 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
   return { sessions: page, hasMore };
 }
 
+/**
+ * Fast session listing from EventStore (SQLite) — no filesystem scan.
+ * Returns the same shape as getSessions() for API compatibility.
+ */
+export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
+  const all = eventStore.listSessions().filter((m) => {
```
dimakis, line 989: 🔵 **unsafe_assumptions**: `getSessionsCached` calls `eventStore.listSessions()` with no limit, loading ALL non-hidden sessions into memory, then filters and slices in JS. For a personal tool this is fine, but if the session count grows large, pushing the filter and pagination into the SQL query would be more efficient. Not urgent — just flagging for future awareness. `[fixable]`

server/chat.ts
```diff
@@ -979,6 +981,114 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
...
+    .finally(() => {
+      _reconciling = false;
+    });
+}
+
+/**
+ * Full timestamp sync: scan filesystem and update EventStore timestamps
+ * for all sessions to match their actual lastModified time.
+ */
+async function syncSessionTimestamps(): Promise<void> {
```
dimakis: 🔵 **missing_tests**: `syncSessionTimestamps` has no unit tests. Key behaviors worth testing: new session insertion, timestamp drift sync (>60s threshold), `manuallyRenamed` summary preservation, and the `hiddenSessionIds` filter. The `reconcileSessionsBackground` concurrency guard (`_reconciling` flag) is also untested. `[fixable]`

server/chat.ts
```diff
@@ -979,6 +981,114 @@ export async function getSessions(offset = 0, limit = SESSION_PAGE_SIZE) {
   return { sessions: page, hasMore };
 }
 
+/**
+ * Fast session listing from EventStore (SQLite) — no filesystem scan.
+ * Returns the same shape as getSessions() for API compatibility.
+ */
+export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
```
dimakis, line 988: 🔵 **missing_tests**: `getSessionsCached` has no unit test. The filtering logic (numTurns=0 + promptCount=0 + !isActive → hidden) is meaningful business logic that should have a test to prevent regressions. `[fixable]`

frontend/src/hooks/useDraft.ts
```diff
@@ -0,0 +1,95 @@
...
+
+function draftKey(sessionId: string | undefined): string {
+  return `${KEY_PREFIX}${sessionId ?? 'new'}`;
+}
+
+/**
+ * Persists draft prompt text to localStorage per session.
+ * Returns [text, setText, clearDraft].
+ */
+export function useDraft(
```
dimakis, line 18: 🔵 **style**: Multi-line JSDoc comment (`/** ... */`) on the hook — project convention per CLAUDE.md is to default to no comments or one short line max. The behavior is self-documenting from the return type. `[fixable]`

frontend/src/hooks/useDraft.ts
```diff
@@ -0,0 +1,95 @@
...
+    } catch {
+      // localStorage unavailable — ignore
+    }
+  }, [sessionId]);
+
+  // Debounced save to localStorage on text change
+  useEffect(() => {
+    clearTimeout(timerRef.current);
+    timerRef.current = setTimeout(() => {
+      try {
```
dimakis, line 72: 🔵 **bugs**: The debounced-save effect fires on every `text` change, including the initial render. When a draft is restored from localStorage, this writes the same value back after the debounce — a harmless no-op but unnecessary I/O. Could skip the first render with an `isMounted` ref. `[fixable]`

PR #246, opened 2026-04-19
fix(mobile): unified layout + cherry-picked keeper features from PR #242

dimakis:
## Summary

- Cherry-picked 8 keeper commits from PR #242 (copy buttons, keyboard dismiss, camera permissions, header collapse, direction-locked swipe)
- Dropped 8 CSS safe-area hack commits that were fighting each other
- Implemented a single consistent layout pattern across all pages: fixed container → header with safe-area padding → scroll wrapper
- `contentInset: 'never'` — CSS owns all safe-area spacing, no double-application
- Chat: scroll-through-messages no longer dismisses keyboard; keyboard open scrolls to bottom
- Logo bumped to 28x28 for better touch target

Measured `env(safe-area-inset-top) = 62px` on iPhone 17 Pro via Safari Web Inspector.

## Test plan

- [ ] Session list: header below Dynamic Island, scrolling doesn't shift header
- [ ] Chat: header background extends into status bar, keyboard dismiss only on tap (not scroll)
- [ ] Inbox: filter pills visible, cards scroll independently of header
- [ ] Todos: category pills visible, swipe works, header stays pinned
- [ ] Todo detail: scrolls content, header pinned
- [ ] Task board: same pattern
- [ ] Calendar: header with safe area, body scrolls
- [ ] File viewer: header extends into status bar
- [ ] All pages: no rubber-band gap at top on overscroll


dimakis:
## Centaur Review

### `frontend/.env.ios`
_Solid layout unification (fixed positioning + explicit safe-area handling), well-tested copy feature, and good keyboard management — but `.env.ios` is tracked despite gitignore, `onEdit` is dead code, and the chat header transition won't animate due to conditional rendering._

- 🟡 **unsafe_assumptions** (L1): This file contains a machine-specific hostname and appears tracked in git (present in the diff) despite the .gitignore entry. Once force-added, .gitignore stops protecting it. The project convention explicitly forbids hardcoded machine-specific configuration. Run `git rm --cached frontend/.env.ios` to untrack it and rely on a `.env.ios.example` template instead. `[fixable]`


### `frontend/ios/App/App.xcodeproj/project.pbxproj`
_Solid layout unification (fixed positioning + explicit safe-area handling), well-tested copy feature, and good keyboard management — but `.env.ios` is tracked despite gitignore, `onEdit` is dead code, and the chat header transition won't animate due to conditional rendering._

- 🟡 **unsafe_assumptions** (L303): DEVELOPMENT_TEAM changed from build variable `$(DEVELOPMENT_TEAM)` to a hardcoded team ID. While `local.xcconfig` can override this, the fallback is now a specific developer's team ID instead of a build-time variable, reducing portability for other contributors. Consider keeping `$(DEVELOPMENT_TEAM)` and setting the value only in `local.xcconfig`. `[fixable]`


_Solid layout unification (fixed positioning + explicit safe-area handling), well-tested copy feature, and good keyboard management — but `.env.ios` is tracked despite gitignore, `onEdit` is dead code, and the chat header transition won't animate due to conditional rendering._

- 🟡 **bugs** (L18): The `useEffect` at line 18 references `scrollRef` (line 69) which is declared 50 lines later. This works at runtime because the callback only executes after mount (when the ref is initialized), but the variable is in the temporal dead zone when the effect is registered. Moving this `useEffect` block below the `scrollRef` declaration (after line 69) would make the dependency clear and prevent confusion. `[fixable]`
- 🔵 **bugs** (L200): The chat header CSS declares `transition: max-height 0.2s, padding 0.2s, opacity 0.2s` but the mode pills and VoiceSettings are toggled via React conditional rendering (`{!keyboardOpen && ...}`), causing instant mount/unmount rather than animated transitions. To get smooth animation, use CSS-based hiding (e.g., `visibility`/`max-height: 0`/`overflow: hidden`) instead of conditional rendering. `[fixable]`


_Solid layout unification (fixed positioning + explicit safe-area handling), well-tested copy feature, and good keyboard management — but `.env.ios` is tracked despite gitignore, `onEdit` is dead code, and the chat header transition won't animate due to conditional rendering._

- 🟡 **regressions** (L15): The `onEdit` prop is added to `UserBubble` (with click handler, role='button', tabIndex, and CSS styling) but is never passed from `ChatArea` or any other consumer. This is dead code shipped to production — the editable styling and accessibility attributes are defined but never activated. Either wire it up or remove it from this PR. `[fixable]`
- 🔵 **missing_tests** (L56): The new collapsible message behavior (`isLong`/`collapsed` state, COLLAPSE_HEIGHT threshold, 'Show more'/'Show less' toggle) has no test coverage. Key scenarios: messages under the threshold shouldn't show the toggle, collapsed messages should limit height, clicking the toggle should expand/collapse, and streaming messages should never collapse. `[fixable]`


### `frontend/src/App.tsx`
_Solid layout unification (fixed positioning + explicit safe-area handling), well-tested copy feature, and good keyboard management — but `.env.ios` is tracked despite gitignore, `onEdit` is dead code, and the chat header transition won't animate due to conditional rendering._

- 🔵 **style** (L44): The `dismissKeyboard` onClick handler on the root div partially overlaps with the touch-based keyboard dismiss in `ChatArea` (handleTouchStart/handleTouchEnd). On mobile, a tap triggers both handlers — harmless since blur is idempotent, but the redundancy adds complexity. Consider documenting that the App-level handler covers non-chat pages while ChatArea's is chat-specific.

frontend/src/pages/ChatView.tsx
```diff
@@ -16,7 +16,18 @@ import type { ImageAttachment } from '../types/chat';
...
   const [keyboardOpen, setKeyboardOpen] = useState(false);
 
-  useEffect(() => onKeyboardToggle(setKeyboardOpen), []);
+  useEffect(
+    () =>
+      onKeyboardToggle((visible) => {
+        setKeyboardOpen(visible);
+        if (visible) {
+          requestAnimationFrame(() => {
+            scrollRef.current?.scrollTo({ top: scrollRef.current!.scrollHeight });
```
dimakis, line 24: 🟡 **bugs**: Non-null assertion `scrollRef.current!.scrollHeight` is unsafe. The optional chain on `scrollTo` guards against null, but if `scrollRef.current` is null, the `!` assertion on the same expression for `scrollHeight` will throw a TypeError. Use optional chaining consistently: `scrollRef.current?.scrollTo({ top: scrollRef.current?.scrollHeight ?? 0 })`. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -16,7 +16,18 @@ import type { ImageAttachment } from '../types/chat';
 export function ChatView() {
   const [keyboardOpen, setKeyboardOpen] = useState(false);
 
-  useEffect(() => onKeyboardToggle(setKeyboardOpen), []);
+  useEffect(
```
dimakis, line 18: 🟡 **bugs**: `scrollRef` (defined at line 70) is referenced in the `useEffect` callback at line 25 but declared after it. This works at runtime because the callback executes asynchronously and refs are stable, but ESLint's `react-hooks/exhaustive-deps` may flag `scrollRef` as missing from the deps array. More importantly, the variable hoisting makes the code harder to follow — consider moving `scrollRef` above this effect. `[fixable]`

frontend/src/components/MessageBubble.tsx
```diff
@@ -14,14 +14,20 @@ interface UserBubbleProps {
...
+export function UserBubble({ text, images, contextBlocks, onEdit, timestamp }: UserBubbleProps) {
   const time = formatTime(timestamp);
   return (
     <div className="msg-bubble-group msg-bubble-group--user">
-      <div className="msg-bubble msg-bubble--user">
+      <div
+        className={`msg-bubble msg-bubble--user${onEdit ? ' msg-bubble--editable' : ''}`}
+        onClick={() => text && onEdit?.(text)}
+        role={onEdit ? 'button' : undefined}
+        tabIndex={onEdit ? 0 : undefined}
```
dimakis, line 29: 🔵 **regressions**: `role="button"` with `tabIndex={0}` but no `onKeyDown` handler means keyboard users (and screen readers) see a focusable button they cannot activate with Enter or Space. Add `onKeyDown={(e) => { if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); text && onEdit?.(text); } }}` for a11y compliance. `[fixable]`

frontend/src/components/MessageBubble.tsx
```diff
@@ -14,14 +14,20 @@ interface UserBubbleProps {
   text?: string;
   images?: string[];
   contextBlocks?: string[];
+  onEdit?: (text: string) => void;
```
dimakis: 🔵 **style**: `onEdit` prop is added to `UserBubble` but never passed by any consumer (`ChatArea.tsx` renders `UserBubble` without it). This is dead scaffolding. If it's intentional prep for a follow-up PR, that's fine — but it should be wired up or removed to avoid confusion.

frontend/src/App.tsx
```diff
@@ -41,7 +41,7 @@ function ChatRoute() {
   return isDesktop ? <DesktopChatView /> : <ChatView />;
 }
 
-function dismissKeyboard(e: React.MouseEvent) {
+function dismissKeyboard(e: React.MouseEvent | React.TouchEvent) {
```
dimakis, line 44: 🔵 **style**: `dismissKeyboard` signature was widened to accept `React.TouchEvent` but it is still only attached to `onClickCapture` (not any touch handler). The type widening is dead code — either attach it to a touch event or revert the signature change. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid mobile layout unification — fixed-position pattern is consistent with codebase conventions. Main concerns are a non-null assertion that can throw and a missing keyboard handler on the new editable bubble._

- 🟡 **bugs** (L25): Non-null assertion `scrollRef.current!.scrollHeight` is unsafe. The optional chain on `scrollTo` guards against null, but if `scrollRef.current` is null, the `!` assertion on the same expression for `scrollHeight` will throw a TypeError. Use optional chaining consistently: `scrollRef.current?.scrollTo({ top: scrollRef.current?.scrollHeight ?? 0 })`. `[fixable]`
- 🟡 **bugs** (L19): `scrollRef` (defined at line 70) is referenced in the `useEffect` callback at line 25 but declared after it. This works at runtime because the callback executes asynchronously and refs are stable, but ESLint's `react-hooks/exhaustive-deps` may flag `scrollRef` as missing from the deps array. More importantly, the variable hoisting makes the code harder to follow — consider moving `scrollRef` above this effect. `[fixable]`


_Solid mobile layout unification — fixed-position pattern is consistent with codebase conventions. Main concerns are a non-null assertion that can throw and a missing keyboard handler on the new editable bubble._

- 🔵 **regressions** (L29): `role="button"` with `tabIndex={0}` but no `onKeyDown` handler means keyboard users (and screen readers) see a focusable button they cannot activate with Enter or Space. Add `onKeyDown={(e) => { if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); text && onEdit?.(text); } }}` for a11y compliance. `[fixable]`
- 🔵 **style** (L17): `onEdit` prop is added to `UserBubble` but never passed by any consumer (`ChatArea.tsx` renders `UserBubble` without it). This is dead scaffolding. If it's intentional prep for a follow-up PR, that's fine — but it should be wired up or removed to avoid confusion.


### `frontend/src/App.tsx`
_Solid mobile layout unification — fixed-position pattern is consistent with codebase conventions. Main concerns are a non-null assertion that can throw and a missing keyboard handler on the new editable bubble._

- 🔵 **style** (L44): `dismissKeyboard` signature was widened to accept `React.TouchEvent` but it is still only attached to `onClickCapture` (not any touch handler). The type widening is dead code — either attach it to a touch event or revert the signature change. `[fixable]`

frontend/.env.ios
```diff
@@ -0,0 +1,2 @@
+VITE_API_BASE_URL=https://dimitris-macbook-pro.taildfe858.ts.net:3100
```
dimakis: 🟡 **unsafe_assumptions**: This commits a machine-specific Tailnet hostname into the repo. CLAUDE.md explicitly forbids hardcoded machine-specific configuration, and other developers/CI will not be able to use this endpoint. `[fixable]`

frontend/src/pages/SessionList.tsx
```diff
@@ -281,117 +260,72 @@ export function SessionList() {
...
-          className="session-search-input"
-          type="text"
-          placeholder="Search sessions..."
-          value={search.query}
-          onChange={(e) => search.setQuery(e.target.value)}
-        />
-        {search.active && (
-          <button className="session-search-clear" onClick={search.clear}>
-            &times;
+      <div className="session-list-scroll">
```
dimakis: 🟡 **regressions**: This removes the session search UI and hook entirely from the mobile session list, which is a user-visible feature regression rather than a layout-only cleanup. The deleted search code and styles indicate the capability was intentionally removed, not relocated. `[fixable]`

frontend/src/pages/SessionList.tsx
```diff
@@ -1,13 +1,12 @@
 import { useState, useEffect, useRef, useCallback } from 'react';
```
dimakis, line 1: 🟡 **missing_tests**: This file removes the search flow and rewrites the main rendering path, but there are no accompanying tests covering the new behavior or guarding against accidental feature removal. Given the project’s TDD requirement, this UI regression risk should be covered.

frontend/src/pages/TaskBoard.tsx
```diff
@@ -28,7 +27,6 @@ export function TaskBoard() {
     refresh,
   } = useTaskBoard();
   const [creating, setCreating] = useState<{ parentId?: string } | null>(null);
-  const [creatingWorkflow, setCreatingWorkflow] = useState(false);
```
dimakis, line 32: 🟡 **regressions**: The workflow creation entry point was removed (`WorkflowCreateForm` import, state, button, and render path), so users can no longer create workflows from the task board. This is a functional regression unless the feature was removed everywhere else intentionally.

frontend/src/App.tsx
```diff
@@ -57,7 +57,7 @@ function dismissKeyboard(e: React.MouseEvent) {
 export function App() {
   return (
     <ErrorBoundary>
-      <div onClickCapture={dismissKeyboard} style={{ height: '100%' }}>
+      <div onClick={dismissKeyboard}>
```
dimakis: 🟡 **bugs**: Changing the global keyboard-dismiss handler from `onClickCapture` to bubbling `onClick` means child click handlers run before the blur. Components that depend on the input being blurred before their own click logic may now behave differently, and touch interactions are no longer handled here despite widening the event type to include `TouchEvent`. `[fixable]`

dimakis (commented):
## Codex Review

Found **5** issue(s) (5 warning).


### `frontend/.env.ios`
_PR improves mobile layout but introduces feature regressions and commits machine-specific iOS config._

- 🟡 **unsafe_assumptions** (L1): This commits a machine-specific Tailnet hostname into the repo. CLAUDE.md explicitly forbids hardcoded machine-specific configuration, and other developers/CI will not be able to use this endpoint. `[fixable]`


### `frontend/src/pages/SessionList.tsx`
_PR improves mobile layout but introduces feature regressions and commits machine-specific iOS config._

- 🟡 **regressions** (L263): This removes the session search UI and hook entirely from the mobile session list, which is a user-visible feature regression rather than a layout-only cleanup. The deleted search code and styles indicate the capability was intentionally removed, not relocated. `[fixable]`
- 🟡 **missing_tests** (L1): This file removes the search flow and rewrites the main rendering path, but there are no accompanying tests covering the new behavior or guarding against accidental feature removal. Given the project’s TDD requirement, this UI regression risk should be covered.


### `frontend/src/pages/TaskBoard.tsx`
_PR improves mobile layout but introduces feature regressions and commits machine-specific iOS config._

- 🟡 **regressions** (L30): The workflow creation entry point was removed (`WorkflowCreateForm` import, state, button, and render path), so users can no longer create workflows from the task board. This is a functional regression unless the feature was removed everywhere else intentionally.


### `frontend/src/App.tsx`
_PR improves mobile layout but introduces feature regressions and commits machine-specific iOS config._

- 🟡 **bugs** (L60): Changing the global keyboard-dismiss handler from `onClickCapture` to bubbling `onClick` means child click handlers run before the blur. Components that depend on the input being blurred before their own click logic may now behave differently, and touch interactions are no longer handled here despite widening the event type to include `TouchEvent`. `[fixable]`

dimakis (commented):
## Codex Review

_PR improves mobile layout, but it appears to remove still-used CSS and introduces untested keyboard/interaction regressions._

- 🟡 **regressions**: This PR deletes several still-used CSS selectors (`.session-search-bar`, `.session-search-input`, `.session-search-clear`, `.search-result-snippet`, `.msg-bubble-footer`, `.msg-bubble-footer--user`, `.msg-bubble-copy--user`, `.msg-bubble-group--user`, `.msg-timestamp`, `.msg-timestamp--user`, plus task-node/workflow classes). Those class names are still referenced by current components such as `SessionList` and `MessageBubble`, so the change silently strips existing UI styling/features rather than just cleaning up dead CSS. `[fixable]`


### `frontend/src/App.tsx`
_PR improves mobile layout, but it appears to remove still-used CSS and introduces untested keyboard/interaction regressions._

- 🟡 **bugs**: `dismissKeyboard` was widened to accept touch events, but the root element still only wires `onClickCapture`. On mobile Safari/Capacitor, taps that do not synthesize a click consistently will no longer dismiss the keyboard, while `ChatArea` now handles only that page. Other pages with inputs (login, file editor, etc.) lose the intended global behavior unless an `onTouchEndCapture`/equivalent is also attached. `[fixable]`


_PR improves mobile layout, but it appears to remove still-used CSS and introduces untested keyboard/interaction regressions._

- 🟡 **missing_tests**: The keyboard behavior changed materially: `onKeyboardToggle` now triggers an automatic scroll-to-bottom on show, `ScrollFab` was removed, and touch-based blur logic was added in `ChatArea`, but no frontend interaction tests were added/updated to cover these regressions. Existing keyboard tests only cover `configureKeyboard`, so the new mobile-specific behavior is effectively untested. `[fixable]`

PR #249, opened 2026-04-19
chore: add gitleaks pre-commit hook for secret scanning

dimakis:
## Summary
- Add gitleaks to pre-commit hook — scans staged changes for secrets before every commit
- Add `.gitleaks.toml` config with allowlists for package-lock.json and worktree paths

Prevents the kind of accidental exposure we just had to BFG-scrub (Tailscale hostname, Apple DEVELOPMENT_TEAM).

## Test plan
- [x] Verified gitleaks runs on commit and reports "no leaks found"

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -1,0 +1,7 @@
+title = "Mitzo gitleaks config"
+
+[allowlist]
+paths = [
+  '''package-lock\.json''',
+  '''\.claude/worktrees/''',
+]
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -4,4 +4,5 @@ title = "Mitzo gitleaks config"
 paths = [
   '''package-lock\.json''',
   '''\.claude/worktrees/''',
+  '''\.cursor/worktrees/''',
 ]
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -6,9 +6,11 @@ fi
 
 npx lint-staged
 
-# Scan staged changes for secrets (skip if gitleaks not installed)
+# Optional: scan staged changes for secrets with gitleaks.
+# gitleaks is not an npm dependency — install it separately if you want
+# pre-commit secret scanning. See README § Development for details.
 if command -v gitleaks >/dev/null 2>&1; then
   gitleaks git --pre-commit --staged --verbose
 else
-  echo "⚠ gitleaks not installed — skipping secret scan (install: https://github.com/gitleaks/gitleaks#installing)"
+  echo "note: gitleaks not found — skipping secret scan (https://github.com/gitleaks/gitleaks#installing)"
 fi
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -6,5 +6,9 @@ fi
 
 npx lint-staged
 
-# Scan staged changes for secrets
-gitleaks git --pre-commit --staged --verbose
+# Scan staged changes for secrets (skip if gitleaks not installed)
+if command -v gitleaks >/dev/null 2>&1; then
+  gitleaks git --pre-commit --staged --verbose
+else
+  echo "WARNING: gitleaks not found — skipping secret scan. Install: brew install gitleaks"
+fi
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -5,3 +5,6 @@ if [ "$branch" = "main" ]; then
 fi
 
 npx lint-staged
+
+# Scan staged changes for secrets
+gitleaks git --pre-commit --staged --verbose
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -10,5 +10,5 @@ npx lint-staged
 if command -v gitleaks >/dev/null 2>&1; then
   gitleaks git --pre-commit --staged --verbose
 else
-  echo "WARNING: gitleaks not found — skipping secret scan. Install: brew install gitleaks"
+  echo "⚠ gitleaks not installed — skipping secret scan (install: https://github.com/gitleaks/gitleaks#installing)"
 fi
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -1547,19 +1547,20 @@ textarea:focus {
 .msg-bubble-footer {
   display: flex;
   align-items: center;
-  gap: var(--space-1);
+  gap: var(--space-2);
   margin-top: var(--space-1);
   opacity: 0;
   transition: opacity 0.15s;
-  justify-content: flex-end;
 }
 
-.msg-bubble--assistant:hover .msg-bubble-footer {
+.msg-bubble--assistant:hover .msg-bubble-footer,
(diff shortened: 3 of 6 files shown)


.husky/pre-commit
```diff
@@ -5,3 +5,6 @@ if [ "$branch" = "main" ]; then
 fi
 
 npx lint-staged
+
+# Scan staged changes for secrets
+gitleaks git --pre-commit --staged --verbose
```
dimakis: 🟡 **regressions**: gitleaks is invoked as a bare command, but it is not an npm dependency or otherwise installed by the project. Contributors who don't have gitleaks on their PATH will get a hard failure on every commit. Guard with a `command -v gitleaks` check and either skip with a warning or document the install requirement. `[fixable]`

.gitleaks.toml
```diff
@@ -0,0 +1,7 @@
+title = "Mitzo gitleaks config"
+
+[allowlist]
+paths = [
+  '''package-lock\.json''',
+  '''\.claude/worktrees/''',
```
dimakis, line 6: 🔵 **unsafe_assumptions**: The allowlist covers `.claude/worktrees/` but the codebase also uses `.cursor/worktrees/` (see worktree-guard.ts and cleanup logic). If worktree content under `.cursor/worktrees/` ever gets staged, gitleaks could flag false positives. Consider adding `'''\.cursor/worktrees/'''` to the allowlist for consistency. `[fixable]`

dimakis (commented):
## Centaur Review

### `.husky/pre-commit`
_Useful secret-scanning addition, but gitleaks is assumed installed without a guard — contributors without it will have broken commits._

- 🟡 **regressions** (L10): gitleaks is invoked as a bare command, but it is not an npm dependency or otherwise installed by the project. Contributors who don't have gitleaks on their PATH will get a hard failure on every commit. Guard with a `command -v gitleaks` check and either skip with a warning or document the install requirement. `[fixable]`


### `.gitleaks.toml`
_Useful secret-scanning addition, but gitleaks is assumed installed without a guard — contributors without it will have broken commits._

- 🔵 **unsafe_assumptions** (L6): The allowlist covers `.claude/worktrees/` but the codebase also uses `.cursor/worktrees/` (see worktree-guard.ts and cleanup logic). If worktree content under `.cursor/worktrees/` ever gets staged, gitleaks could flag false positives. Consider adding `'''\.cursor/worktrees/'''` to the allowlist for consistency. `[fixable]`


### `.github/workflows/ci.yml`
_Useful secret-scanning addition, but gitleaks is assumed installed without a guard — contributors without it will have broken commits._

- 🔵 **regressions**: The pre-commit hook now runs gitleaks, but CI does not. A contributor could skip the hook (--no-verify) and push secrets. Consider adding a gitleaks step to the CI workflow for defense-in-depth, consistent with the enterprise-plan.md recommendation for secret detection on every push. `[fixable]`

dimakis:
## Centaur Review

_Failing test due to removed `msg-timestamp--user` class, dead CSS, and gitleaks hook needs an install guard — the UI refactor should also be a separate commit._

- 🔴 **regressions** (L145): Test asserts `msg-timestamp--user` class exists in UserBubble output, but this class was removed in the PR. The `UserBubble timestamp` → `renders timestamp when provided` test will fail. The assertion should be updated to check for `msg-bubble-footer` instead. `[fixable]`


_Failing test due to removed `msg-timestamp--user` class, dead CSS, and gitleaks hook needs an install guard — the UI refactor should also be a separate commit._

- 🔵 **style** (L1580): The `.msg-timestamp--user` CSS rule (lines 1580-1584) is now dead code — no component renders this class after the UserBubble refactor. Should be removed. `[fixable]`


### `.husky/pre-commit`
_Failing test due to removed `msg-timestamp--user` class, dead CSS, and gitleaks hook needs an install guard — the UI refactor should also be a separate commit._

- 🟡 **unsafe_assumptions** (L10): The hook calls `gitleaks` without checking if it's installed. Contributors without gitleaks will get a hard failure on every commit. Consider guarding with `command -v gitleaks >/dev/null 2>&1 && gitleaks git --pre-commit --staged --verbose` or adding installation instructions. `[fixable]`
- 🔵 **style**: PR title is 'chore: add gitleaks pre-commit hook' but the diff also includes a MessageBubble UI refactor (copy button + footer layout changes). These are unrelated changes that should be in separate commits for clean history and easier bisecting.

frontend/src/styles/global.css
```diff
@@ -1547,19 +1547,20 @@ textarea:focus {
...
-  gap: var(--space-1);
+  gap: var(--space-2);
   margin-top: var(--space-1);
   opacity: 0;
   transition: opacity 0.15s;
-  justify-content: flex-end;
 }
 
-.msg-bubble--assistant:hover .msg-bubble-footer {
+.msg-bubble--assistant:hover .msg-bubble-footer,
```
dimakis, line 1612: 🟡 **regressions**: The footer is now revealed only on `:hover` for both assistant and user bubbles. On touch devices (a primary target for this mobile-first app), hover is unavailable, so the new copy button/timestamp footer in `UserBubble` and the existing assistant footer can remain effectively inaccessible/invisible. Consider making the footer always visible on coarse pointers or exposing it via a non-hover interaction. `[fixable]`

dimakis (commented):
## Codex Review

_Mostly fine, but the new hover-only footer behavior introduces a mobile usability regression for message actions._

- 🟡 **regressions** (L1556): The footer is now revealed only on `:hover` for both assistant and user bubbles. On touch devices (a primary target for this mobile-first app), hover is unavailable, so the new copy button/timestamp footer in `UserBubble` and the existing assistant footer can remain effectively inaccessible/invisible. Consider making the footer always visible on coarse pointers or exposing it via a non-hover interaction. `[fixable]`

.husky/pre-commit
```diff
@@ -5,3 +5,10 @@ if [ "$branch" = "main" ]; then
 fi
 
 npx lint-staged
+
+# Scan staged changes for secrets (skip if gitleaks not installed)
+if command -v gitleaks >/dev/null 2>&1; then
```
dimakis, line 12: 🟡 **regressions**: The hook silently skips secret scanning unless contributors have a globally installed `gitleaks`, so the new protection is not actually enforced for most clones. Since this repo already relies on local devDependencies via `npx lint-staged`, wire gitleaks through the repo (e.g. add it as a devDependency and invoke via `npx`) or document/bootstrap installation in the project setup. `[fixable]`

dimakis (commented):
## Codex Review

### `.husky/pre-commit`
_PR is mostly fine, but the pre-commit secret scan is not reliably enforced because it depends on a global gitleaks install._

- 🟡 **regressions** (L10): The hook silently skips secret scanning unless contributors have a globally installed `gitleaks`, so the new protection is not actually enforced for most clones. Since this repo already relies on local devDependencies via `npx lint-staged`, wire gitleaks through the repo (e.g. add it as a devDependency and invoke via `npx`) or document/bootstrap installation in the project setup. `[fixable]`

PR #250, opened 2026-04-20
feat(harness): worktree guard observability + name-based cleanup

dimakis:
## Summary

- **Deny logging**: `log.warn()` on every worktree policy denial with tool name, attempted path, suggested correction, and session ID — both in the guard itself and the permission handler
- **Allow logging**: `log.debug()` on successful checks for full audit trail (gated by `LOG_LEVEL`)
- **Stats counter**: In-memory allowed/denied counts, logged every 5 minutes and reset
- **Startup inventory**: Logs worktree count per repo at server boot
- **Fix stale cleanup**: Parses creation date from worktree directory name (`YYYY-MM-DD-XXXXXX`) instead of `mtime` — git operations keep refreshing mtime, defeating the 96h cutoff and causing worktree sprawl (28 in mitzo, 17 in mgmt)
- **Periodic cleanup**: Runs every 6h via `setInterval`, not just at startup
- **10 new tests**: Stats tracking (5 cases), `parseWorktreeAge` (4 cases), name-based cleanup (1 case)

## Context

Worktree isolation was completely silent on denials — no logs, no traces, no counters. Impossible to tell if agents were hitting the guard or sailing through clean. This makes the system observable.

The sprawl fix is a bonus: April 15 worktrees were accumulating because `mtime` kept getting refreshed. Name-based age parsing is immune to this.

## Test plan

- [x] All 27 worktree-related tests pass (harness + server)
- [x] All 1617 tests in full suite pass
- [x] Harness package type-checks clean
- [ ] Verify deny logs appear in `server-stdout.log` when an agent targets a main repo path
- [ ] Verify stale worktrees (>4 days old by name) get cleaned on next server restart

### DIFF
--- a/packages/harness/src/permission-handler.ts
+++ b/packages/harness/src/permission-handler.ts
@@ -9,10 +9,7 @@ import { getToolTier, shouldAutoAllow } from './tool-tiers.js';
 import { summarizeToolInput } from '@mitzo/protocol';
 import { checkSkillPolicy } from './skill-policy.js';
 import { checkWorktreePolicy } from './worktree-guard.js';
-import { createLogger } from './logger.js';
 import { PERMISSION_TIMEOUT_MS, NTFY_NOTIFICATION_DELAY_MS } from './constants.js';
-
-const log = createLogger('permission-handler');
 import type { SessionRegistry } from './session-registry.js';
 import type { SessionTransport } from './session-transport.js';
 
@@ -69,12 +66,6 @@ export function buildPermissionHandler(clientId: string, registry: SessionRegist
 
     const worktreeViolation = checkWorktreePolicy(session, toolName, _toolInput);
     if (worktreeViolation) {
-      log.warn('worktree violation denied', {
-        clientId,
-        toolName,
-        violation: worktreeViolation,
-        sessionId: session.sessionId,
-      });
       return { behavior: 'deny', message: worktreeViolation };
     }
 
--- a/packages/harness/src/permission-handler.ts
+++ b/packages/harness/src/permission-handler.ts
@@ -9,7 +9,10 @@ import { getToolTier, shouldAutoAllow } from './tool-tiers.js';
 import { summarizeToolInput } from '@mitzo/protocol';
 import { checkSkillPolicy } from './skill-policy.js';
 import { checkWorktreePolicy } from './worktree-guard.js';
+import { createLogger } from './logger.js';
 import { PERMISSION_TIMEOUT_MS, NTFY_NOTIFICATION_DELAY_MS } from './constants.js';
+
+const log = createLogger('permission-handler');
 import type { SessionRegistry } from './session-registry.js';
 import type { SessionTransport } from './session-transport.js';
 
@@ -66,6 +69,12 @@ export function buildPermissionHandler(clientId: string, registry: SessionRegist
 
     const worktreeViolation = checkWorktreePolicy(session, toolName, _toolInput);
     if (worktreeViolation) {
+      log.warn('worktree violation denied', {
+        clientId,
+        toolName,
+        violation: worktreeViolation,
+        sessionId: session.sessionId,
+      });
       return { behavior: 'deny', message: worktreeViolation };
     }
 
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -1,8 +1,22 @@
 import type { ManagedSession } from './session-registry.js';
+import { createLogger } from './logger.js';
+
+const log = createLogger('worktree-guard');
 
 const WRITE_TOOLS = new Set(['Write', 'Edit', 'StrReplace', 'EditNotebook', 'MultiEdit']);
 const SHELL_TOOLS = new Set(['Bash', 'Shell']);
 const PATH_FIELDS = ['file_path', 'path', 'target_notebook'];
+
+const stats = { allowed: 0, denied: 0 };
+
+export function getWorktreeGuardStats() {
+  return { ...stats };
+}
+
+export function resetWorktreeGuardStats() {
+  stats.allowed = 0;
+  stats.denied = 0;
+}
 
 /**
  * Extract absolute paths from a shell command string. Heuristic — catches
@@ -79,10 +93,17 @@ export function checkWorktreePolicy(
       if (allowed) continue;
 
       const suggestion = suggestWorktreePath(filePath, session.worktreePaths);
-      if (suggestion) {
-        return `Path ${filePath} is outside session worktrees. ` + `Use ${suggestion} instead.`;
-      }
-      return `Path ${filePath} is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+      const message = suggestion
+        ? `Path ${filePath} is outside session worktrees. Use ${suggestion} instead.`
+        : `Path ${filePath} is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+      stats.denied++;
+      log.warn('worktree policy denied', {
+        toolName,
+        attemptedPath: filePath,
+        suggestedPath: suggestion,
+        sessionId: session.sessionId,
+      });
+      return message;
     }
   }
 
@@ -95,14 +116,21 @@ export function checkWorktreePolicy(
       if (findAllowedWorktree(p, session.worktreePaths)) continue;
 
       const suggestion = suggestWorktreePath(p, session.worktreePaths);
-      if (suggestion) {
-        return (
-          `Shell command references ${p} which is outside session worktrees. ` +
-          `Use ${suggestion} instead.`
-        );
-      }
+      const message = suggestion
+        ? `Shell command references ${p} which is outside session worktrees. Use ${suggestion} instead.`
+        : `Shell command references ${p} which is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+      stats.denied++;
+      log.warn('worktree policy denied', {
+        toolName,
+        attemptedPath: p,
+        suggestedPath: suggestion,
+        sessionId: session.sessionId,
+      });
+      return message;
     }
   }
 
+  stats.allowed++;
+  log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
   return null;
 }
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -105,6 +105,9 @@ export function checkWorktreePolicy(
       });
       return message;
     }
+    stats.allowed++;
+    log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
+    return null;
   }
 
   if (SHELL_TOOLS.has(toolName)) {
@@ -128,9 +131,9 @@ export function checkWorktreePolicy(
       });
       return message;
     }
+    stats.allowed++;
+    log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
   }
 
-  stats.allowed++;
-  log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
   return null;
 }
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -85,9 +85,11 @@ export function checkWorktreePolicy(
   if (session.worktreePaths.size === 0) return null;
 
   if (WRITE_TOOLS.has(toolName)) {
+    let checkedAnyPath = false;
     for (const field of PATH_FIELDS) {
       const filePath = toolInput[field];
       if (typeof filePath !== 'string' || !filePath.startsWith('/')) continue;
+      checkedAnyPath = true;
 
       const allowed = findAllowedWorktree(filePath, session.worktreePaths);
       if (allowed) continue;
@@ -105,8 +107,10 @@ export function checkWorktreePolicy(
       });
       return message;
     }
-    stats.allowed++;
-    log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
+    if (checkedAnyPath) {
+      stats.allowed++;
+      log.debug('worktree policy allowed', { toolName, sessionId: session.sessionId });
+    }
     return null;
   }
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -26,7 +26,13 @@ import {
 } from './chat.js';
 import { cleanupStaleWorktrees, countWorktrees } from './worktree.js';
 import { getWorktreeGuardStats, resetWorktreeGuardStats } from '@mitzo/harness';
-import { HEARTBEAT_INTERVAL_MS, PORT_DEFAULT, SHUTDOWN_GRACE_MS } from './constants.js';
+import {
+  HEARTBEAT_INTERVAL_MS,
+  PORT_DEFAULT,
+  SHUTDOWN_GRACE_MS,
+  GUARD_STATS_INTERVAL_MS,
+  WORKTREE_CLEANUP_INTERVAL_MS,
+} from './constants.js';
 import { createLogger } from './logger.js';
 import {
   app,
@@ -803,18 +809,14 @@ checkPort(PORT).then((inUse) => {
       }
     }
 
-    // Periodic worktree guard stats (every 5 minutes)
-    const GUARD_STATS_INTERVAL_MS = 5 * 60 * 1000;
     setInterval(() => {
       const stats = getWorktreeGuardStats();
       if (stats.denied > 0 || stats.allowed > 0) {
         log.info('worktree guard stats', stats);
-        resetWorktreeGuardStats();
       }
+      resetWorktreeGuardStats();
     }, GUARD_STATS_INTERVAL_MS);
 
-    // Periodic stale worktree cleanup (every 6 hours)
-    const WORKTREE_CLEANUP_INTERVAL_MS = 6 * 60 * 60 * 1000;
     setInterval(() => {
       for (const [label, repoPath] of repoEntries) {
         try {
--- a/server/index.ts
+++ b/server/index.ts
@@ -24,7 +24,8 @@ import {
   setConnectionRegistry,
   reconcileSessionsBackground,
 } from './chat.js';
-import { cleanupStaleWorktrees } from './worktree.js';
+import { cleanupStaleWorktrees, countWorktrees } from './worktree.js';
+import { getWorktreeGuardStats, resetWorktreeGuardStats } from '@mitzo/harness';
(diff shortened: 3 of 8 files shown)


packages/harness/src/worktree-guard.ts
```diff
@@ -95,14 +116,21 @@ export function checkWorktreePolicy(
...
+        toolName,
+        attemptedPath: p,
+        suggestedPath: suggestion,
+        sessionId: session.sessionId,
+      });
+      return message;
     }
   }
 
+  stats.allowed++;
```
dimakis: 🟡 **bugs**: `stats.allowed++` fires for every tool that isn't Write/Edit/Shell — including Read, Grep, Glob, etc. that aren't subject to worktree policy. This inflates the 'allowed' metric (read-heavy sessions will dominate) and makes the allowed/denied ratio misleading for understanding enforcement effectiveness. Consider only counting when the tool is actually in WRITE_TOOLS or SHELL_TOOLS. `[fixable]`

packages/harness/src/permission-handler.ts
```diff
@@ -9,7 +9,10 @@ import { getToolTier, shouldAutoAllow } from './tool-tiers.js';
 import { summarizeToolInput } from '@mitzo/protocol';
 import { checkSkillPolicy } from './skill-policy.js';
 import { checkWorktreePolicy } from './worktree-guard.js';
+import { createLogger } from './logger.js';
 import { PERMISSION_TIMEOUT_MS, NTFY_NOTIFICATION_DELAY_MS } from './constants.js';
+
```
dimakis: 🟡 **style**: `const log = createLogger(...)` is placed between import statements (after `constants.js`, before `type SessionRegistry`). ESM hoists imports so it works, but it breaks visual grouping. Move `const log` below all imports. `[fixable]`

packages/harness/src/permission-handler.ts
```diff
@@ -66,6 +69,12 @@ export function buildPermissionHandler(clientId: string, registry: SessionRegist
 
     const worktreeViolation = checkWorktreePolicy(session, toolName, _toolInput);
     if (worktreeViolation) {
+      log.warn('worktree violation denied', {
```
dimakis: 🔵 **style**: Worktree violations are now logged at `warn` level in both `worktree-guard.ts` (line 100/123) and `permission-handler.ts` (line 72). Both fire on the same deny path with overlapping fields. Consider removing the permission-handler log since the guard already emits a structured log with more detail (attemptedPath, suggestedPath). `[fixable]`

server/worktree.ts
```diff
@@ -278,6 +295,20 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
   }
 }
 
+/**
+ * Count worktrees in a repo's .claude/worktrees/ directory.
+ * Used for inventory logging at startup and periodic health checks.
+ */
+export function countWorktrees(baseRepo: string): number {
```
dimakis, line 310: 🔵 **missing_tests**: `countWorktrees` is exported and used in startup inventory logging but has no test coverage. It's simple, but a test would cover the edge cases (missing directory, empty directory, error path). `[fixable]`

server/worktree.ts
```diff
@@ -18,6 +18,18 @@ function worktreesDir(baseRepo: string): string {
...
 }
 
+/**
+ * Parse creation age from a worktree directory name (YYYY-MM-DD-XXXXXX format).
+ * Returns age in milliseconds, or null if the name doesn't match the convention.
+ */
+export function parseWorktreeAge(entry: string): number | null {
+  const match = entry.match(/^(\d{4}-\d{2}-\d{2})-/);
+  if (!match) return null;
+  const created = new Date(match[1] + 'T00:00:00Z').getTime();
```
dimakis: 🔵 **unsafe_assumptions**: `parseWorktreeAge` anchors to UTC midnight (`T00:00:00Z`) while `Date.now()` uses real time. A worktree created at 23:59 UTC on day X gets a name with day X but is nearly 24h younger than what the name-based age implies. With a 96h stale threshold, a worst-case ~24h overshoot is unlikely to cause problems, but it's worth noting if the threshold ever shrinks.

dimakis (commented):
## Centaur Review

_Solid observability addition. Main concern: the `allowed` stat counter fires for all tools (including reads) rather than only write/shell tools that are actually policy-checked, making the metric noisy._

- 🟡 **bugs** (L133): `stats.allowed++` fires for every tool that isn't Write/Edit/Shell — including Read, Grep, Glob, etc. that aren't subject to worktree policy. This inflates the 'allowed' metric (read-heavy sessions will dominate) and makes the allowed/denied ratio misleading for understanding enforcement effectiveness. Consider only counting when the tool is actually in WRITE_TOOLS or SHELL_TOOLS. `[fixable]`


### `packages/harness/src/permission-handler.ts`
_Solid observability addition. Main concern: the `allowed` stat counter fires for all tools (including reads) rather than only write/shell tools that are actually policy-checked, making the metric noisy._

- 🟡 **style** (L14): `const log = createLogger(...)` is placed between import statements (after `constants.js`, before `type SessionRegistry`). ESM hoists imports so it works, but it breaks visual grouping. Move `const log` below all imports. `[fixable]`
- 🔵 **style** (L72): Worktree violations are now logged at `warn` level in both `worktree-guard.ts` (line 100/123) and `permission-handler.ts` (line 72). Both fire on the same deny path with overlapping fields. Consider removing the permission-handler log since the guard already emits a structured log with more detail (attemptedPath, suggestedPath). `[fixable]`


### `server/worktree.ts`
_Solid observability addition. Main concern: the `allowed` stat counter fires for all tools (including reads) rather than only write/shell tools that are actually policy-checked, making the metric noisy._

- 🔵 **missing_tests** (L302): `countWorktrees` is exported and used in startup inventory logging but has no test coverage. It's simple, but a test would cover the edge cases (missing directory, empty directory, error path). `[fixable]`
- 🔵 **unsafe_assumptions** (L28): `parseWorktreeAge` anchors to UTC midnight (`T00:00:00Z`) while `Date.now()` uses real time. A worktree created at 23:59 UTC on day X gets a name with day X but is nearly 24h younger than what the name-based age implies. With a 96h stale threshold, a worst-case ~24h overshoot is unlikely to cause problems, but it's worth noting if the threshold ever shrinks.

server/worktree.ts
```diff
@@ -278,6 +295,20 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
...
 
+/**
+ * Count worktrees in a repo's .claude/worktrees/ directory.
+ * Used for inventory logging at startup and periodic health checks.
+ */
+export function countWorktrees(baseRepo: string): number {
+  const dir = worktreesDir(baseRepo);
+  if (!existsSync(dir)) return 0;
+  try {
+    return readdirSync(dir).length;
```
dimakis: 🟡 **bugs**: countWorktrees uses readdirSync(dir).length which counts all filesystem entries (files, symlinks, .DS_Store, etc.), not just directories. On macOS, .DS_Store files will inflate the count. Consider filtering with statSync().isDirectory() or at minimum filtering out dotfiles. `[fixable]`

server/worktree.ts
```diff
@@ -278,6 +295,20 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
   }
 }
 
+/**
```
dimakis, line 306: 🔵 **style**: The JSDoc on countWorktrees says 'Count worktrees' but actually counts all directory entries. Either rename to countWorktreeEntries or adjust the implementation to only count directories — the name creates an expectation of accuracy. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_Solid observability additions — name-based age is a good improvement over mtime. Main concern is countWorktrees counting all fs entries including non-directories, and a minor stats-counting edge case in the worktree guard._

- 🟡 **bugs** (L306): countWorktrees uses readdirSync(dir).length which counts all filesystem entries (files, symlinks, .DS_Store, etc.), not just directories. On macOS, .DS_Store files will inflate the count. Consider filtering with statSync().isDirectory() or at minimum filtering out dotfiles. `[fixable]`
- 🔵 **style** (L298): The JSDoc on countWorktrees says 'Count worktrees' but actually counts all directory entries. Either rename to countWorktreeEntries or adjust the implementation to only count directories — the name creates an expectation of accuracy. `[fixable]`


_Solid observability additions — name-based age is a good improvement over mtime. Main concern is countWorktrees counting all fs entries including non-directories, and a minor stats-counting edge case in the worktree guard._

- 🟡 **bugs** (L108): Write tool stats.allowed++ is reached after the for-loop over PATH_FIELDS completes without denial, but a Write call with NO path fields (all fields missing or non-absolute) also reaches this line and increments allowed. This inflates the allowed count for no-op checks. The early return at line 85 (worktreePaths.size === 0) already returns null without counting — same gap applies here. `[fixable]`
- 🔵 **unsafe_assumptions** (L10): Global mutable stats object is a process-wide singleton. While safe in single-threaded Node.js, it conflates counts across all sessions. If per-session observability is ever needed, this design won't support it. For now this is fine since the stats are just periodic aggregate logging, but worth noting the limitation.


_Solid observability additions — name-based age is a good improvement over mtime. Main concern is countWorktrees counting all fs entries including non-directories, and a minor stats-counting edge case in the worktree guard._

- 🔵 **missing_tests**: countWorktrees has no test for a directory containing non-directory entries (e.g. a stray file). If the intent is to count only worktree directories, this case should be tested and the implementation adjusted. `[fixable]`


### `server/index.ts`
_Solid observability additions — name-based age is a good improvement over mtime. Main concern is countWorktrees counting all fs entries including non-directories, and a minor stats-counting edge case in the worktree guard._

- 🔵 **style** (L807): The guard stats interval (5 min) and cleanup interval (6 hours) are inline magic numbers. Other intervals in this file (HEARTBEAT_INTERVAL_MS, UPDATE_CHECK_INTERVAL_MS) are named constants. Consider moving these to constants.ts for consistency, or at least co-locating them with the existing interval constants. `[fixable]`

packages/harness/src/worktree-guard.ts
```diff
@@ -79,11 +93,21 @@ export function checkWorktreePolicy(
...
+      stats.denied++;
+      log.warn('worktree policy denied', {
+        toolName,
+        attemptedPath: filePath,
+        suggestedPath: suggestion,
+        sessionId: session.sessionId,
+      });
+      return message;
     }
+    stats.allowed++;
```
dimakis: 🟡 **bugs**: Write tool stats.allowed++ is reached after the for-loop over PATH_FIELDS completes without denial, but a Write call with NO path fields (all fields missing or non-absolute) also reaches this line and increments allowed. This inflates the allowed count for no-op checks. The early return at line 85 (worktreePaths.size === 0) already returns null without counting — same gap applies here. `[fixable]`

packages/harness/src/worktree-guard.ts
```diff
@@ -1,9 +1,23 @@
 import type { ManagedSession } from './session-registry.js';
+import { createLogger } from './logger.js';
+
+const log = createLogger('worktree-guard');
 
 const WRITE_TOOLS = new Set(['Write', 'Edit', 'StrReplace', 'EditNotebook', 'MultiEdit']);
 const SHELL_TOOLS = new Set(['Bash', 'Shell']);
 const PATH_FIELDS = ['file_path', 'path', 'target_notebook'];
 
+const stats = { allowed: 0, denied: 0 };
```
dimakis, line 10: 🔵 **unsafe_assumptions**: Global mutable stats object is a process-wide singleton. While safe in single-threaded Node.js, it conflates counts across all sessions. If per-session observability is ever needed, this design won't support it. For now this is fine since the stats are just periodic aggregate logging, but worth noting the limitation.

server/index.ts
```diff
@@ -791,5 +792,39 @@ checkPort(PORT).then((inUse) => {
...
+    // Log worktree inventory at startup
+    for (const [label, repoPath] of repoEntries) {
+      const count = countWorktrees(repoPath);
+      if (count > 0) {
+        log.info('worktree inventory', { repo: label, count });
+      }
+    }
+
+    // Periodic worktree guard stats (every 5 minutes)
+    const GUARD_STATS_INTERVAL_MS = 5 * 60 * 1000;
```
dimakis: 🔵 **style**: The guard stats interval (5 min) and cleanup interval (6 hours) are inline magic numbers. Other intervals in this file (HEARTBEAT_INTERVAL_MS, UPDATE_CHECK_INTERVAL_MS) are named constants. Consider moving these to constants.ts for consistency, or at least co-locating them with the existing interval constants. `[fixable]`

server/worktree.ts
```diff
@@ -18,6 +18,18 @@ function worktreesDir(baseRepo: string): string {
   return join(baseRepo, '.claude', 'worktrees');
 }
 
+/**
+ * Parse creation age from a worktree directory name (YYYY-MM-DD-XXXXXX format).
+ * Returns age in milliseconds, or null if the name doesn't match the convention.
+ */
```
dimakis, line 24: 🟡 **bugs**: parseWorktreeAge accepts impossible calendar dates like 2026-02-31 because Date normalizes them instead of producing NaN. cleanupStaleWorktrees/listWorktrees will therefore treat malformed names as valid dated worktrees and compute the wrong age instead of falling back to mtime/null as the tests imply. `[fixable]`

server/worktree.ts
```diff
@@ -236,8 +248,13 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
   for (const entry of readdirSync(dir)) {
```
dimakis, line 255: 🟡 **regressions**: cleanupStaleWorktrees now prefers the directory name over mtime for any YYYY-MM-DD-* entry. Existing manually named worktrees or non-session directories under .claude/worktrees that happen to match that pattern can now be deleted based solely on their name, even if they were recently touched. The previous behavior only pruned by actual filesystem age. `[fixable]`

packages/harness/src/worktree-guard.ts
```diff
@@ -1,9 +1,23 @@
 import type { ManagedSession } from './session-registry.js';
+import { createLogger } from './logger.js';
+
+const log = createLogger('worktree-guard');
 
 const WRITE_TOOLS = new Set(['Write', 'Edit', 'StrReplace', 'EditNotebook', 'MultiEdit']);
 const SHELL_TOOLS = new Set(['Bash', 'Shell']);
 const PATH_FIELDS = ['file_path', 'path', 'target_notebook'];
```
dimakis, line 8: 🟡 **unsafe_assumptions**: Worktree guard stats are stored in a process-global singleton and reset on a timer in server/index.ts. In this app the harness package is shared infrastructure, so any future multi-process deployment, tests running in parallel, or other consumers importing the guard will see partial/non-authoritative counters. If these logs are meant as observability, the implementation should make the scope explicit or aggregate per process/session. `[fixable]`

dimakis (commented):
## Codex Review

### `server/worktree.ts`
_Useful observability additions, but the new name-based age parsing/pruning introduces a real cleanup risk and the date parser is too permissive._

- 🟡 **bugs** (L24): parseWorktreeAge accepts impossible calendar dates like 2026-02-31 because Date normalizes them instead of producing NaN. cleanupStaleWorktrees/listWorktrees will therefore treat malformed names as valid dated worktrees and compute the wrong age instead of falling back to mtime/null as the tests imply. `[fixable]`
- 🟡 **regressions** (L248): cleanupStaleWorktrees now prefers the directory name over mtime for any YYYY-MM-DD-* entry. Existing manually named worktrees or non-session directories under .claude/worktrees that happen to match that pattern can now be deleted based solely on their name, even if they were recently touched. The previous behavior only pruned by actual filesystem age. `[fixable]`


_Useful observability additions, but the new name-based age parsing/pruning introduces a real cleanup risk and the date parser is too permissive._

- 🟡 **unsafe_assumptions** (L8): Worktree guard stats are stored in a process-global singleton and reset on a timer in server/index.ts. In this app the harness package is shared infrastructure, so any future multi-process deployment, tests running in parallel, or other consumers importing the guard will see partial/non-authoritative counters. If these logs are meant as observability, the implementation should make the scope explicit or aggregate per process/session. `[fixable]`

dimakis (commented):
## Codex Review

### `server/worktree.ts`
_Good test coverage overall, but there are a couple of observability edge cases around worktree age parsing and stats window semantics._

- 🟡 **bugs** (L26): `parseWorktreeAge()` can return a negative age for future-dated directory names, which means stale cleanup and age display silently trust malformed or clock-skewed names instead of rejecting them. A future-named worktree will appear fresh indefinitely until that date passes. `[fixable]`


### `server/index.ts`
_Good test coverage overall, but there are a couple of observability edge cases around worktree age parsing and stats window semantics._

- 🟡 **regressions** (L804): The new periodic worktree-guard stats logger only resets counters when at least one event occurred. Because the counters are process-global and cumulative, a single denied/allowed event can be reported repeatedly across intervals if logging throws or if future code reads stats without resetting. Resetting unconditionally after sampling would preserve fixed-window semantics more reliably. `[fixable]`

server/worktree.ts
```diff
@@ -18,6 +18,23 @@ function worktreesDir(baseRepo: string): string {
   return join(baseRepo, '.claude', 'worktrees');
 }
 
+/**
+ * Parse creation age from a worktree directory name (YYYY-MM-DD-XXXXXX format).
+ * Returns age in milliseconds, or null if the name doesn't match the convention.
+ */
+export function parseWorktreeAge(entry: string): number | null {
+  const match = entry.match(/^(\d{4})-(\d{2})-(\d{2})-[a-f0-9]{6}$/);
```
dimakis, line 26: 🟡 **bugs**: `parseWorktreeAge()` can return a negative age for future-dated directory names, which means stale cleanup and age display silently trust malformed or clock-skewed names instead of rejecting them. A future-named worktree will appear fresh indefinitely until that date passes. `[fixable]`

server/index.ts
```diff
@@ -791,5 +792,39 @@ checkPort(PORT).then((inUse) => {
...
+
+    // --- Worktree observability ---
+
+    // Log worktree inventory at startup
+    for (const [label, repoPath] of repoEntries) {
+      const count = countWorktrees(repoPath);
+      if (count > 0) {
+        log.info('worktree inventory', { repo: label, count });
+      }
+    }
```
dimakis, line 810: 🟡 **regressions**: The new periodic worktree-guard stats logger only resets counters when at least one event occurred. Because the counters are process-global and cumulative, a single denied/allowed event can be reported repeatedly across intervals if logging throws or if future code reads stats without resetting. Resetting unconditionally after sampling would preserve fixed-window semantics more reliably. `[fixable]`

PR #251, opened 2026-04-20
fix(auto-rename): trigger session naming on initial prompt

dimakis:
## Summary

- **Bug:** `tryAutoRename` was only called from `sendToChat()` (follow-up messages), never for the initial prompt processed in `runQueryLoop()`. The prompt count was incremented to 1 but no rename attempt was made, so sessions stayed "Untitled" until the 4th message (count 4 is the first to pass `shouldAutoRename` after the skip zone at 2-3).
- **Fix:** Store the initial prompt as a `user_message` event (so `extractRecentPrompts()` has content to work with) and add an `onInitialPrompt` callback that fires `tryAutoRename` when the session ID resolves.
- **Result:** Sessions get named immediately after the first assistant response, not after the 4th user message.

## Test plan

- [x] Existing auto-rename tests pass (62/62)
- [x] Updated test: initial prompt now stored as `user_message` event
- [x] New test: `onInitialPrompt` callback fires with session ID
- [x] New test: callback does not fire when no initial prompt (resume case)
- [ ] Manual: start a new Mitzo session, verify it gets named after the first message

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -566,7 +566,7 @@ export async function startChat(
         connRegistry: _connRegistry ?? undefined,
         onSessionResolved: options.onSessionResolved,
         onInitialPrompt: (sessionId: string) => {
-          tryAutoRename(sessionId, clientId).catch(() => {});
+          tryAutoRename(sessionId, clientId).catch(() => { /* errors logged internally */ });
         },
       },
     );
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -566,7 +566,9 @@ export async function startChat(
         connRegistry: _connRegistry ?? undefined,
         onSessionResolved: options.onSessionResolved,
         onInitialPrompt: (sessionId: string) => {
-          tryAutoRename(sessionId, clientId).catch(() => { /* errors logged internally */ });
+          tryAutoRename(sessionId, clientId).catch(() => {
+            /* errors logged internally */
+          });
         },
       },
     );
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -565,6 +565,9 @@ export async function startChat(
       {
         connRegistry: _connRegistry ?? undefined,
         onSessionResolved: options.onSessionResolved,
+        onInitialPrompt: (sessionId: string) => {
+          tryAutoRename(sessionId, clientId).catch(() => {});
+        },
       },
     );
   } catch (err: unknown) {
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -309,9 +309,7 @@ export async function runQueryLoop(
                 messageId: `umsg-${Date.now()}-init`,
                 text: initialPrompt,
               });
-              // Count the initial prompt for auto-rename tracking
-              store.incrementPromptCount(resolvedSessionId);
-              // Trigger auto-rename on prompt 1 (immediate naming)
+              // Trigger auto-rename — tryAutoRename handles the increment
               onInitialPrompt?.(resolvedSessionId);
             }
           }
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -108,6 +108,8 @@ function v2(type: string, rest: Record<string, unknown> = {}): Record<string, un
 export interface QueryLoopOptions {
   connRegistry?: ConnectionRegistry;
   onSessionResolved?: (sessionId: string) => void;
+  /** Called after the initial prompt is registered, enabling auto-rename on prompt 1. */
+  onInitialPrompt?: (sessionId: string) => void;
 }
 
 export async function runQueryLoop(
@@ -121,6 +123,7 @@ export async function runQueryLoop(
 ) {
   const connRegistry = options?.connRegistry;
   const onSessionResolved = options?.onSessionResolved;
+  const onInitialPrompt = options?.onInitialPrompt;
   // Tool input buffers keyed by content block index (reset per message_start).
   const toolInputBuffers = new Map<
     number,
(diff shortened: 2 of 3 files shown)


server/query-loop.ts
```diff
@@ -297,8 +300,19 @@ export async function runQueryLoop(
               ...(initialPrompt ? { initialPrompt } : {}),
             });
```
dimakis, line 301: 🔴 **bugs**: Double-increment bug: `store.incrementPromptCount()` is called here (incrementing to 1), then `onInitialPrompt` fires `tryAutoRename()` which calls `incrementPromptCount` again (incrementing to 2). Since `shouldAutoRename(2, false)` returns `false`, auto-rename never actually triggers on the initial prompt — defeating the entire purpose of this PR. The fix is to remove this `incrementPromptCount` call from `runQueryLoop`, since `tryAutoRename` (called via the `onInitialPrompt` callback) already handles the increment. This matches the pattern in `sendToChat()` which does NOT call `incrementPromptCount` separately — it delegates entirely to `tryAutoRename`. `[fixable]`

server/chat.ts
```diff
@@ -565,6 +565,9 @@ export async function startChat(
       {
         connRegistry: _connRegistry ?? undefined,
         onSessionResolved: options.onSessionResolved,
```
dimakis, line 567: 🔵 **style**: The `onInitialPrompt` callback is defined inline as `(sessionId: string) => { tryAutoRename(sessionId, clientId).catch(() => {}); }`. The empty `.catch(() => {})` silently swallows errors. While `tryAutoRename` has its own internal try/catch that logs, if `tryAutoRename` itself throws before reaching its try block (unlikely but possible), the error is lost. The existing pattern in `sendToChat` (line 655) uses `.catch(() => { /* errors logged internally */ })` — matching that comment would clarify intent. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Critical double-increment bug: `incrementPromptCount` is called in both `runQueryLoop` and `tryAutoRename` (via the `onInitialPrompt` callback), so the count reaches 2 instead of 1, and `shouldAutoRename(2)` returns false — the initial-prompt rename never fires._

- 🔴 **bugs** (L301): Double-increment bug: `store.incrementPromptCount()` is called here (incrementing to 1), then `onInitialPrompt` fires `tryAutoRename()` which calls `incrementPromptCount` again (incrementing to 2). Since `shouldAutoRename(2, false)` returns `false`, auto-rename never actually triggers on the initial prompt — defeating the entire purpose of this PR. The fix is to remove this `incrementPromptCount` call from `runQueryLoop`, since `tryAutoRename` (called via the `onInitialPrompt` callback) already handles the increment. This matches the pattern in `sendToChat()` which does NOT call `incrementPromptCount` separately — it delegates entirely to `tryAutoRename`. `[fixable]`


_Critical double-increment bug: `incrementPromptCount` is called in both `runQueryLoop` and `tryAutoRename` (via the `onInitialPrompt` callback), so the count reaches 2 instead of 1, and `shouldAutoRename(2)` returns false — the initial-prompt rename never fires._

- 🟡 **missing_tests**: The new tests verify the callback fires but don't test the integration with `tryAutoRename` end-to-end. A test that wires up a real `EventStore` and verifies the prompt count is exactly 1 (not 2) after the initial prompt + callback would have caught the double-increment bug. Consider adding an integration test that asserts `store.getSession(sessionId).promptCount === 1` after the full flow. `[fixable]`


### `server/chat.ts`
_Critical double-increment bug: `incrementPromptCount` is called in both `runQueryLoop` and `tryAutoRename` (via the `onInitialPrompt` callback), so the count reaches 2 instead of 1, and `shouldAutoRename(2)` returns false — the initial-prompt rename never fires._

- 🔵 **style** (L567): The `onInitialPrompt` callback is defined inline as `(sessionId: string) => { tryAutoRename(sessionId, clientId).catch(() => {}); }`. The empty `.catch(() => {})` silently swallows errors. While `tryAutoRename` has its own internal try/catch that logs, if `tryAutoRename` itself throws before reaching its try block (unlikely but possible), the error is lost. The existing pattern in `sendToChat` (line 655) uses `.catch(() => { /* errors logged internally */ })` — matching that comment would clarify intent. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `server/query-loop.ts`
_Clean fix: stores the initial prompt as a user_message event (enabling extractRecentPrompts for auto-rename) and adds an onInitialPrompt callback to trigger naming on the first turn. Replay dedup in replayEventsToMessages correctly handles the new event. Test coverage is solid. One minor style nit._

- 🔵 **bugs** (L308): Two separate `Date.now()` calls on lines 308 and 309 could return different millisecond values, making `ts` and the timestamp embedded in `messageId` diverge. Extract to a `const now = Date.now()` and reuse it — matches the pattern used by `sendToChat` (which uses a single expression for `messageId` and `ts` separately but they're in different statements there too, so this is consistent, though a shared variable would be cleaner). `[fixable]`

server/query-loop.ts
```diff
@@ -297,8 +300,17 @@ export async function runQueryLoop(
...
             if (initialPrompt) {
-              // Count the initial prompt for auto-rename tracking
-              store.incrementPromptCount(resolvedSessionId);
+              // Store the initial prompt as a user_message event so
+              // extractRecentPrompts() can find it for auto-rename.
+              store.append(resolvedSessionId, 'user_message', {
+                v: 2,
+                type: 'user_message',
+                ts: Date.now(),
```
dimakis: 🔵 **bugs**: Two separate `Date.now()` calls on lines 308 and 309 could return different millisecond values, making `ts` and the timestamp embedded in `messageId` diverge. Extract to a `const now = Date.now()` and reuse it — matches the pattern used by `sendToChat` (which uses a single expression for `messageId` and `ts` separately but they're in different statements there too, so this is consistent, though a shared variable would be cleaner). `[fixable]`

PR #252, opened 2026-04-20
feat(task-board): workflow layer with stage types, signals, and templates

dimakis:
## Summary

- **Workflow stage types**: Tasks can now be `agent_work`, `wait_for_signal`, or `human_review` — the orchestrator dispatches differently for each (agent session, signal polling, or immediate pending_review)
- **SignalProcessor**: Polls external systems (GitHub CI/review via `gh` CLI, Centaur API) for gate resolution, with retry logic that resets preceding agent work stages on failure
- **Workflow templates**: Reusable multi-stage workflow definitions with Mustache-style variable substitution; CRUD endpoints + instantiation that creates a goal tree from a template
- **TaskArtifact**: New MCP tool + REST endpoint for storing structured key-value artifacts per task, injected into sibling context as `<sibling-artifacts>` XML
- **Frontend**: Stage type badges (signal/review), retry count indicator, artifacts display on TaskNode
- **Full backwards compatibility**: All new fields nullable/defaulted, existing tasks unaffected

### New files
- `server/signal-processor.ts` — gate checker + polling engine
- `server/workflow-templates.ts` — template CRUD + instantiation
- 4 new test files (132 workflow-specific tests)

### Modified
- `task-store.ts` — 6 new columns + migration
- `task-orchestrator.ts` — stage-type dispatch in `tick()`
- `task-context.ts` — sibling artifact injection
- `task-tools.ts` + `task-mcp-server.ts` — TaskArtifact tool
- `app.ts` + `index.ts` — REST endpoints + server wiring
- Frontend types + components + CSS

## Test plan

- [x] 132 workflow-specific tests pass (task-store, workflow-templates, signal-processor, task-orchestrator, task-tools, task-context)
- [x] Full test suite passes (2023/2025 — 2 pre-existing chat.test.ts timeouts)
- [ ] Built-in template seeding (PR Triage, Upstream Issue) — follow-up commit
- [ ] WorkflowCreateForm UI — follow-up commit
- [ ] End-to-end test with real PR

### DIFF
--- a/frontend/src/types/task.ts
+++ b/frontend/src/types/task.ts
@@ -9,6 +9,8 @@ export type TaskStatus =
 
 export type SessionPolicy = 'reuse' | 'spawn' | 'auto';
 
+export type StageType = 'agent_work' | 'wait_for_signal' | 'human_review';
+
 export interface LoopStatus {
   state: 'idle' | 'running' | 'paused';
   goalId: string | null;
@@ -37,5 +39,11 @@ export interface Task {
   createdAt: number;
   updatedAt: number;
   completedAt: number | null;
+  stageType: StageType | null;
+  gateConfig: Record<string, unknown> | null;
+  artifacts: Record<string, unknown> | null;
+  retryCount: number;
+  maxRetries: number;
+  templateId: string | null;
   children: Task[];
 }
--- a/server/app.ts
+++ b/server/app.ts
@@ -57,10 +57,11 @@ import {
   TaskUpdateBody,
   LoopStartBody,
   WorkflowInstantiateBody,
+  TemplateCreateBody,
   SignalBody,
 } from './api-schemas.js';
 import type { TaskOrchestrator } from './task-orchestrator.js';
-import type { WorkflowTemplateStore } from './workflow-templates.js';
+import type { WorkflowTemplateStore, TemplateCreateInput } from './workflow-templates.js';
 import { instantiateTemplate } from './workflow-templates.js';
 import type { SignalProcessor } from './signal-processor.js';
 import {
@@ -677,7 +678,7 @@ app.get('/api/templates', (_req, res) => {
     res.status(503).json({ error: 'Template store not initialized' });
     return;
   }
-  res.json({ templates: templateStore.list() });
+  res.json(templateStore.list());
 });
 
 app.get('/api/templates/:id', (req, res) => {
@@ -691,6 +692,20 @@ app.get('/api/templates/:id', (req, res) => {
     return;
   }
   res.json({ template: tmpl });
+});
+
+app.post('/api/templates', (req, res) => {
+  if (!templateStore) {
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  const body = TemplateCreateBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  const tmpl = templateStore.create(body.data as TemplateCreateInput);
+  res.status(201).json(tmpl);
 });
 
 app.delete('/api/templates/:id', (req, res) => {
@@ -747,6 +762,14 @@ app.post('/api/tasks/:id/signal', (req, res) => {
   const task = taskStore.get(req.params.id);
   if (!task) {
     res.status(404).json({ error: 'Task not found' });
+    return;
+  }
+  if (task.stageType !== 'wait_for_signal') {
+    res.status(400).json({ error: 'Task is not a wait_for_signal stage' });
+    return;
+  }
+  if (task.status !== 'active') {
+    res.status(400).json({ error: `Task is ${task.status}, not active` });
     return;
   }
   signalProcessor.resolveSignal(req.params.id, {
--- a/server/app.ts
+++ b/server/app.ts
@@ -41,6 +41,7 @@ import {
   handleTaskComplete,
   handleTaskStatus,
   handleTaskBlock,
+  handleTaskArtifact,
 } from './task-tools.js';
 import { setTaskStore } from './chat.js';
 import {
@@ -55,8 +56,12 @@ import {
   TaskCreateBody,
   TaskUpdateBody,
   LoopStartBody,
+  WorkflowInstantiateBody,
+  SignalBody,
 } from './api-schemas.js';
 import type { TaskOrchestrator } from './task-orchestrator.js';
+import type { WorkflowTemplateStore } from './workflow-templates.js';
+import type { SignalProcessor } from './signal-processor.js';
 import {
   listInboxItems,
   readInboxItem,
@@ -195,9 +200,19 @@ let onUpdateAvailable: (() => void) | null = null;
 let onInboxUpdated: (() => void) | null = null;
 let onTaskBroadcast: ((event: Record<string, unknown>) => void) | null = null;
 let orchestrator: TaskOrchestrator | null = null;
+let templateStore: WorkflowTemplateStore | null = null;
+let signalProcessor: SignalProcessor | null = null;
 
 export function setOrchestrator(o: TaskOrchestrator): void {
   orchestrator = o;
+}
+
+export function setTemplateStore(ts: WorkflowTemplateStore): void {
+  templateStore = ts;
+}
+
+export function setSignalProcessor(sp: SignalProcessor): void {
+  signalProcessor = sp;
 }
 
 // --- Task store ---
@@ -532,6 +547,24 @@ app.post('/api/internal/task-tools/block', (req, res) => {
   });
 });
 
+app.post('/api/internal/task-tools/artifact', (req, res) => {
+  if (!verifyInternalToken(req)) {
+    res.status(401).json({ error: 'Internal token required' });
+    return;
+  }
+  const taskId = resolveTaskContext(req);
+  if (!taskId) {
+    res.status(400).json({ error: 'No active task context' });
+    return;
+  }
+  const result = handleTaskArtifact(taskStore, taskId, req.body.key ?? '', req.body.value ?? '');
+  res.json({ result });
+  onTaskBroadcast?.({
+    type: 'task_state',
+    tasks: taskStore.getTree(),
+  });
+});
+
 // --- Loop orchestrator API ---
 
 app.get('/api/loop/status', (req, res) => {
@@ -634,6 +667,94 @@ app.post('/api/loop/spec/reject', (_req, res) => {
     return;
   }
   res.json(orchestrator.rejectSpec());
+});
+
+// --- Workflow Templates ---
+
+app.get('/api/templates', (_req, res) => {
+  if (!templateStore) {
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  res.json({ templates: templateStore.list() });
+});
+
+app.get('/api/templates/:id', (req, res) => {
+  if (!templateStore) {
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  const tmpl = templateStore.get(req.params.id);
+  if (!tmpl) {
+    res.status(404).json({ error: 'Template not found' });
+    return;
+  }
+  res.json({ template: tmpl });
+});
+
+app.delete('/api/templates/:id', (req, res) => {
+  if (!templateStore) {
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  const ok = templateStore.delete(req.params.id);
+  if (!ok) {
+    res.status(404).json({ error: 'Template not found' });
+    return;
+  }
+  res.json({ ok: true });
+});
+
+app.post('/api/workflows/instantiate', (req, res) => {
+  if (!templateStore) {
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  const body = WorkflowInstantiateBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  try {
+    const { instantiateTemplate } = require('./workflow-templates.js');
+    const goal = instantiateTemplate(
+      taskStore,
+      templateStore,
+      body.data.templateId,
+      body.data.title,
+      body.data.variables,
+    );
+    res.status(201).json({ task: goal });
+    onTaskBroadcast?.({ type: 'task_state', tasks: taskStore.getTree() });
+  } catch (err: unknown) {
+    const message = err instanceof Error ? err.message : 'Unknown error';
+    res.status(400).json({ error: message });
+  }
+});
+
+// --- Signal injection ---
+
+app.post('/api/tasks/:id/signal', (req, res) => {
+  if (!signalProcessor) {
+    res.status(503).json({ error: 'Signal processor not initialized' });
+    return;
+  }
+  const body = SignalBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  const task = taskStore.get(req.params.id);
+  if (!task) {
+    res.status(404).json({ error: 'Task not found' });
+    return;
+  }
+  signalProcessor.resolveSignal(req.params.id, {
+    status: body.data.status,
+    artifacts: body.data.artifacts,
+  });
+  res.json({ ok: true });
+  onTaskBroadcast?.({ type: 'task_state', tasks: taskStore.getTree() });
 });
 
 app.post('/api/auth/login', loginLimiter, async (req, res) => {
--- a/server/app.ts
+++ b/server/app.ts
@@ -61,6 +61,7 @@ import {
 } from './api-schemas.js';
 import type { TaskOrchestrator } from './task-orchestrator.js';
 import type { WorkflowTemplateStore } from './workflow-templates.js';
+import { instantiateTemplate } from './workflow-templates.js';
 import type { SignalProcessor } from './signal-processor.js';
 import {
   listInboxItems,
@@ -74,7 +75,7 @@ import { SkillRegistry } from './skills.js';
 
 import { mkdirSync } from 'fs';
 import { homedir } from 'os';
-import { TaskStore } from './task-store.js';
+import { TaskStore, type TaskCreateInput, type TaskUpdateInput } from './task-store.js';
 
 const log = createLogger('server');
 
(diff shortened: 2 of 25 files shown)


server/app.ts
```diff
@@ -636,6 +669,94 @@ app.post('/api/loop/spec/reject', (_req, res) => {
...
+    res.status(503).json({ error: 'Template store not initialized' });
+    return;
+  }
+  const body = WorkflowInstantiateBody.safeParse(req.body);
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  try {
+    const { instantiateTemplate } = require('./workflow-templates.js');
```
dimakis: 🔴 **bugs**: `require('./workflow-templates.js')` is a CommonJS call inside an ESM module. All other imports in this file use ESM `import` with `.js` extensions. This will fail at runtime in a strict ESM environment. `instantiateTemplate` should be imported at the top of the file alongside the existing `import type { WorkflowTemplateStore }` on line 63. `[fixable]`

server/app.ts
```diff
@@ -636,6 +669,94 @@ app.post('/api/loop/spec/reject', (_req, res) => {
   res.json(orchestrator.rejectSpec());
 });
 
+// --- Workflow Templates ---
+
+app.get('/api/templates', (_req, res) => {
```
dimakis, line 676: 🟡 **missing_tests**: No REST endpoint tests for the five new routes: GET/GET/:id/DELETE /api/templates, POST /api/workflows/instantiate, POST /api/tasks/:id/signal. The existing `task-routes.test.ts` covers task CRUD but none of these new endpoints. Error paths (503 when store not initialized, 404 for missing resources, 400 for invalid payloads) are untested. `[fixable]`

dimakis, line 676: 🟡 **regressions**: There is no `POST /api/templates` endpoint for creating templates. The `WorkflowTemplateStore` has a `create()` method and `GET`/`DELETE` routes exist, but there is no way to create templates via the REST API. The only consumer is `instantiateTemplate` which reads existing templates. Templates can never be created through the API. `[fixable]`

server/app.ts
```diff
@@ -636,6 +669,94 @@ app.post('/api/loop/spec/reject', (_req, res) => {
...
+  if (!body.success) {
+    res.status(400).json({ error: body.error.issues[0]?.message ?? 'Invalid input' });
+    return;
+  }
+  const task = taskStore.get(req.params.id);
+  if (!task) {
+    res.status(404).json({ error: 'Task not found' });
+    return;
+  }
+  signalProcessor.resolveSignal(req.params.id, {
```
dimakis, line 775: 🟡 **unsafe_assumptions**: The `/api/tasks/:id/signal` endpoint calls `signalProcessor.resolveSignal()` without checking whether the task is actually a `wait_for_signal` task or whether it's in `active` status. Any authenticated user can resolve any task as a signal, potentially corrupting non-signal tasks. Should validate `task.stageType === 'wait_for_signal'` and `task.status === 'active'`. `[fixable]`

server/signal-processor.ts
```diff
@@ -0,0 +1,297 @@
...
+    }
+  }
+
+  /** Reset the most recent agent_work sibling before this task to pending. */
+  private resetPrecedingSibling(task: { id: string; parentId: string | null }): void {
+    if (!task.parentId) return;
+
+    const siblings = this.store.getChildren(task.parentId);
+    // Siblings are sorted by priority DESC — find the one just before this task
+    let found = false;
```
dimakis: 🟡 **bugs**: Dead code: the first loop (lines 166-175) sets `found = true` but never uses it and has no side effects. The actual work is done by the second loop (lines 179-185). The dead loop and its misleading comments should be removed. `[fixable]`

server/signal-processor.ts
```diff
@@ -0,0 +1,297 @@
...
+        summary: `Signal resolved: ${task.gateConfig?.type ?? 'unknown'} passed`,
+        artifacts: result.artifacts,
+      });
+      this.store.cascadeStatus(taskId);
+      log.info('signal passed', { taskId, type: task.gateConfig?.type });
+    } else {
+      this.handleFailure(task.id, result.artifacts);
+    }
+
+    this.onSignalResolved(taskId);
```
dimakis, line 106: 🔵 **style**: The `onSignalResolved` callback name suggests it's a notification, but in `handleFailure` with retries the task is reset to `pending` (not resolved). The callback still fires, which is correct (it triggers `tick()`), but the naming is misleading. Consider `onSignalProcessed` or `onTaskStateChanged`. `[fixable]`

server/signal-processor.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (interval && gateConfig.type !== 'human_approval') {
+      intervalId = setInterval(() => this.poll(taskId), interval);
+      log.info('started polling', { taskId, type: gateConfig.type, intervalMs: interval });
+    } else {
+      log.info('watching for manual signal', { taskId, type: gateConfig.type });
+    }
+
+    this.watches.set(taskId, { taskId, gateConfig, intervalId });
+  }
+
```
dimakis, line 61: 🔵 **unsafe_assumptions**: Polling intervals for `gh_ci` (30s) and `gh_review` (60s) use `setInterval` with `execFile('gh', ...)`. If the GitHub CLI is slow or rate-limited, polls can stack up since there's no guard against concurrent poll execution for the same task. Consider using `setTimeout` chains or a mutex per watch entry. `[fixable]`

server/index.ts
```diff
@@ -107,9 +111,22 @@ setTaskBroadcast((event) => {
   connRegistry.broadcastAll(event as Record<string, unknown>);
 });
 
+// --- Workflow layer ---
+const wfTemplateStore = new WorkflowTemplateStore(join(BASE_REPO, '.mitzo', 'tasks.db'));
```
dimakis: 🟡 **unsafe_assumptions**: `WorkflowTemplateStore` opens a second `better-sqlite3` connection to the same `tasks.db` file that `TaskStore` already has open. While WAL mode supports concurrent readers, two in-process write connections to the same file can cause SQLITE_BUSY errors under contention. Consider either sharing the db instance or using a separate database file for templates. `[fixable]`

dimakis (commented):
## Centaur Review

Found **10** issue(s) (1 critical) (5 warning).


### `server/app.ts`
_Solid feature addition with good test coverage on the core stores and orchestrator dispatch, but the `require()` call in ESM is a runtime bug, the signal REST endpoint lacks task-type validation, and the new API routes have no integration tests._

- 🔴 **bugs** (L719): `require('./workflow-templates.js')` is a CommonJS call inside an ESM module. All other imports in this file use ESM `import` with `.js` extensions. This will fail at runtime in a strict ESM environment. `instantiateTemplate` should be imported at the top of the file alongside the existing `import type { WorkflowTemplateStore }` on line 63. `[fixable]`
- 🟡 **missing_tests** (L674): No REST endpoint tests for the five new routes: GET/GET/:id/DELETE /api/templates, POST /api/workflows/instantiate, POST /api/tasks/:id/signal. The existing `task-routes.test.ts` covers task CRUD but none of these new endpoints. Error paths (503 when store not initialized, 404 for missing resources, 400 for invalid payloads) are untested. `[fixable]`
- 🟡 **regressions** (L674): There is no `POST /api/templates` endpoint for creating templates. The `WorkflowTemplateStore` has a `create()` method and `GET`/`DELETE` routes exist, but there is no way to create templates via the REST API. The only consumer is `instantiateTemplate` which reads existing templates. Templates can never be created through the API. `[fixable]`
- 🟡 **unsafe_assumptions** (L752): The `/api/tasks/:id/signal` endpoint calls `signalProcessor.resolveSignal()` without checking whether the task is actually a `wait_for_signal` task or whether it's in `active` status. Any authenticated user can resolve any task as a signal, potentially corrupting non-signal tasks. Should validate `task.stageType === 'wait_for_signal'` and `task.status === 'active'`. `[fixable]`


_Solid feature addition with good test coverage on the core stores and orchestrator dispatch, but the `require()` call in ESM is a runtime bug, the signal REST endpoint lacks task-type validation, and the new API routes have no integration tests._

- 🟡 **bugs** (L166): Dead code: the first loop (lines 166-175) sets `found = true` but never uses it and has no side effects. The actual work is done by the second loop (lines 179-185). The dead loop and its misleading comments should be removed. `[fixable]`
- 🔵 **style** (L106): The `onSignalResolved` callback name suggests it's a notification, but in `handleFailure` with retries the task is reset to `pending` (not resolved). The callback still fires, which is correct (it triggers `tick()`), but the naming is misleading. Consider `onSignalProcessed` or `onTaskStateChanged`. `[fixable]`
- 🔵 **unsafe_assumptions** (L61): Polling intervals for `gh_ci` (30s) and `gh_review` (60s) use `setInterval` with `execFile('gh', ...)`. If the GitHub CLI is slow or rate-limited, polls can stack up since there's no guard against concurrent poll execution for the same task. Consider using `setTimeout` chains or a mutex per watch entry. `[fixable]`


### `server/index.ts`
_Solid feature addition with good test coverage on the core stores and orchestrator dispatch, but the `require()` call in ESM is a runtime bug, the signal REST endpoint lacks task-type validation, and the new API routes have no integration tests._

- 🟡 **unsafe_assumptions** (L115): `WorkflowTemplateStore` opens a second `better-sqlite3` connection to the same `tasks.db` file that `TaskStore` already has open. While WAL mode supports concurrent readers, two in-process write connections to the same file can cause SQLITE_BUSY errors under contention. Consider either sharing the db instance or using a separate database file for templates. `[fixable]`


_Solid feature addition with good test coverage on the core stores and orchestrator dispatch, but the `require()` call in ESM is a runtime bug, the signal REST endpoint lacks task-type validation, and the new API routes have no integration tests._

- 🔵 **style** (L12): `StageType`, `Task`, `TaskStatus`, `SessionPolicy`, and `LoopStatus` are now defined identically in both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`. These should live in `@mitzo/protocol` as a single source of truth to prevent future divergence. This is a pre-existing pattern but the PR widens the duplication surface by adding `StageType` and 6 new `Task` fields to both. `[fixable]`


_Solid feature addition with good test coverage on the core stores and orchestrator dispatch, but the `require()` call in ESM is a runtime bug, the signal REST endpoint lacks task-type validation, and the new API routes have no integration tests._

- 🔵 **missing_tests**: The `checkGate` function and its sub-checkers (`checkGhCi`, `checkGhReview`, `checkCentaurReview`, `checkCompound`) are exported but have no unit tests. These are the core gate evaluation functions. While they shell out to `gh` CLI, they could be tested with mocked `execFile`. `[fixable]`

frontend/src/types/task.ts
```diff
@@ -9,6 +9,8 @@ export type TaskStatus =
 
 export type SessionPolicy = 'reuse' | 'spawn' | 'auto';
 
+export type StageType = 'agent_work' | 'wait_for_signal' | 'human_review';
```
dimakis, line 12: 🔵 **style**: `StageType`, `Task`, `TaskStatus`, `SessionPolicy`, and `LoopStatus` are now defined identically in both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`. These should live in `@mitzo/protocol` as a single source of truth to prevent future divergence. This is a pre-existing pattern but the PR widens the duplication surface by adding `StageType` and 6 new `Task` fields to both. `[fixable]`

server/app.ts
dimakis, line 676: 🔴 **bugs**: GET /api/templates returns `{ templates: [...] }` but frontend WorkflowCreateForm.tsx line 33 expects a plain array (`data: WorkflowTemplate[]`). `data.length` will be `undefined` and `data[0]` will be `undefined`, so the template selector will always be empty. Fix: return `res.json(templateStore.list())` or unwrap on the frontend. `[fixable]`

server/app.ts
dimakis: 🔴 **bugs**: `require('./workflow-templates.js')` in an ESM project (`"type": "module"` in package.json). This will throw `ReferenceError: require is not defined` at runtime. Use a top-level `import { instantiateTemplate } from './workflow-templates.js'` instead — the function is already importable and `templateStore` is already imported as a type. `[fixable]`

server/signal-processor.ts
dimakis: 🟡 **style**: Dead code: the first loop (lines 166-175) sets `found = true` but never uses it, and the loop body after the `continue` is empty. The second loop (lines 179-185) is the one that actually does the work. Delete the first loop entirely. `[fixable]`

server/signal-processor.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// --- Gate checker implementations ---
+
+export async function checkGate(config: GateConfig): Promise<GateResult> {
+  switch (config.type) {
+    case 'gh_ci':
+      return checkGhCi(config as GateConfig & { repo: string; pr: number | string });
+    case 'gh_review':
+      return checkGhReview(config as GateConfig & { repo: string; pr: number | string });
```
dimakis, line 187: 🟡 **unsafe_assumptions**: `checkGhCi` calls `execFileAsync('gh', [...])` which assumes the `gh` CLI is installed and authenticated. If `gh` is not on PATH or not authenticated, every poll will log an error and return `{ resolved: false }` indefinitely. Consider a startup check or a more informative error on first failure so the user knows the gate can never resolve. `[fixable]`

server/signal-processor.ts
```diff
@@ -0,0 +1,297 @@
...
+      log.warn('already watching task', { taskId });
+      return;
+    }
+
+    const interval = POLL_INTERVALS[gateConfig.type];
+    let intervalId: NodeJS.Timeout | null = null;
+
+    if (interval && gateConfig.type !== 'human_approval') {
+      intervalId = setInterval(() => this.poll(taskId), interval);
+      log.info('started polling', { taskId, type: gateConfig.type, intervalMs: interval });
```
dimakis, line 54: 🔵 **unsafe_assumptions**: Polling intervals are set via `setInterval` but never cleaned up if the task is externally deleted or the orchestrator stops. The `unwatchAll()` in shutdown handles graceful exit, but if a goal is stopped mid-run via `orchestrator.stop()`, active signal watches for that goal's tasks are leaked. Consider having the orchestrator call `unwatch()` for signal tasks when stopping. `[fixable]`

server/task-store.ts
```diff
@@ -102,13 +130,36 @@ const SCHEMA = `
     claimed_at      REAL,
```
dimakis, line 130: 🔵 **regressions**: The CREATE TABLE schema now includes the workflow columns, but the MIGRATIONS array also adds them. For new databases, both paths run. This is safe because `CREATE TABLE IF NOT EXISTS` runs first and includes the columns, so the migration check (`pragma_table_info`) finds them and skips the ALTER. But it's fragile — if another migration is added later, the schema and migrations could drift. Consider a comment noting this coupling.

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Two critical bugs: GET /api/templates response shape mismatches the frontend (will render an empty template list), and `require()` in an ESM module will crash at runtime. Dead code in signal-processor and missing test coverage for gate checkers and the new frontend form._

- 🔴 **bugs** (L674): GET /api/templates returns `{ templates: [...] }` but frontend WorkflowCreateForm.tsx line 33 expects a plain array (`data: WorkflowTemplate[]`). `data.length` will be `undefined` and `data[0]` will be `undefined`, so the template selector will always be empty. Fix: return `res.json(templateStore.list())` or unwrap on the frontend. `[fixable]`
- 🔴 **bugs** (L719): `require('./workflow-templates.js')` in an ESM project (`"type": "module"` in package.json). This will throw `ReferenceError: require is not defined` at runtime. Use a top-level `import { instantiateTemplate } from './workflow-templates.js'` instead — the function is already importable and `templateStore` is already imported as a type. `[fixable]`


_Two critical bugs: GET /api/templates response shape mismatches the frontend (will render an empty template list), and `require()` in an ESM module will crash at runtime. Dead code in signal-processor and missing test coverage for gate checkers and the new frontend form._

- 🟡 **style** (L166): Dead code: the first loop (lines 166-175) sets `found = true` but never uses it, and the loop body after the `continue` is empty. The second loop (lines 179-185) is the one that actually does the work. Delete the first loop entirely. `[fixable]`
- 🟡 **unsafe_assumptions** (L201): `checkGhCi` calls `execFileAsync('gh', [...])` which assumes the `gh` CLI is installed and authenticated. If `gh` is not on PATH or not authenticated, every poll will log an error and return `{ resolved: false }` indefinitely. Consider a startup check or a more informative error on first failure so the user knows the gate can never resolve. `[fixable]`
- 🔵 **unsafe_assumptions** (L54): Polling intervals are set via `setInterval` but never cleaned up if the task is externally deleted or the orchestrator stops. The `unwatchAll()` in shutdown handles graceful exit, but if a goal is stopped mid-run via `orchestrator.stop()`, active signal watches for that goal's tasks are leaked. Consider having the orchestrator call `unwatch()` for signal tasks when stopping. `[fixable]`


_Two critical bugs: GET /api/templates response shape mismatches the frontend (will render an empty template list), and `require()` in an ESM module will crash at runtime. Dead code in signal-processor and missing test coverage for gate checkers and the new frontend form._

- 🟡 **missing_tests**: No tests for the gate checker functions (`checkGate`, `checkGhCi`, `checkGhReview`, `checkCentaurReview`, `checkCompound`). These execute external commands (`gh pr checks`) and make HTTP calls — they should have unit tests with mocked child_process/fetch to verify parsing of CI states, review decisions, and compound logic. `[fixable]`


### `frontend/src/components/WorkflowCreateForm.tsx`
_Two critical bugs: GET /api/templates response shape mismatches the frontend (will render an empty template list), and `require()` in an ESM module will crash at runtime. Dead code in signal-processor and missing test coverage for gate checkers and the new frontend form._

- 🟡 **missing_tests**: No test file exists for WorkflowCreateForm. The project's existing components (TaskNode, TaskBoard) all have test files. This form has non-trivial logic (template loading, variable initialization, API submission) that should be tested — especially given the API contract mismatch already identified. `[fixable]`


### `server/task-store.ts`
_Two critical bugs: GET /api/templates response shape mismatches the frontend (will render an empty template list), and `require()` in an ESM module will crash at runtime. Dead code in signal-processor and missing test coverage for gate checkers and the new frontend form._

- 🔵 **regressions** (L130): The CREATE TABLE schema now includes the workflow columns, but the MIGRATIONS array also adds them. For new databases, both paths run. This is safe because `CREATE TABLE IF NOT EXISTS` runs first and includes the columns, so the migration check (`pragma_table_info`) finds them and skips the ALTER. But it's fragile — if another migration is added later, the schema and migrations could drift. Consider a comment noting this coupling.

PR #253, opened 2026-04-20
fix: atomic session-index writes, inbox pruning, type-safe loop status

dimakis:
## Summary

Addresses remaining review items from #230 that weren't covered by #233 and #234:

- **Atomic session-index writes**: `upsertEntry` writes to a temp file then renames, preventing corrupt YAML from concurrent session mutations (e.g. two sessions starting simultaneously).
- **InboxView pendingRemovals leak**: Prunes entries no longer present server-side so the optimistic-removal set converges instead of growing monotonically.
- **Type-safe getLoopStatus**: Expanded API client return type eliminates verbose `Record<string, unknown>` casts in `loadLoopStatus`.
- **wtId format safety**: `registerSession` validates the expected `YYYY-MM-DD-hex` format instead of blindly slicing/splitting.
- **updateSessionTitle double-read**: Single `upsertEntry` call with conditional fields instead of read-then-upsert.

## Test plan

- [x] 1987 tests pass locally
- [x] Type-check clean
- [x] Full build clean

### DIFF
--- a/frontend/src/pages/InboxView.tsx
+++ b/frontend/src/pages/InboxView.tsx
@@ -186,7 +186,17 @@ export function InboxView() {
 
   // When the store's inbox updates (via WS), sync to local state — but
   // filter out items that were optimistically removed to prevent flicker.
+  // Prune pendingRemovals to items still present server-side so the set
+  // doesn't grow monotonically and hide legitimately re-added items.
   useEffect(() => {
+    const serverFilenames = new Set((storeInbox as InboxItem[]).map((i) => i.filename));
+    setPendingRemovals((prev) => {
+      const pruned = new Set<string>();
+      for (const f of prev) {
+        if (serverFilenames.has(f)) pruned.add(f);
+      }
+      return pruned.size === prev.size ? prev : pruned;
+    });
     const filtered = (storeInbox as InboxItem[]).filter(
       (item) => !pendingRemovals.has(item.filename),
     );
--- a/packages/client/src/api-client.ts
+++ b/packages/client/src/api-client.ts
@@ -274,7 +274,14 @@ export class MitzoApiClient {
     );
   }
 
-  async getLoopStatus(): Promise<{ state: string; goalId: string | null }> {
+  async getLoopStatus(): Promise<{
+    state: string;
+    goalId: string | null;
+    activeTaskId?: string | null;
+    progress?: { completed: number; total: number } | null;
+    specMode?: boolean;
+    awaitingApproval?: boolean;
+  }> {
     const res = await this.assertOk(
       await this.fetch('/api/loop/status', { credentials: 'include' }),
     );
--- a/server/session-index.ts
+++ b/server/session-index.ts
@@ -1,5 +1,6 @@
-import { readFileSync, writeFileSync, mkdirSync, existsSync } from 'fs';
+import { readFileSync, writeFileSync, mkdirSync, existsSync, renameSync } from 'fs';
 import { join, dirname } from 'path';
+import { randomBytes } from 'crypto';
 import yaml from 'js-yaml';
 
 export interface SessionIndexRepo {
@@ -70,7 +71,10 @@ export function upsertEntry(
   }
 
   const out: IndexFile = { sessions: entries };
-  writeFileSync(path, yaml.dump(out, { lineWidth: 120, noRefs: true }));
+  const content = yaml.dump(out, { lineWidth: 120, noRefs: true });
+  const tmp = `${path}.${randomBytes(4).toString('hex')}.tmp`;
+  writeFileSync(tmp, content);
+  renameSync(tmp, path);
 }
 
 /**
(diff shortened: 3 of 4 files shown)


packages/client/src/api-client.ts
```diff
@@ -274,7 +274,14 @@ export class MitzoApiClient {
     );
   }
 
-  async getLoopStatus(): Promise<{ state: string; goalId: string | null }> {
+  async getLoopStatus(): Promise<{
+    state: string;
+    goalId: string | null;
+    activeTaskId?: string | null;
+    progress?: { completed: number; total: number } | null;
```
dimakis, line 281: 🟡 **bugs**: Type declares `completed` but server returns `done` (`task-orchestrator.ts:322` returns `{ done, total }`). Currently harmless because `store.ts` casts to `LoopStatus['progress']` (which uses `done`), but any future consumer of `getLoopStatus()` that accesses `.completed` will get `undefined`. Fix: change `completed` → `done`. `[fixable]`

dimakis (commented):
## Centaur Review

_One inaccurate type (`completed` vs `done` in api-client progress field) that is currently masked by a cast; the rest are minor style and test coverage gaps — overall a clean fix PR._

- 🟡 **bugs** (L281): Type declares `completed` but server returns `done` (`task-orchestrator.ts:322` returns `{ done, total }`). Currently harmless because `store.ts` casts to `LoopStatus['progress']` (which uses `done`), but any future consumer of `getLoopStatus()` that accesses `.completed` will get `undefined`. Fix: change `completed` → `done`. `[fixable]`


### `frontend/src/pages/InboxView.tsx`
_One inaccurate type (`completed` vs `done` in api-client progress field) that is currently masked by a cast; the rest are minor style and test coverage gaps — overall a clean fix PR._

- 🔵 **style** (L200): The filter on line 200 reads `pendingRemovals` from the closure, but `setPendingRemovals` on line 193 is batched — if pruning shrinks the set, the filter still uses the stale value for this render. The effect self-corrects on re-run (since `pendingRemovals` is a dependency), but the intermediate render is wasted. Computing a local `effectivePendingRemovals` set and using it for both the `setPendingRemovals` call and the filter would avoid the extra cycle. `[fixable]`
- 🔵 **missing_tests** (L191): No tests exist for InboxView. The new pruning logic — especially the interaction between pruning and the optimistic-removal filter — is subtle enough to warrant a unit test for the pruning/filter logic extracted into a pure function.


### `server/session-index.ts`
_One inaccurate type (`completed` vs `done` in api-client progress field) that is currently masked by a cast; the rest are minor style and test coverage gaps — overall a clean fix PR._

- 🔵 **unsafe_assumptions** (L76): If `renameSync` throws (e.g., permissions error), the tmp file is leaked. A `try/finally` that unlinks the tmp on failure would be more robust, though the practical risk is low since both files are in the same directory. `[fixable]`
- 🔵 **missing_tests** (L75): The atomic write (write-to-tmp + rename) is a correctness improvement but has no dedicated test. Consider a test that verifies the target file is not corrupted if `writeFileSync` throws partway through (e.g., by mocking fs).

frontend/src/pages/InboxView.tsx
```diff
@@ -186,7 +186,17 @@ export function InboxView() {
...
   useEffect(() => {
+    const serverFilenames = new Set((storeInbox as InboxItem[]).map((i) => i.filename));
+    setPendingRemovals((prev) => {
+      const pruned = new Set<string>();
+      for (const f of prev) {
+        if (serverFilenames.has(f)) pruned.add(f);
+      }
+      return pruned.size === prev.size ? prev : pruned;
+    });
     const filtered = (storeInbox as InboxItem[]).filter(
```
dimakis, line 200: 🔵 **style**: The filter on line 200 reads `pendingRemovals` from the closure, but `setPendingRemovals` on line 193 is batched — if pruning shrinks the set, the filter still uses the stale value for this render. The effect self-corrects on re-run (since `pendingRemovals` is a dependency), but the intermediate render is wasted. Computing a local `effectivePendingRemovals` set and using it for both the `setPendingRemovals` call and the filter would avoid the extra cycle. `[fixable]`

frontend/src/pages/InboxView.tsx
```diff
@@ -186,7 +186,17 @@ export function InboxView() {
 
   // When the store's inbox updates (via WS), sync to local state — but
   // filter out items that were optimistically removed to prevent flicker.
+  // Prune pendingRemovals to items still present server-side so the set
+  // doesn't grow monotonically and hide legitimately re-added items.
   useEffect(() => {
```
dimakis, line 191: 🔵 **missing_tests**: No tests exist for InboxView. The new pruning logic — especially the interaction between pruning and the optimistic-removal filter — is subtle enough to warrant a unit test for the pruning/filter logic extracted into a pure function.

server/session-index.ts
```diff
@@ -70,7 +71,10 @@ export function upsertEntry(
   }
 
   const out: IndexFile = { sessions: entries };
-  writeFileSync(path, yaml.dump(out, { lineWidth: 120, noRefs: true }));
+  const content = yaml.dump(out, { lineWidth: 120, noRefs: true });
+  const tmp = `${path}.${randomBytes(4).toString('hex')}.tmp`;
+  writeFileSync(tmp, content);
```
dimakis, line 76: 🔵 **unsafe_assumptions**: If `renameSync` throws (e.g., permissions error), the tmp file is leaked. A `try/finally` that unlinks the tmp on failure would be more robust, though the practical risk is low since both files are in the same directory. `[fixable]`

server/session-index.ts
```diff
@@ -70,7 +71,10 @@ export function upsertEntry(
   }
 
   const out: IndexFile = { sessions: entries };
-  writeFileSync(path, yaml.dump(out, { lineWidth: 120, noRefs: true }));
+  const content = yaml.dump(out, { lineWidth: 120, noRefs: true });
+  const tmp = `${path}.${randomBytes(4).toString('hex')}.tmp`;
```
dimakis, line 75: 🔵 **missing_tests**: The atomic write (write-to-tmp + rename) is a correctness improvement but has no dedicated test. Consider a test that verifies the target file is not corrupted if `writeFileSync` throws partway through (e.g., by mocking fs).

PR #254, opened 2026-04-20
fix(ui): fix PermissionBanner buttons unresponsive on iOS

dimakis:
## Summary

- **iOS hit-test bug**: PermissionBanner used `position: fixed` + CSS `animation: forwards` which causes iOS WebKit to keep the touch hit-test region at the pre-animation position (off-screen). Buttons were visually present but untouchable. Replaced with state-driven CSS transition.
- **Silent WS send failure**: `respondToPermission()` ignored `connection.send()` return value. On WS disconnect (common during iOS background/resume), the response was silently lost but the banner cleared anyway — user had to wait for 120s auto-deny. Now only clears the banner if the message was actually sent.
- **Visual polish**: Added hover/active states, cursor:pointer, replaced hard-coded greens with `var(--success)` design system variable.

## Test plan

- [ ] Run `npx vitest run frontend/src/components/__tests__/PermissionBanner.test.tsx` — 9 tests pass
- [ ] Run `npx vitest run packages/client/__tests__/store.test.ts` — 48 tests pass
- [ ] On iOS: trigger a permission request, verify buttons respond to taps
- [ ] Verify touch feedback (scale + darken on press)
- [ ] Test with WS disconnect: banner should stay visible if send fails

### DIFF
--- a/frontend/src/components/__tests__/PermissionBanner.test.tsx
+++ b/frontend/src/components/__tests__/PermissionBanner.test.tsx
@@ -75,15 +75,15 @@ describe('PermissionBanner', () => {
   });
 
   it('adds perm-banner--visible class after mount', () => {
-    // rAF fires synchronously in jsdom with fake timers
-    vi.useRealTimers();
     const rafSpy = vi.spyOn(globalThis, 'requestAnimationFrame').mockImplementation((cb) => {
       cb(0);
       return 0;
     });
-    const { container } = render(<PermissionBanner {...defaultProps} />);
-    expect(container.querySelector('.perm-banner--visible')).toBeTruthy();
-    rafSpy.mockRestore();
-    vi.useFakeTimers();
+    try {
+      const { container } = render(<PermissionBanner {...defaultProps} />);
+      expect(container.querySelector('.perm-banner--visible')).toBeTruthy();
+    } finally {
+      rafSpy.mockRestore();
+    }
   });
 });
--- a/frontend/src/components/__tests__/PermissionBanner.test.tsx
+++ b/frontend/src/components/__tests__/PermissionBanner.test.tsx
@@ -66,4 +66,24 @@ describe('PermissionBanner', () => {
     });
     expect(onRespond).toHaveBeenCalledWith('p1', 'deny', 'Bash');
   });
+
+  it('renders all action buttons with correct CSS classes', () => {
+    const { container } = render(<PermissionBanner {...defaultProps} />);
+    expect(container.querySelector('.perm-banner-btn--once')).toBeTruthy();
+    expect(container.querySelector('.perm-banner-btn--always')).toBeTruthy();
+    expect(container.querySelector('.perm-banner-btn--deny')).toBeTruthy();
+  });
+
+  it('adds perm-banner--visible class after mount', () => {
+    // rAF fires synchronously in jsdom with fake timers
+    vi.useRealTimers();
+    const rafSpy = vi.spyOn(globalThis, 'requestAnimationFrame').mockImplementation((cb) => {
+      cb(0);
+      return 0;
+    });
+    const { container } = render(<PermissionBanner {...defaultProps} />);
+    expect(container.querySelector('.perm-banner--visible')).toBeTruthy();
+    rafSpy.mockRestore();
+    vi.useFakeTimers();
+  });
 });
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -302,7 +302,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         decision,
       });
       if (sent) {
-        set({ permissions: { pending: null } });
+        set((s) => ({
+          permissions: { pending: null },
+          messages: { ...s.messages, permission: null },
+        }));
       }
       // If not sent, leave the banner visible so user can retry
     },
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -295,13 +295,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     respondToPermission(permId: string, decision: 'once' | 'always' | 'deny') {
-      connection.send({
+      const sent = connection.send({
         type: 'permission_response',
         ...(parserState.currentSessionId ? { sessionId: parserState.currentSessionId } : {}),
         permId,
         decision,
       });
-      set({ permissions: { pending: null } });
+      if (sent) {
+        set({ permissions: { pending: null } });
+      }
+      // If not sent, leave the banner visible so user can retry
     },
 
     setMode(mode: MitzoMode) {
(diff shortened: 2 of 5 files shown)


dimakis (commented):
## Centaur Review

_Solid iOS fix — correct diagnosis of the WebKit hit-test bug and clean transition to state-driven animations. The store-level `send()` guard is a good resilience improvement but needs test coverage for the failure path._

- 🟡 **missing_tests** (L304): The new `if (sent)` guard in `respondToPermission` is not tested. Existing store tests use a connected WebSocket (via `createReadyStore`), so `send()` always returns `true`. A test should verify that when `send()` returns `false` (e.g., after `ws.close()`), `permissions.pending` is NOT cleared. `[fixable]`


_Solid iOS fix — correct diagnosis of the WebKit hit-test bug and clean transition to state-driven animations. The store-level `send()` guard is a good resilience improvement but needs test coverage for the failure path._

- 🟡 **regressions**: The `.perm-banner-dismiss` class and `perm-slide-down` keyframes were removed, eliminating the slide-down exit animation. Although `perm-banner-dismiss` was never applied in the component code (confirmed by grep), if this was dead CSS then removing it is fine. However, if there was an intent to add a dismiss animation later, the approach for it would need to change. As-is, the banner now disappears instantly on dismiss rather than sliding down — verify this is acceptable UX. `[fixable]`
- 🔵 **style**: The `.perm-banner-btn--always` hover/active states use hardcoded `rgba(76, 175, 80, ...)` values that derive from the same color as `var(--success)` (#4caf50). Since the base style already references `var(--success)`, the hover/active states could use `color-mix()` or at least a comment noting these are `--success` variants, so they stay in sync if the variable changes. `[fixable]`


### `frontend/src/components/__tests__/PermissionBanner.test.tsx`
_Solid iOS fix — correct diagnosis of the WebKit hit-test bug and clean transition to state-driven animations. The store-level `send()` guard is a good resilience improvement but needs test coverage for the failure path._

- 🔵 **style** (L76): The test manually toggles between `vi.useRealTimers()` and `vi.useFakeTimers()` and mocks `requestAnimationFrame`. This is fragile — if the test file's `beforeEach` changes timer mode, this test could break silently. Consider isolating the timer restore in an `afterEach` or wrapping in a try/finally. `[fixable]`

packages/client/src/store.ts
```diff
@@ -295,13 +295,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
     respondToPermission(permId: string, decision: 'once' | 'always' | 'deny') {
-      connection.send({
+      const sent = connection.send({
         type: 'permission_response',
         ...(parserState.currentSessionId ? { sessionId: parserState.currentSessionId } : {}),
         permId,
         decision,
       });
-      set({ permissions: { pending: null } });
+      if (sent) {
```
dimakis, line 304: 🟡 **missing_tests**: The new `if (sent)` guard in `respondToPermission` is not tested. Existing store tests use a connected WebSocket (via `createReadyStore`), so `send()` always returns `true`. A test should verify that when `send()` returns `false` (e.g., after `ws.close()`), `permissions.pending` is NOT cleared. `[fixable]`

frontend/src/components/__tests__/PermissionBanner.test.tsx
```diff
@@ -66,4 +66,24 @@ describe('PermissionBanner', () => {
...
     expect(onRespond).toHaveBeenCalledWith('p1', 'deny', 'Bash');
   });
+
+  it('renders all action buttons with correct CSS classes', () => {
+    const { container } = render(<PermissionBanner {...defaultProps} />);
+    expect(container.querySelector('.perm-banner-btn--once')).toBeTruthy();
+    expect(container.querySelector('.perm-banner-btn--always')).toBeTruthy();
+    expect(container.querySelector('.perm-banner-btn--deny')).toBeTruthy();
+  });
+
```
dimakis, line 76: 🔵 **style**: The test manually toggles between `vi.useRealTimers()` and `vi.useFakeTimers()` and mocks `requestAnimationFrame`. This is fragile — if the test file's `beforeEach` changes timer mode, this test could break silently. Consider isolating the timer restore in an `afterEach` or wrapping in a try/finally. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid fix for a real iOS WebKit hit-test bug. The approach (state-driven transition replacing CSS animation:forwards) is correct and well-tested. Minor concerns around hardcoded hover colors and the removed dismiss animation._

- 🟡 **regressions**: The `perm-banner-dismiss` class and `perm-slide-down` keyframes were removed, but there is no replacement dismiss/slide-down animation. When the banner is dismissed (permission responded to), it will now disappear instantly instead of animating out. If the old slide-down animation was intentionally used, this is a UX regression. If it was never wired up in the component (no reference found in PermissionBanner.tsx on either branch), this is a no-op cleanup — but worth confirming intent. `[fixable]`
- 🔵 **style**: The `:hover` and `:active` states for `--once`, `--always`, and `--deny` buttons use hardcoded hex colors (#43a047, #388e3c, #d32f2f, #c62828) instead of deriving from the CSS custom properties (`var(--success)`, `var(--danger)`) used for the base state. If the theme variables are ever changed, the hover/active states will be out of sync. Consider using `color-mix()`, `filter: brightness()`, or additional CSS variables for the hover/active variants. `[fixable]`
- 🔵 **style**: The `--always` button hover/active rules have inline comments `/* --success variants */` explaining the relationship to the CSS variable. These comments will rot if the variable value changes and add noise. The relationship is clear from the rgba values matching the `--success` hex (#4caf50 = rgb(76,175,80)). `[fixable]`


_Solid fix for a real iOS WebKit hit-test bug. The approach (state-driven transition replacing CSS animation:forwards) is correct and well-tested. Minor concerns around hardcoded hover colors and the removed dismiss animation._

- 🔵 **missing_tests**: The new disconnected-send test verifies that the permission stays pending when send fails, but doesn't test that a subsequent reconnection + retry succeeds (i.e., that the user can actually recover by retrying after reconnect). This is the full user scenario the fix enables — the happy-path retry after reconnect is untested. `[fixable]`

packages/client/src/store.ts
```diff
@@ -295,13 +295,19 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     respondToPermission(permId: string, decision: 'once' | 'always' | 'deny') {
-      connection.send({
+      const sent = connection.send({
```
dimakis, line 298: 🟡 **regressions**: `respondToPermission()` now clears the banner only when `connection.send()` returns truthy. In this client, `send()` returns false while the socket is reconnecting but still queues the message for later delivery (as your new test in `store.test.ts` documents). That means permission responses made during a transient disconnect will leave the banner visible and allow duplicate taps, while the queued response is still sent on reconnect. The implementation contradicts the intended queued-send behavior and the added test expectations. `[fixable]`

dimakis (commented):
## Codex Review

_PR fixes the iOS hit-target issue in the banner UI, but introduces a client-side regression in queued permission responses during reconnect._

- 🟡 **regressions** (L298): `respondToPermission()` now clears the banner only when `connection.send()` returns truthy. In this client, `send()` returns false while the socket is reconnecting but still queues the message for later delivery (as your new test in `store.test.ts` documents). That means permission responses made during a transient disconnect will leave the banner visible and allow duplicate taps, while the queued response is still sent on reconnect. The implementation contradicts the intended queued-send behavior and the added test expectations. `[fixable]`

PR #257, opened 2026-04-20
fix: zombie session reconnect + pendingSend timeout safety net

dimakis:
## Summary

- **Zombie session fix** (`ws-handler-v2.ts`): `handleReconnect` now cross-references the in-memory SessionRegistry with the durable EventStore. Sessions whose query loop has ended are correctly reported as `running: false` — prevents the client getting stuck with a spinning indicator after laptop sleep/wake.
- **pendingSend timeout** (`client/store.ts`): 5-second safety net for queued messages waiting on `session_end`. If the event never arrives (stale running state after reconnect), the message flushes directly to the WebSocket instead of silently disappearing.
- **Test**: verifies `handleReconnect` reports `running=false` when EventStore says inactive.

## Test plan

- [x] All tests pass
- [ ] Sleep laptop mid-session → wake → verify session shows as not running
- [ ] Send message during stale running state → verify it delivers within 5s

### DIFF
--- a/frontend/src/components/MessageBubble.tsx
+++ b/frontend/src/components/MessageBubble.tsx
@@ -34,10 +34,12 @@ export function UserBubble({ text, images, contextBlocks, timestamp }: UserBubbl
         )}
         {text && <div className="msg-bubble-content">{text}</div>}
       </div>
-      <div className="msg-bubble-footer msg-bubble-footer--user">
-        {time && <span className="msg-timestamp msg-timestamp--user">{time}</span>}
-        {text && <CopyButton text={text} className="msg-bubble-copy msg-bubble-copy--user" />}
-      </div>
+      {(time || text) && (
+        <div className="msg-bubble-footer msg-bubble-footer--user">
+          {time && <span className="msg-timestamp msg-timestamp--user">{time}</span>}
+          {text && <CopyButton text={text} className="msg-bubble-copy msg-bubble-copy--user" />}
+        </div>
+      )}
     </div>
   );
 }
--- a/frontend/src/components/MessageBubble.tsx
+++ b/frontend/src/components/MessageBubble.tsx
@@ -34,7 +34,10 @@ export function UserBubble({ text, images, contextBlocks, timestamp }: UserBubbl
         )}
         {text && <div className="msg-bubble-content">{text}</div>}
       </div>
-      {time && <span className="msg-timestamp msg-timestamp--user">{time}</span>}
+      <div className="msg-bubble-footer msg-bubble-footer--user">
+        {time && <span className="msg-timestamp msg-timestamp--user">{time}</span>}
+        {text && <CopyButton text={text} className="msg-bubble-copy msg-bubble-copy--user" />}
+      </div>
     </div>
   );
 }
--- a/ipv4-preload.cjs
+++ b/ipv4-preload.cjs
@@ -8,8 +8,10 @@
 'use strict';
 
 const dns = require('dns');
-const origLookup = dns.lookup;
 
+dns.setDefaultResultOrder('ipv4first');
+
+const origLookup = dns.lookup;
 dns.lookup = function patchedLookup(hostname, options, callback) {
   if (typeof options === 'function') {
     callback = options;
@@ -25,4 +27,18 @@ dns.lookup = function patchedLookup(hostname, options, callback) {
   }
 
   return origLookup.call(dns, hostname, options, callback);
+};
+
+const origPromiseLookup = dns.promises.lookup;
+dns.promises.lookup = function patchedPromiseLookup(hostname, options) {
+  if (typeof options === 'number') {
+    options = { family: options };
+  }
+  if (!options) options = {};
+
+  if (!options.family) {
+    options = { ...options, family: 4 };
+  }
+
+  return origPromiseLookup.call(dns.promises, hostname, options);
 };
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -138,16 +138,27 @@ function removeTaskFromTree(tasks: Task[], id: string): Task[] {
     });
 }
 
+const PENDING_SEND_TIMEOUT_MS = 5_000;
+
 // ─── Factory ─────────────────────────────────────────────────────────────────
 
 export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStoreState> {
   const api = new MitzoApiClient(options.transport.fetch.bind(options.transport));
   const connection = new MitzoConnection(options.wsConfig);
 
-  const parserState: ProtocolParserState = {
+  const parserState: ProtocolParserState & {
+    pendingSendTimer?: ReturnType<typeof setTimeout>;
+  } = {
     currentSessionId: undefined,
     pendingSend: null,
   };
+
+  function clearPendingSendTimer() {
+    if (parserState.pendingSendTimer) {
+      clearTimeout(parserState.pendingSendTimer);
+      parserState.pendingSendTimer = undefined;
+    }
+  }
 
   const store = createStore<MitzoStoreState>((set, get) => ({
     // ── Initial state ────────────────────────────────────────────────────
@@ -198,6 +209,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     newSession() {
       parserState.currentSessionId = undefined;
       parserState.pendingSend = null;
+      clearPendingSendTimer();
       connection.send({ type: 'switch_session', sessionId: null });
       set({
         sessions: { ...get().sessions, active: null },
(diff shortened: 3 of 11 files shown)


server/chat.ts
```diff
@@ -122,12 +123,24 @@ function send(transport: SessionTransport, data: Record<string, unknown>) {
   if (transport.isOpen()) transport.send(data);
 }
 
+const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
```
dimakis: 🔴 **bugs**: `IPV4_PRELOAD` resolves to `dirname(import.meta.url) + 'ipv4-preload.cjs'`, but `tsconfig.build.json` only includes `server/**/*.ts`. The `.cjs` file is never copied to `dist/`. In production (`npm start` / `npm run deploy`), the path will resolve to `dist/ipv4-preload.cjs` which doesn't exist. NODE_OPTIONS will point to a missing file, causing the SDK child process to fail on startup. Either add a post-build copy step or reference the file from a fixed absolute path. `[fixable]`

server/chat.ts
```diff
@@ -422,15 +435,25 @@ export async function startChat(
...
     if (sessionMeta?.cwd) {
-      baseCwd = sessionMeta.cwd;
-      log.info('resume: using original session CWD', {
-        sessionId: options.resume,
-        cwd: baseCwd,
-      });
+      if (existsSync(sessionMeta.cwd)) {
+        baseCwd = sessionMeta.cwd;
+        log.info('resume: using original session CWD', {
+          sessionId: options.resume,
```
dimakis, line 446: 🔵 **missing_tests**: The CWD fallback logic (existsSync check before restoring original CWD) is untested. A test verifying that a missing worktree CWD falls back to BASE_REPO would guard against regressions in session resumption. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_The `ipv4-preload.cjs` file will be missing from `dist/` in production builds — this is a critical deployment bug. The zombie session fix and pendingSend timeout are well-designed but the timeout path lacks test coverage._

- 🔴 **bugs** (L126): `IPV4_PRELOAD` resolves to `dirname(import.meta.url) + 'ipv4-preload.cjs'`, but `tsconfig.build.json` only includes `server/**/*.ts`. The `.cjs` file is never copied to `dist/`. In production (`npm start` / `npm run deploy`), the path will resolve to `dist/ipv4-preload.cjs` which doesn't exist. NODE_OPTIONS will point to a missing file, causing the SDK child process to fail on startup. Either add a post-build copy step or reference the file from a fixed absolute path. `[fixable]`
- 🔵 **missing_tests** (L447): The CWD fallback logic (existsSync check before restoring original CWD) is untested. A test verifying that a missing worktree CWD falls back to BASE_REPO would guard against regressions in session resumption. `[fixable]`


_The `ipv4-preload.cjs` file will be missing from `dist/` in production builds — this is a critical deployment bug. The zombie session fix and pendingSend timeout are well-designed but the timeout path lacks test coverage._

- 🟡 **bugs** (L37): The footer div renders unconditionally. When both `time` and `text` are falsy (e.g. image-only messages), an empty `<div class="msg-bubble-footer ...">` is rendered, adding `margin-top: var(--space-1)` of dead space. Wrap with `{(time || text) && (<div ...>...</div>)}`. `[fixable]`


_The `ipv4-preload.cjs` file will be missing from `dist/` in production builds — this is a critical deployment bug. The zombie session fix and pendingSend timeout are well-designed but the timeout path lacks test coverage._

- 🟡 **missing_tests** (L260): The 5-second pendingSend timeout safety net is untested. Key scenarios to cover: (1) timeout fires and flushes the queued message when no session_end arrives, (2) timer is cancelled when session_end arrives before timeout, (3) timer is cleared on newSession() and onSessionExpired(). These are timing-sensitive paths — a test with fake timers would prevent regressions. `[fixable]`
- 🔵 **style** (L140): `PENDING_SEND_TIMEOUT_MS` is a magic constant with no way to configure or override it in tests. Consider accepting it as a store option (with default) so tests can use a shorter value without relying on fake timers. `[fixable]`


### `server/ipv4-preload.cjs`
_The `ipv4-preload.cjs` file will be missing from `dist/` in production builds — this is a critical deployment bug. The zombie session fix and pendingSend timeout are well-designed but the timeout path lacks test coverage._

- 🟡 **unsafe_assumptions** (L13): The monkey-patched `dns.lookup` doesn't handle the promise-based overload (`dns.promises.lookup`). If the SDK or undici uses the promise API instead of the callback API, this patch has no effect. The server-side `dns.setDefaultResultOrder('ipv4first')` in index.ts covers that for the parent process, but the child process only gets this preload script. `[fixable]`

frontend/src/components/MessageBubble.tsx
```diff
@@ -34,7 +34,10 @@ export function UserBubble({ text, images, contextBlocks, timestamp }: UserBubbl
         )}
         {text && <div className="msg-bubble-content">{text}</div>}
       </div>
-      {time && <span className="msg-timestamp msg-timestamp--user">{time}</span>}
+      <div className="msg-bubble-footer msg-bubble-footer--user">
```
dimakis: 🟡 **bugs**: The footer div renders unconditionally. When both `time` and `text` are falsy (e.g. image-only messages), an empty `<div class="msg-bubble-footer ...">` is rendered, adding `margin-top: var(--space-1)` of dead space. Wrap with `{(time || text) && (<div ...>...</div>)}`. `[fixable]`

packages/client/src/store.ts
```diff
@@ -245,6 +257,19 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
       if (wasRunning) {
         parserState.pendingSend = msg;
+        // Safety net: if no session_end arrives within 5s (e.g. stale running
```
dimakis, line 260: 🟡 **missing_tests**: The 5-second pendingSend timeout safety net is untested. Key scenarios to cover: (1) timeout fires and flushes the queued message when no session_end arrives, (2) timer is cancelled when session_end arrives before timeout, (3) timer is cleared on newSession() and onSessionExpired(). These are timing-sensitive paths — a test with fake timers would prevent regressions. `[fixable]`

packages/client/src/store.ts
```diff
@@ -138,17 +138,28 @@ function removeTaskFromTree(tasks: Task[], id: string): Task[] {
     });
 }
```
dimakis, line 140: 🔵 **style**: `PENDING_SEND_TIMEOUT_MS` is a magic constant with no way to configure or override it in tests. Consider accepting it as a store option (with default) so tests can use a shorter value without relying on fake timers. `[fixable]`

ipv4-preload.cjs
```diff
@@ -0,0 +1,28 @@
...
+// Works around a known undici bug where autoSelectFamily (RFC 8305) fails to
+// fall back from IPv6 when EHOSTUNREACH is returned on networks with a
+// Tailscale IPv6 ULA address but no routable IPv6 to Google's OAuth servers.
+// See: https://github.com/nodejs/node/issues/48145
+'use strict';
+
+const dns = require('dns');
+const origLookup = dns.lookup;
+
+dns.lookup = function patchedLookup(hostname, options, callback) {
```
dimakis, line 15: 🟡 **unsafe_assumptions**: The monkey-patched `dns.lookup` doesn't handle the promise-based overload (`dns.promises.lookup`). If the SDK or undici uses the promise API instead of the callback API, this patch has no effect. The server-side `dns.setDefaultResultOrder('ipv4first')` in index.ts covers that for the parent process, but the child process only gets this preload script. `[fixable]`

dimakis:
## Centaur Review

_Solid fixes for real production issues (zombie sessions, pendingSend stalls, IPv4 DNS). Two functional bugs found: copy button invisible on mobile due to missing touch-device CSS override, and `switchSession()` not clearing the pendingSend timer which can route a queued message to the wrong session._

- 🟡 **bugs** (L1625): Copy button on user messages is invisible on touch devices. `.msg-bubble-footer--user .msg-bubble-copy` has `opacity: 0` and the only rule that reveals it (`.msg-bubble-group--user:hover ...`) requires hover, which doesn't fire on touch. The `@media (hover: none)` block sets `.msg-bubble-footer--user` to `opacity: 1` but doesn't override the copy button's opacity. For a mobile-first app, add `.msg-bubble-footer--user .msg-bubble-copy { opacity: 1; }` (or similar) inside the `@media (hover: none)` block, matching how assistant copy buttons become visible on touch via the `.msg-bubble-footer` override. `[fixable]`


_Solid fixes for real production issues (zombie sessions, pendingSend stalls, IPv4 DNS). Two functional bugs found: copy button invisible on mobile due to missing touch-device CSS override, and `switchSession()` not clearing the pendingSend timer which can route a queued message to the wrong session._

- 🟡 **bugs** (L184): `switchSession()` does not clear `pendingSend` or `pendingSendTimer`. If a user sends a message while a session is running (queueing it as pendingSend), then switches to a different session within 5s, the timeout fires and sends the queued message into the wrong session. `newSession()` and `onSessionExpired()` both call `clearPendingSendTimer()` — `switchSession()` should do the same. `[fixable]`


_Solid fixes for real production issues (zombie sessions, pendingSend stalls, IPv4 DNS). Two functional bugs found: copy button invisible on mobile due to missing touch-device CSS override, and `switchSession()` not clearing the pendingSend timer which can route a queued message to the wrong session._

- 🔵 **missing_tests**: Tests cover pendingSend timeout cancellation (session_end and newSession paths) but there is no test verifying the timeout actually fires and sends the pending message after 5s when no session_end arrives. This is the core happy-path of the safety net feature. `[fixable]`


### `server/chat.ts`
_Solid fixes for real production issues (zombie sessions, pendingSend stalls, IPv4 DNS). Two functional bugs found: copy button invisible on mobile due to missing touch-device CSS override, and `switchSession()` not clearing the pendingSend timer which can route a queued message to the wrong session._

- 🔵 **unsafe_assumptions** (L125): `IPV4_PRELOAD` is resolved via `process.cwd()` at module load time. This works when the server is started from the repo root but is fragile — launchd `WorkingDirectory` misconfiguration or running from a different directory would silently break DNS patching in child processes. Consider `import.meta.dirname` with a relative `..` join, or `fileURLToPath(import.meta.url)` for a path anchored to the source file rather than the working directory. `[fixable]`
- 🔵 **missing_tests** (L443): The new `existsSync` fallback for resume CWD (when a worktree has been cleaned up) has no test coverage. The `event-store.test.ts` tests verify `getSession` returns the CWD, but the `startChat` branching logic (existsSync → use original vs. fall back to BASE_REPO) is untested. `[fixable]`

PR #258, opened 2026-04-20
feat(isolation): configurable worktree isolation

dimakis:
## Summary
- Adds `isolation` field to `.mitzo.json` — defaults to `true`, set to `false` to disable globally
- Per-session override via WS `send` message — UI lock toggle visible on new sessions
- Precedence chain: per-session toggle > `.mitzo.json` > `WORKTREE_ENABLED` env var > default (true)
- Hot-reloadable: `.mitzo.json` changes picked up within 5s (existing config TTL), no restart needed

## Changes
- **Protocol**: `isolation?: boolean` on v1 + v2 send schemas
- **Server**: `RepoConfig.isolation`, `isIsolationEnabled(perSession?)`, `/api/sessions` respects config
- **Client**: `SendMessageOptions.isolation`, payload builder
- **Frontend**: Lock icon toggle in header (mobile + desktop), new-session only
- **Tests**: 3 new isolation config tests, session-dirs mock updated

## Companion
mgmt hook changes (`.claude/hooks/session-isolate.sh`, `.cursor/hooks/session-isolate.sh`) handle `isolation: false` responses gracefully — will be committed separately in mgmt.

## Test plan
- [ ] Start new session with isolation ON (default) — worktrees created as before
- [ ] Toggle lock off, start session — no worktrees, works in main repo
- [ ] Set `"isolation": false` in `.mitzo.json`, start session — no worktrees
- [ ] Remove `"isolation"` from `.mitzo.json` — defaults back to true
- [ ] Claude Code / Cursor hooks handle disabled isolation without errors


server/chat.ts
```diff
@@ -69,7 +70,14 @@ export function getMcpServerNames(): string[] {
 }
 
 export const BASE_REPO = process.env.REPO_PATH || '';
-const WORKTREE_ENABLED = process.env.WORKTREE_ENABLED !== 'false';
+
+/** Check whether worktree isolation is enabled.
+ *  Per-session override > .mitzo.json config > WORKTREE_ENABLED env var. */
+function isIsolationEnabled(perSession?: boolean): boolean {
```
dimakis: 🟡 **bugs**: `isIsolationEnabled` lets a per-session `true` override bypass the `WORKTREE_ENABLED=false` ceiling. CLAUDE.md documents the env var as "the ceiling — if `false`, worktrees are disabled entirely." The `WORKTREE_ENABLED` check should come before the `perSession` check so the env var always wins when set to `false`. `[fixable]`

server/chat.ts
```diff
@@ -69,7 +70,14 @@ export function getMcpServerNames(): string[] {
 }
 
 export const BASE_REPO = process.env.REPO_PATH || '';
-const WORKTREE_ENABLED = process.env.WORKTREE_ENABLED !== 'false';
+
+/** Check whether worktree isolation is enabled.
+ *  Per-session override > .mitzo.json config > WORKTREE_ENABLED env var. */
```
dimakis: 🟡 **bugs**: Docstring says priority is "Per-session override > .mitzo.json config > WORKTREE_ENABLED env var" but the code implements "Per-session > WORKTREE_ENABLED > .mitzo.json". The code order is correct per CLAUDE.md; the comment is backwards and will mislead future readers. `[fixable]`

server/chat.ts
dimakis: 🟡 **missing_tests**: `isIsolationEnabled()` has no unit tests. The three-way priority cascade (per-session, env var, config) and its edge cases (e.g., `perSession=true` with `WORKTREE_ENABLED=false`) should be tested directly — especially since the current implementation has a ceiling-bypass bug. `[fixable]`

server/app.ts
```diff
@@ -374,9 +374,16 @@ app.post('/api/sessions', (req, res) => {
     return;
   }
 
-  const wtId = generateWtId();
   const config = getRepoConfig();
 
+  if (!config.isolation) {
+    log.info('session isolation disabled via .mitzo.json', { source });
```
dimakis: 🟡 **bugs**: `POST /api/sessions` only checks `config.isolation` but ignores the `WORKTREE_ENABLED` env var. When `WORKTREE_ENABLED=false` and `.mitzo.json` has `isolation: true`, this endpoint creates worktrees while `chat.ts` skips them — inconsistent behavior between the two entry points. `[fixable]`

dimakis: 🟡 **missing_tests**: The new `config.isolation` early-return path in `POST /api/sessions` has no test coverage. A test should verify the `{ sessionId: null, worktrees: null, isolation: false }` response when isolation is disabled. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -54,6 +54,7 @@ export function ChatView() {
   const [mode, setMode] = useState<'ask' | 'agent' | 'auto'>(
     searchParams.get('extraTools') ? 'auto' : 'agent',
   );
+  const [isolation, setIsolation] = useState(true);
```
dimakis, line 67: 🔵 **unsafe_assumptions**: Isolation toggle defaults to `true` without querying the server's config. If `.mitzo.json` has `isolation: false`, the UI shows a locked icon but the server won't create worktrees (since the frontend omits the field when `true`, falling through to server default). Consider reading the effective isolation setting from `GET /api/config` to sync the toggle's initial state. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -132,6 +133,7 @@ export function ChatView() {
       mode,
       cwd: searchParams.get('cwd') ?? undefined,
       extraTools: searchParams.get('extraTools') ?? undefined,
```
dimakis, line 145: 🔵 **style**: The spread `...(!activeSessionId && !isolation ? { isolation: false } : {})` is clever but easy to misread. A plain conditional would be clearer: `isolation: !activeSessionId && !isolation ? false : undefined` — though this is consistent with the existing `cwd`/`extraTools` pattern above it. `[fixable]`

dimakis (commented):
## Centaur Review

Found **7** issue(s) (5 warning).


### `server/chat.ts`
_Solid feature implementation with correct data flow end-to-end. Main issue: `isIsolationEnabled` lets per-session overrides bypass the `WORKTREE_ENABLED=false` ceiling, and `POST /api/sessions` doesn't check the env var at all — both violate the documented ceiling semantics. Fix the priority order and add tests for the cascade._

- 🟡 **bugs** (L76): `isIsolationEnabled` lets a per-session `true` override bypass the `WORKTREE_ENABLED=false` ceiling. CLAUDE.md documents the env var as "the ceiling — if `false`, worktrees are disabled entirely." The `WORKTREE_ENABLED` check should come before the `perSession` check so the env var always wins when set to `false`. `[fixable]`
- 🟡 **bugs** (L75): Docstring says priority is "Per-session override > .mitzo.json config > WORKTREE_ENABLED env var" but the code implements "Per-session > WORKTREE_ENABLED > .mitzo.json". The code order is correct per CLAUDE.md; the comment is backwards and will mislead future readers. `[fixable]`
- 🟡 **missing_tests** (L76): `isIsolationEnabled()` has no unit tests. The three-way priority cascade (per-session, env var, config) and its edge cases (e.g., `perSession=true` with `WORKTREE_ENABLED=false`) should be tested directly — especially since the current implementation has a ceiling-bypass bug. `[fixable]`


### `server/app.ts`
_Solid feature implementation with correct data flow end-to-end. Main issue: `isIsolationEnabled` lets per-session overrides bypass the `WORKTREE_ENABLED=false` ceiling, and `POST /api/sessions` doesn't check the env var at all — both violate the documented ceiling semantics. Fix the priority order and add tests for the cascade._

- 🟡 **bugs** (L380): `POST /api/sessions` only checks `config.isolation` but ignores the `WORKTREE_ENABLED` env var. When `WORKTREE_ENABLED=false` and `.mitzo.json` has `isolation: true`, this endpoint creates worktrees while `chat.ts` skips them — inconsistent behavior between the two entry points. `[fixable]`
- 🟡 **missing_tests** (L380): The new `config.isolation` early-return path in `POST /api/sessions` has no test coverage. A test should verify the `{ sessionId: null, worktrees: null, isolation: false }` response when isolation is disabled. `[fixable]`


_Solid feature implementation with correct data flow end-to-end. Main issue: `isIsolationEnabled` lets per-session overrides bypass the `WORKTREE_ENABLED=false` ceiling, and `POST /api/sessions` doesn't check the env var at all — both violate the documented ceiling semantics. Fix the priority order and add tests for the cascade._

- 🔵 **unsafe_assumptions** (L57): Isolation toggle defaults to `true` without querying the server's config. If `.mitzo.json` has `isolation: false`, the UI shows a locked icon but the server won't create worktrees (since the frontend omits the field when `true`, falling through to server default). Consider reading the effective isolation setting from `GET /api/config` to sync the toggle's initial state. `[fixable]`
- 🔵 **style** (L135): The spread `...(!activeSessionId && !isolation ? { isolation: false } : {})` is clever but easy to misread. A plain conditional would be clearer: `isolation: !activeSessionId && !isolation ? false : undefined` — though this is consistent with the existing `cwd`/`extraTools` pattern above it. `[fixable]`

server/chat.ts
```diff
@@ -122,12 +131,24 @@ function send(transport: SessionTransport, data: Record<string, unknown>) {
   if (transport.isOpen()) transport.send(data);
 }
 
+const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
```
dimakis: 🔴 **bugs**: `IPV4_PRELOAD` resolves relative to `import.meta.url`. In dev (`tsx`), this is `server/chat.ts` so the path resolves to `server/ipv4-preload.cjs` — correct. In production (`node dist/index.js`), it resolves to `dist/ipv4-preload.cjs` — but `tsc` only compiles `.ts` files and does not copy `.cjs` files to `dist/`. The child process will get `NODE_OPTIONS=--require=<nonexistent path>` and fail to start. Fix: resolve relative to the project root (e.g., `join(__filename, '..', '..', 'server', 'ipv4-preload.cjs')`) or add a post-build copy step. `[fixable]`

server/chat.ts
```diff
@@ -422,15 +444,25 @@ export async function startChat(
...
     const sessionMeta = eventStore.getSession(options.resume);
     if (sessionMeta?.cwd) {
-      baseCwd = sessionMeta.cwd;
-      log.info('resume: using original session CWD', {
-        sessionId: options.resume,
-        cwd: baseCwd,
-      });
+      if (existsSync(sessionMeta.cwd)) {
+        baseCwd = sessionMeta.cwd;
+        log.info('resume: using original session CWD', {
```
dimakis: 🟡 **missing_tests**: The new `existsSync` guard for resumed session CWD (falling back to `BASE_REPO` when the original worktree has been cleaned up) is untested. This is a correctness-critical path — if the fallback fails, resumed sessions won't find their conversation history. `[fixable]`

server/app.ts
```diff
@@ -374,9 +375,16 @@ app.post('/api/sessions', (req, res) => {
     return;
   }
 
-  const wtId = generateWtId();
+  if (!isIsolationEnabled()) {
```
dimakis, line 395: 🟡 **missing_tests**: The `POST /api/sessions` endpoint gained an early-return path when `isIsolationEnabled()` is false (returning `{ sessionId: null, worktrees: null, isolation: false }`), but there are no integration tests for this route. An existing session-creation test or a supertest-based test against `app.ts` would catch regressions if this logic changes. `[fixable]`

frontend/src/pages/ChatView.tsx
dimakis, line 67: 🔵 **style**: The isolation toggle state, conditional spread (`...(!activeSessionId && !isolation ? { isolation: false } : {})`), and the button JSX are duplicated verbatim between `ChatView.tsx` and `DesktopChatView.tsx`. Consider extracting the toggle into a shared component (e.g., `IsolationToggle`) or a shared hook to keep the two views in sync. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Core isolation-toggle logic and precedence are sound, but ipv4-preload.cjs will not be found in production builds (path resolved relative to dist/ where tsc doesn't copy .cjs files) — this is a ship-blocker. Secondary gaps: no tests for the new POST /api/sessions early-return path, the resume CWD fallback, or the preload script itself._

- 🔴 **bugs** (L134): `IPV4_PRELOAD` resolves relative to `import.meta.url`. In dev (`tsx`), this is `server/chat.ts` so the path resolves to `server/ipv4-preload.cjs` — correct. In production (`node dist/index.js`), it resolves to `dist/ipv4-preload.cjs` — but `tsc` only compiles `.ts` files and does not copy `.cjs` files to `dist/`. The child process will get `NODE_OPTIONS=--require=<nonexistent path>` and fail to start. Fix: resolve relative to the project root (e.g., `join(__filename, '..', '..', 'server', 'ipv4-preload.cjs')`) or add a post-build copy step. `[fixable]`
- 🟡 **missing_tests** (L455): The new `existsSync` guard for resumed session CWD (falling back to `BASE_REPO` when the original worktree has been cleaned up) is untested. This is a correctness-critical path — if the fallback fails, resumed sessions won't find their conversation history. `[fixable]`


### `server/app.ts`
_Core isolation-toggle logic and precedence are sound, but ipv4-preload.cjs will not be found in production builds (path resolved relative to dist/ where tsc doesn't copy .cjs files) — this is a ship-blocker. Secondary gaps: no tests for the new POST /api/sessions early-return path, the resume CWD fallback, or the preload script itself._

- 🟡 **missing_tests** (L378): The `POST /api/sessions` endpoint gained an early-return path when `isIsolationEnabled()` is false (returning `{ sessionId: null, worktrees: null, isolation: false }`), but there are no integration tests for this route. An existing session-creation test or a supertest-based test against `app.ts` would catch regressions if this logic changes. `[fixable]`


_Core isolation-toggle logic and precedence are sound, but ipv4-preload.cjs will not be found in production builds (path resolved relative to dist/ where tsc doesn't copy .cjs files) — this is a ship-blocker. Secondary gaps: no tests for the new POST /api/sessions early-return path, the resume CWD fallback, or the preload script itself._

- 🔵 **style** (L57): The isolation toggle state, conditional spread (`...(!activeSessionId && !isolation ? { isolation: false } : {})`), and the button JSX are duplicated verbatim between `ChatView.tsx` and `DesktopChatView.tsx`. Consider extracting the toggle into a shared component (e.g., `IsolationToggle`) or a shared hook to keep the two views in sync. `[fixable]`


### `server/ipv4-preload.cjs`
_Core isolation-toggle logic and precedence are sound, but ipv4-preload.cjs will not be found in production builds (path resolved relative to dist/ where tsc doesn't copy .cjs files) — this is a ship-blocker. Secondary gaps: no tests for the new POST /api/sessions early-return path, the resume CWD fallback, or the preload script itself._

- 🔵 **missing_tests**: No tests for `ipv4-preload.cjs`. A unit test that imports the script, calls `dns.lookup` for a hostname, and asserts `family: 4` was passed through would prevent regressions in this workaround. `[fixable]`

PR #259, opened 2026-04-20
fix: prevent session bleed between concurrent clients

dimakis:
## Summary

- **Session bleed fix**: When phone and Theia (desktop) connect simultaneously, sessions no longer leak messages between devices
- **Connection ownership**: Active sessions reject sends from other connections with `active_elsewhere` error; detached sessions still allow takeover
- **Default mode/model**: Default to `agent` mode and `opus 4.6` model

## Root Cause

Three interacting bugs:

1. **`session_id` re-emission on every SDK `result`** — broadcast to all watchers, hijacking `currentSessionId` on clients in "new chat" state (the `session_id` event passed the `isEarlySessionEvent` filter when `currentSessionId` was null)
2. **Stale watches on `switch_session(null)`** — `setActive(null)` didn't unwatch the previous session, so connections accumulated watches and kept receiving broadcast events for sessions they'd moved away from
3. **No connection ownership on `handleSendV2`** — any connection could inject messages into any active SDK driver, causing cross-device message injection

## Changes

| File | Change |
|------|--------|
| `server/query-loop.ts` | Remove redundant `session_id` emission on `result` events |
| `server/ws-handler-v2.ts` | Unwatch previous session on `switch_session(null)`; add connection ownership guard to `handleSendV2` and `handleReconnect` |
| `packages/client/src/protocol-parser.ts` | Handle `active_elsewhere` error gracefully — show message, unblock input |
| `packages/client/src/slices/config.ts` | Default mode `auto` → `agent` |

## Test plan

- [x] All 69 ws-handler-v2 tests pass (including 8 new tests for ownership guard, unwatch-on-clear, and reconnect guard)
- [x] All 220 client package tests pass
- [ ] Manual: open Mitzo on phone + Theia on laptop, verify sessions don't bleed
- [ ] Manual: start session on phone, open same session on Theia → should see `active_elsewhere` error
- [ ] Manual: background phone (kill WS), send from Theia → should take over session
- [ ] Manual: switch to new chat on phone while Theia is active → phone should NOT populate with Theia's messages

### DIFF
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -1558,11 +1558,6 @@ textarea:focus {
   opacity: 1;
 }
 
-.msg-bubble-footer--user {
-  opacity: 0;
-  justify-content: flex-end;
-}
-
 .msg-bubble-group--user:hover .msg-bubble-footer--user {
   opacity: 1;
 }
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -13,7 +13,11 @@
   --accent: #6c63ff;
   --accent-hover: #5a52e0;
   --danger: #f44336;
+  --danger-hover: #d32f2f;
+  --danger-active: #c62828;
   --success: #4caf50;
+  --success-hover: #43a047;
+  --success-active: #388e3c;
   --code-bg: #141425;
   --code-font: 'SF Mono', 'Fira Code', 'Cascadia Code', 'Menlo', monospace;
 
@@ -1558,7 +1562,16 @@ textarea:focus {
   opacity: 1;
 }
 
-.msg-bubble-group--user:hover .msg-bubble-footer--user {
+.msg-bubble-footer--user {
+  opacity: 1;
+}
+
+.msg-bubble-footer--user .msg-bubble-copy {
+  opacity: 0;
+  transition: opacity 0.15s;
+}
+
+.msg-bubble-group--user:hover .msg-bubble-footer--user .msg-bubble-copy {
   opacity: 1;
 }
 
@@ -1617,6 +1630,9 @@ textarea:focus {
   .msg-bubble-footer,
   .msg-bubble-footer--user {
     opacity: 1;
+  }
+  .msg-bubble-footer--user .msg-bubble-copy {
+    opacity: 0.7;
   }
   .code-block-copy {
     opacity: 0.5;
@@ -1974,33 +1990,15 @@ textarea:focus {
   padding-bottom: max(1rem, env(safe-area-inset-bottom));
   z-index: 200;
   transform: translateY(100%);
-  animation: perm-slide-up 0.25s ease-out forwards;
+  transition: transform 0.25s ease-out;
   display: flex;
   flex-direction: column;
   max-height: 60vh;
   overflow: hidden;
 }
 
-@keyframes perm-slide-up {
-  from {
-    transform: translateY(100%);
-  }
-  to {
-    transform: translateY(0);
-  }
-}
-
-.perm-banner-dismiss {
-  animation: perm-slide-down 0.2s ease-in forwards;
-}
-
-@keyframes perm-slide-down {
-  from {
-    transform: translateY(0);
-  }
-  to {
-    transform: translateY(100%);
-  }
+.perm-banner--visible {
+  transform: translateY(0);
 }
 
 .perm-banner-title {
@@ -2066,24 +2064,55 @@ textarea:focus {
   font-weight: 600;
   border-radius: 10px;
   border: none;
+  cursor: pointer;
   touch-action: manipulation;
   -webkit-tap-highlight-color: transparent;
+  transition:
+    background 0.12s,
+    transform 0.1s;
 }
 
 .perm-banner-btn--once {
-  background: #2e7d32;
+  background: var(--success);
   color: #fff;
 }
 
+.perm-banner-btn--once:hover {
+  background: var(--success-hover);
+}
+
+.perm-banner-btn--once:active {
+  background: var(--success-active);
+  transform: scale(0.97);
+}
+
 .perm-banner-btn--always {
-  background: #1b5e20;
+  background: rgba(76, 175, 80, 0.25);
   color: #fff;
-  border: 2px solid #4caf50;
+  border: 2px solid var(--success);
+}
+
+.perm-banner-btn--always:hover {
+  background: rgba(76, 175, 80, 0.35);
+}
+
+.perm-banner-btn--always:active {
+  background: rgba(76, 175, 80, 0.45);
+  transform: scale(0.97);
 }
 
 .perm-banner-btn--deny {
   background: var(--danger);
   color: #fff;
+}
+
+.perm-banner-btn--deny:hover {
+  background: var(--danger-hover);
+}
+
+.perm-banner-btn--deny:active {
+  background: var(--danger-active);
+  transform: scale(0.97);
 }
 
 .perm-banner--elevated {
@@ -2119,7 +2148,7 @@ textarea:focus {
 
 .perm-banner-tier--standard {
   background: rgba(76, 175, 80, 0.15);
-  color: #4caf50;
+  color: var(--success);
 }
 
 /* ===== Chat Input ===== */
@@ -3299,13 +3328,8 @@ textarea:focus {
     border-bottom: none;
   }
 
-  @keyframes perm-slide-up {
-    from {
-      transform: translateX(-50%) translateY(100%);
-    }
-    to {
-      transform: translateX(-50%) translateY(0);
-    }
+  .perm-banner--visible {
+    transform: translateX(-50%) translateY(0);
   }
 }
 
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -526,6 +526,58 @@ textarea:focus {
   opacity: 0.85;
 }
 
+.session-search-bar {
+  position: relative;
+  padding: 0 0 var(--space-2);
+  flex-shrink: 0;
+}
+
+.session-search-input {
+  width: 100%;
+  padding: 0.5rem 2rem 0.5rem 0.75rem;
+  font-size: var(--text-sm);
+  background: var(--bg-secondary);
+  color: var(--text);
+  border: 1px solid var(--border);
+  border-radius: 8px;
+  outline: none;
+  box-sizing: border-box;
+}
+
+.session-search-input:focus {
+  border-color: var(--accent);
+}
+
+.session-search-input::placeholder {
+  color: var(--text-dim);
+}
+
+.session-search-clear {
+  position: absolute;
+  right: 0.5rem;
+  top: 50%;
+  transform: translateY(calc(-50% - var(--space-1)));
+  background: none;
+  border: none;
+  color: var(--text-dim);
+  font-size: 1.2rem;
+  cursor: pointer;
+  padding: 0.2rem 0.4rem;
+  line-height: 1;
+}
+
+.search-result-snippet {
+  font-size: var(--text-xs);
+  color: var(--text-dim);
+  margin-top: 0.2rem;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
+  word-break: break-word;
+}
+
 .refresh-ui-btn {
   font-size: 1.1rem;
   padding: 0.4rem 0.5rem;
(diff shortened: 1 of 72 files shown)


server/chat.ts
```diff
@@ -122,12 +123,24 @@ function send(transport: SessionTransport, data: Record<string, unknown>) {
   if (transport.isOpen()) transport.send(data);
 }
 
+const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
```
dimakis, line 159: 🔴 **bugs**: `IPV4_PRELOAD` is computed relative to `import.meta.url`, which resolves to `dist/chat.js` in production builds. But `ipv4-preload.cjs` is a plain JS file that `tsc` does not compile or copy to `dist/`. The build script (`tsc -p tsconfig.build.json`) only processes `server/**/*.ts`. Result: in production (`node dist/index.js`), `sdkEnv()` sets `NODE_OPTIONS=--require=<missing file>`, and every SDK child process crashes with `MODULE_NOT_FOUND`. Works in dev (`tsx`) but breaks production. Fix: add a `cp server/ipv4-preload.cjs dist/` post-build step in `build:server`, or reference the file from `server/` using a path that's stable across both modes. `[fixable]`

server/chat.ts
```diff
@@ -422,15 +435,25 @@ export async function startChat(
...
   let baseCwd = options.cwd || BASE_REPO;
   if (options.resume && !options.cwd) {
     const sessionMeta = eventStore.getSession(options.resume);
     if (sessionMeta?.cwd) {
-      baseCwd = sessionMeta.cwd;
-      log.info('resume: using original session CWD', {
-        sessionId: options.resume,
-        cwd: baseCwd,
-      });
+      if (existsSync(sessionMeta.cwd)) {
```
dimakis: 🟡 **missing_tests**: The resume CWD fallback logic (if `sessionMeta.cwd` no longer exists, fall back to `BASE_REPO`) is untested. This is a real-world scenario when worktrees get cleaned up. A test asserting that `startChat` falls back to `BASE_REPO` when the stored CWD is missing would prevent regressions. `[fixable]`

frontend/src/styles/global.css
```diff
@@ -1558,10 +1558,28 @@ textarea:focus {
   opacity: 1;
 }
 
+.msg-bubble-footer--user {
```
dimakis, line 1617: 🔵 **style**: `.msg-bubble-footer--user` declares `opacity: 0` and `justify-content: flex-end` which are already set identically in the base `.msg-bubble-footer` class. Since the element uses both classes (`className="msg-bubble-footer msg-bubble-footer--user"`), these declarations are redundant. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -127,14 +127,21 @@ export function handleReconnect(
...
-          reattachChat(found.clientId, conn.transport);
-          log.info('reattached detached session on reconnect', {
-            connectionId,
-            sessionId: entry.sessionId,
-            clientId: found.clientId,
-          });
+        // Only reattach if this connection owns the driver (or the driver
+        // was started by a connection that is now gone). Prevents a
+        // reconnecting phone from hijacking a session Theia is driving.
+        const ownerConnection = found.clientId.split(':')[0];
```
dimakis: 🟡 **unsafe_assumptions**: Ownership is determined by `found.clientId.split(':')[0]`. This relies on the invariant that `connectionId` never contains a colon. Currently safe (format is `conn-<timestamp>-<random>`), but the contract is implicit. If `connectionId` generation ever changes to include a colon, the ownership check silently breaks. Consider extracting a helper (e.g. `getOwnerConnection(clientId)`) with a comment documenting the format invariant, used at both call sites (lines 133 and 344). `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -198,8 +205,13 @@ export async function handleSwitchSession(
   span.setAttribute('ws.sessionId', msg.sessionId ?? 'null');
 
   try {
-    // null sessionId = clear active session
+    // null sessionId = clear active session and stop watching it so
+    // broadcast events from the old session don't leak into a new chat.
     if (msg.sessionId === null) {
```
dimakis, line 247: 🔵 **regressions**: Unwatching the previous active session on `switch_session(null)` means a running background session's events (including permission prompts) will no longer reach this client. Push notifications (ntfy/Pushover) still fire so the user is notified, but this is a behavior change worth documenting — previously, switching away from a session kept the watch active.

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Core session-bleed fix is well-designed with solid test coverage, but ipv4-preload.cjs is missing from the production build pipeline and will break SDK child process spawning in deployed builds._

- 🔴 **bugs** (L126): `IPV4_PRELOAD` is computed relative to `import.meta.url`, which resolves to `dist/chat.js` in production builds. But `ipv4-preload.cjs` is a plain JS file that `tsc` does not compile or copy to `dist/`. The build script (`tsc -p tsconfig.build.json`) only processes `server/**/*.ts`. Result: in production (`node dist/index.js`), `sdkEnv()` sets `NODE_OPTIONS=--require=<missing file>`, and every SDK child process crashes with `MODULE_NOT_FOUND`. Works in dev (`tsx`) but breaks production. Fix: add a `cp server/ipv4-preload.cjs dist/` post-build step in `build:server`, or reference the file from `server/` using a path that's stable across both modes. `[fixable]`
- 🟡 **missing_tests** (L444): The resume CWD fallback logic (if `sessionMeta.cwd` no longer exists, fall back to `BASE_REPO`) is untested. This is a real-world scenario when worktrees get cleaned up. A test asserting that `startChat` falls back to `BASE_REPO` when the stored CWD is missing would prevent regressions. `[fixable]`


_Core session-bleed fix is well-designed with solid test coverage, but ipv4-preload.cjs is missing from the production build pipeline and will break SDK child process spawning in deployed builds._

- 🔵 **style** (L1561): `.msg-bubble-footer--user` declares `opacity: 0` and `justify-content: flex-end` which are already set identically in the base `.msg-bubble-footer` class. Since the element uses both classes (`className="msg-bubble-footer msg-bubble-footer--user"`), these declarations are redundant. `[fixable]`


### `server/ws-handler-v2.ts`
_Core session-bleed fix is well-designed with solid test coverage, but ipv4-preload.cjs is missing from the production build pipeline and will break SDK child process spawning in deployed builds._

- 🟡 **unsafe_assumptions** (L133): Ownership is determined by `found.clientId.split(':')[0]`. This relies on the invariant that `connectionId` never contains a colon. Currently safe (format is `conn-<timestamp>-<random>`), but the contract is implicit. If `connectionId` generation ever changes to include a colon, the ownership check silently breaks. Consider extracting a helper (e.g. `getOwnerConnection(clientId)`) with a comment documenting the format invariant, used at both call sites (lines 133 and 344). `[fixable]`
- 🔵 **regressions** (L210): Unwatching the previous active session on `switch_session(null)` means a running background session's events (including permission prompts) will no longer reach this client. Push notifications (ntfy/Pushover) still fire so the user is notified, but this is a behavior change worth documenting — previously, switching away from a session kept the watch active.

dimakis:
## Centaur Review

### `server/ws-handler-v2.ts`
_Well-structured fix for session bleed via composite clientIds, ownership guards in send/reconnect, and explicit unwatch on session clear. The core mechanism is sound and well-tested. Minor gap: `handleInterruptV2` lacks the same ownership guard, which could re-introduce cross-device bleed through the interrupt path._

- 🔵 **missing_tests** (L86): `getOwnerConnection` has no direct unit tests. It's tested indirectly through handler tests, but edge cases (clientId with no colon, empty string, multiple colons like `conn-1:sess:extra`) rely on integration coverage. A focused unit test would guard against regressions in this parsing logic. `[fixable]`
- 🟡 **regressions** (L464): `handleInterruptV2` does not check connection ownership — any client can interrupt a session driven by another device, and it calls `setActive`/`watch` which shifts event delivery to the interrupting client. This is inconsistent with the ownership model introduced in `handleSendV2`. If a phone interrupts a session the laptop is driving, both connections end up with the same active session, potentially re-introducing bleed. Consider adding the same `getOwnerConnection`/`active_elsewhere` guard, or documenting this as intentional takeover behavior. `[fixable]`


### `server/chat.ts`
_Well-structured fix for session bleed via composite clientIds, ownership guards in send/reconnect, and explicit unwatch on session clear. The core mechanism is sound and well-tested. Minor gap: `handleInterruptV2` lacks the same ownership guard, which could re-introduce cross-device bleed through the interrupt path._

- 🔵 **style** (L471): Double blank line after `const baseCwd = resolveResumeCwd(options);` — the formatter should catch this, but it wasn't flagged. `[fixable]`
- 🔵 **unsafe_assumptions** (L138): `nodeOpts.includes('ipv4-preload')` is a substring match on the full `NODE_OPTIONS` string. While unlikely in practice, a different preload script whose path happens to contain 'ipv4-preload' would cause the guard to skip adding the actual preload. A more precise check like matching the exact filename would be more robust. `[fixable]`


### `server/query-loop.ts`
_Well-structured fix for session bleed via composite clientIds, ownership guards in send/reconnect, and explicit unwatch on session clear. The core mechanism is sound and well-tested. Minor gap: `handleInterruptV2` lacks the same ownership guard, which could re-introduce cross-device bleed through the interrupt path._

- 🔵 **regressions** (L319): Removed `session_id` emit in the `result` handler. Verified this is safe: the primary `session_id` event is still emitted in the `assistant` handler (line 279) when the session is first resolved. The `result` handler's emission was redundant — by the time `result` fires, the `assistant` handler has already emitted `session_id`. The client protocol parser also has fallback paths (`message_end`, `session_end`) that call `onSessionAssigned`. No regression.

packages/client/src/slices/config.ts
```diff
@@ -22,6 +22,6 @@ export interface ConfigState {
 export const INITIAL_CONFIG_STATE: ConfigState = {
   contextBlocks: {},
   skills: [],
-  mode: 'auto',
+  mode: 'agent',
```
dimakis: 🟡 **regressions**: Changing the default mode from `auto` to `agent` is unrelated to the session-bleed fix and changes product behavior globally: new sessions will now prompt for elevated tools instead of auto-allowing them. I couldn't find corresponding PR context or tests covering this intentional behavior change. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -140,14 +153,21 @@ export function handleReconnect(
         }
```
dimakis, line 153: 🟡 **bugs**: `handleReconnect()` now refuses to reattach detached sessions unless the reconnecting socket's `connectionId` matches the original driver prefix. After a real WebSocket disconnect/reconnect, the server assigns a new `connectionId`, so the same client/device will no longer satisfy `isOwner` and its detached session won't be reattached. The new tests only cover synthetic reuse of the same connection id, not the actual reconnect path described in the architecture. `[fixable]`

dimakis (commented):
## Codex Review

### `packages/client/src/slices/config.ts`
_PR mostly addresses cross-client session bleed, but it introduces one likely reconnect regression and includes an unrelated default-mode behavior change._

- 🟡 **regressions** (L25): Changing the default mode from `auto` to `agent` is unrelated to the session-bleed fix and changes product behavior globally: new sessions will now prompt for elevated tools instead of auto-allowing them. I couldn't find corresponding PR context or tests covering this intentional behavior change. `[fixable]`


### `server/ws-handler-v2.ts`
_PR mostly addresses cross-client session bleed, but it introduces one likely reconnect regression and includes an unrelated default-mode behavior change._

- 🟡 **bugs** (L153): `handleReconnect()` now refuses to reattach detached sessions unless the reconnecting socket's `connectionId` matches the original driver prefix. After a real WebSocket disconnect/reconnect, the server assigns a new `connectionId`, so the same client/device will no longer satisfy `isOwner` and its detached session won't be reattached. The new tests only cover synthetic reuse of the same connection id, not the actual reconnect path described in the architecture. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -140,14 +153,26 @@ export function handleReconnect(
...
```
dimakis, line 153: 🟡 **bugs**: `handleReconnect()` now blocks reattach whenever the original owner connection is still registered, even if that connection has already switched away from this session. Because `ConnectionRegistry` keeps idle browser tabs/devices registered until the socket closes, a detached session can become impossible to reclaim from another client after the original owner intentionally leaves it. The guard should check whether the owner connection is still actively attached/watching this session, not merely whether the connection ID exists. `[fixable]`

dimakis (commented):
## Codex Review

### `server/ws-handler-v2.ts`
_PR mostly addresses cross-client session bleed, but the reconnect ownership guard appears too strict and can strand detached sessions when the original connection remains open._

- 🟡 **bugs** (L153): `handleReconnect()` now blocks reattach whenever the original owner connection is still registered, even if that connection has already switched away from this session. Because `ConnectionRegistry` keeps idle browser tabs/devices registered until the socket closes, a detached session can become impossible to reclaim from another client after the original owner intentionally leaves it. The guard should check whether the owner connection is still actively attached/watching this session, not merely whether the connection ID exists. `[fixable]`

PR #263, opened 2026-04-21
fix(chat): preserve primary worktree on session cleanup

dimakis:
## Summary

- **Primary worktree no longer deleted on query loop end or stop** — `cleanupSessionWorktrees` now skips the `'primary'` entry, only removing secondary repo worktrees. Stale GC (96h TTL) handles primary lifecycle.
- **Resume CWD re-creates missing directory** instead of falling back to BASE_REPO — the SDK encodes conversation data by CWD path, so a different CWD means "No conversation found" / "Session expired".

## Root cause

Every `startChat` call cleaned up ALL worktrees (including primary) in its `finally` block. On resume, the original CWD no longer existed, the fallback to BASE_REPO produced a different SDK project encoding, and the SDK returned "No conversation found".

## Test plan

- [x] New unit tests for `cleanupSessionWorktrees` — verifies primary is preserved, secondaries removed
- [x] Existing tests pass (`npx vitest run server/__tests__/chat.test.ts`)
- [ ] Manual: send multiple messages in a Mitzo session — no more "Session expired" on follow-up messages
- [ ] Manual: stop a session mid-response, then send another message — session resumes correctly

### DIFF
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -1,4 +1,5 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
+import type { ManagedSession } from '@mitzo/harness';
 
 describe('chat module exports', () => {
   it('exports expected functions', async () => {
@@ -50,5 +51,66 @@ describe('getMessages', () => {
     const messages = await getMessages('nonexistent-session-id');
     expect(Array.isArray(messages)).toBe(true);
     expect(messages.length).toBe(0);
+  });
+});
+
+describe('cleanupSessionWorktrees', () => {
+  it('skips primary worktree and only removes secondaries', async () => {
+    const { removeWorktree } = await import('../worktree.js');
+    vi.spyOn(await import('../worktree.js'), 'removeWorktree').mockImplementation(() => {});
+
+    const { cleanupSessionWorktrees } = await import('../chat.js');
+
+    const session = {
+      worktreePaths: new Map([
+        ['primary', { path: '/repo/.claude/worktrees/abc', wtId: 'abc' }],
+        ['mitzo', { path: '/tools/mitzo/.claude/worktrees/abc', wtId: 'abc' }],
+        ['centaur', { path: '/projects/centaur/.claude/worktrees/abc', wtId: 'abc' }],
+      ]),
+    } as unknown as ManagedSession;
+
+    cleanupSessionWorktrees(session);
+
+    // Primary should be preserved in the map
+    expect(session.worktreePaths.has('primary')).toBe(true);
+    expect(session.worktreePaths.get('primary')?.wtId).toBe('abc');
+
+    // Secondary entries should be removed from the map
+    expect(session.worktreePaths.has('mitzo')).toBe(false);
+    expect(session.worktreePaths.has('centaur')).toBe(false);
+    expect(session.worktreePaths.size).toBe(1);
+
+    vi.restoreAllMocks();
+  });
+
+  it('handles session with only primary worktree', async () => {
+    vi.spyOn(await import('../worktree.js'), 'removeWorktree').mockImplementation(() => {});
+
+    const { cleanupSessionWorktrees } = await import('../chat.js');
+
+    const session = {
+      worktreePaths: new Map([
+        ['primary', { path: '/repo/.claude/worktrees/abc', wtId: 'abc' }],
+      ]),
+    } as unknown as ManagedSession;
+
+    cleanupSessionWorktrees(session);
+
+    expect(session.worktreePaths.has('primary')).toBe(true);
+    expect(session.worktreePaths.size).toBe(1);
+
+    vi.restoreAllMocks();
+  });
+
+  it('handles session with no worktrees', async () => {
+    const { cleanupSessionWorktrees } = await import('../chat.js');
+
+    const session = {
+      worktreePaths: new Map(),
+    } as unknown as ManagedSession;
+
+    cleanupSessionWorktrees(session);
+
+    expect(session.worktreePaths.size).toBe(0);
   });
 });
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -56,7 +56,6 @@ describe('getMessages', () => {
 
 describe('cleanupSessionWorktrees', () => {
   it('skips primary worktree and only removes secondaries', async () => {
-    const { removeWorktree } = await import('../worktree.js');
     vi.spyOn(await import('../worktree.js'), 'removeWorktree').mockImplementation(() => {});
 
     const { cleanupSessionWorktrees } = await import('../chat.js');
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -1,5 +1,18 @@
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
 import type { ManagedSession } from '@mitzo/harness';
+
+vi.mock('../worktree.js', async (importOriginal) => {
+  const actual = (await importOriginal()) as Record<string, unknown>;
+  return { ...actual, removeWorktree: vi.fn() };
+});
+
+vi.mock('../repo-config.js', async (importOriginal) => {
+  const actual = (await importOriginal()) as Record<string, unknown>;
+  return {
+    ...actual,
+    loadRepoConfig: vi.fn(actual.loadRepoConfig as (...args: unknown[]) => unknown),
+  };
+});
 
 describe('chat module exports', () => {
   it('exports expected functions', async () => {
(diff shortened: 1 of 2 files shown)


server/__tests__/chat.test.ts
```diff
@@ -52,3 +53,64 @@ describe('getMessages', () => {
     expect(messages.length).toBe(0);
   });
 });
+
+describe('cleanupSessionWorktrees', () => {
+  it('skips primary worktree and only removes secondaries', async () => {
```
dimakis, line 79: 🟡 **missing_tests**: The primary test case ('skips primary worktree and only removes secondaries') verifies the map state but never asserts that `removeWorktree` was called for the secondary repos ('mitzo', 'centaur') and NOT called for 'primary'. The mock is set up but never checked — this means the test would pass even if `removeWorktree` was called for all entries (or none). Add `expect(removeWorktreeSpy).toHaveBeenCalledTimes(2)` and `expect(removeWorktreeSpy).not.toHaveBeenCalledWith('abc', ...)` for primary's repo path. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -52,3 +53,64 @@ describe('getMessages', () => {
...
+    const { removeWorktree } = await import('../worktree.js');
```
dimakis: 🔵 **style**: Line 59 destructures `removeWorktree` from a dynamic import but the binding is never used — the spy is created from a separate `await import(...)` on line 60. This is dead code. Remove line 59 and capture the spy from line 60 instead (which would also enable the missing assertion noted above). `[fixable]`

dimakis (commented):
## Centaur Review

_Solid fix for preserving primary worktrees on session cleanup. The main gap is that the test verifies map state but never asserts `removeWorktree` was called correctly for secondaries and skipped for primary — the core behavioral contract is untested._

- 🟡 **missing_tests** (L58): The primary test case ('skips primary worktree and only removes secondaries') verifies the map state but never asserts that `removeWorktree` was called for the secondary repos ('mitzo', 'centaur') and NOT called for 'primary'. The mock is set up but never checked — this means the test would pass even if `removeWorktree` was called for all entries (or none). Add `expect(removeWorktreeSpy).toHaveBeenCalledTimes(2)` and `expect(removeWorktreeSpy).not.toHaveBeenCalledWith('abc', ...)` for primary's repo path. `[fixable]`
- 🔵 **style** (L59): Line 59 destructures `removeWorktree` from a dynamic import but the binding is never used — the spy is created from a separate `await import(...)` on line 60. This is dead code. Remove line 59 and capture the spy from line 60 instead (which would also enable the missing assertion noted above). `[fixable]`
- 🔵 **missing_tests** (L86): The 'handles session with only primary worktree' test mocks `removeWorktree` but doesn't assert it was never called. Since there are no secondary worktrees, `removeWorktree` should not be invoked at all. `[fixable]`


### `server/chat.ts`
_Solid fix for preserving primary worktrees on session cleanup. The main gap is that the test verifies map state but never asserts `removeWorktree` was called correctly for secondaries and skipped for primary — the core behavioral contract is untested._

- 🟡 **bugs** (L453): The resume fallback creates a bare directory via `mkdirSync` when the original CWD is gone, but this is no longer a valid git worktree — just an empty directory. If the SDK only needs the path to exist for its internal path-based lookup (conversation data encoding), this is fine. But if any downstream code (e.g. git operations, worktree guard path checks) assumes the CWD is a real git worktree, this could cause subtle failures. The comment explains the intent well, but worth verifying the SDK's actual requirement.
- 🔵 **style** (L738): The 3-line comment block at lines 738-740 repeats the same information as the JSDoc on lines 728-731 and the inline comment on line 753. Three locations saying 'SDK encodes conversation data by CWD path' is redundant — the JSDoc is sufficient context, and the `continue` statement is self-explanatory given the function name. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -52,3 +53,64 @@ describe('getMessages', () => {
...
+
+    // Secondary entries should be removed from the map
+    expect(session.worktreePaths.has('mitzo')).toBe(false);
+    expect(session.worktreePaths.has('centaur')).toBe(false);
+    expect(session.worktreePaths.size).toBe(1);
+
+    vi.restoreAllMocks();
+  });
+
+  it('handles session with only primary worktree', async () => {
```
dimakis, line 116: 🔵 **missing_tests**: The 'handles session with only primary worktree' test mocks `removeWorktree` but doesn't assert it was never called. Since there are no secondary worktrees, `removeWorktree` should not be invoked at all. `[fixable]`

server/chat.ts
```diff
@@ -447,10 +447,14 @@ export async function startChat(
           cwd: baseCwd,
         });
       } else {
-        log.warn('resume: original CWD no longer exists, falling back to BASE_REPO', {
+        // The worktree directory was cleaned up, but the SDK encodes
+        // conversation data by CWD path. Re-create the directory so
+        // the SDK can find the conversation using the original path.
+        mkdirSync(sessionMeta.cwd, { recursive: true });
```
dimakis, line 454: 🟡 **bugs**: The resume fallback creates a bare directory via `mkdirSync` when the original CWD is gone, but this is no longer a valid git worktree — just an empty directory. If the SDK only needs the path to exist for its internal path-based lookup (conversation data encoding), this is fine. But if any downstream code (e.g. git operations, worktree guard path checks) assumes the CWD is a real git worktree, this could cause subtle failures. The comment explains the intent well, but worth verifying the SDK's actual requirement.

server/chat.ts
```diff
@@ -720,14 +725,21 @@ export async function interruptChat(
...
  */
-function cleanupSessionWorktrees(session: import('./session-registry.js').ManagedSession): void {
+export function cleanupSessionWorktrees(
+  session: import('./session-registry.js').ManagedSession,
+): void {
   const config = getRepoConfig();
   for (const [repoName, { wtId }] of session.worktreePaths) {
-    // For 'primary', use BASE_REPO; for secondary repos, look up in config
-    const repoPath = repoName === 'primary' ? BASE_REPO : config.repos[repoName];
+    // Skip the primary worktree — it must persist for session resume.
```
dimakis: 🔵 **style**: The 3-line comment block at lines 738-740 repeats the same information as the JSDoc on lines 728-731 and the inline comment on line 753. Three locations saying 'SDK encodes conversation data by CWD path' is redundant — the JSDoc is sufficient context, and the `continue` statement is self-explanatory given the function name. `[fixable]`

dimakis (commented):
## Centaur Review

_Correct fix for a real resume-breaking bug. The logic is sound and idempotent across all call sites (catch+finally double-call, stopChat). Main gap is that the test verifies Map state but not that removeWorktree was actually invoked for secondaries._

- 🟡 **missing_tests** (L59): The test mocks `removeWorktree` but never asserts it was called with the correct arguments for secondary repos. The test only verifies the Map mutation (entries removed), not that the actual git worktree removal happened. Add `expect(removeWorktreeSpy).toHaveBeenCalledWith('abc', ...)` for both secondary repos and `expect(removeWorktreeSpy).not.toHaveBeenCalledWith('abc', <primary-repo-path>)` to fully verify the behavior. `[fixable]`


### `server/chat.ts`
_Correct fix for a real resume-breaking bug. The logic is sound and idempotent across all call sites (catch+finally double-call, stopChat). Main gap is that the test verifies Map state but not that removeWorktree was actually invoked for secondaries._

- 🔵 **style** (L743): The same rationale ("primary must persist because the SDK encodes conversation data by CWD path") is repeated four times: the JSDoc (line 733–736), the inline comment (line 744–746), a second inline comment (line 758), and the resume-path comment (line 449–451). Per project conventions, comments should explain the WHY only once — the JSDoc already covers it. Consider removing the redundant inline comments. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -52,3 +53,63 @@ describe('getMessages', () => {
     expect(messages.length).toBe(0);
   });
 });
+
+describe('cleanupSessionWorktrees', () => {
+  it('skips primary worktree and only removes secondaries', async () => {
+    vi.spyOn(await import('../worktree.js'), 'removeWorktree').mockImplementation(() => {});
```
dimakis: 🟡 **missing_tests**: The test mocks `removeWorktree` but never asserts it was called with the correct arguments for secondary repos. The test only verifies the Map mutation (entries removed), not that the actual git worktree removal happened. Add `expect(removeWorktreeSpy).toHaveBeenCalledWith('abc', ...)` for both secondary repos and `expect(removeWorktreeSpy).not.toHaveBeenCalledWith('abc', <primary-repo-path>)` to fully verify the behavior. `[fixable]`

server/chat.ts
```diff
@@ -725,14 +730,21 @@ export async function interruptChat(
...
```
dimakis: 🔵 **style**: The same rationale ("primary must persist because the SDK encodes conversation data by CWD path") is repeated four times: the JSDoc (line 733–736), the inline comment (line 744–746), a second inline comment (line 758), and the resume-path comment (line 449–451). Per project conventions, comments should explain the WHY only once — the JSDoc already covers it. Consider removing the redundant inline comments. `[fixable]`

PR #264, opened 2026-04-21
fix(test): increase timeout for flaky abort test in hook-bridge

dimakis:
## Summary

- The `returns empty object when aborted` test in `hook-bridge.test.ts` spawns a `sleep 10` command with a 10s callback timeout, but vitest's default test timeout is 5s. The abort races with the test timeout on slower CI runners, causing consistent failures.
- Increases the test timeout to 15s so the abort has time to complete.

## Test plan

- [x] This test was failing on every CI run for PRs #259 and #263
- [ ] CI passes on this branch

### DIFF
--- a/server/__tests__/hook-bridge.test.ts
+++ b/server/__tests__/hook-bridge.test.ts
@@ -204,7 +204,7 @@ describe('createCommandCallback', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #265, opened 2026-04-21
chore: stop tracking .env.ios (already gitignored)

dimakis:
The file was committed before the gitignore rule existed. This removes it from tracking so machine-specific Tailnet hostnames don't get pushed.


dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #266, opened 2026-04-21
fix(ws): allow session takeover when owner connection is gone

dimakis:
## Summary

- After a server restart or reconnect race, the ownership guard rejected messages with "Session is active on another device" because the `connectionId` changed but the old owner connection was already gone.
- Adds `ownerGone` check to both send paths — if the original owner connection no longer exists in the registry, allow the new connection to take over instead of rejecting.

## Test plan

- [x] Verified on production: phone could not send to existing session after server restart, now works
- [ ] CI passes


dimakis:
Closing — fully superseded by PRs #271, #274, #276, #278 which landed a comprehensive auto-takeover rewrite with rekey, stale-session EventStore cross-referencing, and session_takeover notifications. The ownerGone check from this PR is already present on main as part of that work.

PR #271, opened 2026-04-21
fix(ws): prevent false active_elsewhere on stale sessions

dimakis:
## Summary

- `handleSendV2()` and `handleInterruptV2()` now cross-reference the durable `EventStore` before rejecting with `active_elsewhere`, matching the existing safeguard in `handleReconnect()`
- Fixes false "session is active on another device" errors when the in-memory `SessionRegistry` has stale entries (48h detach TTL outliving the actual query loop)
- Adds 2 regression tests for both send and interrupt paths

## Root cause

The in-memory `SessionRegistry` keeps entries alive during the detach TTL window even after the query loop ends. `handleReconnect()` already cross-referenced the `EventStore` (ground truth from `markSessionInactive()` in the query loop's `finally` block), but the send and interrupt handlers did not.

## Test plan

- [x] All 80 ws-handler-v2 tests pass (78 existing + 2 new)
- [x] TypeScript compiles cleanly
- [ ] Manual: verify mobile reconnect no longer shows false active_elsewhere

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -47,6 +47,9 @@ export function ChatView() {
   const storeSetModel = useMitzoStore((s) => s.setModel);
   const storeDispatchMessages = useMitzoStore((s) => s.dispatchMessages);
   const storeFetchSessionMeta = useMitzoStore((s) => s.fetchSessionMeta);
+  const pendingSession = useMitzoStore((s) => s.pendingSession);
+  const clearPendingSession = useMitzoStore((s) => s.clearPendingSession);
+  const sessionContext = useMitzoStore((s) => s.messages.sessionContext);
 
   const connected = connection.status === 'connected';
 
@@ -115,6 +118,19 @@ export function ChatView() {
       storeFetchSessionMeta(sessionId);
     }
   }, [sessionId]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  // Auto-send pending session (from "Start Session" on inbox/todo items)
+  const pendingConsumed = useRef(false);
+  useEffect(() => {
+    if (!pendingSession || pendingConsumed.current) return;
+    pendingConsumed.current = true;
+    // Set the context block for display
+    storeDispatchMessages({ type: 'SET_SESSION_CONTEXT', context: pendingSession.context });
+    // Auto-send the prompt
+    storeSendMessage(pendingSession.prompt, { model: modelState, mode });
+    clearPendingSession();
+    forceScrollToBottom();
+  }, [pendingSession]); // eslint-disable-line react-hooks/exhaustive-deps
 
   useAutoSpeak({
     messages: messages.messages,
@@ -241,6 +257,7 @@ export function ChatView() {
         permission={messages.permission}
         onPermissionRespond={handlePermission}
         scrollRef={scrollRef}
+        sessionContext={sessionContext}
       />
 
       <ChatInput
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -120,10 +120,13 @@ export function ChatView() {
   }, [sessionId]); // eslint-disable-line react-hooks/exhaustive-deps
 
   // Auto-send pending session (from "Start Session" on inbox/todo items)
-  const pendingConsumed = useRef(false);
+  const pendingConsumed = useRef<string | null>(null);
   useEffect(() => {
-    if (!pendingSession || pendingConsumed.current) return;
-    pendingConsumed.current = true;
+    if (!pendingSession) return;
+    // Guard against double-consumption of the same pending session
+    const key = pendingSession.prompt;
+    if (pendingConsumed.current === key) return;
+    pendingConsumed.current = key;
     // Set the context block for display
     storeDispatchMessages({ type: 'SET_SESSION_CONTEXT', context: pendingSession.context });
     // Auto-send the prompt
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1620,9 +1620,7 @@ describe('handleSendV2 stale session via EventStore', () => {
       ctx,
     );
 
-    expect(transport.sent).toContainEqual(
-      expect.objectContaining({ code: 'active_elsewhere' }),
-    );
+    expect(transport.sent).toContainEqual(expect.objectContaining({ code: 'active_elsewhere' }));
 
     (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
   });
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1593,6 +1593,39 @@ describe('handleSendV2 stale session via EventStore', () => {
 
     (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
   });
+
+  it('still rejects when EventStore confirms session IS active', () => {
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValueOnce(true);
+
+    const sessionReg = mockSessionRegistry();
+    sessionReg.findBySessionId.mockReturnValue({ clientId: 'other-conn:sess-1', session: {} });
+    sessionReg.isActive.mockReturnValue(true);
+    sessionReg.isAttached.mockReturnValue(true);
+
+    const eventStore = mockEventStore();
+    // EventStore ground truth: session IS active
+    eventStore.getSession.mockReturnValue({ sessionId: 'sess-1', isActive: true });
+
+    const ctx = createContext({
+      sessionRegistry: sessionReg as unknown as V2HandlerContext['sessionRegistry'],
+      eventStore: eventStore as unknown as V2HandlerContext['eventStore'],
+    });
+    const transport = mockTransport();
+    ctx.connRegistry.register('c1', transport);
+
+    handleSendV2(
+      'c1',
+      transport,
+      { type: 'send' as const, sessionId: 'sess-1', prompt: 'hello', clientMsgId: 'cmsg-active' },
+      ctx,
+    );
+
+    expect(transport.sent).toContainEqual(
+      expect.objectContaining({ code: 'active_elsewhere' }),
+    );
+
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
+  });
 });
 
 describe('handleInterruptV2 stale session via EventStore', () => {
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1655,6 +1655,8 @@ describe('handleInterruptV2 stale session via EventStore', () => {
     expect(transport.sent).not.toContainEqual(
       expect.objectContaining({ code: 'active_elsewhere' }),
     );
+    // Verify interruptChat was actually called (positive outcome)
+    expect(interruptChat).toHaveBeenCalled();
 
     (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
   });
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1554,6 +1554,81 @@ describe('handleSendV2 connection ownership', () => {
   });
 });
 
+// ─── stale session — EventStore cross-reference ─────────────────────────────
+
+describe('handleSendV2 stale session via EventStore', () => {
+  it('allows send when in-memory registry is stale but EventStore says inactive', () => {
+    (startChat as ReturnType<typeof vi.fn>).mockClear();
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValueOnce(true);
+
+    const sessionReg = mockSessionRegistry();
+    sessionReg.findBySessionId.mockReturnValue({ clientId: 'old-conn:sess-1', session: {} });
+    sessionReg.isActive.mockReturnValue(true);
+    sessionReg.isAttached.mockReturnValue(true); // would normally reject
+
+    const eventStore = mockEventStore();
+    // EventStore ground truth: session is NOT active (query loop ended)
+    eventStore.getSession.mockReturnValue({ sessionId: 'sess-1', isActive: false });
+
+    const ctx = createContext({
+      sessionRegistry: sessionReg as unknown as V2HandlerContext['sessionRegistry'],
+      eventStore: eventStore as unknown as V2HandlerContext['eventStore'],
+    });
+    const transport = mockTransport();
+    ctx.connRegistry.register('c1', transport);
+
+    handleSendV2(
+      'c1',
+      transport,
+      { type: 'send' as const, sessionId: 'sess-1', prompt: 'hello', clientMsgId: 'cmsg-stale' },
+      ctx,
+    );
+
+    // Should NOT get active_elsewhere error
+    expect(transport.sent).not.toContainEqual(
+      expect.objectContaining({ code: 'active_elsewhere' }),
+    );
+    // Should fall through to resume path (startChat for resume)
+    expect(startChat).toHaveBeenCalled();
+
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
+  });
+});
+
+describe('handleInterruptV2 stale session via EventStore', () => {
+  it('skips ownership guard when EventStore says session is inactive', () => {
+    (interruptChat as ReturnType<typeof vi.fn>).mockClear();
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValueOnce(true);
+
+    const sessionReg = mockSessionRegistry();
+    sessionReg.findBySessionId.mockReturnValue({ clientId: 'old-conn:sess-1', session: {} });
+    sessionReg.isAttached.mockReturnValue(true); // would normally reject
+
+    const eventStore = mockEventStore();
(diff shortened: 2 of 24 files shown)


dimakis:
## Centaur Review

### `frontend/src/components/PageHeader.tsx`
_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🔴 **regressions** (L12): The `onBack` prop is silently discarded (renamed to `_onBack`), but 5 callers still pass it. Two existing tests in `PageHeader.test.tsx` (`renders a back button when onBack is provided` and `uses lsaquo character for back button`) will fail because the back button no longer renders. Either update the tests and remove the prop from the interface, or restore the back button alongside the logo. `[fixable]`
- 🔴 **regressions** (L6): `TodoDetailView` passes `onBack={() => navigate('/todos')}` to navigate from the detail page back to the list. With the back button removed, users on `/todos/:id` have no in-app way to return to the todo list. Level-1 pages (Inbox, Tasks, Calendar, Todos) also lose their home navigation, though the MitzoLogo may serve that purpose if it links to `/`. `[fixable]`
- 🔵 **style** (L6): `onBack` remains in `PageHeaderProps` but is intentionally unused (`_onBack`). If the back button is permanently removed, remove the prop from the interface and clean up the 5 callers. If the removal is intentional, this is dead code that misleads callers. `[fixable]`


### `frontend/src/lib/inbox-utils.ts`
_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🟡 **missing_tests**: No test file exists for `inbox-utils.ts`. `stripFrontmatter`, `buildInboxContext`, and `buildInboxPrompt` are untested. Per the project's TDD requirements, tests should be written first. `[fixable]`


### `frontend/src/lib/todo-utils.ts`
_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🟡 **missing_tests** (L48): The new `buildTodoContext` function has no tests. The existing `todo-utils.test.ts` covers `sourceIcon` and `buildPrompt` but not this new export. `[fixable]`


_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🟡 **missing_tests** (L100): The new `SET_SESSION_CONTEXT` action has no test coverage in `messages-slice.test.ts`. The existing `CLEAR` test also doesn't assert that `sessionContext` resets to `null`. `[fixable]`


_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🟡 **missing_tests** (L1598): The `handleInterruptV2` stale-session test verifies no `active_elsewhere` error is sent, but does not assert that `interruptChat` was actually called. The mock is cleared (line 1600) but never checked. Add `expect(interruptChat).toHaveBeenCalled()` to confirm the interrupt flows through. `[fixable]`


### `frontend/src/components/ContextBlock.tsx`
_The core WS fix (EventStore cross-reference for stale sessions) is well-implemented and tested. However, the PageHeader refactor silently breaks 2 existing tests and removes back-button navigation from 5 pages — particularly TodoDetailView which has no alternative navigation back to the list. Several new utility functions lack test coverage despite the project's TDD mandate._

- 🔵 **style** (L7): The `if (!content) return null` guard on line 10 is redundant — the parent (`ChatArea`) already guards with `{sessionContext && <ContextBlock content={sessionContext} />}`, so `content` will always be a non-empty string when rendered. `[fixable]`

dimakis:
## Centaur Review

Found **8** issue(s) (1 critical) (5 warning).


### `frontend/src/components/PageHeader.tsx`
_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🔴 **regressions** (L12): The `onBack` prop is accepted but silently ignored (renamed to `_onBack`), and the back button was replaced with `<MitzoLogo />`. This breaks at least 3 existing tests in `PageHeader.test.tsx` that assert on `screen.getByTitle('Back')` and the `onBack` callback. Five callers (InboxView, TodoView, TodoDetailView, CalendarView, TaskBoard) still pass `onBack` expecting navigation — those back buttons are now gone with no replacement. `[fixable]`
- 🔵 **style** (L6): `onBack` remains in the `PageHeaderProps` interface despite being unused. Either remove it from the interface and all 5 call sites, or restore the back button functionality. Keeping a dead prop in the API is confusing. `[fixable]`


_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🟡 **bugs** (L123): `pendingConsumed` is a ref that is set to `true` once and never reset. If the ChatView component is never unmounted (e.g. stays mounted across route changes in a layout), a second "Start Session" navigation will be silently dropped because `pendingConsumed.current` is still `true`. Consider clearing it when the ref's guard is no longer needed, or removing the ref and relying on `clearPendingSession()` to null out `pendingSession` (which already prevents re-fire). `[fixable]`


### `frontend/src/lib/inbox-utils.ts`
_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🟡 **missing_tests**: No test file exists for `inbox-utils.ts`. The `stripFrontmatter`, `buildInboxContext`, and `buildInboxPrompt` functions are untested. The project requires TDD — tests should accompany the implementation. `[fixable]`


### `frontend/src/lib/todo-utils.ts`
_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🟡 **missing_tests** (L48): The new `buildTodoContext` function has no tests. The existing `todo-utils.test.ts` only covers `sourceIcon` and `buildPrompt`. `[fixable]`


_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🟡 **missing_tests**: The new EventStore cross-reference tests don't cover the case where `eventStore.getSession()` returns `null` (session never registered in EventStore). In that scenario `staleInMemory` is `false` and the code proceeds to the ownership check — this is likely correct behavior but should be explicitly tested to lock it in. `[fixable]`


### `server/ws-handler-v2.ts`
_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🟡 **unsafe_assumptions** (L384): `ctx.eventStore.getSession(sessionId)` could throw (e.g. SQLite error). The call is not wrapped in try/catch. A database error would bubble up as an unhandled exception and crash the WS handler for this message. Consider wrapping in try/catch and falling through to the ownership check on failure (fail-open to existing behavior). `[fixable]`


_The core WS fix (EventStore cross-reference for stale sessions) is sound and well-tested. However, the PageHeader change silently removes all back buttons across 5 pages and breaks 3 existing tests — this is the most urgent issue to address before merge._

- 🔵 **style**: The diff removes `.inbox-scroll` at one location (line ~834) and re-adds it at another (line ~924) with different properties (now includes `display: flex`, `flex-direction: column`, `gap`). Meanwhile `InboxView.tsx` renames the container from `inbox-list` to `inbox-scroll`. This is fine functionally, but the class was previously used for something else — verify no other files reference the old `.inbox-list` class.

dimakis:
## Round 1 Review — Claude Code

### Summary
The core WS fix (EventStore cross-reference in send/interrupt handlers) is solid and matches the existing pattern in `handleReconnect`. The 2 regression tests cover the right paths. However, this PR introduces **14 new test failures** that must be fixed.

### Findings

**1. [Critical] PageHeader: back button removed, 9/9 tests broken**
`PageHeader.tsx` replaces the back button with `<MitzoLogo />` and renames `onBack` to `_onBack` (unused). All 9 PageHeader tests fail because they assert on the back button element (`screen.getByTitle('Back')`). Additionally, 5 callers still pass `onBack` — it's silently discarded.

Fix: Either restore the back button alongside the logo, or update the tests + remove `onBack` from the interface + clean up callers.

**2. [Critical] TodoView: 5 new test failures**
TodoView went from 1 failing test (pre-existing) to 6. The `onStartSession` prop was added to `TodoCard` but the TodoView test stubs likely weren't updated.

**3. [Minor] `_onBack` is a backward-compat hack**
Renaming unused params to `_var` is explicitly called out as an anti-pattern in this codebase. Either use the prop or remove it.

**4. [Observation] Scope**
The PR title says "fix(ws): prevent false active_elsewhere" but includes unrelated frontend features (Start Session, ContextBlock, UI polish). Consider splitting in the future — not blocking.

### Verdict
Fix the 14 new test failures (PageHeader + TodoView), then re-request review.

dimakis:
## Round 2 Review — Claude Code (deeper pass)

### Summary
Round 1 critical issues are resolved — CI green, 14 new test failures fixed, `onBack` cleanup complete. This round focuses on edge cases and coverage gaps.

### Findings

**1. [Medium] Missing negative test for EventStore cross-reference**
Tests only cover the "stale → allow" path (EventStore says inactive → bypass guard). There's no test for the counterpart: EventStore says active → ownership check still applies. Without this, a regression that accidentally inverts the condition wouldn't be caught. Recommend adding a test where `storeMeta.isActive === true` and verifying `active_elsewhere` is still sent.

**2. [Low] `fullContent ?? item.preview` fallback in InboxView:163**
If the inbox card was never expanded, `fullContent` is null and the "Start Session" button falls back to `item.preview` — which may be truncated. The prompt/context sent to the agent would be incomplete. Consider fetching full content on session start, or noting the limitation.

**3. [Low] No unit tests for `inbox-utils.ts`**
`stripFrontmatter`, `buildInboxContext`, and `buildInboxPrompt` are pure functions — easy to test. The `stripFrontmatter` regex handles the happy path but could miss edge cases (e.g., `---` inside YAML values, Windows line endings).

### Verdict
Finding 1 is actionable — add the negative test. Findings 2-3 are improvements for a follow-up. Approve after addressing finding 1.

frontend/src/pages/ChatView.tsx
```diff
@@ -116,6 +119,19 @@ export function ChatView() {
     }
   }, [sessionId]); // eslint-disable-line react-hooks/exhaustive-deps
 
+  // Auto-send pending session (from "Start Session" on inbox/todo items)
+  const pendingConsumed = useRef(false);
```
dimakis: 🟡 **bugs**: pendingConsumed ref is initialized once per mount and never reset. If the ChatView component instance survives across multiple pending sessions (e.g., the route stays mounted while store state changes), the second pending session will be silently dropped. Currently safe because setPendingSession callers always navigate from different routes (causing remount), but this is a fragile implicit assumption. Consider resetting the ref when pendingSession changes to null, or deriving the guard from store state instead of a local ref. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -377,42 +377,58 @@ export function handleSendV2(
...
-        // so the v1 fallback path in sendOrBuffer can still deliver events.
-        if (isDetached) {
-          reattachChat(found.clientId, transport);
-          log.info('reattached detached session on send', {
+        // Cross-reference with durable EventStore: the in-memory registry
+        // keeps entries alive during the detach TTL window even after the
+        // query loop has ended. markSessionInactive() in the query loop's
+        // finally block is ground truth.
+        const storeMeta = ctx.eventStore.getSession(sessionId);
+        const staleInMemory = storeMeta && !storeMeta.isActive;
```
dimakis, line 385: 🔵 **unsafe_assumptions**: When staleInMemory is detected and the code falls through to the resume path (line 434+), the old clientId (e.g., 'old-conn:sess-1') remains registered as 'active' in the SessionRegistry. The resume creates a new clientId ('c1:sess-1') via startChat, leaving a ghost entry. This is benign (TTL cleanup will eventually remove it), but the next reconnect or send for the same session could hit the stale entry again until it expires. Consider explicitly cleaning up the stale registry entry. `[fixable]`

server/__tests__/ws-handler-v2.test.ts
```diff
@@ -1554,6 +1554,112 @@ describe('handleSendV2 connection ownership', () => {
...
+      ctx,
+    );
+
+    expect(transport.sent).toContainEqual(expect.objectContaining({ code: 'active_elsewhere' }));
+
+    (isActive as ReturnType<typeof vi.fn>).mockReturnValue(false);
+  });
+});
+
+describe('handleInterruptV2 stale session via EventStore', () => {
```
dimakis, line 1629: 🔵 **missing_tests**: handleInterruptV2 stale session test only asserts the negative (active_elsewhere NOT sent) but doesn't verify the positive outcome — that interruptChat was actually called. The mock is cleared at line 1631, so asserting expect(interruptChat).toHaveBeenCalled() would strengthen the test. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/lib/inbox-utils.ts`
_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🟡 **missing_tests**: New file with three exported functions (stripFrontmatter, buildInboxContext, buildInboxPrompt) but no test file exists. The sibling todo-utils.ts has a test file that covers its original functions but also doesn't test the newly added buildTodoContext. Project enforces TDD — tests should be the first artifact. `[fixable]`


### `frontend/src/lib/todo-utils.ts`
_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🟡 **missing_tests**: buildTodoContext is a new export (39 lines of logic with conditional formatting) but is not covered in the existing todo-utils.test.ts, which only tests sourceIcon and buildPrompt. `[fixable]`


_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🟡 **bugs** (L123): pendingConsumed ref is initialized once per mount and never reset. If the ChatView component instance survives across multiple pending sessions (e.g., the route stays mounted while store state changes), the second pending session will be silently dropped. Currently safe because setPendingSession callers always navigate from different routes (causing remount), but this is a fragile implicit assumption. Consider resetting the ref when pendingSession changes to null, or deriving the guard from store state instead of a local ref. `[fixable]`


### `server/ws-handler-v2.ts`
_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🔵 **unsafe_assumptions** (L385): When staleInMemory is detected and the code falls through to the resume path (line 434+), the old clientId (e.g., 'old-conn:sess-1') remains registered as 'active' in the SessionRegistry. The resume creates a new clientId ('c1:sess-1') via startChat, leaving a ghost entry. This is benign (TTL cleanup will eventually remove it), but the next reconnect or send for the same session could hit the stale entry again until it expires. Consider explicitly cleaning up the stale registry entry. `[fixable]`


_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🔵 **missing_tests** (L1629): handleInterruptV2 stale session test only asserts the negative (active_elsewhere NOT sent) but doesn't verify the positive outcome — that interruptChat was actually called. The mock is cleared at line 1631, so asserting expect(interruptChat).toHaveBeenCalled() would strengthen the test. `[fixable]`


### `None`
_Core ws-handler fix is correct and well-tested — EventStore cross-reference properly prevents false active_elsewhere on zombie sessions. Main gaps are missing tests for the new frontend utility functions (inbox-utils, buildTodoContext), and the PR bundles significant unrelated UI work._

- 🔵 **style**: PR title is 'fix(ws): prevent false active_elsewhere on stale sessions' but the diff includes substantial unrelated work: Start Session flow for inbox/todos, PageHeader onBack removal, ContextBlock component, MessageBubble DOM restructuring, quick-action CSS restyle, and a new Cursor rule file. Consider splitting into focused PRs for easier review and bisectability.

PR #272, opened 2026-04-22
feat(ui): persist queued messages across session switches

dimakis:
## Summary
- Queued messages (typed while agent is running) were plain React state — lost on session switch
- New `useQueuedMessages` hook persists the queue to `localStorage` per session, mirroring the existing `useDraft` pattern
- Handles sessionId migration (new → assigned), quota-safe writes, and loads correct queue when switching back

## Test plan
- [ ] Queue 2-3 messages while agent is running
- [ ] Switch to a different session and back — queue should be intact
- [ ] Edit/remove/fire-as-interrupt queued messages — all operations should persist
- [ ] Refresh the page while messages are queued — they should survive

### DIFF
--- a/frontend/src/hooks/__tests__/useQueuedMessages.test.ts
+++ b/frontend/src/hooks/__tests__/useQueuedMessages.test.ts
@@ -1,0 +1,199 @@
+// @vitest-environment jsdom
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { renderHook, act } from '@testing-library/react';
+import { useQueuedMessages, type QueuedMessage } from '../useQueuedMessages';
+
+function msg(text: string): QueuedMessage {
+  return { text, images: [], contextBlocks: [] };
+}
+
+beforeEach(() => localStorage.clear());
+afterEach(() => localStorage.clear());
+
+describe('useQueuedMessages', () => {
+  it('initializes with empty queue when nothing is stored', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-1'));
+    expect(result.current.queue).toEqual([]);
+  });
+
+  it('loads queue from localStorage on init', () => {
+    localStorage.setItem(
+      'mitzo-queue-sess-2',
+      JSON.stringify([{ text: 'hello', contextBlocks: [] }]),
+    );
+    const { result } = renderHook(() => useQueuedMessages('sess-2'));
+    expect(result.current.queue).toHaveLength(1);
+    expect(result.current.queue[0].text).toBe('hello');
+    // images should be reconstructed as empty array
+    expect(result.current.queue[0].images).toEqual([]);
+  });
+
+  it('enqueues a message and persists it', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-3'));
+
+    act(() => {
+      const added = result.current.enqueue(msg('first'));
+      expect(added).toBe(true);
+    });
+
+    expect(result.current.queue).toHaveLength(1);
+    expect(result.current.queue[0].text).toBe('first');
+
+    const stored = JSON.parse(localStorage.getItem('mitzo-queue-sess-3')!);
+    expect(stored).toEqual([{ text: 'first', contextBlocks: [] }]);
+  });
+
+  it('respects maxQueued limit', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-4', 2));
+
+    act(() => result.current.enqueue(msg('one')));
+    act(() => result.current.enqueue(msg('two')));
+
+    let added: boolean = false;
+    act(() => {
+      added = result.current.enqueue(msg('three'));
+    });
+
+    expect(added).toBe(false);
+    expect(result.current.queue).toHaveLength(2);
+  });
+
+  it('dequeues the first message', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-5'));
+
+    act(() => result.current.enqueue(msg('first')));
+    act(() => result.current.enqueue(msg('second')));
+
+    let item: QueuedMessage | undefined;
+    act(() => {
+      item = result.current.dequeue();
+    });
+
+    expect(item?.text).toBe('first');
+    expect(result.current.queue).toHaveLength(1);
+    expect(result.current.queue[0].text).toBe('second');
+  });
+
+  it('dequeue returns undefined for empty queue', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-6'));
+
+    let item: QueuedMessage | undefined;
+    act(() => {
+      item = result.current.dequeue();
+    });
+
+    expect(item).toBeUndefined();
+  });
+
+  it('removes a message by index', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-7'));
+
+    act(() => result.current.enqueue(msg('a')));
+    act(() => result.current.enqueue(msg('b')));
+    act(() => result.current.enqueue(msg('c')));
+    act(() => result.current.remove(1));
+
+    expect(result.current.queue.map((q) => q.text)).toEqual(['a', 'c']);
+  });
+
+  it('edits (removes and returns) a message by index', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-8'));
+
+    act(() => result.current.enqueue(msg('x')));
+    act(() => result.current.enqueue(msg('y')));
+
+    let item: QueuedMessage | undefined;
+    act(() => {
+      item = result.current.edit(0);
+    });
+
+    expect(item?.text).toBe('x');
+    expect(result.current.queue).toHaveLength(1);
+    expect(result.current.queue[0].text).toBe('y');
+  });
+
+  it('edit returns undefined for invalid index', () => {
+    const { result } = renderHook(() => useQueuedMessages('sess-9'));
+
+    let item: QueuedMessage | undefined;
+    act(() => {
+      item = result.current.edit(5);
+    });
+
+    expect(item).toBeUndefined();
+  });
+
+  it('uses "new" key when sessionId is undefined', () => {
+    localStorage.setItem(
+      'mitzo-queue-new',
+      JSON.stringify([{ text: 'pending', contextBlocks: [] }]),
+    );
+    const { result } = renderHook(() => useQueuedMessages(undefined));
+    expect(result.current.queue[0].text).toBe('pending');
+  });
+
+  it('migrates queue when sessionId changes from undefined to real ID', () => {
+    localStorage.setItem(
+      'mitzo-queue-new',
+      JSON.stringify([{ text: 'queued', contextBlocks: ['ctx'] }]),
+    );
+    const { rerender } = renderHook(({ id }: { id: string | undefined }) => useQueuedMessages(id), {
+      initialProps: { id: undefined as string | undefined },
+    });
+
+    rerender({ id: 'sess-real' });
+
+    expect(localStorage.getItem('mitzo-queue-sess-real')).toContain('queued');
+    expect(localStorage.getItem('mitzo-queue-new')).toBeNull();
+  });
+
+  it('loads existing queue for new sessionId during migration', () => {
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,128 @@
+import { useState, useEffect, useRef, useCallback } from 'react';
+import type { ImageAttachment } from '../types/chat';
+
+const KEY_PREFIX = 'mitzo-queue-';
+
+export interface QueuedMessage {
+  text: string;
+  images: ImageAttachment[];
+  contextBlocks: string[];
```
dimakis, line 9: 🟡 **unsafe_assumptions**: ImageAttachment contains `data` (base64 string) and `preview` (full data URL) — each image is stored twice. A single 1600×1600 JPEG at 0.85 quality can be ~300–500 KB base64. With 5 queued messages containing images, this can easily exceed localStorage's 5 MB limit, causing silent write failures (caught by the try/catch). Consider either stripping `preview`/`data` before persisting and reconstructing on load, or skipping image persistence entirely. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useQueuedMessages.ts`
_Clean extraction of queued-message state into a reusable hook with localStorage persistence, correctly mirroring the useDraft pattern. Main gaps: no tests (required by project TDD policy), and base64 image data in localStorage risks hitting the 5 MB storage limit._

- 🟡 **missing_tests**: No test file exists for useQueuedMessages. The project requires TDD (per CLAUDE.md) and has 11 other hook tests in `frontend/src/hooks/__tests__/`. The analogous `useDraft` hook has comprehensive tests covering init, persistence, migration, and clear. At minimum, tests are needed for: load from localStorage, enqueue/dequeue/remove/edit operations, max-queue cap, session ID migration (undefined → real ID), and persist-on-change. `[fixable]`
- 🟡 **unsafe_assumptions** (L9): ImageAttachment contains `data` (base64 string) and `preview` (full data URL) — each image is stored twice. A single 1600×1600 JPEG at 0.85 quality can be ~300–500 KB base64. With 5 queued messages containing images, this can easily exceed localStorage's 5 MB limit, causing silent write failures (caught by the try/catch). Consider either stripping `preview`/`data` before persisting and reconstructing on load, or skipping image persistence entirely. `[fixable]`
- 🟡 **bugs** (L88): `enqueue`, `dequeue`, and `edit` capture return values via closure variables set inside setState updaters. React docs warn against side effects in updaters, and in StrictMode React may invoke them twice. While this works today because React calls functional updaters synchronously, it's fragile. Consider splitting into a read-then-write pattern: check `queue` length from the ref/state first, then call the setter, rather than extracting the return value from inside the updater. `[fixable]`
- 🔵 **style** (L50): `clear` is returned from the hook but never used by ChatInput. Consider omitting it from the return type until it has a consumer, to keep the API surface minimal. `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,128 @@
...
+  }, [sessionId]);
+
+  // Persist whenever queue changes
+  useEffect(() => {
+    saveQueue(sessionRef.current, queue);
+  }, [queue]);
+
+  const enqueue = useCallback(
+    (msg: QueuedMessage): boolean => {
+      let added = false;
```
dimakis: 🟡 **bugs**: `enqueue`, `dequeue`, and `edit` capture return values via closure variables set inside setState updaters. React docs warn against side effects in updaters, and in StrictMode React may invoke them twice. While this works today because React calls functional updaters synchronously, it's fragile. Consider splitting into a read-then-write pattern: check `queue` length from the ref/state first, then call the setter, rather than extracting the return value from inside the updater. `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,128 @@
...
+export function useQueuedMessages(
+  sessionId: string | undefined,
+  maxQueued: number = 5,
+): {
+  queue: QueuedMessage[];
+  enqueue: (msg: QueuedMessage) => boolean;
+  dequeue: () => QueuedMessage | undefined;
+  remove: (index: number) => void;
+  edit: (index: number) => QueuedMessage | undefined;
+  clear: () => void;
```
dimakis: 🔵 **style**: `clear` is returned from the hook but never used by ChatInput. Consider omitting it from the return type until it has a consumer, to keep the API surface minimal. `[fixable]`

dimakis (commented):
## Centaur Review

Found **6** issue(s) (2 critical) (2 warning).


### `frontend/src/hooks/useQueuedMessages.ts`
_The migration logic in useQueuedMessages runs unconditionally on every sessionId change, but was designed only for the undefined→real-id case. On desktop (component stays mounted during session switches), this deletes the old session's persisted queue and can copy it into the wrong session. Needs a guard to distinguish migration from session switching._

- 🔴 **bugs** (L94): Unconditional `localStorage.removeItem(oldKey)` runs on every sessionId change, not just the intended `undefined → real-id` migration. On desktop, `DesktopChatView` stays mounted across session switches (no `key` prop), so switching from sess-A to sess-B deletes sess-A's persisted queue. Switching back shows an empty queue. The migration block (lines 80-97) should guard `removeItem` and the copy logic to only run when `prev` is `undefined`. For real session switches (`prev` is a real ID), just load the new session's queue via `setQueueRaw(loadQueue(sessionId))`. `[fixable]`
- 🔴 **bugs** (L88): When switching sessions (not migrating from 'new'), the `else` branch copies the OLD session's queue into the NEW session's localStorage key if the new session has no stored queue. This causes sess-A's queued messages to appear in sess-B. Same root cause as the unconditional migration — this code path should only run for `prev === undefined`. `[fixable]`
- 🟡 **bugs** (L75): When sessionId changes to a session with no stored queue AND the old session also has no stored queue, `setQueueRaw` is never called. The in-memory state retains the previous session's queued messages. On desktop (component stays mounted), stale messages from session A bleed into session B's UI. A catch-all `setQueueRaw(loadQueue(sessionId))` for non-migration transitions would fix this. `[fixable]`
- 🔵 **unsafe_assumptions** (L114): `dequeue()` reads from `queueRef.current` (synced via useEffect, one render behind state) and returns the item synchronously while scheduling a setState. If called twice before the ref syncs, both calls return the same item. Not a practical issue in current usage (auto-send only fires on running→idle transition) but fragile for future callers. `[fixable]`


### `frontend/src/hooks/__tests__/useQueuedMessages.test.ts`
_The migration logic in useQueuedMessages runs unconditionally on every sessionId change, but was designed only for the undefined→real-id case. On desktop (component stays mounted during session switches), this deletes the old session's persisted queue and can copy it into the wrong session. Needs a guard to distinguish migration from session switching._

- 🟡 **missing_tests**: No test covers switching between two real session IDs (e.g., 'sess-A' → 'sess-B'). All migration tests only cover `undefined → real-id`. A test with `rerender({ id: 'sess-B' })` from `initialProps: { id: 'sess-A' }` would have caught the queue deletion and cross-contamination bugs. `[fixable]`
- 🔵 **missing_tests** (L142): The migration test ('migrates queue when sessionId changes from undefined to real ID') only asserts on localStorage state, not on `result.current.queue`. It should verify the in-memory queue is correct after migration — currently the `else` branch (line 88-91 in the hook) doesn't call `setQueueRaw`, so the assertion would pass by accident (state carried over from init). `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (existing) {
+        setQueueRaw(fromStored(JSON.parse(existing)));
+      } else {
+        const old = localStorage.getItem(oldKey);
+        if (old) {
+          localStorage.setItem(newKey, old);
+          // Queue state stays the same — just moved the key
+        }
+      }
+      localStorage.removeItem(oldKey);
```
dimakis, line 94: 🔴 **bugs**: Unconditional `localStorage.removeItem(oldKey)` runs on every sessionId change, not just the intended `undefined → real-id` migration. On desktop, `DesktopChatView` stays mounted across session switches (no `key` prop), so switching from sess-A to sess-B deletes sess-A's persisted queue. Switching back shows an empty queue. The migration block (lines 80-97) should guard `removeItem` and the copy logic to only run when `prev` is `undefined`. For real session switches (`prev` is a real ID), just load the new session's queue via `setQueueRaw(loadQueue(sessionId))`. `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    // Migrate queue from 'new' key when session gets assigned an ID
+    const oldKey = queueKey(prev);
+    const newKey = queueKey(sessionId);
+    try {
+      const existing = localStorage.getItem(newKey);
+      if (existing) {
+        setQueueRaw(fromStored(JSON.parse(existing)));
+      } else {
+        const old = localStorage.getItem(oldKey);
```
dimakis, line 88: 🔴 **bugs**: When switching sessions (not migrating from 'new'), the `else` branch copies the OLD session's queue into the NEW session's localStorage key if the new session has no stored queue. This causes sess-A's queued messages to appear in sess-B. Same root cause as the unconditional migration — this code path should only run for `prev === undefined`. `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,134 @@
...
+  const queueRef = useRef(queue);
+  const sessionRef = useRef(sessionId);
+
+  // Keep ref in sync with state
+  useEffect(() => {
+    queueRef.current = queue;
+  }, [queue]);
+
+  // When sessionId changes, load queue for new session
+  useEffect(() => {
```
dimakis, line 75: 🟡 **bugs**: When sessionId changes to a session with no stored queue AND the old session also has no stored queue, `setQueueRaw` is never called. The in-memory state retains the previous session's queued messages. On desktop (component stays mounted), stale messages from session A bleed into session B's UI. A catch-all `setQueueRaw(loadQueue(sessionId))` for non-migration transitions would fix this. `[fixable]`

frontend/src/hooks/useQueuedMessages.ts
```diff
@@ -0,0 +1,134 @@
...
+  const enqueue = useCallback(
+    (msg: QueuedMessage): boolean => {
+      if (queueRef.current.length >= maxQueued) return false;
+      setQueueRaw((prev) => [...prev, msg]);
+      return true;
+    },
+    [maxQueued],
+  );
+
+  const dequeue = useCallback((): QueuedMessage | undefined => {
```
dimakis, line 114: 🔵 **unsafe_assumptions**: `dequeue()` reads from `queueRef.current` (synced via useEffect, one render behind state) and returns the item synchronously while scheduling a setState. If called twice before the ref syncs, both calls return the same item. Not a practical issue in current usage (auto-send only fires on running→idle transition) but fragile for future callers. `[fixable]`

frontend/src/hooks/__tests__/useQueuedMessages.test.ts
```diff
@@ -0,0 +1,199 @@
...
+  });
+
+  it('migrates queue when sessionId changes from undefined to real ID', () => {
+    localStorage.setItem(
+      'mitzo-queue-new',
+      JSON.stringify([{ text: 'queued', contextBlocks: ['ctx'] }]),
+    );
+    const { rerender } = renderHook(({ id }: { id: string | undefined }) => useQueuedMessages(id), {
+      initialProps: { id: undefined as string | undefined },
+    });
```
dimakis, line 142: 🔵 **missing_tests**: The migration test ('migrates queue when sessionId changes from undefined to real ID') only asserts on localStorage state, not on `result.current.queue`. It should verify the in-memory queue is correct after migration — currently the `else` branch (line 88-91 in the hook) doesn't call `setQueueRaw`, so the assertion would pass by accident (state carried over from init). `[fixable]`

PR #274, opened 2026-04-22
fix(ws): rekey session after reattach to prevent false active_elsewhere

dimakis:
## Summary

- **Root cause:** When mobile WS reconnects (new connectionId), `handleReconnect` reattaches the session transport but the clientId still carries the old connectionId prefix. `handleSendV2` then checks `getOwnerConnection(clientId)` → old conn ≠ new conn → false `active_elsewhere` rejection.
- **Fix:** Call `SessionRegistry.rekey()` after every reattach to transfer ownership. Applied in `handleReconnect`, `handleSendV2` (detached takeover), and `handleInterruptV2`.
- The `rekey()` method already existed (added in an earlier PR) but was never called — this wires it into the three code paths that perform reattach.

## Test plan

- [x] New tests: reconnect rekeys to new connection, skips rekey when same connection
- [x] New tests: send rekeys after detached takeover, uses new clientId for `sendToChat`
- [x] New tests: interrupt rekeys after detached takeover, uses new clientId for `interruptChat`
- [x] All 85 ws-handler-v2 tests pass
- [x] All server tests pass (pre-existing failures in token-update and chat unrelated)
- [ ] Manual: mobile WS reconnect after background → send message → no active_elsewhere error

### DIFF
--- a/frontend/src/components/SessionSearchBar.tsx
+++ b/frontend/src/components/SessionSearchBar.tsx
@@ -21,14 +21,15 @@ export function SessionSearchBar({
   clear,
   onSelectSession,
 }: SessionSearchBarProps) {
-  const [open, setOpen] = useState(active);
+  const [open, setOpen] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
+  const visible = open || active;
 
   useEffect(() => {
-    if (open && inputRef.current) {
+    if (visible && inputRef.current) {
       inputRef.current.focus();
     }
-  }, [open]);
+  }, [visible]);
 
   function handleClose() {
     setOpen(false);
@@ -40,7 +41,7 @@ export function SessionSearchBar({
     onSelectSession(sessionId);
   }
 
-  if (!open) {
+  if (!visible) {
     return (
       <button
         className="session-search-toggle"
--- a/frontend/src/components/SessionSearchBar.tsx
+++ b/frontend/src/components/SessionSearchBar.tsx
@@ -1,0 +1,93 @@
+import { useState, useRef, useEffect } from 'react';
+import type { SessionSearchResult } from '../types/chat';
+import { formatRelativeTime } from '../lib/formatTime';
+
+export interface SessionSearchBarProps {
+  query: string;
+  setQuery: (q: string) => void;
+  results: SessionSearchResult[];
+  searching: boolean;
+  active: boolean;
+  clear: () => void;
+  onSelectSession: (sessionId: string) => void;
+}
+
+export function SessionSearchBar({
+  query,
+  setQuery,
+  results,
+  searching,
+  active,
+  clear,
+  onSelectSession,
+}: SessionSearchBarProps) {
+  const [open, setOpen] = useState(active);
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    if (open && inputRef.current) {
+      inputRef.current.focus();
+    }
+  }, [open]);
+
+  function handleClose() {
+    setOpen(false);
+    clear();
+  }
+
+  function handleSelect(sessionId: string) {
+    handleClose();
+    onSelectSession(sessionId);
+  }
+
+  if (!open) {
+    return (
+      <button
+        className="session-search-toggle"
+        onClick={() => setOpen(true)}
+        title="Search sessions"
+      >
+        ⌕
+      </button>
+    );
+  }
+
+  return (
+    <div className="session-search-bar">
+      <input
+        ref={inputRef}
+        className="session-search-input"
+        type="text"
+        placeholder="Search sessions..."
+        value={query}
+        onChange={(e) => setQuery(e.target.value)}
+      />
+      <button className="session-search-close" onClick={handleClose} title="Close search">
+        ✕
+      </button>
+      {active && (
+        <div className="session-search-results">
+          {searching && <div className="session-search-status">Searching...</div>}
+          {!searching && results.length === 0 && (
+            <div className="session-search-status">No matches</div>
+          )}
+          {results.map((r) => (
+            <button
+              key={r.sessionId}
+              className="session-search-result"
+              onClick={() => handleSelect(r.sessionId)}
+            >
+              <div className="session-search-result-summary">{r.summary || 'Untitled session'}</div>
+              <div className="session-search-result-snippet">{r.snippet}</div>
+              <div className="session-search-result-meta">
+                <span className="session-search-result-time">
+                  {formatRelativeTime(r.updatedAt)}
+                </span>
+              </div>
+            </button>
+          ))}
+        </div>
+      )}
+    </div>
+  );
+}
--- a/frontend/src/components/__tests__/SessionSearchBar.test.ts
+++ b/frontend/src/components/__tests__/SessionSearchBar.test.ts
@@ -1,0 +1,104 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { createElement } from 'react';
+import { render, screen, cleanup } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { SessionSearchBar } from '../SessionSearchBar';
+import type { SessionSearchResult } from '../../types/chat';
+
+afterEach(() => {
+  cleanup();
+});
+
+const mockResults: SessionSearchResult[] = [
+  {
+    sessionId: 'abc123',
+    summary: 'Fix auth bug',
+    snippet: '...looking at the auth middleware...',
+    matchedAt: Date.now() - 3600_000,
+    updatedAt: Date.now() - 3600_000,
+  },
+  {
+    sessionId: 'def456',
+    summary: null,
+    snippet: '...deploy pipeline changes...',
+    matchedAt: Date.now() - 86400_000,
+    updatedAt: Date.now() - 86400_000,
+  },
+];
+
+function renderBar(props: Partial<Parameters<typeof SessionSearchBar>[0]> = {}) {
+  const defaults = {
+    query: '',
+    setQuery: vi.fn(),
+    results: [] as SessionSearchResult[],
+    searching: false,
+    active: false,
+    clear: vi.fn(),
+    onSelectSession: vi.fn(),
+  };
+  return render(createElement(SessionSearchBar, { ...defaults, ...props }));
+}
+
+describe('SessionSearchBar', () => {
+  it('renders search toggle button', () => {
+    renderBar();
+    expect(screen.getByTitle('Search sessions')).toBeTruthy();
+  });
+
+  it('shows input when toggle is clicked', async () => {
+    const user = userEvent.setup();
+    renderBar();
+    await user.click(screen.getByTitle('Search sessions'));
+    expect(screen.getByPlaceholderText('Search sessions...')).toBeTruthy();
+  });
+
+  it('calls setQuery on input change', async () => {
+    const setQuery = vi.fn();
+    const user = userEvent.setup();
+    renderBar({ setQuery, active: false });
+    // Open the search bar
+    await user.click(screen.getByTitle('Search sessions'));
+    await user.type(screen.getByPlaceholderText('Search sessions...'), 'auth');
+    expect(setQuery).toHaveBeenCalled();
+  });
+
+  it('renders search results when active', () => {
+    renderBar({ active: true, query: 'auth', results: mockResults });
+    expect(screen.getByText('Fix auth bug')).toBeTruthy();
+    expect(screen.getByText('Untitled session')).toBeTruthy();
+  });
+
+  it('shows snippet text in results', () => {
+    renderBar({ active: true, query: 'auth', results: mockResults });
+    expect(screen.getByText('...looking at the auth middleware...')).toBeTruthy();
+  });
+
+  it('calls onSelectSession when result is clicked', async () => {
+    const onSelectSession = vi.fn();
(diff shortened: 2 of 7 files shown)


server/ws-handler-v2.ts
```diff
@@ -531,6 +560,22 @@ export function handleInterruptV2(
...
       }
+
+      // Transfer ownership if session was detached and we're reclaiming it
+      if (isDetached) {
+        reattachChat(found.clientId, transport);
+        const newClientId = `${connectionId}:${msg.sessionId}`;
+        if (found.clientId !== newClientId) {
+          rekeyChat(found.clientId, newClientId);
+          // Update found reference for interruptChat below
+          found.clientId = newClientId;
```
dimakis: 🔵 **style**: Mutates `found.clientId` directly to propagate the rekeyed ID to the `interruptChat` call at line 584. The `handleSendV2` path uses a separate `let activeClientId` variable instead. Using a local variable would be safer and consistent — `found` is a fresh wrapper from `findBySessionId`, but the mutation pattern is fragile if that ever changes. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -531,6 +560,22 @@ export function handleInterruptV2(
         });
         return;
       }
+
+      // Transfer ownership if session was detached and we're reclaiming it
```
dimakis, line 568: 🟡 **regressions**: In `handleInterruptV2`, the rekey logic only runs inside the `else` branch (when EventStore does NOT mark the session stale). When the stale-correction path is taken (lines 542-548), the code skips rekey but still calls `interruptChat(found.clientId)` with the old clientId at line 584. If the session was previously rekeyed by a `handleReconnect` or `handleSendV2` call, `found.clientId` from `findBySessionId` would already reflect the new key so this is fine. But if the stale-correction path and detached-reattach path ever need to combine, the interrupt handler lacks the rekey step that `handleSendV2` applies. Currently not a bug — just an asymmetry to be aware of.

frontend/src/components/__tests__/SessionSearchBar.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  it('shows searching indicator', () => {
+    renderBar({ active: true, query: 'auth', searching: true });
+    expect(screen.getByText('Searching...')).toBeTruthy();
+  });
+
+  it('shows no results message when active with empty results', () => {
+    renderBar({ active: true, query: 'zzz', results: [], searching: false });
+    expect(screen.getByText('No matches')).toBeTruthy();
+  });
```
dimakis, line 95: 🔵 **missing_tests**: The 'calls clear and hides input on close' test asserts `clear` was called but does not verify the input is actually removed from the DOM after close (e.g., `expect(screen.queryByPlaceholderText('Search sessions...')).toBeNull()`). The test name promises both behaviors but only checks one. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_The rekey-after-reattach fix is correct and well-tested — it closes the ownership split-brain that caused false `active_elsewhere` errors. Minor style inconsistency between the interrupt and send handler rekey patterns, and the PR bundles unrelated session search UI changes._

- 🔵 **style** (L571): Mutates `found.clientId` directly to propagate the rekeyed ID to the `interruptChat` call at line 584. The `handleSendV2` path uses a separate `let activeClientId` variable instead. Using a local variable would be safer and consistent — `found` is a fresh wrapper from `findBySessionId`, but the mutation pattern is fragile if that ever changes. `[fixable]`
- 🟡 **regressions** (L564): In `handleInterruptV2`, the rekey logic only runs inside the `else` branch (when EventStore does NOT mark the session stale). When the stale-correction path is taken (lines 542-548), the code skips rekey but still calls `interruptChat(found.clientId)` with the old clientId at line 584. If the session was previously rekeyed by a `handleReconnect` or `handleSendV2` call, `found.clientId` from `findBySessionId` would already reflect the new key so this is fine. But if the stale-correction path and detached-reattach path ever need to combine, the interrupt handler lacks the rekey step that `handleSendV2` applies. Currently not a bug — just an asymmetry to be aware of.


### `frontend/src/components/__tests__/SessionSearchBar.test.ts`
_The rekey-after-reattach fix is correct and well-tested — it closes the ownership split-brain that caused false `active_elsewhere` errors. Minor style inconsistency between the interrupt and send handler rekey patterns, and the PR bundles unrelated session search UI changes._

- 🔵 **missing_tests** (L95): The 'calls clear and hides input on close' test asserts `clear` was called but does not verify the input is actually removed from the DOM after close (e.g., `expect(screen.queryByPlaceholderText('Search sessions...')).toBeNull()`). The test name promises both behaviors but only checks one. `[fixable]`


### `frontend/src/components/SessionSearchBar.tsx`
_The rekey-after-reattach fix is correct and well-tested — it closes the ownership split-brain that caused false `active_elsewhere` errors. Minor style inconsistency between the interrupt and send handler rekey patterns, and the PR bundles unrelated session search UI changes._

- 🔵 **style**: The session search UI (SessionSearchBar component, hook wiring, CSS) is unrelated to the rekey fix described in the PR title. Splitting into a separate PR would make the review scope clearer and the git history more navigable.

server/ws-handler-v2.ts
```diff
@@ -506,6 +535,8 @@ export function handleInterruptV2(
   const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
   if (!found) return;
 
+  let activeClientId = found.clientId;
+
   // Ownership guard: reject if another connection actively drives this session
   if (isActive(found.clientId)) {
```
dimakis, line 541: 🟡 **bugs**: In `handleInterruptV2`, when `isActive()` is true but the EventStore says the session is stale (`storeMeta && !storeMeta.isActive`), the code skips the ownership guard entirely — but also skips reattach/rekey. `activeClientId` remains `found.clientId` (the dead connection's key). The subsequent `interruptChat(activeClientId, ...)` call on line 585 then targets the old/dead clientId. This is inconsistent with `handleSendV2`, which uses a separate `staleInMemory` branch to fall through to the resume path (where a fresh `startChat` is created). Here, the interrupt just fires against a stale key. This may be benign if the in-memory session still honors interrupts via the old key, but it's a divergence from the send handler's approach. `[fixable]`

frontend/src/components/SessionSearchBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+export function SessionSearchBar({
+  query,
+  setQuery,
+  results,
+  searching,
+  active,
+  clear,
+  onSelectSession,
+}: SessionSearchBarProps) {
+  const [open, setOpen] = useState(active);
```
dimakis: 🟡 **bugs**: `open` is initialized from `active` prop (`useState(active)`) but never syncs if `active` changes after mount. If the parent sets `active=true` (e.g., from a restored search query), the search bar stays closed because `open` is only set once. This is likely fine given that `active` is derived from `query.trim().length > 0` in `useSessionSearch` and starts as `false`, but it creates a subtle coupling — if the hook's initial state changes, the component will be out of sync. `[fixable]`

frontend/src/components/SessionSearchBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+        className="session-search-input"
+        type="text"
+        placeholder="Search sessions..."
+        value={query}
+        onChange={(e) => setQuery(e.target.value)}
+      />
+      <button className="session-search-close" onClick={handleClose} title="Close search">
+        ✕
+      </button>
+      {active && (
```
dimakis, line 69: 🔵 **style**: Results are shown when `active` (query is non-empty) but the results dropdown is inside the `open` conditional branch. If the user closes and reopens the search bar while the hook still has a query, `open` resets but `active` remains true. The `handleClose()` calls `clear()` which resets the query — so this works — but the render logic conflates two independent pieces of state (`open` for UI visibility, `active` for search activity). A simpler design would remove the `open` state entirely and derive visibility from `active`. `[fixable]`

frontend/src/components/SessionSearchBar.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <div className="session-search-status">No matches</div>
+          )}
+          {results.map((r) => (
+            <button
+              key={r.sessionId}
+              className="session-search-result"
+              onClick={() => handleSelect(r.sessionId)}
+            >
+              <div className="session-search-result-summary">{r.summary || 'Untitled session'}</div>
+              <div className="session-search-result-snippet">{r.snippet}</div>
```
dimakis, line 82: 🔵 **unsafe_assumptions**: The `r.snippet` value is rendered as raw text content. If the search backend ever returns HTML or user-generated content with angle brackets, React's JSX escaping will handle it safely. However, if this is later changed to use `dangerouslySetInnerHTML` for highlighting, it would become an XSS vector. Current implementation is safe.

frontend/src/styles/global.css
```diff
@@ -476,6 +476,132 @@ textarea:focus {
...
+  padding: 0.3rem 0.45rem;
+  border-radius: 6px;
+  background: transparent;
+  border: none;
+  color: var(--text-dim);
+  cursor: pointer;
+  line-height: 1;
+  -webkit-tap-highlight-color: transparent;
+}
+
```
dimakis, line 530: 🔵 **style**: `.session-search-results` uses `left: -60px` as a magic number to widen the dropdown beyond the input. This is fragile and will break on different screen widths or if the parent layout changes. Consider using a `min-width` or `width: max-content` approach instead. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Core rekey fix is well-structured and correctly addresses the false `active_elsewhere` problem across reconnect/send/interrupt paths. The `handleInterruptV2` stale-EventStore branch has a minor inconsistency where the old clientId is used without rekey. Session search UI is clean and well-tested for a v1._

- 🟡 **bugs** (L541): In `handleInterruptV2`, when `isActive()` is true but the EventStore says the session is stale (`storeMeta && !storeMeta.isActive`), the code skips the ownership guard entirely — but also skips reattach/rekey. `activeClientId` remains `found.clientId` (the dead connection's key). The subsequent `interruptChat(activeClientId, ...)` call on line 585 then targets the old/dead clientId. This is inconsistent with `handleSendV2`, which uses a separate `staleInMemory` branch to fall through to the resume path (where a fresh `startChat` is created). Here, the interrupt just fires against a stale key. This may be benign if the in-memory session still honors interrupts via the old key, but it's a divergence from the send handler's approach. `[fixable]`


_Core rekey fix is well-structured and correctly addresses the false `active_elsewhere` problem across reconnect/send/interrupt paths. The `handleInterruptV2` stale-EventStore branch has a minor inconsistency where the old clientId is used without rekey. Session search UI is clean and well-tested for a v1._

- 🔵 **missing_tests**: No test covers the `handleInterruptV2` path where the EventStore corrects stale in-memory state (`storeMeta && !storeMeta.isActive`). In this branch, no rekey happens and `activeClientId` remains the old dead connection's key. A test should verify correct behavior (or document this as intentional). `[fixable]`


### `frontend/src/components/SessionSearchBar.tsx`
_Core rekey fix is well-structured and correctly addresses the false `active_elsewhere` problem across reconnect/send/interrupt paths. The `handleInterruptV2` stale-EventStore branch has a minor inconsistency where the old clientId is used without rekey. Session search UI is clean and well-tested for a v1._

- 🟡 **bugs** (L24): `open` is initialized from `active` prop (`useState(active)`) but never syncs if `active` changes after mount. If the parent sets `active=true` (e.g., from a restored search query), the search bar stays closed because `open` is only set once. This is likely fine given that `active` is derived from `query.trim().length > 0` in `useSessionSearch` and starts as `false`, but it creates a subtle coupling — if the hook's initial state changes, the component will be out of sync. `[fixable]`
- 🔵 **style** (L68): Results are shown when `active` (query is non-empty) but the results dropdown is inside the `open` conditional branch. If the user closes and reopens the search bar while the hook still has a query, `open` resets but `active` remains true. The `handleClose()` calls `clear()` which resets the query — so this works — but the render logic conflates two independent pieces of state (`open` for UI visibility, `active` for search activity). A simpler design would remove the `open` state entirely and derive visibility from `active`. `[fixable]`
- 🔵 **unsafe_assumptions** (L81): The `r.snippet` value is rendered as raw text content. If the search backend ever returns HTML or user-generated content with angle brackets, React's JSX escaping will handle it safely. However, if this is later changed to use `dangerouslySetInnerHTML` for highlighting, it would become an XSS vector. Current implementation is safe.


### `frontend/src/components/__tests__/SessionSearchBar.test.ts`
_Core rekey fix is well-structured and correctly addresses the false `active_elsewhere` problem across reconnect/send/interrupt paths. The `handleInterruptV2` stale-EventStore branch has a minor inconsistency where the old clientId is used without rekey. Session search UI is clean and well-tested for a v1._

- 🔵 **missing_tests**: No test verifies keyboard interaction (e.g., Escape to close, Enter to select first result). Also no test for the `active=true` prop being passed at mount time — the `open` state initializes from `active`, so the search bar should render open immediately, but this path is untested. `[fixable]`


_Core rekey fix is well-structured and correctly addresses the false `active_elsewhere` problem across reconnect/send/interrupt paths. The `handleInterruptV2` stale-EventStore branch has a minor inconsistency where the old clientId is used without rekey. Session search UI is clean and well-tested for a v1._

- 🔵 **style** (L530): `.session-search-results` uses `left: -60px` as a magic number to widen the dropdown beyond the input. This is fragile and will break on different screen widths or if the parent layout changes. Consider using a `min-width` or `width: max-content` approach instead. `[fixable]`

PR #276, opened 2026-04-22
fix(server): resolve active_elsewhere false positive and worktree collision

dimakis:
## Summary
- **active_elsewhere false positive**: `handleSendV2`/`handleInterruptV2` rejected with "active on another device" when the same device reconnected with a new `connectionId` before the old WS `onclose` fired. Added `ownerGone` check (matching `handleReconnect` pattern) so a dead owner connection triggers takeover instead of rejection.
- **worktree collision**: `createWorktree` crashed when the worktree path already existed from a stale session. Now checks for existing paths — reuses valid worktrees, removes stale directories + prunes before retrying.

## Test plan
- [x] New test: `should allow send when owner connection is gone (ownerGone takeover)`
- [x] New test: worktree path collision handling
- [ ] Manual: reconnect from phone after backgrounding app — no "active on another device" error
- [ ] Manual: create session, kill server, restart — worktree reuses correctly

### DIFF
--- a/server/__tests__/ws-handler-v2.test.ts
+++ b/server/__tests__/ws-handler-v2.test.ts
@@ -1480,6 +1480,8 @@ describe('handleSendV2 connection ownership', () => {
     });
     const transport = mockTransport();
     ctx.connRegistry.register('c1', transport);
+    // Register the other connection so it's truly alive (not gone)
+    ctx.connRegistry.register('other-conn', mockTransport());
 
     handleSendV2(
       'c1',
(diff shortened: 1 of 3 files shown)


dimakis:
## Centaur Review

### `server/ws-handler-v2.ts`
_The active_elsewhere false-positive fix is correct and well-tested for handleSendV2/handleInterruptV2, but handleStopV2 is missing the same ownership guard (pre-existing), and the worktree collision fix lacks unit test coverage._

- 🟡 **regressions** (L525): handleStopV2 has no ownership check at all — any connection can stop any session by sessionId. handleSendV2 and handleInterruptV2 both received the ownerGone fix in this PR, but handleStopV2 was not updated. A client on one device could stop a session actively running on another device. This is a pre-existing issue, not introduced by this PR, but it's the same pattern being fixed here and should be addressed. `[fixable]`


### `server/worktree.ts`
_The active_elsewhere false-positive fix is correct and well-tested for handleSendV2/handleInterruptV2, but handleStopV2 is missing the same ownership guard (pre-existing), and the worktree collision fix lacks unit test coverage._

- 🟡 **missing_tests** (L61): createWorktree has no unit tests at all — the existing worktree.test.ts only covers parseWorktreeAge, countWorktrees, and cleanupStaleWorktrees. The new collision-handling logic (reuse valid worktree, remove stale directory, prune) has three code paths and none are tested. Given the TDD mandate in CLAUDE.md, this is a gap. `[fixable]`
- 🔵 **unsafe_assumptions** (L62): The reuse path returns early (line 69) when git rev-parse succeeds, assuming the existing worktree is correct for this session. However, rev-parse --git-dir only validates that the directory is *some* valid worktree — it doesn't verify the worktree is on the expected branch or belongs to the expected session. If a different session's worktree occupies the same path (unlikely with UUID session IDs, but possible with user-provided opts), this silently reuses the wrong worktree. `[fixable]`
- 🔵 **style** (L79): The empty catch block on git worktree prune (line 79) silently swallows all errors. A debug-level log would help diagnose cases where prune fails and subsequent git worktree add also fails due to stale metadata. `[fixable]`


_The active_elsewhere false-positive fix is correct and well-tested for handleSendV2/handleInterruptV2, but handleStopV2 is missing the same ownership guard (pre-existing), and the worktree collision fix lacks unit test coverage._

- 🔵 **missing_tests** (L1527): The new 'owner connection is gone' tests for both handleSendV2 and handleInterruptV2 verify reattachChat but do not assert that rekeyChat was called. The implementation does call rekeyChat to transfer ownership after reattach. A separate test ('rekey after detached reattach') covers this, but the new tests should verify the full ownership-transfer contract for the ownerGone path specifically, since rekey behavior after a dead-WS reconnect is the core of this fix. `[fixable]`

PR #278, opened 2026-04-22
feat: session isolation overhaul — Phase 1

dimakis:
## Summary

Implements the session isolation overhaul Phase 1 from `docs/design/session-isolation-overhaul.md`, fixing the 6 highest-impact WS plumbing bugs plus collision-proof worktree IDs from Phase 2b.

- **P5 Stale Registry Entries**: `registry.remove()` called in all 3 stale-detection sites instead of only correcting a local variable
- **P1 Session Bleed**: server unwatches previous session on switch; client clears `seqBySession` on switch/new; null-session filter tightened to only pass `session_id`
- **Takeover (design doc 1.2)**: send/interrupt from another device performs full ownership transfer (session_takeover event, unwatch, deny pending permissions, reattach+rekey) instead of `active_elsewhere` rejection. Reconnect stays passive (R1)
- **P2 Reconnect Storm**: handshake timeout reduced to 1s, dead sockets and invalid JSON closed immediately instead of routing to v1
- **P6 Resume Breaks**: new `validateResumable()` checks git dir validity before SDK query, attempts worktree recreation, falls back to fresh start with friendly error
- **Client recovery**: `onReconnected` callback always re-fetches messages; foreground handler always re-fetches (removes empty-store guard); `session_takeover` handler; `active_elsewhere` removed; session-expired no longer wipes client state
- **Observability**: `session` span wrapping query loop, `session.detach` span on WS close, LOKI_HOST uncommented
- **P3 Worktree ID Collision**: `randomUUID().slice(0, 12)` (was `randomBytes(3)`, 6 chars); `.mitzo-session` lockfile in worktrees

## Test plan

- [x] 3 new tests: stale session removal from registry (reconnect, send, interrupt)
- [x] 2 new tests: switch_session unwatches previous (different session, same session)
- [x] 3 new tests: getTrackedSessions (returns tracked, empty, excludes cleared)
- [x] 2 new tests: null-session filter drops session_end and permission_request
- [x] 3 updated tests: send/interrupt takeover replaces active_elsewhere rejection
- [x] 1 updated test: EventStore-confirmed active → takeover instead of reject
- [x] 4 new tests: validateResumable (valid, recreated, invalid path, recreation failure)
- [x] 2 new tests: generateWtId uniqueness and format
- [x] 1 new test: reconnect recovery re-fetches messages
- [x] 1 new test: session_takeover produces SET_RUNNING false + ERROR
- [x] 1 new test: onReconnected callback called
- [x] 1 new test: "No conversation found" no longer calls onSessionExpired
- [x] 3 updated tests: foreground recovery, session-expired, clobber guard reflect new behavior
- [x] All 14,418+ targeted tests pass

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -596,6 +596,22 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       connection.send(msg);
     },
 
+    onReconnected() {
+      const activeId = parserState.currentSessionId;
+      if (activeId) {
+        api
+          .getSessionMessages(activeId)
+          .then((msgs) => {
+            if (Array.isArray(msgs) && msgs.length > 0) {
+              store.setState((s) => ({
+                messages: messagesReducer(s.messages, { type: 'RESTORE', messages: msgs }),
+              }));
+            }
+          })
+          .catch(() => {});
+      }
+    },
+
     onTokensHydrated(tokens: Record<string, unknown>) {
       store.setState((s) => ({
         tokens: {
@@ -615,19 +631,15 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     // evicted it from memory, losing in-memory state), re-fetch messages from
     // the REST API if we have an active session but no messages in the store.
     if (msg.type === '_foreground') {
-      const { sessions, messages: msgs } = store.getState();
-      if (sessions.active && msgs.messages.length === 0 && !msgs.current) {
+      const { sessions } = store.getState();
+      if (sessions.active) {
         api
           .getSessionMessages(sessions.active)
           .then((restored) => {
             if (Array.isArray(restored) && restored.length > 0) {
-              // Only restore if store is still empty (avoid clobbering live data)
-              const current = store.getState().messages;
-              if (current.messages.length === 0 && !current.current) {
-                store.setState((s) => ({
-                  messages: messagesReducer(s.messages, { type: 'RESTORE', messages: restored }),
-                }));
-              }
+              store.setState((s) => ({
+                messages: messagesReducer(s.messages, { type: 'RESTORE', messages: restored }),
+              }));
             }
           })
           .catch((err) => {
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -196,6 +196,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     async switchSession(id: string) {
+      const oldId = parserState.currentSessionId;
+      if (oldId) connection.clearSession(oldId);
       parserState.currentSessionId = id;
 
       set((s) => ({
@@ -221,6 +223,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     newSession() {
+      for (const sid of connection.getTrackedSessions()) {
+        connection.clearSession(sid);
+      }
       parserState.currentSessionId = undefined;
       parserState.pendingSend = null;
       clearPendingSendTimer();
@@ -645,15 +650,12 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       if (parserState.currentSessionId) {
         if (eventSessionId !== parserState.currentSessionId) return;
       } else {
-        // Allow session assignment, session end, and permission requests through
-        // before session_id arrives. Permission requests can arrive for brand-new
-        // sessions before the session_id event — dropping them blocks the user
-        // from answering tool prompts.
-        const isEarlySessionEvent =
-          msg.type === 'session_id' ||
-          msg.type === 'session_end' ||
-          msg.type === 'permission_request';
-        if (!isEarlySessionEvent) return;
+        // Only allow session_id through when no active session — this is the
+        // assignment event for a newly started session. All other session-scoped
+        // events (session_end, permission_request) are dropped to prevent
+        // foreign session bleed during the window between newSession() and
+        // receiving the new session_id.
+        if (msg.type !== 'session_id') return;
       }
     }
 
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -166,6 +166,30 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     pendingSend: null,
   };
 
+  let recoveryInFlight = false;
+
+  function fetchAndRestoreMessages(sessionId: string) {
+    if (recoveryInFlight) return;
+    recoveryInFlight = true;
+    api
+      .getSessionMessages(sessionId)
+      .then((msgs) => {
+        if (Array.isArray(msgs) && msgs.length > 0) {
+          store.setState((s) => ({
+            messages: messagesReducer(s.messages, { type: 'RESTORE', messages: msgs }),
+          }));
+        }
+      })
+      .catch((err) => {
+        if (typeof console !== 'undefined') {
+          console.warn('[mitzo] message recovery fetch failed', err);
+        }
+      })
+      .finally(() => {
+        recoveryInFlight = false;
+      });
+  }
+
   function clearPendingSendTimer() {
     if (parserState.pendingSendTimer) {
       clearTimeout(parserState.pendingSendTimer);
@@ -560,13 +584,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     },
 
     onSessionExpired() {
-      parserState.currentSessionId = undefined;
-      parserState.pendingSend = null;
-      clearPendingSendTimer();
-      store.setState((s) => ({
-        sessions: { ...s.sessions, active: null },
-        messages: INITIAL_MESSAGES_STATE,
-      }));
+      // No-op: server-side resume validation handles expired sessions now.
+      // Kept to satisfy ProtocolCallbacks interface.
     },
 
     onSessionRenamed(name: string) {
@@ -598,18 +617,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
     onReconnected() {
       const activeId = parserState.currentSessionId;
-      if (activeId) {
-        api
-          .getSessionMessages(activeId)
-          .then((msgs) => {
-            if (Array.isArray(msgs) && msgs.length > 0) {
-              store.setState((s) => ({
-                messages: messagesReducer(s.messages, { type: 'RESTORE', messages: msgs }),
-              }));
-            }
-          })
-          .catch(() => {});
-      }
+      if (activeId) fetchAndRestoreMessages(activeId);
     },
 
     onTokensHydrated(tokens: Record<string, unknown>) {
@@ -632,22 +640,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     // the REST API if we have an active session but no messages in the store.
     if (msg.type === '_foreground') {
       const { sessions } = store.getState();
-      if (sessions.active) {
-        api
-          .getSessionMessages(sessions.active)
-          .then((restored) => {
-            if (Array.isArray(restored) && restored.length > 0) {
-              store.setState((s) => ({
-                messages: messagesReducer(s.messages, { type: 'RESTORE', messages: restored }),
-              }));
-            }
-          })
-          .catch((err) => {
-            if (typeof console !== 'undefined') {
-              console.warn('[mitzo] foreground recovery fetch failed', err);
-            }
-          });
-      }
+      if (sessions.active) fetchAndRestoreMessages(sessions.active);
       return;
     }
 
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -174,9 +174,12 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     api
       .getSessionMessages(sessionId)
       .then((msgs) => {
-        if (Array.isArray(msgs) && msgs.length > 0) {
+        if (Array.isArray(msgs)) {
           store.setState((s) => ({
-            messages: messagesReducer(s.messages, { type: 'RESTORE', messages: msgs }),
+            messages:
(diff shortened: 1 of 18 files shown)


dimakis:
## Centaur Review

### `packages/harness/src/permission-handler.ts`
_Session takeover logic is well-designed, but `denyPendingBySession()` is dead code because `registerPending()` never stores the sessionId, and the 'No conversation found' retry path logs a retry but doesn't actually retry — both are critical bugs that undermine the PR's stated goals._

- 🔴 **bugs** (L111): `registerPending()` is called without the `sessionId` parameter (6th arg), so `denyPendingBySession(sessionId)` in ws-handler-v2.ts will never match any entries — all pending entries have `sessionId: undefined`. The takeover permission cleanup is a no-op. Pass `session.sessionId` as the 6th argument. `[fixable]`


### `server/chat.ts`
_Session takeover logic is well-designed, but `denyPendingBySession()` is dead code because `registerPending()` never stores the sessionId, and the 'No conversation found' retry path logs a retry but doesn't actually retry — both are critical bugs that undermine the PR's stated goals._

- 🔴 **bugs** (L681): The log says 'retrying without resume' but no retry happens — execution falls through to `registry.abort(clientId)` unconditionally. The session is cleaned up and aborted, so the user gets an error with no actual retry. Either implement the retry (call `startChat` again with `resume` removed) or fix the log message to say 'starting fresh requires user re-send'. `[fixable]`
- 🔵 **style** (L301): Writing `.mitzo-session` into the worktree is a new side effect that isn't mentioned in the PR description or tested. If the primary worktree creation fails after this write, the file is left behind. Consider documenting the purpose of this marker file and adding it to `.gitignore` if it shouldn't be committed.


_Session takeover logic is well-designed, but `denyPendingBySession()` is dead code because `registerPending()` never stores the sessionId, and the 'No conversation found' retry path logs a retry but doesn't actually retry — both are critical bugs that undermine the PR's stated goals._

- 🟡 **regressions** (L662): The null-session filter now only allows `session_id` through when no active session. Previously `permission_request` was also allowed through for brand-new sessions where the permission fires before `session_id` arrives from the SDK. If a tool permission is requested on the very first SDK event (before `session_id`), the client will drop it and the user can never approve the tool — the session blocks forever. The test at line 782 validates the new behavior but this may be a real-world regression. `[fixable]`
- 🟡 **unsafe_assumptions** (L633): Foreground recovery now always re-fetches messages even when the store already has data, removing the `messages.length === 0` guard and the 'only restore if store is still empty' check. If the user is mid-conversation and the app briefly backgrounds/foregrounds (common on iOS), this will fire a REST call and RESTORE action that could reset scroll position or cause a flash. The old guard existed to prevent this. `[fixable]`
- 🔵 **style** (L610): `onReconnected()` silently swallows fetch errors with `.catch(() => {})`. Unlike the `_foreground` handler (which logs via `console.warn`), reconnect failures are completely invisible. Add at minimum a `console.warn` for debuggability. `[fixable]`


### `server/index.ts`
_Session takeover logic is well-designed, but `denyPendingBySession()` is dead code because `registerPending()` never stores the sessionId, and the 'No conversation found' retry path logs a retry but doesn't actually retry — both are critical bugs that undermine the PR's stated goals._

- 🟡 **unsafe_assumptions** (L238): Handshake timeout reduced from 5s to 1s. On high-latency mobile networks (e.g., poor cellular), the WS may open but the first message can take >1s to arrive, causing legitimate connections to be killed with code 4000. Consider 2-3s as a safer middle ground. `[fixable]`


### `packages/harness/src/permissions.ts`
_Session takeover logic is well-designed, but `denyPendingBySession()` is dead code because `registerPending()` never stores the sessionId, and the 'No conversation found' retry path logs a retry but doesn't actually retry — both are critical bugs that undermine the PR's stated goals._

- 🟡 **missing_tests** (L63): `denyPendingBySession()` has no dedicated unit test. The ws-handler tests mock it, but there's no test verifying it actually finds and denies entries by sessionId — especially important given that `registerPending` doesn't currently pass sessionId (bug above). `[fixable]`

dimakis:
## Centaur Review

### `packages/harness/src/permission-handler.ts`
_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🔴 **bugs** (L111): `registerPending` is called without `sessionId` (only 5 args: permId, toolName, wrappedResolve, _toolInput, tier). The new `denyPendingBySession(sessionId)` iterates entries looking for `entry.sessionId === sessionId`, but since no caller passes sessionId to `registerPending`, `entry.sessionId` is always `undefined`. The function will always return 0 and never deny any permissions during session takeover. Fix: pass the session's sessionId as the 6th argument to `registerPending` in `permission-handler.ts`. `[fixable]`


### `server/chat.ts`
_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🟡 **bugs** (L681): Log message says 'SDK rejected resume, retrying without resume' but no retry occurs. The catch block sends an error to the client and falls through to `registry.abort(clientId)` + `cleanupSessionWorktrees`, terminating the session. The user receives 'Session expired. Starting fresh.' but nothing actually starts fresh — they'd need to manually send another message. Either implement the retry (re-call startChat without resume) or fix the misleading log message. `[fixable]`


_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🟡 **unsafe_assumptions** (L596): Both `onReconnected` (line 596) and the `_foreground` handler (line 633) independently fetch messages and dispatch RESTORE for the same active session. When the app returns from background AND the WS reconnects simultaneously (common on iOS), both fire and race. Two concurrent fetches can resolve in either order, each overwriting `state.messages`. The RESTORE reducer's dedup check (line 348-353 in messages.ts) helps if the second response is identical, but if the server state changed between the two fetches, the final state depends on resolution order. Consider deduplicating — e.g., skip `_foreground` fetch if a reconnect-triggered fetch is already in flight. `[fixable]`
- 🔵 **style** (L611): Silent `.catch(() => {})` in `onReconnected` swallows all errors with no logging. The adjacent `_foreground` handler (line 645) logs failures via `console.warn`. These two recovery paths should have consistent error handling. `[fixable]`


_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🟡 **regressions** (L328): `onSessionExpired` is now dead code — the only call site (the 'No conversation found' error branch) was removed, but the callback definition (protocol-parser.ts:31) and implementation (store.ts:562-570) remain. This is harmless but should be cleaned up to avoid confusion about whether session expiration is still a supported flow. `[fixable]`


### `packages/harness/__tests__/permissions.test.ts`
_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🟡 **missing_tests**: `denyPendingBySession` has no unit tests in the harness permissions test suite. The ws-handler-v2 tests only verify it's called (mocked), not that it works. Add tests covering: denying matching entries, skipping non-matching entries, and the current bug where no entries have sessionId set. `[fixable]`


_The session takeover mechanism is well-designed but `denyPendingBySession` is silently broken — `registerPending` never receives a `sessionId`, so takeover cannot deny stale permissions. The misleading retry log and the reconnect/foreground race condition are secondary concerns._

- 🔵 **missing_tests**: The `startChat` catch block for 'No conversation found' (the non-retry path at chat.ts:680-682) has no test coverage. The `validateResumable` tests cover pre-SDK validation but not the SDK-level error path. `[fixable]`

packages/client/src/store.ts
```diff
@@ -645,15 +662,12 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       if (parserState.currentSessionId) {
```
dimakis, line 658: 🟡 **regressions**: Tightening the null-session filter to allow only `session_id` drops `permission_request` events that arrive before the first `session_id` on a brand-new session. The previous comment documents this ordering can happen, and if it does the user can no longer answer the tool prompt because the request is discarded. The added test only covers a foreign-session request with no active session, not the legitimate first-turn race this code used to handle. `[fixable]`

server/chat.ts
```diff
@@ -615,8 +677,13 @@ export async function startChat(
     );
```
dimakis, line 677: 🟡 **bugs**: When `startChat()` catches an SDK resume failure containing "No conversation found", it now only sends an error and aborts cleanup, but it never retries without `options.resume`. The client-side changes also stopped clearing the session / resending fresh, so this path leaves the user with a dead session instead of actually "starting fresh" as the message claims. `[fixable]`

server/query-loop.ts
```diff
@@ -121,6 +123,9 @@ export async function runQueryLoop(
   initialPrompt?: string,
```
dimakis, line 123: 🟡 **bugs**: The new OpenTelemetry span started at the top of `runQueryLoop()` is only ended in the `finally` success path. If any exception escapes before that `finally` block completes normally, or if the function returns early before reaching the end-of-loop cleanup, the span can be left open. This should be wrapped in a `try/finally` that always ends the span. `[fixable]`

dimakis (commented):
## Codex Review

_PR adds useful session-takeover coverage, but it introduces a first-turn permission regression and an incomplete resume-recovery path._

- 🟡 **regressions** (L662): Tightening the null-session filter to allow only `session_id` drops `permission_request` events that arrive before the first `session_id` on a brand-new session. The previous comment documents this ordering can happen, and if it does the user can no longer answer the tool prompt because the request is discarded. The added test only covers a foreign-session request with no active session, not the legitimate first-turn race this code used to handle. `[fixable]`


### `server/chat.ts`
_PR adds useful session-takeover coverage, but it introduces a first-turn permission regression and an incomplete resume-recovery path._

- 🟡 **bugs** (L677): When `startChat()` catches an SDK resume failure containing "No conversation found", it now only sends an error and aborts cleanup, but it never retries without `options.resume`. The client-side changes also stopped clearing the session / resending fresh, so this path leaves the user with a dead session instead of actually "starting fresh" as the message claims. `[fixable]`


### `server/query-loop.ts`
_PR adds useful session-takeover coverage, but it introduces a first-turn permission regression and an incomplete resume-recovery path._

- 🟡 **bugs** (L123): The new OpenTelemetry span started at the top of `runQueryLoop()` is only ended in the `finally` success path. If any exception escapes before that `finally` block completes normally, or if the function returns early before reaching the end-of-loop cleanup, the span can be left open. This should be wrapped in a `try/finally` that always ends the span. `[fixable]`

packages/client/src/store.ts
```diff
@@ -166,6 +166,30 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     pendingSend: null,
   };
 
+  let recoveryInFlight = false;
```
dimakis, line 169: 🟡 **bugs**: `fetchAndRestoreMessages()` only dispatches RESTORE when the fetched array is non-empty. If the active session was deleted/reset server-side and now has zero messages, reconnect/foreground recovery leaves stale local messages visible forever instead of clearing them to match server truth. `[fixable]`

server/index.ts
```diff
@@ -229,15 +229,18 @@ server.on('upgrade', async (req, socket, head) => {
...
  * v1 clients send a regular message (send, reattach, etc.).
- * Timeout fallback: if no message within 5s, assume v1.
+ *
+ * Dead sockets (no message within 1s) and invalid JSON are closed
+ * immediately instead of falling through to v1. This prevents reconnect
+ * storms from stale/dead connections consuming resources.
  */
 function routeWsClient(ws: WebSocket, assignedId: string) {
-  const HANDSHAKE_TIMEOUT_MS = 5_000;
+  const HANDSHAKE_TIMEOUT_MS = 1_000;
```
dimakis, line 239: 🟡 **regressions**: The websocket handshake timeout was reduced from 5s to 1s and now closes non-hello sockets instead of falling back to v1. The repo docs still describe supporting legacy v1 clients, so this change effectively removes that compatibility path and may break slower/mobile clients or any remaining v1 consumers. `[fixable]`

server/query-loop.ts
```diff
@@ -121,6 +123,9 @@ export async function runQueryLoop(
...
   options?: QueryLoopOptions,
 ) {
+  const span = tracer.startSpan('session');
```
dimakis, line 126: 🟡 **bugs**: The new OpenTelemetry span started at the top of `runQueryLoop()` is only ended in the normal `finally` path after the main loop. If an exception escapes before reaching that block, the span is leaked because there is no outer try/finally guarding the whole function. `[fixable]`

dimakis (commented):
## Codex Review

_Solid test coverage overall, but there are a few behavioral regressions and one observability leak to address before merge._

- 🟡 **bugs** (L169): `fetchAndRestoreMessages()` only dispatches RESTORE when the fetched array is non-empty. If the active session was deleted/reset server-side and now has zero messages, reconnect/foreground recovery leaves stale local messages visible forever instead of clearing them to match server truth. `[fixable]`


### `server/index.ts`
_Solid test coverage overall, but there are a few behavioral regressions and one observability leak to address before merge._

- 🟡 **regressions** (L239): The websocket handshake timeout was reduced from 5s to 1s and now closes non-hello sockets instead of falling back to v1. The repo docs still describe supporting legacy v1 clients, so this change effectively removes that compatibility path and may break slower/mobile clients or any remaining v1 consumers. `[fixable]`


### `server/query-loop.ts`
_Solid test coverage overall, but there are a few behavioral regressions and one observability leak to address before merge._

- 🟡 **bugs** (L126): The new OpenTelemetry span started at the top of `runQueryLoop()` is only ended in the normal `finally` path after the main loop. If an exception escapes before reaching that block, the span is leaked because there is no outer try/finally guarding the whole function. `[fixable]`

PR #279, opened 2026-04-22
docs: Phase 2 session isolation handoff

dimakis:
Handoff document for the next agent implementing Phase 2 of the session isolation overhaul. Contains implementation specifics, file locations, dependency order, test plan, and warnings.

### DIFF
--- a/docs/design/session-isolation-phase2-handoff.md
+++ b/docs/design/session-isolation-phase2-handoff.md
@@ -1,0 +1,220 @@
+# Session Isolation Phase 2 — Handoff for Next Agent
+
+**Context:** Phase 1 of the session isolation overhaul was merged in PR #278 (squash-merged to main, deployed via launchd). This document is the complete handoff for Phase 2 implementation.
+
+**Design doc:** `docs/design/session-isolation-overhaul.md` — Phase 2 starts at the "## Phase 2: Bulletproof Worktree Isolation" section (L411). Read that first, then come back here for implementation specifics.
+
+**Prior art:** PR #278 commit history shows the Phase 1 TDD pattern. The centaur + codex review findings and fixes are instructive for what reviewers will flag.
+
+---
+
+## What Phase 1 Already Shipped
+
+These are done and deployed — don't re-implement:
+
+- **2b (Collision-proof IDs):** `generateWtId()` in `server/chat.ts` now uses `randomUUID().slice(0, 12)`. `.mitzo-session` lockfile written inside primary worktree at creation time.
+- **Stale registry cleanup:** `registry.remove()` in all 3 handler stale-detection sites
+- **Session bleed fix:** server unwatches on switch, client clears `seqBySession`, null-session filter tightened (allows `session_id` + `permission_request` only)
+- **Auto-takeover:** send/interrupt from another device does full ownership transfer instead of `active_elsewhere` rejection
+- **Reconnect storm:** 1s handshake timeout, dead sockets closed
+- **Resume validation:** `validateResumable()` checks git dir before SDK query
+- **Client recovery:** `onReconnected` + `_foreground` both go through shared `fetchAndRestoreMessages()` with in-flight dedup guard
+- **OTel spans:** `session` span wraps query loop, `session.detach` span on WS close, LOKI_HOST enabled
+
+---
+
+## Phase 2 Scope: 5 Steps (2b is done)
+
+### 2a. Lazy Worktree Creation
+
+**Problem (P7):** `createSessionWorktrees()` in `server/chat.ts` L317-332 creates worktrees for ALL configured secondary repos at session start. With 11 repos and 5 sessions = 55 worktree dirs, most never used.
+
+**Files to modify:**
+
+1. `server/chat.ts` `createSessionWorktrees()` (L278-338) — remove the secondary repo loop at L317-332. Only keep primary worktree creation (L296-315). The function signature and return type stay the same.
+
+2. `packages/harness/src/worktree-guard.ts` `checkWorktreePolicy()` (L80-143) — currently sync, returns `string | null`. Needs to become async or the on-demand creation needs to happen in the permission handler wrapper (`packages/harness/src/permission-handler.ts` L44-146, the `canUseTool` callback is already async).
+
+   The on-demand flow when a write tool targets a configured repo with no worktree:
+   - Identify the repo from the file path (compare against `config.repos` values from `server/repo-config.ts`)
+   - Call async `createWorktreeAsync(session.wtId, repoPath)` (new, see 2f)
+   - Add to `session.worktreePaths` Map
+   - Broadcast `worktree_opened` event to client
+   - Return deny with redirect message (existing behavior — agent retries with the worktree path)
+
+3. `server/chat.ts` `buildWorktreeSystemPrompt()` (L380-402) — update text to note secondary repos are lazy: "Worktrees for secondary repos are created on first write. Use `$MITZO_REPO_<NAME>` env vars."
+
+**Key constraint:** `checkWorktreePolicy` is called from `buildPermissionHandler` which already returns `Promise<PermissionResult>`. The guard itself is sync — either make it async (changing the return type to `Promise<string | null>`) or do the creation in the permission handler wrapper before calling the guard.
+
+**Test plan:**
+
+- `server/__tests__/chat.test.ts`: Session start creates only primary worktree (verify `createWorktree` called once)
+- `packages/harness/__tests__/worktree-guard.test.ts`: Write to uncovered configured repo triggers on-demand creation
+
+### 2c. Runtime Symlinks (opt-in)
+
+**Problem (P4):** `.venv`, `node_modules` don't exist in worktrees. `PATH` injection handles executables, but CWD-relative tool resolution (e.g. `./node_modules/.bin/vitest`) still fails.
+
+**Files to modify:**
+
+1. `server/worktree.ts` — new `symlinkRuntimeDirs(repoPath: string, worktreePath: string, dirs: string[]): void`. For each dir in `dirs`, if `join(repoPath, dir)` exists, create `symlink(join(repoPath, dir), join(worktreePath, dir))`.
+
+2. `server/repo-config.ts` — add `runtimeSymlinks: string[]` to `RepoConfig` interface (L30-36) and parser (around L120). Default: empty array `[]`.
+
+3. Integration: call `symlinkRuntimeDirs` after worktree creation (both eager primary and lazy secondary).
+
+**Design decision (R3):** Symlinks are opt-in escape hatch, NOT the primary mechanism. `PATH` injection via `sdkEnv()` (L169-193) remains primary. System prompt must warn: "Symlinked runtime dirs are shared mutable state across sessions."
+
+**Config example:**
+
+```json
+{ "runtimeSymlinks": [".venv", "node_modules"] }
+```
+
+**Test plan:**
+
+- `server/__tests__/worktree.test.ts`: `symlinkRuntimeDirs` creates symlinks for existing dirs, skips missing
+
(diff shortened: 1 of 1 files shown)


docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+**Context:** Phase 1 of the session isolation overhaul was merged in PR #278 (squash-merged to main, deployed via launchd). This document is the complete handoff for Phase 2 implementation.
+
+**Design doc:** `docs/design/session-isolation-overhaul.md` — Phase 2 starts at the "## Phase 2: Bulletproof Worktree Isolation" section (L411). Read that first, then come back here for implementation specifics.
+
+**Prior art:** PR #278 commit history shows the Phase 1 TDD pattern. The centaur + codex review findings and fixes are instructive for what reviewers will flag.
+
+---
+
+## What Phase 1 Already Shipped
+
```
dimakis, line 12: 🟡 **bugs**: Phase 1 claim that `generateWtId()` uses `randomUUID().slice(0, 12)` is inaccurate. The actual code does `randomUUID().replace(/-/g, '').slice(0, 12)` — it strips dashes first, then slices. The result is 12 hex chars, not a UUID substring with dashes. This distinction matters if the next agent tries to verify or extend the function. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+    cleanupSessionWorktrees(session);
+}
+```
+
+**New code:**
+
+```typescript
+} finally {
+    // Worktrees survive until explicit close or stale GC.
+    // cleanupSessionWorktrees removed — see design doc 2e.
```
dimakis, line 119: 🟡 **bugs**: Section 2e describes the problem as `cleanupSessionWorktrees(session)` "destroys worktrees after every query loop, breaking resume" — but the function explicitly skips the primary worktree (`if (repoName === 'primary') continue;` at chat.ts:828). Resume uses the primary worktree CWD, so it's not broken today. The real problem is that Phase 2 lazy-created secondaries would be lost, which is a valid concern but different from what's stated. An agent reading this literally might think resume is currently broken. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+| `server/chat.ts`                             | `startChat()` finally block   | L693-698                              |
+| `server/chat.ts`                             | `validateResumable()`         | L116-157                              |
+| `server/chat.ts`                             | `cleanupSessionWorktrees()`   | L803+                                 |
+| `server/worktree.ts`                         | `createWorktree()` (sync)     | L46-102                               |
+| `server/worktree.ts`                         | `cleanupStaleWorktrees()`     | L271+                                 |
+| `server/constants.ts`                        | `WORKTREE_STALE_HOURS`        | L29 (currently 96)                    |
+| `server/repo-config.ts`                      | `RepoConfig` interface        | L30-36 (`repos` field)                |
+| `packages/harness/src/worktree-guard.ts`     | `checkWorktreePolicy()`       | L80-143 (sync)                        |
+| `packages/harness/src/permission-handler.ts` | `buildPermissionHandler()`    | L44-146 (async `canUseTool`)          |
+
```
dimakis, line 170: 🟡 **unsafe_assumptions**: `buildWorktreeSystemPrompt()` is listed at L380-402, but the actual range is L377-399. `RepoConfig` interface is listed at L30-36 but actually starts at L23. `cleanupSessionWorktrees()` is listed at L803+ but actually starts at L823. While minor, the next agent will use these line references to navigate — stale references waste time or cause confusion, especially in an automated agent workflow. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+   - Return deny with redirect message (existing behavior — agent retries with the worktree path)
+
+3. `server/chat.ts` `buildWorktreeSystemPrompt()` (L380-402) — update text to note secondary repos are lazy: "Worktrees for secondary repos are created on first write. Use `$MITZO_REPO_<NAME>` env vars."
+
+**Key constraint:** `checkWorktreePolicy` is called from `buildPermissionHandler` which already returns `Promise<PermissionResult>`. The guard itself is sync — either make it async (changing the return type to `Promise<string | null>`) or do the creation in the permission handler wrapper before calling the guard.
+
+**Test plan:**
+
+- `server/__tests__/chat.test.ts`: Session start creates only primary worktree (verify `createWorktree` called once)
+- `packages/harness/__tests__/worktree-guard.test.ts`: Write to uncovered configured repo triggers on-demand creation
```
dimakis, line 52: 🔵 **unsafe_assumptions**: Section 2a says `checkWorktreePolicy()` needs to trigger on-demand worktree creation and then 'return deny with redirect message (existing behavior — agent retries with the worktree path).' But the on-demand flow also needs to handle the case where `createWorktreeAsync` fails (disk full, git lock, branch conflict). The test plan doesn't cover this failure path — only the happy path is specified. A failed creation that still returns a redirect would send the agent into a retry loop. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+**Test plan:**
+
+- `server/__tests__/chat.test.ts`: `finally` block does NOT call `cleanupSessionWorktrees`
+
+### 2f. Async Worktree Creation
+
+**Problem:** `createWorktree` in `server/worktree.ts` (L46-102) uses `execFileSync`, blocking the Node event loop during git operations. With lazy on-demand creation (2a), this blocks ALL sessions.
+
```
dimakis, line 130: 🔵 **unsafe_assumptions**: Section 2e proposes moving `cleanupSessionWorktrees` out of `finally` but keeping it in `catch`. However, the `catch` block at L691 already calls it conditionally (`if (failedSession)`), and the `finally` block at L697 calls it unconditionally. If both catch and finally run (which they do when catch doesn't rethrow), secondary worktrees get cleaned twice. This double-call is harmless today (`removeWorktree` is idempotent) but the doc should mention it as context for the next agent removing the finally call. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+2. `server/chat.ts` — in the resume path of `startChat()` (the block starting at L292 `if (!isIsolationEnabled(options.isolation) || options.resume ...)`), when `options.resume` is set, call `discoverSessionWorktrees` to rebuild `session.worktreePaths`. Recreate primary if gone. Don't recreate secondaries (lazy creation handles it).
+
+**Note:** Phase 1 added `validateResumable()` (L116-157) which already handles the CWD validation and worktree recreation for the primary. This step extends that to populate `worktreePaths` for the system prompt and guard.
+
+**Test plan:**
+
+- `server/__tests__/worktree.test.ts`: `discoverSessionWorktrees` finds worktrees across repos
+- `server/__tests__/chat.test.ts`: Resume rebuilds `worktreePaths` from disk
+
+### 2e. Cleanup Only on Close
```
dimakis, line 95: 🔵 **missing_tests**: Section 2c (Runtime Symlinks) test plan only covers `symlinkRuntimeDirs` unit behavior. Missing: test for symlink-already-exists (re-run on resume), symlink pointing to a path that's been deleted, and the system prompt warning injection. Since this is an opt-in escape hatch with shared mutable state, the edge cases matter more than usual. `[fixable]`

docs/design/session-isolation-phase2-handoff.md
```diff
@@ -0,0 +1,220 @@
...
+- [ ] Agent runs `python` in worktree -> found via PATH (or symlink if opted in)
+- [ ] 3 concurrent session starts -> no wtId collision (already shipped in Phase 1)
+- [ ] Resume after stale GC -> worktrees rebuilt from disk
+- [ ] Work in 3 repos, disconnect, resume -> all 3 worktrees survive
+- [ ] Close session -> all worktrees cleaned up, branches preserved
+- [ ] 5 parallel sessions -> zero conflicts, zero bleed
+
+---
+
+## Warnings for Next Agent
```
dimakis, line 210: 🔵 **style**: Warning #4 includes a hardcoded ngrok URL (`beamily-operable-shenita.ngrok-free.dev`) and a specific webhook ID (`604995403`). These are ephemeral values that will be stale by the time Phase 2 starts. The ngrok tunnel is likely already dead. Consider replacing with generic instructions ('check the current ngrok tunnel URL' or 'find the webhook ID via `gh api repos/.../hooks`'). `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/session-isolation-phase2-handoff.md`
_Documentation-only PR with a well-structured handoff, but contains several inaccurate line references and a misleading characterization of the resume-breaking problem in 2e that could misdirect the next agent._

- 🟡 **bugs** (L12): Phase 1 claim that `generateWtId()` uses `randomUUID().slice(0, 12)` is inaccurate. The actual code does `randomUUID().replace(/-/g, '').slice(0, 12)` — it strips dashes first, then slices. The result is 12 hex chars, not a UUID substring with dashes. This distinction matters if the next agent tries to verify or extend the function. `[fixable]`
- 🟡 **bugs** (L119): Section 2e describes the problem as `cleanupSessionWorktrees(session)` "destroys worktrees after every query loop, breaking resume" — but the function explicitly skips the primary worktree (`if (repoName === 'primary') continue;` at chat.ts:828). Resume uses the primary worktree CWD, so it's not broken today. The real problem is that Phase 2 lazy-created secondaries would be lost, which is a valid concern but different from what's stated. An agent reading this literally might think resume is currently broken. `[fixable]`
- 🟡 **unsafe_assumptions** (L170): `buildWorktreeSystemPrompt()` is listed at L380-402, but the actual range is L377-399. `RepoConfig` interface is listed at L30-36 but actually starts at L23. `cleanupSessionWorktrees()` is listed at L803+ but actually starts at L823. While minor, the next agent will use these line references to navigate — stale references waste time or cause confusion, especially in an automated agent workflow. `[fixable]`
- 🔵 **unsafe_assumptions** (L52): Section 2a says `checkWorktreePolicy()` needs to trigger on-demand worktree creation and then 'return deny with redirect message (existing behavior — agent retries with the worktree path).' But the on-demand flow also needs to handle the case where `createWorktreeAsync` fails (disk full, git lock, branch conflict). The test plan doesn't cover this failure path — only the happy path is specified. A failed creation that still returns a redirect would send the agent into a retry loop. `[fixable]`
- 🔵 **unsafe_assumptions** (L130): Section 2e proposes moving `cleanupSessionWorktrees` out of `finally` but keeping it in `catch`. However, the `catch` block at L691 already calls it conditionally (`if (failedSession)`), and the `finally` block at L697 calls it unconditionally. If both catch and finally run (which they do when catch doesn't rethrow), secondary worktrees get cleaned twice. This double-call is harmless today (`removeWorktree` is idempotent) but the doc should mention it as context for the next agent removing the finally call. `[fixable]`
- 🔵 **missing_tests** (L95): Section 2c (Runtime Symlinks) test plan only covers `symlinkRuntimeDirs` unit behavior. Missing: test for symlink-already-exists (re-run on resume), symlink pointing to a path that's been deleted, and the system prompt warning injection. Since this is an opt-in escape hatch with shared mutable state, the edge cases matter more than usual. `[fixable]`
- 🔵 **style** (L210): Warning #4 includes a hardcoded ngrok URL (`beamily-operable-shenita.ngrok-free.dev`) and a specific webhook ID (`604995403`). These are ephemeral values that will be stale by the time Phase 2 starts. The ngrok tunnel is likely already dead. Consider replacing with generic instructions ('check the current ngrok tunnel URL' or 'find the webhook ID via `gh api repos/.../hooks`'). `[fixable]`

PR #280, opened 2026-04-23
feat: Phase 2 session isolation — bulletproof worktree isolation

dimakis:
## Summary

Implements Phase 2 of the session isolation overhaul (design doc: `docs/design/session-isolation-overhaul.md` §Phase 2, handoff: PR #279).

- **2f: Async worktree creation** — `createWorktreeAsync` using promisified `execFile` so on-demand secondary creation doesn't block the event loop
- **2a: Lazy secondary worktrees** — remove eager creation of all 11 repos at session start; secondary worktrees created on-demand when the worktree guard detects a write to a configured repo. `checkWorktreePolicy` is now async with an `onDemandCreate` callback. On creation failure, hard deny (no redirect) prevents retry loops
- **2e: Cleanup only on close** — remove `cleanupSessionWorktrees` from the `finally` block so worktrees survive across turns for reliable resume. Stale threshold reduced from 96h to 24h
- **2c: Runtime symlinks** — opt-in `runtimeSymlinks` field in `.mitzo.json` symlinks `.venv`/`node_modules` from source repo into worktrees. Idempotent on resume, handles dangling targets. System prompt warns about shared mutable state
- **2d: Resume worktree rebuild** — `discoverSessionWorktrees` scans disk for existing worktrees on resume after server restart, rebuilding `session.worktreePaths` for the guard and system prompt

Addresses review findings from PR #279 (line reference corrections, lockfile extension to secondaries, creation failure hard deny, symlink edge cases).

## Test plan

- [x] `createWorktreeAsync` — creates, reuses, handles branch-exists fallback (3 tests)
- [x] `checkWorktreePolicy` — on-demand creation triggers on write, hard deny on failure, all existing tests updated to async (24 tests)
- [x] `createSessionWorktrees` — creates only primary worktree (1 test)
- [x] `finally` block — does NOT call `cleanupSessionWorktrees` (1 structural test)
- [x] `symlinkRuntimeDirs` — creates/skips/idempotent/dangling/empty (5 tests)
- [x] `discoverSessionWorktrees` — finds across repos, .cursor prefix, empty, partial (4 tests + 1 integration)

### DIFF
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -160,6 +160,24 @@ export async function checkWorktreePolicy(
     for (const p of paths) {
       if (findAllowedWorktree(p, session.worktreePaths)) continue;
 
+      if (opts?.onDemandCreate) {
+        const created = await opts.onDemandCreate(p);
+        if (created) {
+          session.worktreePaths.set(created.repoName, {
+            path: created.worktreePath,
+            wtId: session.wtId ?? '',
+          });
+          const suggestion = suggestWorktreePath(p, session.worktreePaths);
+          stats.denied++;
+          log.info('on-demand worktree created, redirecting', {
+            toolName,
+            repo: created.repoName,
+            sessionId: session.sessionId,
+          });
+          return denyMessage('Shell command references', p, suggestion);
+        }
+      }
+
       const suggestion = suggestWorktreePath(p, session.worktreePaths);
       stats.denied++;
       log.warn('worktree policy denied', {
--- a/packages/harness/src/worktree-guard.ts
+++ b/packages/harness/src/worktree-guard.ts
@@ -19,18 +19,30 @@ export function resetWorktreeGuardStats() {
 }
 
 /**
+ * Callback for on-demand worktree creation. Given the absolute path the agent
+ * tried to write to, creates a worktree for the matching repo and returns its
+ * name + path. Returns null if creation failed or the path doesn't match any
+ * configured repo.
+ */
+export type OnDemandCreateFn = (
+  absolutePath: string,
+) => Promise<{ repoName: string; worktreePath: string } | null>;
+
+export interface CheckWorktreePolicyOptions {
+  onDemandCreate?: OnDemandCreateFn;
+}
+
+/**
  * Extract absolute paths from a shell command string. Heuristic — catches
  * explicit absolute paths but can't catch every indirect construction.
  */
 function extractAbsolutePaths(command: string): string[] {
   const results: string[] = [];
-  // Match quoted absolute paths (handles spaces)
   const quotedRe = /["'](\/[^"']+)["']/g;
   let m: RegExpExecArray | null;
   while ((m = quotedRe.exec(command)) !== null) {
     results.push(m[1]);
   }
-  // Match unquoted absolute paths (supports tildes and broader char set)
   const unquotedRe = /(?:^|\s|=)(\/[\w/.@~-]+)/g;
   while ((m = unquotedRe.exec(command)) !== null) {
     results.push(m[1]);
@@ -59,8 +71,6 @@ function suggestWorktreePath(
   worktreePaths: Map<string, { path: string; wtId: string }>,
 ): string | null {
   for (const [, { path: wtPath }] of worktreePaths) {
-    // worktree path format: /repo-root/.claude/worktrees/<id>
-    // extract repo root: everything before /.claude/worktrees/ or /.cursor/worktrees/
     const worktreeMarker = wtPath.match(/^(.+?)\/(\.claude|\.cursor)\/worktrees\/.+$/);
     if (!worktreeMarker) continue;
 
@@ -73,15 +83,27 @@ function suggestWorktreePath(
   return null;
 }
 
+function denyMessage(prefix: string, attemptedPath: string, suggestion: string | null): string {
+  return suggestion
+    ? `${prefix} ${attemptedPath} is outside session worktrees. Use ${suggestion} instead.`
+    : `${prefix} ${attemptedPath} is outside session worktrees. Check $MITZO_REPO_* env vars for correct paths.`;
+}
+
 /**
  * Check if a tool invocation violates worktree isolation.
  * Returns null if allowed, or a deny message with the correct path.
+ *
+ * When `onDemandCreate` is provided and the path maps to a configured repo
+ * without a worktree, creation is attempted. On success the worktree is added
+ * to the session and a redirect is returned. On failure a hard deny is returned
+ * (no redirect) to prevent retry loops.
  */
-export function checkWorktreePolicy(
+export async function checkWorktreePolicy(
   session: ManagedSession,
   toolName: string,
   toolInput: Record<string, unknown>,
-): string | null {
+  opts?: CheckWorktreePolicyOptions,
+): Promise<string | null> {
   if (session.worktreePaths.size === 0) return null;
 
   if (WRITE_TOOLS.has(toolName)) {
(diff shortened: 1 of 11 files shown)


packages/harness/src/worktree-guard.ts
```diff
@@ -94,18 +116,34 @@ export function checkWorktreePolicy(
       const allowed = findAllowedWorktree(filePath, session.worktreePaths);
       if (allowed) continue;
 
+      // Try on-demand creation if a callback is provided
+      if (opts?.onDemandCreate) {
+        const created = await opts.onDemandCreate(filePath);
```
dimakis, line 121: 🟡 **bugs**: The new on-demand creation path mutates `session.worktreePaths`, but `buildPermissionHandler()` never passes an `onDemandCreate` callback, so writes to configured secondary repos still hard-deny exactly as before. The PR adds the async hook and tests it in isolation, but the production permission flow does not wire it up, so the advertised lazy secondary worktree creation behavior is not actually reachable. `[fixable]`

server/chat.ts
```diff
@@ -546,6 +551,25 @@ export async function startChat(
     options,
   );
 
+  // On resume, rebuild worktreePaths from disk so the system prompt and guard
+  // have the full map even after server restart (Phase 2d).
```
dimakis, line 587: 🟡 **regressions**: Resume-time worktree discovery only runs when `repoWorktrees.size === 0`, but resumed sessions already seed `repoWorktrees` with the current `cwd` as `primary`. That makes this block effectively dead for normal resumed worktree sessions, so secondary worktrees created earlier are never rediscovered after a server restart and the rebuilt session loses its full worktree map. `[fixable]`

server/chat.ts
```diff
@@ -546,6 +551,25 @@ export async function startChat(
...
+  if (options.resume && repoWorktrees.size === 0 && BASE_REPO) {
+    const config = getRepoConfig();
```
dimakis, line 591: 🟡 **bugs**: `wtIdFromCwd` is extracted with `/\/(\.claude|\.cursor)\/worktrees\/([^/]+)$/`, which only matches when `baseCwd` is exactly the worktree root. Resumed sessions commonly use a subdirectory inside the worktree, so discovery will silently fail and no worktrees are rebuilt from disk in those cases. `[fixable]`

server/worktree.ts
```diff
@@ -101,6 +114,123 @@ export function createWorktree(
...
+  }
+
+  return result;
+}
+
+/**
+ * Create symlinks for runtime directories (e.g. .venv, node_modules) from the
+ * source repo into a worktree. Opt-in escape hatch for CWD-relative tool
+ * resolution. Symlinked dirs are shared mutable state across sessions.
+ *
```
dimakis, line 207: 🟡 **unsafe_assumptions**: `symlinkRuntimeDirs()` skips any existing symlink without verifying that it points at the current repo path. If a stale or incorrect symlink is already present in the worktree, resume/idempotent runs will preserve the wrong target and silently keep sharing the wrong runtime directory. `[fixable]`

dimakis (commented):
## Codex Review

Found **4** issue(s) (4 warning).


_PR adds useful worktree-isolation pieces, but lazy secondary creation and resume-time worktree reconstruction are not fully wired and contain logic gaps._

- 🟡 **bugs** (L121): The new on-demand creation path mutates `session.worktreePaths`, but `buildPermissionHandler()` never passes an `onDemandCreate` callback, so writes to configured secondary repos still hard-deny exactly as before. The PR adds the async hook and tests it in isolation, but the production permission flow does not wire it up, so the advertised lazy secondary worktree creation behavior is not actually reachable. `[fixable]`


### `server/chat.ts`
_PR adds useful worktree-isolation pieces, but lazy secondary creation and resume-time worktree reconstruction are not fully wired and contain logic gaps._

- 🟡 **regressions** (L555): Resume-time worktree discovery only runs when `repoWorktrees.size === 0`, but resumed sessions already seed `repoWorktrees` with the current `cwd` as `primary`. That makes this block effectively dead for normal resumed worktree sessions, so secondary worktrees created earlier are never rediscovered after a server restart and the rebuilt session loses its full worktree map. `[fixable]`
- 🟡 **bugs** (L557): `wtIdFromCwd` is extracted with `/\/(\.claude|\.cursor)\/worktrees\/([^/]+)$/`, which only matches when `baseCwd` is exactly the worktree root. Resumed sessions commonly use a subdirectory inside the worktree, so discovery will silently fail and no worktrees are rebuilt from disk in those cases. `[fixable]`


### `server/worktree.ts`
_PR adds useful worktree-isolation pieces, but lazy secondary creation and resume-time worktree reconstruction are not fully wired and contain logic gaps._

- 🟡 **unsafe_assumptions** (L205): `symlinkRuntimeDirs()` skips any existing symlink without verifying that it points at the current repo path. If a stale or incorrect symlink is already present in the worktree, resume/idempotent runs will preserve the wrong target and silently keep sharing the wrong runtime directory. `[fixable]`

packages/harness/src/worktree-guard.ts
```diff
@@ -94,18 +116,34 @@ export function checkWorktreePolicy(
       const allowed = findAllowedWorktree(filePath, session.worktreePaths);
       if (allowed) continue;
 
+      // Try on-demand creation if a callback is provided
```
dimakis, line 119: 🟡 **bugs**: On-demand creation is only attempted for WRITE_TOOLS. Bash commands that reference a configured secondary repo before its worktree exists are still hard-denied, even though Phase 2 moves secondary worktree creation to first write and Bash is one of the enforced write-capable tools. This breaks workflows like `git -C /secondary-repo ...` or `cd /secondary-repo && ...` as the guard never provisions the missing worktree for shell-based writes. `[fixable]`

server/chat.ts
dimakis, line 587: 🟡 **regressions**: Resume-time worktree discovery only runs when `repoWorktrees.size === 0`. If a resumed session already has its primary worktree in memory but is missing lazily-created secondary entries after a restart/reattach, discovery is skipped and the session loses access to those secondary worktrees. The rebuild should merge discovered entries whenever resuming from a worktree cwd, not only when the map is completely empty. `[fixable]`

server/worktree.ts
```diff
@@ -101,6 +114,123 @@ export function createWorktree(
...
+    allRepos.push([name, repoPath]);
+  }
+
+  for (const [name, repoPath] of allRepos) {
+    for (const prefix of prefixes) {
+      const candidate = join(repoPath, prefix, 'worktrees', wtId);
+      if (existsSync(candidate)) {
+        result.set(name, { path: candidate, wtId });
+        break;
+      }
```
dimakis, line 196: 🟡 **bugs**: `symlinkRuntimeDirs()` treats any existing symlink at the target as valid and skips it without checking where it points. If the symlink points to the wrong location (stale repo path, previous worktree, or malicious/manual replacement), sessions will silently use the wrong shared runtime directory. `[fixable]`

dimakis (commented):
## Codex Review

_Solid test coverage overall, but the lazy secondary-worktree implementation has a couple of behavioral gaps around Bash/on-resume handling and symlink validation._

- 🟡 **bugs** (L119): On-demand creation is only attempted for WRITE_TOOLS. Bash commands that reference a configured secondary repo before its worktree exists are still hard-denied, even though Phase 2 moves secondary worktree creation to first write and Bash is one of the enforced write-capable tools. This breaks workflows like `git -C /secondary-repo ...` or `cd /secondary-repo && ...` as the guard never provisions the missing worktree for shell-based writes. `[fixable]`


### `server/chat.ts`
_Solid test coverage overall, but the lazy secondary-worktree implementation has a couple of behavioral gaps around Bash/on-resume handling and symlink validation._

- 🟡 **regressions** (L555): Resume-time worktree discovery only runs when `repoWorktrees.size === 0`. If a resumed session already has its primary worktree in memory but is missing lazily-created secondary entries after a restart/reattach, discovery is skipped and the session loses access to those secondary worktrees. The rebuild should merge discovered entries whenever resuming from a worktree cwd, not only when the map is completely empty. `[fixable]`


### `server/worktree.ts`
_Solid test coverage overall, but the lazy secondary-worktree implementation has a couple of behavioral gaps around Bash/on-resume handling and symlink validation._

- 🟡 **bugs** (L194): `symlinkRuntimeDirs()` treats any existing symlink at the target as valid and skips it without checking where it points. If the symlink points to the wrong location (stale repo path, previous worktree, or malicious/manual replacement), sessions will silently use the wrong shared runtime directory. `[fixable]`

PR #284, opened 2026-04-23
chore(infra): always-on observability stack

dimakis:
## Summary

- Add `restart: always` to Jaeger, Loki, and Grafana containers so they survive podman restarts
- Add a launchd agent (`com.mitzo.podman-machine`) to auto-start the podman machine on login
- `deploy.sh` now ensures podman machine + containers are up before restarting the Mitzo server
- Set Loki `log_level: warn` to suppress noisy query metric and gRPC scheduler logs

## Test plan

- [ ] Reboot Mac, verify podman machine starts automatically
- [ ] Verify Jaeger, Loki, Grafana containers come up with the machine
- [ ] Run `npm run deploy`, confirm observability stack is healthy before server starts
- [ ] Check `podman logs mitzo_loki_1` is quiet (no info-level query spam)

### DIFF
--- a/infra/com.mitzo.podman-machine.plist
+++ b/infra/com.mitzo.podman-machine.plist
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>com.mitzo.podman-machine</string>
+    <key>ProgramArguments</key>
+    <array>
+        <string>/opt/homebrew/bin/podman</string>
+        <string>machine</string>
+        <string>start</string>
+    </array>
+    <key>RunAtLoad</key>
+    <true/>
+    <key>KeepAlive</key>
+    <false/>
+    <key>LaunchOnlyOnce</key>
+    <true/>
+    <key>StandardOutPath</key>
+    <string>__MITZO_HOME__/logs/podman-machine-stdout.log</string>
+    <key>StandardErrorPath</key>
+    <string>__MITZO_HOME__/logs/podman-machine-stderr.log</string>
+    <key>EnvironmentVariables</key>
+    <dict>
+        <key>PATH</key>
+        <string>/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin</string>
+    </dict>
+</dict>
+</plist>
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -15,6 +15,21 @@ npm run build
 echo "Installing launchd plist..."
 sed "s|__MITZO_HOME__|${MITZO_HOME}|g" com.mitzo.server.plist > "$PLIST_DEST"
 
+# Install podman machine launchd agent
+PODMAN_PLIST_DEST="$HOME/Library/LaunchAgents/com.mitzo.podman-machine.plist"
+sed "s|__MITZO_HOME__|${MITZO_HOME}|g" infra/com.mitzo.podman-machine.plist > "$PODMAN_PLIST_DEST"
+launchctl bootout "gui/$(id -u)/com.mitzo.podman-machine" 2>/dev/null || true
+launchctl bootstrap "gui/$(id -u)" "$PODMAN_PLIST_DEST"
+
+# Ensure podman machine is running before bringing up containers
+if ! podman machine inspect --format '{{.State}}' 2>/dev/null | grep -q Running; then
+  echo "Starting podman machine..."
+  podman machine start
+fi
+
+echo "Ensuring observability stack is running..."
+docker compose up -d
+
 echo "Restarting service..."
 # kickstart -k sends SIGTERM and waits for termination before restarting.
 launchctl kickstart -k "gui/$(id -u)/com.mitzo.server"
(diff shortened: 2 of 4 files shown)


infra/com.mitzo.podman-machine.plist
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>Label</key>
+    <string>com.mitzo.podman-machine</string>
+    <key>ProgramArguments</key>
+    <array>
+        <string>/opt/homebrew/bin/podman</string>
```
dimakis, line 9: 🟡 **unsafe_assumptions**: Hardcodes `/opt/homebrew/bin/podman` — only works on Apple Silicon Macs with Homebrew. Intel Macs use `/usr/local/bin`. The existing `com.mitzo.server.plist` has the same assumption in its PATH, so this is consistent with the project, but worth noting if portability is a goal. `[fixable]`

infra/com.mitzo.podman-machine.plist
```diff
@@ -0,0 +1,29 @@
...
+        <string>machine</string>
+        <string>start</string>
+    </array>
+    <key>RunAtLoad</key>
+    <true/>
+    <key>KeepAlive</key>
+    <false/>
+    <key>LaunchOnlyOnce</key>
```
dimakis, line 17: 🔵 **style**: `LaunchOnlyOnce` is not a standard launchd key. The documented key is `LaunchOnlyOnce` under `com.apple.launchd` but its behavior is undocumented in many macOS versions. Since `KeepAlive` is already `false` and `RunAtLoad` is `true`, the agent will run once at login regardless — `LaunchOnlyOnce` may be redundant. `[fixable]`

scripts/deploy.sh
```diff
@@ -15,6 +15,21 @@ npm run build
...
 sed "s|__MITZO_HOME__|${MITZO_HOME}|g" com.mitzo.server.plist > "$PLIST_DEST"
 
+# Install podman machine launchd agent
+PODMAN_PLIST_DEST="$HOME/Library/LaunchAgents/com.mitzo.podman-machine.plist"
+sed "s|__MITZO_HOME__|${MITZO_HOME}|g" infra/com.mitzo.podman-machine.plist > "$PODMAN_PLIST_DEST"
+launchctl bootout "gui/$(id -u)/com.mitzo.podman-machine" 2>/dev/null || true
+launchctl bootstrap "gui/$(id -u)" "$PODMAN_PLIST_DEST"
+
+# Ensure podman machine is running before bringing up containers
+if ! podman machine inspect --format '{{.State}}' 2>/dev/null | grep -q Running; then
```
dimakis, line 25: 🟡 **bugs**: `podman machine inspect --format '{{.State}}'` may return output for the default machine or fail with a non-zero exit code if no machine exists at all. The `2>/dev/null` hides the error, but `podman machine start` (line 27) will also fail if no machine has been initialized (`podman machine init`). Consider adding an `init` fallback or at least a clear error message. `[fixable]`

scripts/deploy.sh
```diff
@@ -15,6 +15,21 @@ npm run build
...
+  echo "Starting podman machine..."
+  podman machine start
+fi
+
+echo "Ensuring observability stack is running..."
+docker compose up -d
```
dimakis, line 31: 🔵 **regressions**: `docker compose up -d` is now run on every deploy. Previously, the observability stack was opt-in (`npm run observability:up`). This is intentional per the PR title ('always-on'), but it means every deploy now requires a running podman machine and pulls container images if not cached — adding latency and a new failure mode to deploys.

dimakis (commented):
## Centaur Review

### `infra/com.mitzo.podman-machine.plist`
_Straightforward infra change making the observability stack always-on. Main concern is the `podman machine inspect` check not handling the 'no machine exists' case, which would cause deploy failures on a fresh setup._

- 🟡 **unsafe_assumptions** (L9): Hardcodes `/opt/homebrew/bin/podman` — only works on Apple Silicon Macs with Homebrew. Intel Macs use `/usr/local/bin`. The existing `com.mitzo.server.plist` has the same assumption in its PATH, so this is consistent with the project, but worth noting if portability is a goal. `[fixable]`
- 🔵 **style** (L17): `LaunchOnlyOnce` is not a standard launchd key. The documented key is `LaunchOnlyOnce` under `com.apple.launchd` but its behavior is undocumented in many macOS versions. Since `KeepAlive` is already `false` and `RunAtLoad` is `true`, the agent will run once at login regardless — `LaunchOnlyOnce` may be redundant. `[fixable]`


### `scripts/deploy.sh`
_Straightforward infra change making the observability stack always-on. Main concern is the `podman machine inspect` check not handling the 'no machine exists' case, which would cause deploy failures on a fresh setup._

- 🟡 **bugs** (L25): `podman machine inspect --format '{{.State}}'` may return output for the default machine or fail with a non-zero exit code if no machine exists at all. The `2>/dev/null` hides the error, but `podman machine start` (line 27) will also fail if no machine has been initialized (`podman machine init`). Consider adding an `init` fallback or at least a clear error message. `[fixable]`
- 🔵 **regressions** (L31): `docker compose up -d` is now run on every deploy. Previously, the observability stack was opt-in (`npm run observability:up`). This is intentional per the PR title ('always-on'), but it means every deploy now requires a running podman machine and pulls container images if not cached — adding latency and a new failure mode to deploys.

PR #288, opened 2026-04-24
feat: proactive session suspend — signal before iOS kills WS

dimakis:
## Summary

- Signals the server BEFORE iOS kills the WebSocket (via `visibilitychange:hidden` + `pagehide`), enabling event buffering and instant resume
- Dual delivery: WS message + `sendBeacon` fallback (belt-and-suspenders for the iOS kill race)
- Server buffers agent responses for suspended sessions; replays them on reconnect
- Grace timer (2 min) transitions to normal detach if client doesn't resume
- Design doc: `docs/design/proactive-session-suspend.md`

### Architecture (4 commits)

1. **Protocol + Registry** — `SessionSuspendMessage` schema, `suspend()`/`isSuspended()`/`bufferEvent()`/`resume()` on SessionRegistry, constants
2. **Server handlers** — `handleSessionSuspend` in ws-handler-v2, query-loop buffer intercept, WS close guard, `POST /api/sessions/suspend` REST endpoint for sendBeacon
3. **Client** — `sendSuspend()` on MitzoConnection (WS + sendBeacon), `visibilitychange`/`pagehide`/Capacitor `onPause` hooks, store action + `switchSession` suspend
4. **Review fixes** — skip idle session suspend, buffer overflow warning, `session_resumed` client handler, withSpan consistency, test mock updates

### Files changed (15 files, +801/-57)

| Layer | Files |
|-------|-------|
| Protocol | `ws-schemas-v2.ts`, `index.ts` |
| Harness | `session-registry.ts`, `constants.ts`, `session-registry.test.ts` |
| Server | `ws-handler-v2.ts`, `query-loop.ts`, `index.ts`, `app.ts`, `ws-handler-v2.test.ts` |
| Client | `connection.ts`, `store.ts` |
| Frontend | `client-store.ts`, `capacitor.ts` |
| Config | `vitest.config.ts` |

## Test plan

- [x] 259 lines of new session-registry tests (suspend/resume/grace/buffer/cleanup)
- [x] 162 lines of new ws-handler-v2 tests (handleSessionSuspend, reconnect resume, dispatch)
- [x] All 2184 passing tests still pass (38 failures are pre-existing)
- [ ] Manual: background Mitzo on iPhone → check server logs for `session suspended`
- [ ] Manual: return to app → check logs for `resumed suspended session` with replayed count
- [ ] Manual: switch sessions while agent running → switch back → response visible

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -13,6 +13,8 @@ export interface MitzoConnectionConfig {
   buildUrl(): string;
   createWebSocket(url: string): WebSocketLike;
   reconnectDelayMs?: number;
+  /** URL for the sendBeacon suspend fallback (POST /api/sessions/suspend). */
+  suspendUrl?: string;
 }
 
 export type ConnectionListener = (msg: Record<string, unknown>) => void;
@@ -32,11 +34,13 @@ export class MitzoConnection {
   private heartbeatTimer: ReturnType<typeof setInterval> | null = null;
   private boundOnVisibility: (() => void) | null = null;
   private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
+  private boundOnPageHide: (() => void) | null = null;
   private config: Required<MitzoConnectionConfig>;
 
   constructor(config: MitzoConnectionConfig) {
     this.config = {
       reconnectDelayMs: 500,
+      suspendUrl: '',
       ...config,
     };
   }
@@ -108,6 +112,47 @@ export class MitzoConnection {
 
   getTrackedSessions(): string[] {
     return Array.from(this.seqBySession.keys());
+  }
+
+  /**
+   * Signal the server that this client is about to be backgrounded (iOS).
+   * Sends via WS if open, and also fires sendBeacon as a belt-and-suspenders
+   * fallback since iOS may kill the socket before the WS send completes.
+   */
+  sendSuspend(): void {
+    if (this.seqBySession.size === 0) return;
+
+    const sessions = Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+      sessionId,
+      lastSeq,
+    }));
+
+    const wsPayload = { type: 'session_suspend', sessions };
+
+    // Try WS first (may already be dying)
+    if (this.ws?.readyState === WS_READY_STATE.OPEN) {
+      try {
+        this.ws.send(JSON.stringify(wsPayload));
+      } catch {
+        // Socket may be transitioning — sendBeacon fallback below
+      }
+    }
+
+    // sendBeacon fallback — works even after visibilitychange:hidden
+    if (
+      this.config.suspendUrl &&
+      this._connectionId &&
+      typeof globalThis.navigator?.sendBeacon === 'function'
+    ) {
+      const beaconPayload = JSON.stringify({
+        connectionId: this._connectionId,
+        sessions,
+      });
+      globalThis.navigator.sendBeacon(
+        this.config.suspendUrl,
+        new Blob([beaconPayload], { type: 'application/json' }),
+      );
+    }
   }
 
   // ─── Internal ──────────────────────────────────────────────────────────────
@@ -241,6 +286,9 @@ export class MitzoConnection {
         // Notify the store so it can re-hydrate messages if the page was
         // evicted from memory (iOS bfcache discard) and state was lost.
         this.listener?.({ type: '_foreground' });
+      } else if (document.visibilityState === 'hidden') {
+        // Proactive suspend: signal the server BEFORE iOS kills the WS
+        this.sendSuspend();
       }
     };
 
@@ -248,8 +296,13 @@ export class MitzoConnection {
       if (e.persisted) this.checkAndReconnect();
     };
 
+    this.boundOnPageHide = () => {
+      this.sendSuspend();
+    };
+
     document.addEventListener('visibilitychange', this.boundOnVisibility);
     globalThis.addEventListener('pageshow', this.boundOnPageShow);
+    globalThis.addEventListener('pagehide', this.boundOnPageHide);
   }
 
   private removeBrowserListeners(): void {
@@ -260,6 +313,10 @@ export class MitzoConnection {
     if (this.boundOnPageShow) {
       globalThis.removeEventListener('pageshow', this.boundOnPageShow);
       this.boundOnPageShow = null;
+    }
+    if (this.boundOnPageHide) {
+      globalThis.removeEventListener('pagehide', this.boundOnPageHide);
+      this.boundOnPageHide = null;
     }
   }
 
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -116,8 +116,10 @@ export class MitzoConnection {
 
   /**
    * Signal the server that this client is about to be backgrounded (iOS).
-   * Sends via WS if open, and also fires sendBeacon as a belt-and-suspenders
-   * fallback since iOS may kill the socket before the WS send completes.
+   * Tries WS first; falls back to sendBeacon only if the WS send fails or
+   * the socket is already closed. This avoids double-sending (which would
+   * refresh the grace timer twice and keep sessions suspended longer than
+   * intended).
    */
   sendSuspend(): void {
     if (this.seqBySession.size === 0) return;
@@ -127,19 +129,21 @@ export class MitzoConnection {
       lastSeq,
     }));
 
-    const wsPayload = { type: 'session_suspend', sessions };
-
     // Try WS first (may already be dying)
+    let wsSent = false;
     if (this.ws?.readyState === WS_READY_STATE.OPEN) {
       try {
-        this.ws.send(JSON.stringify(wsPayload));
+        this.ws.send(JSON.stringify({ type: 'session_suspend', sessions }));
+        wsSent = true;
       } catch {
-        // Socket may be transitioning — sendBeacon fallback below
+        // Socket may be transitioning — fall through to sendBeacon
       }
     }
 
-    // sendBeacon fallback — works even after visibilitychange:hidden
+    // sendBeacon fallback — only when WS send was unavailable or failed.
+    // Works even after visibilitychange:hidden when the socket is already gone.
     if (
+      !wsSent &&
       this.config.suspendUrl &&
       this._connectionId &&
       typeof globalThis.navigator?.sendBeacon === 'function'
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -118,6 +118,7 @@ export interface MitzoStoreState {
 
   // Actions — lifecycle
   forceReconnect(): void;
+  sendSuspend(): void;
 }
 
 // ─── Store options ───────────────────────────────────────────────────────────
@@ -228,7 +229,14 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
     async switchSession(id: string) {
       const oldId = parserState.currentSessionId;
-      if (oldId) connection.clearSession(oldId);
+      if (oldId) {
+        // Suspend the old session so agent responses buffer while we're away
+        connection.send({
+          type: 'session_suspend',
+          sessions: [{ sessionId: oldId, lastSeq: connection.getLastSeq(oldId) }],
+        });
+        connection.clearSession(oldId);
+      }
       parserState.currentSessionId = id;
 
       set((s) => ({
@@ -578,6 +586,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
     forceReconnect() {
       connection.checkAndReconnect();
+    },
+
+    sendSuspend() {
+      connection.sendSuspend();
     },
   }));
 
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -230,13 +230,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     async switchSession(id: string) {
       const oldId = parserState.currentSessionId;
       if (oldId) {
-        const lastSeq = connection.getLastSeq(oldId);
-        if (lastSeq > 0) {
-          connection.send({
-            type: 'session_suspend',
-            sessions: [{ sessionId: oldId, lastSeq }],
-          });
-        }
+        // clearSession stops seq tracking. No suspend needed — session_suspend
+        // is for iOS backgrounding (imminent WS death), not session switching.
+        // Sending suspend here would leave the old session in suspended state
+        // with no resume path, causing it to buffer events until grace expiry.
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -230,11 +230,13 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     async switchSession(id: string) {
       const oldId = parserState.currentSessionId;
       if (oldId) {
-        // Suspend the old session so agent responses buffer while we're away
-        connection.send({
-          type: 'session_suspend',
-          sessions: [{ sessionId: oldId, lastSeq: connection.getLastSeq(oldId) }],
-        });
+        const lastSeq = connection.getLastSeq(oldId);
+        if (lastSeq > 0) {
+          connection.send({
+            type: 'session_suspend',
+            sessions: [{ sessionId: oldId, lastSeq }],
+          });
+        }
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
@@ -662,6 +664,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     if (msg.type === '_foreground') {
       const { sessions } = store.getState();
       if (sessions.active) fetchAndRestoreMessages(sessions.active);
+      return;
+    }
+
+    if (msg.type === 'session_resumed') {
+      if (typeof console !== 'undefined') {
+        console.debug('[mitzo] session resumed', {
+          sessionId: msg.sessionId,
+          replayed: msg.replayed,
+        });
+      }
       return;
     }
 
--- a/packages/harness/src/session-registry.ts
+++ b/packages/harness/src/session-registry.ts
@@ -4,6 +4,8 @@ import {
   CLOSEOUT_LEAD_MS,
   CLOSEOUT_TIMEOUT_MS,
   MAX_OBSERVERS_PER_SESSION,
+  SUSPEND_GRACE_MS,
+  SUSPEND_BUFFER_MAX,
 } from './constants.js';
 import { createLogger } from './logger.js';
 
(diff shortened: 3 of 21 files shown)


dimakis:
## Centaur Review

### `server/ws-handler-v2.ts`
_Well-structured iOS suspend feature with solid server-side test coverage, but has a post-resume zombie session bug (transport/clientId not updated after suspend→resume) and a session-switch lifecycle regression (old sessions now enter closeout after 2 min)._

- 🟡 **bugs** (L185): After suspend/resume, the session remains 'attached' with the OLD dead transport and OLD clientId (no reattach/rekey occurs because `isAttached()` returns true at line 155, skipping the reattach block). Subsequent events flow via connRegistry broadcast (which works), but if the new WS dies unexpectedly (without a suspend signal), the close handler's `session.transport === transport` check fails (old transport ≠ new transport), so no detach is triggered. The session becomes a zombie — permanently 'attached' with a dead transport. The resume block should also call `reattachChat`/`rekeyChat` to update the session's transport and clientId. `[fixable]`


_Well-structured iOS suspend feature with solid server-side test coverage, but has a post-resume zombie session bug (transport/clientId not updated after suspend→resume) and a session-switch lifecycle regression (old sessions now enter closeout after 2 min)._

- 🟡 **bugs** (L199): `rekey()` does not transfer suspend state (`suspended`, `suspendBuffers`, `suspendTimers`). If a session is rekeyed while suspended, the grace timer and buffer remain under the old key, and the session under the new key loses its suspend state. Currently not triggered in the suspend flow (reattach is skipped so rekey doesn't fire), but this is a latent bug if the code paths change. `[fixable]`
- 🔵 **style** (L348): `_lastClientSeq` parameter is accepted but unused. If it's reserved for future seq-based replay, consider removing it now and adding it when needed — YAGNI. It's passed through from the client and WS handler, adding surface area with no current functionality. `[fixable]`


_Well-structured iOS suspend feature with solid server-side test coverage, but has a post-resume zombie session bug (transport/clientId not updated after suspend→resume) and a session-switch lifecycle regression (old sessions now enter closeout after 2 min)._

- 🟡 **regressions** (L232): switchSession now sends `session_suspend` for the old session. This changes session lifecycle: before, switching away left the old session running indefinitely; now, the suspend grace timer (2 min) will expire, triggering detach → closeout. If a user switches away from a long-running session and doesn't return within 2 minutes, the agent will be asked to commit and summarize, then aborted. There is no 'un-suspend' path when switching back — the client never calls resume, so the session will always hit the grace timeout. This may be intentional, but it's a behavior change that should be documented. `[fixable]`
- 🔵 **style** (L670): `if (typeof console !== 'undefined')` guard is unnecessary — `console` is always defined in browser and Node environments. All other `console.debug` calls in this file don't use this guard. `[fixable]`


### `server/app.ts`
_Well-structured iOS suspend feature with solid server-side test coverage, but has a post-resume zombie session bug (transport/clientId not updated after suspend→resume) and a session-switch lifecycle regression (old sessions now enter closeout after 2 min)._

- 🟡 **unsafe_assumptions** (L429): The `/api/sessions/suspend` endpoint has no authentication — it relies on `connectionId` as a shared secret. However, `connectionId` is generated from `Math.random()` (server/index.ts:221) with ~31 bits of entropy (6 base-36 chars), which is weak for a security-bearing token. An attacker who can reach the server and guess/observe a connectionId + sessionId can suspend arbitrary sessions. Consider using `crypto.randomUUID()` for connectionId generation, or adding CORS origin checks on this endpoint. `[fixable]`
- 🟡 **missing_tests** (L429): The REST endpoint `POST /api/sessions/suspend` has no integration tests. Should test: valid suspend, missing connectionId (400), missing sessions (400), non-owner connectionId rejection, and unknown session handling. The WS handler and registry have good test coverage, but the HTTP fallback path is untested. `[fixable]`


_Well-structured iOS suspend feature with solid server-side test coverage, but has a post-resume zombie session bug (transport/clientId not updated after suspend→resume) and a session-switch lifecycle regression (old sessions now enter closeout after 2 min)._

- 🔵 **missing_tests** (L117): No client-side tests for `sendSuspend()`, the `visibilitychange`/`pagehide` listeners, or the `sendBeacon` fallback. The existing `connection.test.ts` and `store.test.ts` files don't cover any suspend functionality. The `switchSession` tests don't verify the new `session_suspend` message is sent for the old session. `[fixable]`

server/query-loop.ts
```diff
@@ -85,6 +85,13 @@ function sendOrBuffer(
     }
   }
 
+  // Suspend buffer: if the session is proactively suspended (iOS backgrounding),
+  // buffer events instead of delivering them. They'll be replayed on resume.
+  if (registry.isSuspended(clientId)) {
```
dimakis, line 93: 🔴 **bugs**: Duplicate event replay on reconnect. Events are appended to the durable EventStore (line 83) BEFORE the suspend buffer check (line 90). The same events end up in both the EventStore and the suspend buffer. In `handleReconnect` (ws-handler-v2.ts:130+186), events are replayed from BOTH sources — first `getEventsAfter()` from EventStore, then `resume()` from the suspend buffer. Every event generated during suspension is sent to the client twice. Fix: either skip EventStore replay when resuming from suspend, or skip the `store.append()` when the session is suspended. `[fixable]`

dimakis, line 93: 🟡 **bugs**: Suspend blocks event broadcast to ALL watchers. The suspend check returns early before `connRegistry.broadcast(sessionId, enriched)` (line 99). Other connections observing the same session via `connRegistry.watch()` will stop receiving events while the session is suspended. The suspend is per-clientId but the broadcast skip affects the entire session's fan-out path. `[fixable]`

packages/client/src/store.ts
```diff
@@ -228,7 +229,16 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
 
     async switchSession(id: string) {
       const oldId = parserState.currentSessionId;
-      if (oldId) connection.clearSession(oldId);
+      if (oldId) {
+        const lastSeq = connection.getLastSeq(oldId);
+        if (lastSeq > 0) {
```
dimakis: 🔴 **bugs**: switchSession sends `session_suspend` when navigating away from a session, but `handleSwitchSession` on the server (ws-handler-v2.ts:246) never resumes suspended sessions when the user navigates back. The session stays suspended — `sendOrBuffer` keeps buffering events, the client sees nothing — until the 2-minute grace timer fires and transitions to detach. This is also a semantic mismatch: session_suspend was designed for iOS backgrounding ('my WS is about to die'), not for 'I'm looking at a different session.' For the switchSession case, `clearSession()` already stops event delivery — the suspend is unnecessary and harmful. `[fixable]`

packages/harness/src/session-registry.ts
```diff
@@ -314,11 +325,80 @@ export class SessionRegistry {
...
+   * enabling event buffering and instant resume. Starts a grace timer — if
+   * the client doesn't resume within SUSPEND_GRACE_MS, the session transitions
+   * to the normal detach flow.
+   */
+  suspend(clientId: string, _lastClientSeq: number): void {
+    const session = this.sessions.get(clientId);
+    if (!session) return;
+
+    this.suspended.add(clientId);
+    this.suspendBuffers.set(clientId, []);
```
dimakis: 🟡 **bugs**: Double-fire resets suspend buffer. Both `visibilitychange:hidden` and `pagehide` call `sendSuspend()` (connection.ts:291,300), and `sendSuspend()` sends both a WS message and a sendBeacon HTTP POST. Each `suspend()` call on the server reinitializes the buffer to `[]` (line 353), discarding events buffered between the two signals. Guard with an early return: `if (this.suspended.has(clientId)) return;` `[fixable]`

packages/harness/src/session-registry.ts
```diff
@@ -314,11 +325,80 @@ export class SessionRegistry {
...
 
+  // ─── Suspend (proactive iOS backgrounding) ───────────────────────────────
+
+  /**
+   * Suspend a session. The client signals this BEFORE iOS kills the WebSocket,
+   * enabling event buffering and instant resume. Starts a grace timer — if
+   * the client doesn't resume within SUSPEND_GRACE_MS, the session transitions
+   * to the normal detach flow.
+   */
+  suspend(clientId: string, _lastClientSeq: number): void {
```
dimakis, line 353: 🔵 **style**: `_lastClientSeq` parameter is accepted but unused (underscore prefix). If it's planned for future seq-based replay optimization, add a brief comment. Otherwise remove it from the API to avoid confusion — callers are passing real values that are silently discarded. `[fixable]`

server/app.ts
```diff
@@ -421,6 +421,46 @@ app.post('/api/sessions', (req, res) => {
   }
 });
 
+// --- Suspend endpoint (sendBeacon — no auth, same-origin only) ---
+// This must be above authMiddleware because sendBeacon cannot set custom headers.
+// The connectionId in the body is sufficient — it's a cryptographically random
```
dimakis: 🟡 **unsafe_assumptions**: Comment claims connectionId is 'a cryptographically random value' but it's generated with `Math.random()` in index.ts:221 (`conn-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`). This gives roughly 31 bits of entropy in the random portion, and the timestamp is predictable. While the blast radius is limited (suspend is non-destructive), the comment is misleading and the endpoint is fully unauthenticated. Either upgrade to `crypto.randomBytes()` or correct the comment. `[fixable]`

server/app.ts
```diff
@@ -421,6 +421,46 @@ app.post('/api/sessions', (req, res) => {
...
+// value known only to the client that established the WebSocket.
+
+app.post('/api/sessions/suspend', (req, res) => {
```
dimakis, line 429: 🟡 **missing_tests**: No supertest/integration tests for the `POST /api/sessions/suspend` REST endpoint. The WS handler tests cover `handleSessionSuspend` but the HTTP fallback path (body validation, connectionId ownership check, 204 response) is not tested. `[fixable]`

dimakis (commented):
## Centaur Review

Found **8** issue(s) (2 critical) (5 warning).


### `server/query-loop.ts`
_Critical duplicate-replay bug: events are stored in both EventStore and suspend buffer, then both are replayed on reconnect. The switchSession suspend also has no matching resume path, causing sessions to go dark when the user navigates back._

- 🔴 **bugs** (L90): Duplicate event replay on reconnect. Events are appended to the durable EventStore (line 83) BEFORE the suspend buffer check (line 90). The same events end up in both the EventStore and the suspend buffer. In `handleReconnect` (ws-handler-v2.ts:130+186), events are replayed from BOTH sources — first `getEventsAfter()` from EventStore, then `resume()` from the suspend buffer. Every event generated during suspension is sent to the client twice. Fix: either skip EventStore replay when resuming from suspend, or skip the `store.append()` when the session is suspended. `[fixable]`
- 🟡 **bugs** (L90): Suspend blocks event broadcast to ALL watchers. The suspend check returns early before `connRegistry.broadcast(sessionId, enriched)` (line 99). Other connections observing the same session via `connRegistry.watch()` will stop receiving events while the session is suspended. The suspend is per-clientId but the broadcast skip affects the entire session's fan-out path. `[fixable]`


_Critical duplicate-replay bug: events are stored in both EventStore and suspend buffer, then both are replayed on reconnect. The switchSession suspend also has no matching resume path, causing sessions to go dark when the user navigates back._

- 🔴 **bugs** (L234): switchSession sends `session_suspend` when navigating away from a session, but `handleSwitchSession` on the server (ws-handler-v2.ts:246) never resumes suspended sessions when the user navigates back. The session stays suspended — `sendOrBuffer` keeps buffering events, the client sees nothing — until the 2-minute grace timer fires and transitions to detach. This is also a semantic mismatch: session_suspend was designed for iOS backgrounding ('my WS is about to die'), not for 'I'm looking at a different session.' For the switchSession case, `clearSession()` already stops event delivery — the suspend is unnecessary and harmful. `[fixable]`


_Critical duplicate-replay bug: events are stored in both EventStore and suspend buffer, then both are replayed on reconnect. The switchSession suspend also has no matching resume path, causing sessions to go dark when the user navigates back._

- 🟡 **bugs** (L353): Double-fire resets suspend buffer. Both `visibilitychange:hidden` and `pagehide` call `sendSuspend()` (connection.ts:291,300), and `sendSuspend()` sends both a WS message and a sendBeacon HTTP POST. Each `suspend()` call on the server reinitializes the buffer to `[]` (line 353), discarding events buffered between the two signals. Guard with an early return: `if (this.suspended.has(clientId)) return;` `[fixable]`
- 🔵 **style** (L348): `_lastClientSeq` parameter is accepted but unused (underscore prefix). If it's planned for future seq-based replay optimization, add a brief comment. Otherwise remove it from the API to avoid confusion — callers are passing real values that are silently discarded. `[fixable]`


### `server/app.ts`
_Critical duplicate-replay bug: events are stored in both EventStore and suspend buffer, then both are replayed on reconnect. The switchSession suspend also has no matching resume path, causing sessions to go dark when the user navigates back._

- 🟡 **unsafe_assumptions** (L426): Comment claims connectionId is 'a cryptographically random value' but it's generated with `Math.random()` in index.ts:221 (`conn-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`). This gives roughly 31 bits of entropy in the random portion, and the timestamp is predictable. While the blast radius is limited (suspend is non-destructive), the comment is misleading and the endpoint is fully unauthenticated. Either upgrade to `crypto.randomBytes()` or correct the comment. `[fixable]`
- 🟡 **missing_tests** (L429): No supertest/integration tests for the `POST /api/sessions/suspend` REST endpoint. The WS handler tests cover `handleSessionSuspend` but the HTTP fallback path (body validation, connectionId ownership check, 204 response) is not tested. `[fixable]`


_Critical duplicate-replay bug: events are stored in both EventStore and suspend buffer, then both are replayed on reconnect. The switchSession suspend also has no matching resume path, causing sessions to go dark when the user navigates back._

- 🟡 **missing_tests**: No tests for `sendSuspend()` on MitzoConnection — the core client-side logic (WS send + sendBeacon fallback, early exit when no sessions tracked, error handling on WS send failure) is entirely untested. `[fixable]`

dimakis (commented):
## Codex Review

### `server/app.ts`
_Good coverage overall, but the unauthenticated suspend endpoint weakens session ownership guarantees and the client currently double-sends suspend signals._

- 🟡 **bugs** (L428): The new unauthenticated `/api/sessions/suspend` endpoint trusts `connectionId` as a bearer secret, but `connectionId` is now sent to every WebSocket client on connect and can be replayed by any same-origin page or script to suspend another user's sessions. Unlike the WS path, this route bypasses auth entirely and does not bind the request to the authenticated user/session, so a leaked or guessed connectionId is sufficient to interfere with active sessions. `[fixable]`


_Good coverage overall, but the unauthenticated suspend endpoint weakens session ownership guarantees and the client currently double-sends suspend signals._

- 🟡 **regressions** (L132): `sendSuspend()` always fires `navigator.sendBeacon()` after a successful WebSocket send. On browsers that deliver both, the server receives duplicate suspend signals and refreshes the suspend grace timer twice. Because `suspend()` is idempotent but extends the timer, this can unintentionally keep sessions suspended longer than intended after backgrounding. Consider only falling back to beacon when WS send is unavailable/throws, or deduplicate server-side. `[fixable]`

server/app.ts
```diff
@@ -421,6 +421,46 @@ app.post('/api/sessions', (req, res) => {
   }
 });
 
+// --- Suspend endpoint (sendBeacon — no auth, same-origin only) ---
+// This must be above authMiddleware because sendBeacon cannot set custom headers.
+// The connectionId in the body is sufficient — it's a cryptographically random
+// value known only to the client that established the WebSocket.
+
```
dimakis, line 428: 🟡 **bugs**: The new unauthenticated `/api/sessions/suspend` endpoint trusts `connectionId` as a bearer secret, but `connectionId` is now sent to every WebSocket client on connect and can be replayed by any same-origin page or script to suspend another user's sessions. Unlike the WS path, this route bypasses auth entirely and does not bind the request to the authenticated user/session, so a leaked or guessed connectionId is sufficient to interfere with active sessions. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -110,6 +114,47 @@ export class MitzoConnection {
...
+    if (this.seqBySession.size === 0) return;
+
+    const sessions = Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+      sessionId,
+      lastSeq,
+    }));
+
+    const wsPayload = { type: 'session_suspend', sessions };
+
+    // Try WS first (may already be dying)
```
dimakis, line 132: 🟡 **regressions**: `sendSuspend()` always fires `navigator.sendBeacon()` after a successful WebSocket send. On browsers that deliver both, the server receives duplicate suspend signals and refreshes the suspend grace timer twice. Because `suspend()` is idempotent but extends the timer, this can unintentionally keep sessions suspended longer than intended after backgrounding. Consider only falling back to beacon when WS send is unavailable/throws, or deduplicate server-side. `[fixable]`

server/app.ts
```diff
@@ -421,6 +421,59 @@ app.post('/api/sessions', (req, res) => {
   }
 });
 
+// --- Suspend endpoint (sendBeacon fallback) ---
+// Above authMiddleware because sendBeacon cannot set custom headers.
+// Auth is verified via the session cookie (sent automatically by sendBeacon
+// on same-origin requests). connectionId ownership is checked per-session.
+
+app.post('/api/sessions/suspend', (req, res) => {
```
dimakis, line 429: 🟡 **bugs**: The suspend endpoint verifies the JWT via `verifyToken(token).then(...)` but never attaches a `.catch(...)` or wraps it in `await`/`try`/`catch`. If token verification rejects (e.g. malformed cookie, jose failure), Express never sends a response and can surface an unhandled rejection instead of a 401. Other auth paths use middleware; this route should handle verification failures explicitly. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -110,6 +114,51 @@ export class MitzoConnection {
...
+   */
+  sendSuspend(): void {
+    if (this.seqBySession.size === 0) return;
+
+    const sessions = Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+      sessionId,
+      lastSeq,
+    }));
+
+    // Try WS first (may already be dying)
```
dimakis, line 132: 🟡 **regressions**: `sendSuspend()` uses `navigator.sendBeacon` with a `Blob` body of type `application/json`, but the server route relies on `req.body` being parsed. In Express, JSON body parsers typically do not parse `sendBeacon` payloads sent as `text/plain;charset=UTF-8` or other browser-specific beacon content types, and browser behavior here is inconsistent. If the body is not parsed, the fallback silently becomes a 400/no-op exactly in the iOS path this feature is meant to protect. This needs an integration test covering the actual beacon content type or server-side raw/text parsing for this endpoint. `[fixable]`

dimakis (commented):
## Codex Review

### `server/app.ts`
_Good coverage overall, but the new suspend path has two reliability issues in auth/error handling and sendBeacon request parsing._

- 🟡 **bugs** (L429): The suspend endpoint verifies the JWT via `verifyToken(token).then(...)` but never attaches a `.catch(...)` or wraps it in `await`/`try`/`catch`. If token verification rejects (e.g. malformed cookie, jose failure), Express never sends a response and can surface an unhandled rejection instead of a 401. Other auth paths use middleware; this route should handle verification failures explicitly. `[fixable]`


_Good coverage overall, but the new suspend path has two reliability issues in auth/error handling and sendBeacon request parsing._

- 🟡 **regressions** (L132): `sendSuspend()` uses `navigator.sendBeacon` with a `Blob` body of type `application/json`, but the server route relies on `req.body` being parsed. In Express, JSON body parsers typically do not parse `sendBeacon` payloads sent as `text/plain;charset=UTF-8` or other browser-specific beacon content types, and browser behavior here is inconsistent. If the body is not parsed, the fallback silently becomes a 400/no-op exactly in the iOS path this feature is meant to protect. This needs an integration test covering the actual beacon content type or server-side raw/text parsing for this endpoint. `[fixable]`

PR #289, opened 2026-04-24
fix(server): persist user messages for resumed sessions

dimakis:
## Summary

- User messages sent through the resume path (`startChat({ resume })`) were never stored in the event store — only assistant responses were persisted
- After any WS reconnect (frequent on iOS), the frontend replays from the event store and user messages vanished
- **Every** message in a typical session hits the resume path (confirmed via Jaeger traces — all `routing.decision=resume`)
- Fix: store the `user_message` event and echo it to the transport before starting the query loop, mirroring `sendToChat()` behavior

## Root cause

`startChat` line 756 passes `undefined` as `initialPrompt` when `options.resume` is set:
```typescript
options.resume ? undefined : fullPrompt,
```
The query loop only stores user messages when `initialPrompt` is truthy. Result: resumed sessions silently drop all user messages from the event store.

## Test plan

- [x] Structural tests verify `eventStore.append(options.resume, 'user_message')` appears before `runQueryLoop` and includes transport echo
- [x] All 29 chat tests pass
- [ ] Manual: send messages on iOS, background app, return — user messages should persist across WS reconnects

### DIFF
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -304,12 +304,31 @@ describe('startChat stores user message for resumed sessions', () => {
   });
 
   it('echoes user_message to transport and broadcasts to observers', () => {
-    const nearAppend = chatSource.indexOf("eventStore.append(options.resume, 'user_message'");
-    expect(nearAppend).toBeGreaterThan(-1);
-    const region = chatSource.slice(nearAppend, nearAppend + 500);
+    // Region between the resume guard and runQueryLoop — bounds the block
+    // without fragile brace-matching or magic byte offsets.
+    const start = chatSource.indexOf(
+      'if (options.resume) {',
+      chatSource.indexOf('session.queryInstance = q'),
+    );
+    const end = chatSource.indexOf('await runQueryLoop(', start);
+    expect(start).toBeGreaterThan(-1);
+    expect(end).toBeGreaterThan(start);
+    const region = chatSource.slice(start, end);
+    expect(region).toContain("eventStore.append(options.resume, 'user_message'");
     expect(region).toContain('send(transport');
     expect(region).toContain("type: 'user_message'");
     expect(region).toContain('broadcastToObservers(session.observers');
+  });
+
+  it('uses clientMsgId with resume fallback for messageId', () => {
+    const start = chatSource.indexOf(
+      'if (options.resume) {',
+      chatSource.indexOf('session.queryInstance = q'),
+    );
+    const end = chatSource.indexOf('await runQueryLoop(', start);
+    const region = chatSource.slice(start, end);
+    expect(region).toContain('options.clientMsgId');
+    expect(region).toMatch(/umsg-.*-resume/);
   });
 });
 
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -282,6 +282,42 @@ describe('startChat finally block does NOT clean up worktrees', () => {
   });
 });
 
+describe('startChat stores user message for resumed sessions', () => {
+  it('appends user_message to eventStore when options.resume is set', async () => {
+    // Structural test: verify that startChat stores the user message before
+    // starting the query loop when resuming. Without this, user messages are
+    // invisible after WS reconnect because sendOrBuffer only persists v2
+    // events emitted by the query loop — the user's prompt is never one of those.
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    const chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+
+    // The pattern: eventStore.append(options.resume, 'user_message', ...)
+    // must appear BEFORE runQueryLoop in startChat
+    const appendIdx = chatSource.indexOf("eventStore.append(options.resume, 'user_message'");
+    const queryLoopIdx = chatSource.indexOf('await runQueryLoop(');
+    expect(appendIdx).toBeGreaterThan(-1);
+    expect(queryLoopIdx).toBeGreaterThan(-1);
+    expect(appendIdx).toBeLessThan(queryLoopIdx);
+  });
+
+  it('echoes user_message to transport on resume', async () => {
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    const chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+
+    // Find the resume block near the eventStore.append call (not the earlier
+    // resumability validation block). Look for the block that contains both
+    // the append and the send.
+    const nearAppend = chatSource.indexOf("eventStore.append(options.resume, 'user_message'");
+    expect(nearAppend).toBeGreaterThan(-1);
+    // The send(transport, ...) echo must be in the same code region
+    const regionAfterAppend = chatSource.slice(nearAppend, nearAppend + 500);
+    expect(regionAfterAppend).toContain('send(transport');
+    expect(regionAfterAppend).toContain("type: 'user_message'");
+  });
+});
+
 describe('isIsolationEnabled', () => {
   let originalEnv: string | undefined;
 
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { describe, it, expect, vi, beforeAll, beforeEach, afterEach } from 'vitest';
 import type { ManagedSession } from '@mitzo/harness';
 
 vi.mock('../worktree.js', async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


server/__tests__/chat.test.ts
```diff
@@ -282,6 +282,42 @@ describe('startChat finally block does NOT clean up worktrees', () => {
   });
 });
 
+describe('startChat stores user message for resumed sessions', () => {
+  it('appends user_message to eventStore when options.resume is set', async () => {
```
dimakis: 🔵 **missing_tests**: The tests are structural (source-code parsing), which is an established pattern in this file but inherently brittle — a refactor that renames the variable or extracts a helper will break the test without changing behavior. Consider also adding (or documenting why it's impractical to add) a behavioral test that exercises the resume path with a mock eventStore and transport, verifying the append and send calls actually happen at runtime.

server/chat.ts
```diff
@@ -747,6 +747,21 @@ async function _startChatInner(
...
+    if (options.resume) {
+      const messageId = options.clientMsgId || `umsg-${Date.now()}-resume`;
+      eventStore.append(options.resume, 'user_message', {
+        v: 2,
+        type: 'user_message',
+        ts: Date.now(),
+        messageId,
+        text: fullPrompt,
+      });
+      send(transport, { type: 'user_message', messageId, text: fullPrompt });
```
dimakis: 🔵 **style**: Both `sendToChat` (line 884) and `interruptChat` (line 914) call `broadcastToObservers(session.observers, echo)` after `send(transport, ...)`. This new code path omits it. At this point in `startChat` the observer set is empty (just registered), so it's a no-op today, but for consistency with the other two call sites and future-proofing, consider adding `broadcastToObservers(session.observers, { type: 'user_message', messageId, text: fullPrompt })` after the send. `[fixable]`

server/chat.ts
```diff
@@ -747,6 +747,21 @@ async function _startChatInner(
...
+    // For resumed sessions the prompt is sent to the SDK but was never stored
+    // in the event store — making user messages invisible after WS reconnect.
+    // Store and echo it here so the frontend can replay it.
+    if (options.resume) {
+      const messageId = options.clientMsgId || `umsg-${Date.now()}-resume`;
+      eventStore.append(options.resume, 'user_message', {
+        v: 2,
+        type: 'user_message',
+        ts: Date.now(),
+        messageId,
```
dimakis, line 759: 🔵 **unsafe_assumptions**: `text: fullPrompt` stores the assembled prompt (including any injected context blocks and image XML tags from `assemblePrompt`), not the raw user input. This is consistent with `sendToChat`/`interruptChat`, but it means the echoed `user_message` shown to the frontend includes context-block XML markup. If the frontend renders this as plain text, it may display raw XML to the user. Verify the frontend strips or hides context-block wrappers.

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Correct fix for a real bug — user messages on resume were lost. Implementation is consistent with existing patterns. Minor suggestions: add observer broadcast for consistency, consider behavioral tests alongside structural ones, and verify frontend rendering of assembled prompts._

- 🔵 **missing_tests** (L286): The tests are structural (source-code parsing), which is an established pattern in this file but inherently brittle — a refactor that renames the variable or extracts a helper will break the test without changing behavior. Consider also adding (or documenting why it's impractical to add) a behavioral test that exercises the resume path with a mock eventStore and transport, verifying the append and send calls actually happen at runtime.


### `server/chat.ts`
_Correct fix for a real bug — user messages on resume were lost. Implementation is consistent with existing patterns. Minor suggestions: add observer broadcast for consistency, consider behavioral tests alongside structural ones, and verify frontend rendering of assembled prompts._

- 🔵 **style** (L762): Both `sendToChat` (line 884) and `interruptChat` (line 914) call `broadcastToObservers(session.observers, echo)` after `send(transport, ...)`. This new code path omits it. At this point in `startChat` the observer set is empty (just registered), so it's a no-op today, but for consistency with the other two call sites and future-proofing, consider adding `broadcastToObservers(session.observers, { type: 'user_message', messageId, text: fullPrompt })` after the send. `[fixable]`
- 🔵 **unsafe_assumptions** (L759): `text: fullPrompt` stores the assembled prompt (including any injected context blocks and image XML tags from `assemblePrompt`), not the raw user input. This is consistent with `sendToChat`/`interruptChat`, but it means the echoed `user_message` shown to the frontend includes context-block XML markup. If the frontend renders this as plain text, it may display raw XML to the user. Verify the frontend strips or hides context-block wrappers.

server/__tests__/chat.test.ts
```diff
@@ -282,6 +282,37 @@ describe('startChat finally block does NOT clean up worktrees', () => {
...
 
+// Structural tests: startChat requires the full Agent SDK query() pipeline,
+// making behavioral mocking impractical. Source-code assertions lock in the
+// key invariant (store + echo + broadcast before runQueryLoop) with minimal
+// coupling to the SDK internals.
+describe('startChat stores user message for resumed sessions', () => {
+  let chatSource: string;
+
+  beforeAll(async () => {
+    const { readFileSync } = await import('fs');
```
dimakis, line 293: 🔵 **style**: The `readFileSync`/`join` import and source-read pattern is duplicated from the test at line 274. Consider hoisting `chatSource` to a shared `beforeAll` in the parent scope or a helper, since both structural test suites read the same file. Low priority — the existing test already has the same pattern. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -282,6 +282,37 @@ describe('startChat finally block does NOT clean up worktrees', () => {
...
+    const queryLoopIdx = chatSource.indexOf('await runQueryLoop(');
+    expect(appendIdx).toBeGreaterThan(-1);
+    expect(queryLoopIdx).toBeGreaterThan(-1);
+    expect(appendIdx).toBeLessThan(queryLoopIdx);
+  });
+
+  it('echoes user_message to transport and broadcasts to observers', () => {
+    const nearAppend = chatSource.indexOf("eventStore.append(options.resume, 'user_message'");
+    expect(nearAppend).toBeGreaterThan(-1);
+    const region = chatSource.slice(nearAppend, nearAppend + 500);
```
dimakis: 🟡 **unsafe_assumptions**: The `slice(nearAppend, nearAppend + 500)` window is a magic number. If a refactor adds lines between `eventStore.append` and `broadcastToObservers`, the test silently passes or fails depending on byte distance. The existing `finally`-block structural test (line 279) uses a regex to extract a bounded code region, which is more resilient. Consider matching the `if (options.resume)` block instead of a fixed byte offset. `[fixable]`

server/chat.ts
```diff
@@ -747,6 +747,23 @@ async function _startChatInner(
 
     session.queryInstance = q;
 
+    // For resumed sessions the prompt is sent to the SDK but was never stored
+    // in the event store — making user messages invisible after WS reconnect.
+    // Store and echo it here so the frontend can replay it.
+    if (options.resume) {
+      const messageId = options.clientMsgId || `umsg-${Date.now()}-resume`;
```
dimakis, line 754: 🔵 **missing_tests**: The structural test verifies ordering and presence but doesn't cover the `messageId` fallback logic (`options.clientMsgId || \`umsg-...\``). This is the same pattern as `sendToChat`/`interruptChat`, so it's low risk, but a unit test for the fallback ID generation would catch accidental breakage of the deduplication contract. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean, well-scoped fix that mirrors the existing sendToChat/interruptChat echo pattern. The structural tests are pragmatic given the SDK mocking constraints. Minor suggestions around test brittleness (magic 500-byte window) and deduplication of the source-read pattern._

- 🔵 **style** (L293): The `readFileSync`/`join` import and source-read pattern is duplicated from the test at line 274. Consider hoisting `chatSource` to a shared `beforeAll` in the parent scope or a helper, since both structural test suites read the same file. Low priority — the existing test already has the same pattern. `[fixable]`
- 🟡 **unsafe_assumptions** (L309): The `slice(nearAppend, nearAppend + 500)` window is a magic number. If a refactor adds lines between `eventStore.append` and `broadcastToObservers`, the test silently passes or fails depending on byte distance. The existing `finally`-block structural test (line 279) uses a regex to extract a bounded code region, which is more resilient. Consider matching the `if (options.resume)` block instead of a fixed byte offset. `[fixable]`


### `server/chat.ts`
_Clean, well-scoped fix that mirrors the existing sendToChat/interruptChat echo pattern. The structural tests are pragmatic given the SDK mocking constraints. Minor suggestions around test brittleness (magic 500-byte window) and deduplication of the source-read pattern._

- 🔵 **missing_tests** (L754): The structural test verifies ordering and presence but doesn't cover the `messageId` fallback logic (`options.clientMsgId || \`umsg-...\``). This is the same pattern as `sendToChat`/`interruptChat`, so it's low risk, but a unit test for the fallback ID generation would catch accidental breakage of the deduplication contract. `[fixable]`

PR #290, opened 2026-04-24
fix(worktree): raise stale TTL to 96h and skip active sessions during cleanup

dimakis:
## Summary

- **Bumps `WORKTREE_STALE_HOURS` from 24h to 96h** — the old 1-day TTL was deleting worktrees from under active Cursor/Claude Code sessions that went idle for even a workday. The 96h value matches what CLAUDE.md already documented.
- **Adds active-session guard** — `cleanupStaleWorktrees` now accepts an optional `activeSessionIds` set (wtIds from the `SessionRegistry`). Worktrees belonging to live sessions are never cleaned up, regardless of age.
- Both startup and periodic cleanup in `index.ts` now collect active wtIds from the registry before each sweep.

## Test plan

- [x] Two new tests: `skips worktrees whose names appear in activeSessionIds` and `still removes stale worktrees not in activeSessionIds`
- [x] All 31 worktree tests pass
- [x] All 94 dependent tests (routes, todos, yapper-proxy) pass
- [x] No linter errors
- [ ] CI green

### DIFF
--- a/server/__tests__/worktree.test.ts
+++ b/server/__tests__/worktree.test.ts
@@ -457,6 +457,49 @@ describe('cleanupStaleWorktrees', () => {
     expect(remaining).not.toContain(sessionId);
   });
 
+  it('skips worktrees whose names appear in activeSessionIds', () => {
+    const wtDir = join(baseRepo, '.claude', 'worktrees');
+    mkdirSync(wtDir, { recursive: true });
+    const sessionId = 'test-active-session';
+    const wtPath = join(wtDir, sessionId);
+    execFileSync('git', ['-C', baseRepo, 'worktree', 'add', '-b', `session/${sessionId}`, wtPath], {
+      stdio: 'pipe',
+    });
+
+    // Age the worktree past the cutoff
+    const fiveDaysAgo = new Date(Date.now() - 5 * 24 * 60 * 60 * 1000);
+    utimesSync(wtPath, fiveDaysAgo, fiveDaysAgo);
+
+    // Pass the session as active — should be protected from cleanup
+    cleanupStaleWorktrees(baseRepo, inboxDir, new Set([sessionId]));
+
+    const remaining = readdirSync(wtDir);
+    expect(remaining).toContain(sessionId);
+  });
+
+  it('still removes stale worktrees not in activeSessionIds', () => {
+    const wtDir = join(baseRepo, '.claude', 'worktrees');
+    mkdirSync(wtDir, { recursive: true });
+
+    const activeId = 'test-active';
+    const staleId = 'test-stale';
+
+    for (const id of [activeId, staleId]) {
+      const wtPath = join(wtDir, id);
+      execFileSync('git', ['-C', baseRepo, 'worktree', 'add', '-b', `session/${id}`, wtPath], {
+        stdio: 'pipe',
+      });
+      const fiveDaysAgo = new Date(Date.now() - 5 * 24 * 60 * 60 * 1000);
+      utimesSync(wtPath, fiveDaysAgo, fiveDaysAgo);
+    }
+
+    cleanupStaleWorktrees(baseRepo, inboxDir, new Set([activeId]));
+
+    const remaining = readdirSync(wtDir);
+    expect(remaining).toContain(activeId);
+    expect(remaining).not.toContain(staleId);
+  });
+
   it('creates unique filenames for multiple dirty worktrees', () => {
     const wtDir = join(baseRepo, '.claude', 'worktrees');
     mkdirSync(wtDir, { recursive: true });
--- a/server/__tests__/worktree.test.ts
+++ b/server/__tests__/worktree.test.ts
@@ -500,6 +500,25 @@ describe('cleanupStaleWorktrees', () => {
     expect(remaining).not.toContain(staleId);
   });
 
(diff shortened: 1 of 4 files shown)


server/worktree.ts
```diff
@@ -424,6 +430,11 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
       const mtimeAge = now - statSync(fullPath).mtimeMs;
       const isStale = nameAge !== null ? nameAge > cutoff && mtimeAge > cutoff : mtimeAge > cutoff;
 
+      if (activeSessionIds?.has(entry)) {
```
dimakis: 🟡 **bugs**: The `activeSessionIds` check is placed after `statSync(fullPath)` (line 430). If `statSync` throws (e.g., directory removed by a concurrent process), the active-session guard never runs and the error falls into the catch block. Move the `activeSessionIds?.has(entry)` check to immediately after `const fullPath = ...` (line 424), before any filesystem calls, so active sessions are skipped without touching disk. `[fixable]`

server/index.ts
```diff
@@ -795,9 +795,17 @@ checkPort(PORT).then((inUse) => {
...
+      for (const [, session] of registry.entries()) {
+        if (session.wtId) ids.add(session.wtId);
+      }
+      return ids;
+    }
+
     for (const [label, repoPath] of repoEntries) {
       try {
-        cleanupStaleWorktrees(repoPath, inboxDir);
+        cleanupStaleWorktrees(repoPath, inboxDir, collectActiveWtIds());
```
dimakis: 🔵 **style**: `collectActiveWtIds()` is called inside the for-loop at startup (line 808, once per repo) but once before the loop in the interval (line 838). At startup the set won't change between iterations — hoist it before the loop for consistency with the interval path and to avoid redundant iteration over the registry. `[fixable]`

server/__tests__/worktree.test.ts
```diff
@@ -457,6 +457,49 @@ describe('cleanupStaleWorktrees', () => {
     expect(remaining).not.toContain(sessionId);
   });
 
+  it('skips worktrees whose names appear in activeSessionIds', () => {
```
dimakis, line 460: 🔵 **missing_tests**: The new tests only cover name-based session IDs (no date prefix). Consider adding a test where the active session's directory name matches the `YYYY-MM-DD-XXXXXX` format used by `parseWorktreeAge`, confirming the active-session guard takes priority over name-based age detection. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_Solid change — TTL bump and active-session guard are correct. The active-session check should be moved before `statSync` to avoid a potential race where the guard is bypassed if stat fails._

- 🟡 **bugs** (L433): The `activeSessionIds` check is placed after `statSync(fullPath)` (line 430). If `statSync` throws (e.g., directory removed by a concurrent process), the active-session guard never runs and the error falls into the catch block. Move the `activeSessionIds?.has(entry)` check to immediately after `const fullPath = ...` (line 424), before any filesystem calls, so active sessions are skipped without touching disk. `[fixable]`


### `server/index.ts`
_Solid change — TTL bump and active-session guard are correct. The active-session check should be moved before `statSync` to avoid a potential race where the guard is bypassed if stat fails._

- 🔵 **style** (L808): `collectActiveWtIds()` is called inside the for-loop at startup (line 808, once per repo) but once before the loop in the interval (line 838). At startup the set won't change between iterations — hoist it before the loop for consistency with the interval path and to avoid redundant iteration over the registry. `[fixable]`


_Solid change — TTL bump and active-session guard are correct. The active-session check should be moved before `statSync` to avoid a potential race where the guard is bypassed if stat fails._

- 🔵 **missing_tests** (L460): The new tests only cover name-based session IDs (no date prefix). Consider adding a test where the active session's directory name matches the `YYYY-MM-DD-XXXXXX` format used by `parseWorktreeAge`, confirming the active-session guard takes priority over name-based age detection. `[fixable]`

server/worktree.ts
```diff
@@ -415,6 +421,11 @@ export function cleanupStaleWorktrees(baseRepo: string, inboxDir?: string): void
   let skipped = 0;
 
   for (const entry of readdirSync(dir)) {
+    if (activeSessionIds?.has(entry)) {
```
dimakis, line 425: 🔵 **style**: Active-session skips are logged individually but not reflected in the summary log on line 479 (`skipped (dirty)`). Consider adding a separate counter (e.g. `protected`) so the summary line reports how many worktrees were kept alive due to active sessions. Minor observability gap — each skip already has its own info log, so this is low priority. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `server/worktree.ts`
_Clean, well-tested change. The TTL bump and active-session guard are both appropriate defenses against premature worktree cleanup. Backward compatibility is preserved via the optional parameter. Only a minor observability suggestion._

- 🔵 **style** (L424): Active-session skips are logged individually but not reflected in the summary log on line 479 (`skipped (dirty)`). Consider adding a separate counter (e.g. `protected`) so the summary line reports how many worktrees were kept alive due to active sessions. Minor observability gap — each skip already has its own info log, so this is low priority. `[fixable]`

PR #291, opened 2026-04-24
fix(tests): update stale test expectations

dimakis:
## Summary
- Fixes 31 test failures across 8 test files where tests had drifted from implementation
- Server: auth cookies for suspend endpoint, sessionTotal removed from message_start, async microtask flush
- Frontend: TabBar tabs reshuffled (Calendar added, Tasks→More), TaskNode CSS selectors + icons, PageHeader adoption, fetch mock for biometric

## Test plan
- [x] All 8 previously-failing test files now pass (147/147 tests)
- [x] No implementation changes except removing `sessionTotal` from message_start token_update (test-driven correction — result updates should be the only source)
- [ ] Full `npm test` suite green

### DIFF
--- a/frontend/src/lib/__tests__/biometric.test.ts
+++ b/frontend/src/lib/__tests__/biometric.test.ts
@@ -105,6 +105,9 @@ describe('biometricLogin', () => {
       password: 'stored-jwt',
     });
 
+    // Mock fetch for server validation
+    global.fetch = vi.fn().mockResolvedValue({ ok: true });
+
     const token = await biometricLogin();
     expect(token).toBe('stored-jwt');
     expect(localStorage.getItem('mitzo_auth_token')).toBe('stored-jwt');
--- a/frontend/src/lib/__tests__/biometric.test.ts
+++ b/frontend/src/lib/__tests__/biometric.test.ts
@@ -105,8 +105,8 @@ describe('biometricLogin', () => {
       password: 'stored-jwt',
     });
 
-    // Mock fetch for server validation
-    global.fetch = vi.fn().mockResolvedValue({ ok: true });
+    // Mock fetch for server validation (vi.spyOn auto-restores on clearAllMocks)
+    vi.spyOn(global, 'fetch').mockResolvedValue({ ok: true } as Response);
 
     const token = await biometricLogin();
     expect(token).toBe('stored-jwt');
--- a/packages/harness/__tests__/permission-handler.test.ts
+++ b/packages/harness/__tests__/permission-handler.test.ts
@@ -112,7 +112,6 @@ describe('buildPermissionHandler', () => {
   });
 
   it('sends permission_request for unknown tools and resolves on response', async () => {
-    vi.useFakeTimers();
     const transport = fakeTransport();
     registry.register('client-1', {
       transport,
@@ -131,6 +130,9 @@ describe('buildPermissionHandler', () => {
       },
     );
 
+    // Flush microtasks to allow async checkWorktreePolicy to complete
+    await Promise.resolve();
+
     // Should have sent a permission_request
     expect(transport.sent.length).toBe(1);
     expect(transport.sent[0].type).toBe('permission_request');
@@ -141,8 +143,6 @@ describe('buildPermissionHandler', () => {
     const result = await promise;
     expect(result.behavior).toBe('allow');
     expect(result.decisionClassification).toBe('user_temporary');
-
-    vi.useRealTimers();
   });
 
   it('denies when aborted', async () => {
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -557,7 +557,6 @@ async function _runQueryLoopInner(
                   type: 'token_update',
                   agentContext: agentContextTokens,
                   contextCeiling: CONTEXT_CEILING_TOKENS,
-                  sessionTotal: liveSessionTokens,
                   turnIndex,
                   ...compactionFields(),
                 });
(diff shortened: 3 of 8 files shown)


frontend/src/lib/__tests__/biometric.test.ts
```diff
@@ -105,6 +105,9 @@ describe('biometricLogin', () => {
       password: 'stored-jwt',
     });
 
+    // Mock fetch for server validation
+    global.fetch = vi.fn().mockResolvedValue({ ok: true });
```
dimakis: 🟡 **bugs**: `global.fetch` is directly reassigned to a mock but never restored. `vi.clearAllMocks()` in `beforeEach` resets call counts but doesn't restore the original function. This leaks a stale mock into subsequent tests. Use `vi.spyOn(global, 'fetch')` instead, which `vi.clearAllMocks()` / `vi.restoreAllMocks()` can properly clean up, or save and restore `global.fetch` in `beforeEach`/`afterEach`. `[fixable]`

server/query-loop.ts
```diff
@@ -557,7 +557,6 @@ async function _runQueryLoopInner(
                   type: 'token_update',
```
dimakis, line 557: 🔵 **regressions**: Removing `sessionTotal` from the mid-turn `token_update` emission is correct — the authoritative total is emitted at end-of-message (line ~424). However, any frontend code that relied on receiving `sessionTotal` on every `token_update` event may briefly show stale data until the end-of-message emission. Verified that the client protocol parser already handles it as optional (`if (msg.sessionTotal != null)`), so this is safe.

dimakis (commented):
## Centaur Review

### `frontend/src/lib/__tests__/biometric.test.ts`
_Straightforward test alignment to match current implementation. All updated expectations verified against source. One minor issue: the biometric test leaks a `global.fetch` mock that should use `vi.spyOn` instead of direct assignment._

- 🟡 **bugs** (L109): `global.fetch` is directly reassigned to a mock but never restored. `vi.clearAllMocks()` in `beforeEach` resets call counts but doesn't restore the original function. This leaks a stale mock into subsequent tests. Use `vi.spyOn(global, 'fetch')` instead, which `vi.clearAllMocks()` / `vi.restoreAllMocks()` can properly clean up, or save and restore `global.fetch` in `beforeEach`/`afterEach`. `[fixable]`


### `server/query-loop.ts`
_Straightforward test alignment to match current implementation. All updated expectations verified against source. One minor issue: the biometric test leaks a `global.fetch` mock that should use `vi.spyOn` instead of direct assignment._

- 🔵 **regressions** (L557): Removing `sessionTotal` from the mid-turn `token_update` emission is correct — the authoritative total is emitted at end-of-message (line ~424). However, any frontend code that relied on receiving `sessionTotal` on every `token_update` event may briefly show stale data until the end-of-message emission. Verified that the client protocol parser already handles it as optional (`if (msg.sessionTotal != null)`), so this is safe.

server/query-loop.ts
dimakis, line 557: 🟡 **regressions**: This is a behavioral change, not a stale test fix. Removing `sessionTotal: liveSessionTokens` from the mid-turn token_update emission means clients no longer receive incremental session token totals during streaming. The field is optional in the protocol and the parser handles its absence, so no breakage — but the PR title ("fix(tests): update stale test expectations") is misleading since this file is production code, not a test. Consider whether this removal is intentional and, if so, should be in its own commit with a more descriptive message.

packages/harness/__tests__/permission-handler.test.ts
```diff
@@ -131,6 +130,9 @@ describe('buildPermissionHandler', () => {
       },
     );
 
+    // Flush microtasks to allow async checkWorktreePolicy to complete
+    await Promise.resolve();
```
dimakis, line 134: 🔵 **unsafe_assumptions**: A single `await Promise.resolve()` to flush microtasks is fragile — it works only if `checkWorktreePolicy` completes in exactly one microtask tick. If the implementation is later refactored to include additional awaits before the permission_request is sent, this test will silently fail. Consider `await vi.waitFor(() => expect(transport.sent.length).toBe(1))` or a small flush helper for robustness. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_All test expectation updates are correct and match current implementations. One non-test behavioral change (removing `sessionTotal` from mid-turn token_update) is bundled under a test-fix commit message — consider separating it._

- 🟡 **regressions** (L557): This is a behavioral change, not a stale test fix. Removing `sessionTotal: liveSessionTokens` from the mid-turn token_update emission means clients no longer receive incremental session token totals during streaming. The field is optional in the protocol and the parser handles its absence, so no breakage — but the PR title ("fix(tests): update stale test expectations") is misleading since this file is production code, not a test. Consider whether this removal is intentional and, if so, should be in its own commit with a more descriptive message.


### `packages/harness/__tests__/permission-handler.test.ts`
_All test expectation updates are correct and match current implementations. One non-test behavioral change (removing `sessionTotal` from mid-turn token_update) is bundled under a test-fix commit message — consider separating it._

- 🔵 **unsafe_assumptions** (L134): A single `await Promise.resolve()` to flush microtasks is fragile — it works only if `checkWorktreePolicy` completes in exactly one microtask tick. If the implementation is later refactored to include additional awaits before the permission_request is sent, this test will silently fail. Consider `await vi.waitFor(() => expect(transport.sent.length).toBe(1))` or a small flush helper for robustness. `[fixable]`

PR #292, opened 2026-04-25
fix(voice): unlock AudioContext on boot when TTS is pre-enabled

dimakis:
## Summary

- **Root cause**: When TTS was already enabled from a previous session (persisted in localStorage), `unlockAudioContext()` was never called on page load — it only ran inside `setTtsEnabled(true)`, which isn't invoked on init. On iOS Safari, this caused `speak()` calls from `useAutoSpeak` to silently hang on `ctx.resume()`, only unblocking when a later user gesture (like pressing the mic button) provided the activation the browser required.
- **Fix 1**: Register a one-shot `click`/`touchstart` listener on mount when `ttsEnabled` is true but the AudioContext is suspended, so the context is unlocked before `useAutoSpeak` needs it.
- **Fix 2**: Call `stopSpeaking()` before `startRecording()` when the mic button is pressed, preventing TTS playback from overlapping with mic recording.

## Test plan

- [x] Existing voice tests pass (53/53)
- [x] 3 new tests added and passing:
  - Interaction listener fires `unlockAudioContext` on click when AudioContext is suspended
  - No listener registered when AudioContext is already running  
  - `stopSpeaking` called when mic recording starts
- [ ] Manual: enable TTS, close app, reopen — verify TTS plays on first assistant response after any tap
- [ ] Manual: while TTS is playing, press mic — verify TTS stops immediately

### DIFF
--- a/frontend/src/components/ChatArea.tsx
+++ b/frontend/src/components/ChatArea.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { UserBubble, TextBubble } from './MessageBubble';
 import { ThinkingBlock } from './ThinkingBlock';
 import { ToolPill } from './ToolPill';
@@ -16,6 +16,13 @@ import type {
 } from '../types/chat';
 import type { ProgressBlock } from '@mitzo/protocol';
 
+export interface ChatAreaVoice {
+  ttsAvailable: boolean;
+  speak: (text: string) => Promise<void>;
+  stopSpeaking: () => void;
+  speaking: boolean;
+}
+
 export interface ChatAreaProps {
   messages: FinishedMessage[];
   current: StreamingMessage | null;
@@ -32,6 +39,8 @@ export interface ChatAreaProps {
   sessionContext?: string | null;
   /** Progress blocks indexed by toolId for rendering ProgressWidget on TodoWrite blocks */
   progressByToolId?: Record<string, ProgressBlock>;
+  /** Voice capabilities for per-block read-aloud */
+  voice?: ChatAreaVoice;
 }
 
 export function ChatArea({
@@ -43,10 +52,36 @@ export function ChatArea({
   scrollRef: externalScrollRef,
   sessionContext,
   progressByToolId,
+  voice,
 }: ChatAreaProps) {
   const internalScrollRef = useRef<HTMLDivElement>(null);
   const scrollRef = externalScrollRef ?? internalScrollRef;
   const prevMessageCount = useRef(0);
+
+  // Track which block is currently being read aloud
+  const [speakingBlockId, setSpeakingBlockId] = useState<string | null>(null);
+
+  // Clear speakingBlockId when voice stops
+  useEffect(() => {
+    if (voice && !voice.speaking) {
+      setSpeakingBlockId(null);
+    }
+  }, [voice?.speaking]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  const speakBlock = useCallback(
+    (blockId: string, text: string) => {
+      if (!voice) return;
+      setSpeakingBlockId(blockId);
+      voice.speak(text);
+    },
+    [voice],
+  );
+
+  const stopBlock = useCallback(() => {
+    if (!voice) return;
+    voice.stopSpeaking();
+    setSpeakingBlockId(null);
+  }, [voice]);
 
   // Scroll to bottom on session restore (messages jump from 0 to N)
   useEffect(() => {
@@ -129,6 +164,15 @@ export function ChatArea({
                 images={msg.images}
                 contextBlocks={msg.contextBlocks}
                 timestamp={msg.timestamp}
+                readAloud={
+                  voice?.ttsAvailable
+                    ? {
+                        active: speakingBlockId === msg.messageId,
+                        onSpeak: (text) => speakBlock(msg.messageId, text),
+                        onStop: stopBlock,
+                      }
+                    : undefined
+                }
               />
             );
           }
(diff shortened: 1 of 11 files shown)


dimakis (commented):
## Centaur Review

LGTM — no issues found.

frontend/src/hooks/useVoice.ts
```diff
@@ -111,6 +112,32 @@ export function useVoice(): UseVoiceReturn {
...
+  // because setTtsEnabled(true) isn't called on load. iOS Safari blocks playback
+  // from non-gesture contexts, so useAutoSpeak's speak() calls silently hang.
+  // Register a one-shot listener to unlock on the user's first tap/click.
+  useEffect(() => {
+    if (!ttsEnabled) return;
+
+    const ctx = getOrCreateAudioContext();
+    if (ctx.state !== 'suspended') return;
+
+    const unlock = () => {
```
dimakis, line 126: 🟡 **bugs**: The `unlock` handler manually removes listeners AND `{ once: true }` is set. When a `touchstart` fires (mobile tap), `once` auto-removes the touchstart listener and `unlock()` manually removes the click listener. However, a tap also synthesizes a `click` event. Since the click listener was already removed inside `unlock()`, no double-call occurs — so this is safe in practice. But the redundant removal pattern (`once` + manual) is slightly confusing. Consider dropping `{ once: true }` since `unlock()` already handles removal, or removing the manual `removeEventListener` calls inside `unlock()` and relying solely on `{ once: true }` plus the cleanup return. Either approach is clearer than both together. `[fixable]`

frontend/src/hooks/__tests__/useVoice.test.ts
```diff
@@ -519,6 +519,54 @@ describe('useVoice', () => {
       });
     });
 
+    it('registers interaction listener to unlock AudioContext when ttsEnabled is pre-set', async () => {
```
dimakis, line 522: 🔵 **missing_tests**: The boot-unlock tests only verify the `click` event. On iOS Safari (the primary target for this fix), `touchstart` fires before `click`. Add a test case dispatching `touchstart` to confirm the listener works for touch devices. `[fixable]`

frontend/src/components/ChatArea.tsx
```diff
@@ -43,11 +52,37 @@ export function ChatArea({
...
 
+  // Track which block is currently being read aloud
+  const [speakingBlockId, setSpeakingBlockId] = useState<string | null>(null);
+
+  // Clear speakingBlockId when voice stops
+  useEffect(() => {
+    if (voice && !voice.speaking) {
+      setSpeakingBlockId(null);
+    }
+  }, [voice?.speaking]); // eslint-disable-line react-hooks/exhaustive-deps
```
dimakis, line 69: 🔵 **style**: The `eslint-disable-line react-hooks/exhaustive-deps` suppresses a valid warning — the effect closure accesses `voice` but it's not in the dep array. The effect body guards with `if (voice && ...)` so it's safe, but a brief inline comment explaining why `voice` is intentionally excluded (only the `speaking` boolean matters) would help future readers understand the suppression. `[fixable]`

frontend/src/components/ChatArea.tsx
```diff
@@ -16,6 +16,13 @@ import type {
 } from '../types/chat';
 import type { ProgressBlock } from '@mitzo/protocol';
```
dimakis, line 17: 🔵 **style**: The `ChatAreaVoice` interface duplicates a subset of `UseVoiceReturn` fields. Consider using `Pick<UseVoiceReturn, 'ttsAvailable' | 'speak' | 'stopSpeaking' | 'speaking'>` to keep the two in sync and avoid drift. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useVoice.ts`
_Solid fix for the iOS Safari AudioContext boot-unlock gap. The implementation is correct and well-tested; findings are limited to minor style/redundancy issues and a missing touchstart test case for the primary target platform._

- 🟡 **bugs** (L126): The `unlock` handler manually removes listeners AND `{ once: true }` is set. When a `touchstart` fires (mobile tap), `once` auto-removes the touchstart listener and `unlock()` manually removes the click listener. However, a tap also synthesizes a `click` event. Since the click listener was already removed inside `unlock()`, no double-call occurs — so this is safe in practice. But the redundant removal pattern (`once` + manual) is slightly confusing. Consider dropping `{ once: true }` since `unlock()` already handles removal, or removing the manual `removeEventListener` calls inside `unlock()` and relying solely on `{ once: true }` plus the cleanup return. Either approach is clearer than both together. `[fixable]`


### `frontend/src/hooks/__tests__/useVoice.test.ts`
_Solid fix for the iOS Safari AudioContext boot-unlock gap. The implementation is correct and well-tested; findings are limited to minor style/redundancy issues and a missing touchstart test case for the primary target platform._

- 🔵 **missing_tests** (L522): The boot-unlock tests only verify the `click` event. On iOS Safari (the primary target for this fix), `touchstart` fires before `click`. Add a test case dispatching `touchstart` to confirm the listener works for touch devices. `[fixable]`


### `frontend/src/components/__tests__/ReadAloudButton.test.tsx`
_Solid fix for the iOS Safari AudioContext boot-unlock gap. The implementation is correct and well-tested; findings are limited to minor style/redundancy issues and a missing touchstart test case for the primary target platform._

- 🔵 **missing_tests**: No test verifies that the ReadAloudButton does NOT render when `readAloud` is undefined (the graceful degradation path in MessageBubble). Consider adding a test on TextBubble/UserBubble that passes no `readAloud` prop and asserts no `.read-aloud-btn` is in the DOM. `[fixable]`


### `frontend/src/components/ChatArea.tsx`
_Solid fix for the iOS Safari AudioContext boot-unlock gap. The implementation is correct and well-tested; findings are limited to minor style/redundancy issues and a missing touchstart test case for the primary target platform._

- 🔵 **style** (L69): The `eslint-disable-line react-hooks/exhaustive-deps` suppresses a valid warning — the effect closure accesses `voice` but it's not in the dep array. The effect body guards with `if (voice && ...)` so it's safe, but a brief inline comment explaining why `voice` is intentionally excluded (only the `speaking` boolean matters) would help future readers understand the suppression. `[fixable]`
- 🔵 **style** (L17): The `ChatAreaVoice` interface duplicates a subset of `UseVoiceReturn` fields. Consider using `Pick<UseVoiceReturn, 'ttsAvailable' | 'speak' | 'stopSpeaking' | 'speaking'>` to keep the two in sync and avoid drift. `[fixable]`

PR #296, opened 2026-04-26
feat(watch): Xcode wiring + WatchConnectivity relay

dimakis:
## Summary

- Programmatically wires the watchOS target into the Xcode project (no manual Xcode UI steps needed)
- Adds MitzoShared as SPM dependency for both iOS app and MitzoWatch targets
- Implements WatchConnectivity relay for when the watch can't reach the server directly

### Xcode Project
- MitzoWatch target: watchOS 10.0, bundle ID `com.mitzo.app.watchkitapp`
- 9 Swift source files wired into Sources build phase
- Code signing, entitlements, deployment target all configured

### WatchConnectivity Relay
- **WatchRelayHost** (iOS): bridges watch messages to/from the active WS connection
- **WatchRelayClient** (watchOS): sends through iPhone when direct WS unavailable
- Auth token sharing: watch can request JWT from phone's Keychain
- **Waterfall connection**: direct WS (5s timeout) → relay via iPhone → error

### What remains manual
- Open Xcode once to let it resolve signing/provisioning
- First build may need to select a watch simulator
- The `setup_watch_target.rb` script (not committed) can re-run if pbxproj is regenerated

## Test plan

- [x] `swift build` and `swift test` pass for MitzoShared
- [ ] Open in Xcode, verify MitzoWatch target appears in scheme selector
- [ ] Build for watchOS simulator
- [ ] Test direct WS connection
- [ ] Test relay fallback (disconnect WiFi on watch)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #297, opened 2026-04-27
fix(watch): review round 2 — reconnect guard, relay errors, full coverage

dimakis:
## Summary

Addresses all 7 findings from the centaur review on PR #296 (post-merge).

- **Reconnect guard**: `isReconnecting` flag prevents duplicate `MitzoWSClient` instances from concurrent `reconnectWithBackoff()` tasks
- **Relay error checking**: `sendMessage` in relay mode now inspects the reply dict for `error` key and throws `RelayResponseError` instead of silently discarding failures
- **Complete relay dict**: `clientMessageToRelayDict` handles all `ClientMessage` cases (`.hello`, `.interrupt`, `.setMode`, `.switchSession`, `.unwatch`, `.reconnect`, `.sessionSuspend`) — no more `default → "unknown"` silent drops
- **Nil-safe dict insertion**: Replaced `params.sessionId as Any` (inserts `Optional.none` that `JSONSerialization` rejects) with conditional `if let` insertion
- **WCSession error handler**: `forwardToWatch` now passes an `errorHandler` to `WCSession.sendMessage` to catch payload-too-large and delivery failures
- **xcconfig for watch**: MitzoWatch Debug config now references `debug.xcconfig` as `baseConfigurationReference` so `local.xcconfig` can override `DEVELOPMENT_TEAM`
- **Full round-trip test coverage**: Added 10 tests for remaining `ServerMessage` cases (`reconnected`, `watched`, `unwatched`, `sessionSwitched`, `sessionCleared`, `sessionResumed`, `sessionTakeover`, `sessionEnd`, `error`, `unknown`). Total: 41 tests passing.

## Test plan

- [x] `swift build` passes
- [x] `swift test` — 41/41 tests pass
- [ ] CI green

### DIFF
--- a/frontend/ios/App/App.xcodeproj/project.pbxproj
+++ b/frontend/ios/App/App.xcodeproj/project.pbxproj
@@ -326,6 +326,7 @@
 /* Begin XCBuildConfiguration section */
 		16EF91AC41A8DC5F7DE4223C /* Debug */ = {
 			isa = XCBuildConfiguration;
+			baseConfigurationReference = 958DCC722DB07C7200EA8C5F /* debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_ENTITLEMENTS = "$(SRCROOT)/../MitzoWatch/MitzoWatch.entitlements";
--- a/frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift
+++ b/frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift
@@ -65,14 +65,23 @@ public final class WatchRelayHost: NSObject, WCSessionDelegate, Sendable {
         }
     }
 
-    /// Forward server messages to the watch
+    /// Forward server messages to the watch.
+    /// WCSession has an undocumented ~64 KB per-message limit; large payloads
+    /// (e.g. file-read block_delta) are silently dropped by the system.
     public func forwardToWatch(_ message: ServerMessage) {
         guard WCSession.default.isReachable else { return }
 
         do {
             let data = try JSONEncoder().encode(message)
             if let dict = try JSONSerialization.jsonObject(with: data) as? [String: Any] {
-                WCSession.default.sendMessage(["_relay": "server_event", "_payload": dict], replyHandler: nil)
+                WCSession.default.sendMessage(
+                    ["_relay": "server_event", "_payload": dict],
+                    replyHandler: nil,
+                    errorHandler: { _ in
+                        // WCSession delivery failure (payload too large, watch unreachable, etc.)
+                        // Non-fatal: the watch will recover via seq-based replay on reconnect.
+                    }
+                )
             }
         } catch {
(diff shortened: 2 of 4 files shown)


dimakis:
## Centaur Review

### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift`
_Solid hardening PR — reconnect guard, exhaustive relay encoding, error propagation, and good test coverage. Main concern is the relay decode asymmetry: the iPhone-side WatchRelayHost.decodeClientMessage now lags behind the watch-side encoder by 5 ClientMessage cases._

- 🟡 **regressions** (L145): Asymmetry: AppState.clientMessageToRelayDict now exhaustively encodes all 11 ClientMessage cases (including unwatch, switchSession, interrupt, setMode, reconnect), but the iPhone-side WatchRelayHost.decodeClientMessage only handles 6. Watch-originated relay messages for the 5 missing actions will hit the default case and throw RelayError.unknownAction, causing the watch to receive an error reply. This was pre-existing but is now exposed by making the send side exhaustive. `[fixable]`


### `frontend/ios/MitzoWatch/Services/AppState.swift`
_Solid hardening PR — reconnect guard, exhaustive relay encoding, error propagation, and good test coverage. Main concern is the relay decode asymmetry: the iPhone-side WatchRelayHost.decodeClientMessage now lags behind the watch-side encoder by 5 ClientMessage cases._

- 🔵 **bugs** (L233): Narrow race in reconnect guard: if connect() succeeds inside the reconnect Task but the WS drops immediately before isReconnecting=false is reached, handleWSEvent will call reconnectWithBackoff() which bails on the guard. When isReconnecting is then set to false, no reconnect is pending — connection stays dead. Consider resetting isReconnecting before calling connect(), or re-checking connection state after the flag is cleared. `[fixable]`


### `frontend/ios/MitzoShared/Tests/MitzoSharedTests/MitzoSharedTests.swift`
_Solid hardening PR — reconnect guard, exhaustive relay encoding, error propagation, and good test coverage. Main concern is the relay decode asymmetry: the iPhone-side WatchRelayHost.decodeClientMessage now lags behind the watch-side encoder by 5 ClientMessage cases._

- 🔵 **missing_tests**: Round-trip tests were added for 11 ServerMessage types (good coverage of lifecycle/control messages), but sessionId, messageStart, blockEnd, messageEnd, tokenUpdate, and modeChanged still lack round-trip tests. The streaming block types (blockStart, blockDelta, toolResult, permissionRequest) have existing decode-only tests but no encode round-trip. Low priority since the new tests cover the relay-critical message types. `[fixable]`


### `frontend/ios/App/App.xcodeproj/project.pbxproj`
_Solid hardening PR — reconnect guard, exhaustive relay encoding, error propagation, and good test coverage. Main concern is the relay decode asymmetry: the iPhone-side WatchRelayHost.decodeClientMessage now lags behind the watch-side encoder by 5 ClientMessage cases._

- 🔵 **style** (L329): The debug.xcconfig file referenced here (path ../debug.xcconfig relative to SOURCE_ROOT) does not exist in the repo. Other build configurations already reference it (lines 356, 465), so this is a consistency fix for a gitignored local config file — just worth confirming the file is documented in onboarding or generated by a setup script.

PR #298, opened 2026-04-27
fix(watch): annotate WCSession errorHandler as @Sendable

dimakis:
## Summary

- Swift 6 strict concurrency flags the `errorHandler` closure passed to `WCSession.sendMessage` as a data race risk
- Annotate with `@Sendable` to satisfy the `sending` parameter requirement

One-liner fix for the Xcode build error in `WatchRelay.swift`.

## Test plan

- [x] `swift build` passes
- [ ] CI green
- [ ] Xcode builds without warnings

### DIFF
--- a/frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift
+++ b/frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift
@@ -3,6 +3,14 @@
 // iPhone side: receives relay messages from watch, forwards to/from WS
 // Watch side: sends messages via WCSession when direct WS is unavailable
 
+/// Wraps a non-Sendable value for use across isolation boundaries.
+/// Safety: WCSession's replyHandler is called exactly once; we transfer
+/// ownership into the Task and never alias it.
+struct UnsafeSendable<T>: @unchecked Sendable {
+    let value: T
+    init(_ value: T) { self.value = value }
+}
+
 #if os(iOS)
 import WatchConnectivity
 import Foundation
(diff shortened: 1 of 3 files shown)


frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift
```diff
@@ -77,10 +77,7 @@ public final class WatchRelayHost: NSObject, WCSessionDelegate, Sendable {
                 WCSession.default.sendMessage(
                     ["_relay": "server_event", "_payload": dict],
                     replyHandler: nil,
-                    errorHandler: { _ in
-                        // WCSession delivery failure (payload too large, watch unreachable, etc.)
-                        // Non-fatal: the watch will recover via seq-based replay on reconnect.
-                    }
+                    errorHandler: { @Sendable _ in }
```
dimakis: 🔵 **style**: The removed comment explained a non-obvious recovery mechanism: errors are silently dropped because the watch recovers via seq-based replay on reconnect. Per the project's own convention ('Only add a comment when the WHY is non-obvious'), this qualified. Consider keeping a shorter version, e.g. `// Non-fatal: watch recovers via seq replay`. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/WatchRelay.swift`
_Clean compiler fix — adds @Sendable to satisfy Swift 6 concurrency. Only concern is the loss of a useful explanatory comment._

- 🔵 **style** (L80): The removed comment explained a non-obvious recovery mechanism: errors are silently dropped because the watch recovers via seq-based replay on reconnect. Per the project's own convention ('Only add a comment when the WHY is non-obvious'), this qualified. Consider keeping a shorter version, e.g. `// Non-fatal: watch recovers via seq replay`. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #295, opened 2026-04-26
fix(client): prevent UI freeze from foreground state wipe + 3 related bugs

dimakis:
## Summary

- **Foreground state wipe** — when iOS briefly backgrounds/foregrounds the app, `_foreground` fires and `fetchAndRestoreMessages` queries REST. If it returns `[]` (timing gap), all messages + streaming state were wiped (`current: null`). Subsequent `block_delta` events silently dropped → UI frozen. Fix: preserve existing state when REST returns empty.
- **`pendingSend` single-slot** — rapid messages while a turn was running overwrote each other. Only the last survived. Fix: changed to FIFO array, `session_end` dequeues one at a time.
- **Error-path `session_end` not persisted** — when the query loop exited via `finally` (errors/aborts), `session_end` was sent directly bypassing EventStore. On reconnect it wasn't replayed → `running` stuck at true. Fix: use `sendOrBuffer()`.
- **`reconnected` ignored server `running` state** — the server sends authoritative `running: false` in the `reconnected` response but the client ignored it. Fix: apply it for the active session.

## Test plan

- [x] Existing 100 tests pass (protocol-parser + store)
- [x] New regression test: `preserves streaming state when REST returns empty array`
- [x] Updated test: `session_end dequeues first pending send` (verifies FIFO)
- [x] Updated test: `error clears pendingSend queue` (verifies array clear)
- [x] 8 pre-existing failures on main, zero new failures introduced

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #300, opened 2026-04-28
fix(client): add runtime validation for reconnected sessions field

dimakis:
## Summary

- Validates `sessions` field in `reconnected` message has the expected array shape before accessing properties
- Guards against `undefined`/missing `running` field with explicit `=== false` check (prevents incorrect `SET_RUNNING false` dispatch from older servers)
- Adds test for invalid shape handling

Follow-up to PR #295 — addresses the second Centaur review round's unsafe type cast finding.

## Test plan

- [x] Protocol parser tests pass (47/47)
- [x] Existing reconnected tests still pass (4 cases)
- [x] New test covers invalid shape no-op behavior

### DIFF
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -115,11 +115,24 @@ export function parseServerMessage(
 
     case 'reconnected': {
       result.connectionUpdate = { status: 'connected' };
-      // Apply authoritative running state from the server for the active session
-      const sessions = msg.sessions as Array<{ sessionId: string; running: boolean }> | undefined;
-      if (sessions && state.currentSessionId) {
+      // Apply authoritative running state from the server for the active session.
+      // Validate runtime shape: sessions must be an array, and each entry must have
+      // sessionId (string) and running (boolean). Explicit running === false check
+      // guards against undefined/missing field.
+      const sessions = msg.sessions as unknown;
+      if (
+        Array.isArray(sessions) &&
+        state.currentSessionId &&
+        sessions.every(
+          (s): s is { sessionId: string; running: boolean } =>
+            typeof s === 'object' &&
+            s !== null &&
+            typeof s.sessionId === 'string' &&
+            typeof s.running === 'boolean',
+        )
+      ) {
         const active = sessions.find((s) => s.sessionId === state.currentSessionId);
-        if (active && !active.running) {
+        if (active && active.running === false) {
           result.messagesActions.push({ type: 'SET_RUNNING', running: false });
         }
       }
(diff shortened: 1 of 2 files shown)


packages/client/src/protocol-parser.ts
```diff
@@ -115,11 +115,24 @@ export function parseServerMessage(
...
-      if (sessions && state.currentSessionId) {
+      // Apply authoritative running state from the server for the active session.
+      // Validate runtime shape: sessions must be an array, and each entry must have
+      // sessionId (string) and running (boolean). Explicit running === false check
+      // guards against undefined/missing field.
+      const sessions = msg.sessions as unknown;
+      if (
+        Array.isArray(sessions) &&
+        state.currentSessionId &&
+        sessions.every(
```
dimakis, line 126: 🔵 **regressions**: The `sessions.every()` validation rejects the entire array if any single entry is malformed, whereas the old code would still find and act on valid entries. If the server ever sends a mix (e.g. one well-formed entry for the active session and one malformed entry for another), the new code silently ignores a valid `running: false` signal. This is a defensible design choice (distrust the whole message if any part is bad), but it is a behavioral change worth being intentional about. Consider using `.filter()` to keep only valid entries instead of `.every()` to reject all-or-nothing. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -115,11 +115,24 @@ export function parseServerMessage(
 
     case 'reconnected': {
       result.connectionUpdate = { status: 'connected' };
-      // Apply authoritative running state from the server for the active session
-      const sessions = msg.sessions as Array<{ sessionId: string; running: boolean }> | undefined;
-      if (sessions && state.currentSessionId) {
+      // Apply authoritative running state from the server for the active session.
```
dimakis, line 118: 🔵 **style**: The 4-line comment block restates what the code already makes clear through its structure (Array.isArray, .every with type guards, === false). Per project conventions ("default to writing no comments", "don't explain WHAT the code does"), this could be reduced to a single line explaining *why* runtime validation is needed here (e.g. `// Server shape is untyped — validate before trusting`) or removed entirely. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Solid defensive fix with good test coverage; minor suggestion to consider `.filter()` over `.every()` to avoid silently dropping valid entries when one sibling is malformed._

- 🔵 **regressions** (L126): The `sessions.every()` validation rejects the entire array if any single entry is malformed, whereas the old code would still find and act on valid entries. If the server ever sends a mix (e.g. one well-formed entry for the active session and one malformed entry for another), the new code silently ignores a valid `running: false` signal. This is a defensible design choice (distrust the whole message if any part is bad), but it is a behavioral change worth being intentional about. Consider using `.filter()` to keep only valid entries instead of `.every()` to reject all-or-nothing. `[fixable]`
- 🔵 **style** (L118): The 4-line comment block restates what the code already makes clear through its structure (Array.isArray, .every with type guards, === false). Per project conventions ("default to writing no comments", "don't explain WHAT the code does"), this could be reduced to a single line explaining *why* runtime validation is needed here (e.g. `// Server shape is untyped — validate before trusting`) or removed entirely. `[fixable]`

PR #303, opened 2026-04-30
feat(observability): record agent output in traces and logs

dimakis:
## Summary

- Adds thinking blocks, text blocks, tool inputs, and tool results to OTel span events (visible in Jaeger)
- Adds info-level log lines for all agent output (visible in Loki/server logs)
- Introduces `TRACE_CONTENT_MAX_CHARS` (16KB) constant to cap payload size

## Test plan

- [x] All 43 query-loop tests pass
- [ ] Deploy, run a session, verify thinking/text/tool content appears in Jaeger spans
- [ ] Check server logs for new `thinking block complete`, `text block complete`, `tool result` entries

### DIFF
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -50,8 +50,12 @@ function createTestSpan(name: string): TestSpan & OTelSpan {
     recordException: () => {},
     spanContext: () => ({ traceId: 'test', spanId: 'test', traceFlags: 1 }),
     // Expose internals for assertions
-    get _events() { return span.events; },
-    get _name() { return span.name; },
+    get _events() {
+      return span.events;
+    },
+    get _name() {
+      return span.name;
+    },
   } as unknown as TestSpan & OTelSpan;
 }
 
@@ -1806,7 +1810,11 @@ describe('runQueryLoop', () => {
         },
         {
           type: 'stream_event',
-          event: { type: 'content_block_delta', index: 0, delta: { type: 'text_delta', text: 'Hello world' } },
+          event: {
+            type: 'content_block_delta',
+            index: 0,
+            delta: { type: 'text_delta', text: 'Hello world' },
+          },
         },
         { type: 'stream_event', event: { type: 'content_block_stop', index: 0 } },
         { type: 'assistant', message: { content: [] }, session_id: 'sess-text' },
@@ -1832,7 +1840,11 @@ describe('runQueryLoop', () => {
         },
         {
           type: 'stream_event',
-          event: { type: 'content_block_delta', index: 0, delta: { type: 'thinking_delta', thinking: 'Let me reason...' } },
+          event: {
+            type: 'content_block_delta',
+            index: 0,
+            delta: { type: 'thinking_delta', thinking: 'Let me reason...' },
+          },
         },
         { type: 'stream_event', event: { type: 'content_block_stop', index: 0 } },
         { type: 'assistant', message: { content: [] }, session_id: 'sess-think' },
@@ -1907,7 +1919,12 @@ describe('runQueryLoop', () => {
           type: 'user',
           message: {
             content: [
-              { type: 'tool_result', tool_use_id: 'tool-2', content: 'file contents here', is_error: false },
+              {
+                type: 'tool_result',
+                tool_use_id: 'tool-2',
+                content: 'file contents here',
+                is_error: false,
+              },
             ],
           },
         },
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -4,6 +4,69 @@ import { ConnectionRegistry } from '../../packages/harness/src/connection-regist
 import { runQueryLoop } from '../query-loop.js';
 import type { SessionRegistry } from '../session-registry.js';
 import { EventStore } from '../event-store.js';
+import type { Span as OTelSpan } from '@opentelemetry/api';
+
+/** Lightweight in-memory span for testing OTel instrumentation. */
+interface RecordedEvent {
+  name: string;
+  attributes?: Record<string, unknown>;
+}
+interface TestSpan {
+  name: string;
+  attributes: Record<string, unknown>;
+  events: RecordedEvent[];
+  status: { code: number; message?: string };
+  ended: boolean;
+}
+
+function createTestSpan(name: string): TestSpan & OTelSpan {
+  const span: TestSpan = { name, attributes: {}, events: [], status: { code: 0 }, ended: false };
+  return {
+    ...span,
+    setAttribute(k: string, v: unknown) {
+      span.attributes[k] = v;
+      return this;
+    },
+    setAttributes(attrs: Record<string, unknown>) {
+      Object.assign(span.attributes, attrs);
+      return this;
+    },
+    addEvent(eName: string, attrs?: Record<string, unknown>) {
+      span.events.push({ name: eName, attributes: attrs });
+      return this;
+    },
+    setStatus(s: { code: number; message?: string }) {
+      span.status = s;
+      return this;
+    },
+    end() {
+      span.ended = true;
+    },
+    updateName(n: string) {
+      span.name = n;
+      return this;
+    },
+    isRecording: () => true,
+    recordException: () => {},
+    spanContext: () => ({ traceId: 'test', spanId: 'test', traceFlags: 1 }),
+    // Expose internals for assertions
+    get _events() { return span.events; },
+    get _name() { return span.name; },
+  } as unknown as TestSpan & OTelSpan;
+}
+
+/** Collects all spans created by the mocked tracer. */
+const recordedSpans: (TestSpan & OTelSpan)[] = [];
+
+vi.mock('../tracing.js', () => ({
+  tracer: {
+    startSpan(name: string) {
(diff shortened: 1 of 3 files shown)


server/query-loop.ts
```diff
@@ -698,11 +699,23 @@ async function _runQueryLoopInner(
               const summarized = summarizeToolInput(toolEntry.name, toolInput);
               const rawInput = getRawInput(toolEntry.name, toolInput);
 
-              log.info('tool call', { clientId, tool: toolEntry.name, toolId: toolEntry.id });
+              log.info('tool call', {
+                clientId,
+                tool: toolEntry.name,
+                toolId: toolEntry.id,
+                input: toolEntry.inputBuf.slice(0, TRACE_CONTENT_MAX_CHARS),
```
dimakis: 🟡 **style**: The raw `inputBuf` is logged at `info` level (up to 16 KB per tool call). With chatty sessions that invoke many tools, this can produce significant log volume on disk and in Loki. Consider using `debug` level for the full content fields (`input`, `content`, `result`) and keeping `info` for the metadata-only lines, or at least gating the content behind a separate log level / env flag. The same applies to lines 778-784 (text/thinking content) and 831-837 (tool result). `[fixable]`

server/query-loop.ts
```diff
@@ -698,11 +699,23 @@ async function _runQueryLoopInner(
...
+              });
 
-              // End tool span
+              // End tool span — record raw input before closing
               const toolSpan = toolSpans.get(blockId);
               if (toolSpan) {
+                toolSpan.addEvent('tool.input', {
```
dimakis, line 715: 🔵 **style**: The truncation + `truncated` flag pattern is repeated three times (tool input at line 712, block content at line 770, tool result at line 823) with identical structure. Consider a small helper like `truncateWithFlag(text: string, max: number): { text: string; truncated?: true }` to DRY these up. This would also make it easier to change the strategy (e.g., adding tail bytes) later. `[fixable]`

server/constants.ts
```diff
@@ -46,3 +46,8 @@ export const WORKTREE_CLEANUP_INTERVAL_MS = 6 * 60 * 60 * 1000; // 6 hours
 // as unreachable (e.g. model unavailable on the configured provider) and
 // surface an error to the client instead of hanging forever.
 export const QUERY_FIRST_EVENT_TIMEOUT_MS = 90_000;
+
+// --- Observability ---
+// Max characters for agent content recorded in OTel span events and log lines.
+// Keeps Jaeger/Loki payloads bounded while preserving enough for debugging.
+export const TRACE_CONTENT_MAX_CHARS = 16_384;
```
dimakis: 🔵 **unsafe_assumptions**: 16 KB is a reasonable default, but there's no way to tune it at runtime (e.g., via env var). If a user needs to reduce log volume or increase it for a debugging session, they'd need a code change. Consider `parseInt(process.env.TRACE_CONTENT_MAX_CHARS ?? '') || 16_384` for zero-cost configurability. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid observability improvement with correct null guards and safe string handling. Main concerns are log volume at `info` level (up to 16 KB per event), missing test coverage for the new span events, and repeated truncation boilerplate._

- 🟡 **style** (L706): The raw `inputBuf` is logged at `info` level (up to 16 KB per tool call). With chatty sessions that invoke many tools, this can produce significant log volume on disk and in Loki. Consider using `debug` level for the full content fields (`input`, `content`, `result`) and keeping `info` for the metadata-only lines, or at least gating the content behind a separate log level / env flag. The same applies to lines 778-784 (text/thinking content) and 831-837 (tool result). `[fixable]`
- 🟡 **missing_tests**: The existing query-loop test suite (1700+ lines) does not assert on any OTel span events or verify that log lines include the new content fields. Since the PR's stated purpose is to record agent output in traces and logs, at least one test per new event type (`tool.input`, `block.text`/`block.thinking`, `tool.result`) should verify the event name and key attributes are emitted. The test already sets up mock spans — extending coverage is straightforward. `[fixable]`
- 🔵 **style** (L712): The truncation + `truncated` flag pattern is repeated three times (tool input at line 712, block content at line 770, tool result at line 823) with identical structure. Consider a small helper like `truncateWithFlag(text: string, max: number): { text: string; truncated?: true }` to DRY these up. This would also make it easier to change the strategy (e.g., adding tail bytes) later. `[fixable]`


### `server/constants.ts`
_Solid observability improvement with correct null guards and safe string handling. Main concerns are log volume at `info` level (up to 16 KB per event), missing test coverage for the new span events, and repeated truncation boilerplate._

- 🔵 **unsafe_assumptions** (L53): 16 KB is a reasonable default, but there's no way to tune it at runtime (e.g., via env var). If a user needs to reduce log volume or increase it for a debugging session, they'd need a code change. Consider `parseInt(process.env.TRACE_CONTENT_MAX_CHARS ?? '') || 16_384` for zero-cost configurability. `[fixable]`

server/constants.ts
```diff
@@ -46,3 +46,10 @@ export const WORKTREE_CLEANUP_INTERVAL_MS = 6 * 60 * 60 * 1000; // 6 hours
 // as unreachable (e.g. model unavailable on the configured provider) and
 // surface an error to the client instead of hanging forever.
 export const QUERY_FIRST_EVENT_TIMEOUT_MS = 90_000;
+
+// --- Observability ---
+// Max characters for agent content recorded in OTel span events and log lines.
+// Keeps Jaeger/Loki payloads bounded while preserving enough for debugging.
+// Configurable via env var for runtime tuning without code changes.
+export const TRACE_CONTENT_MAX_CHARS =
```
dimakis, line 54: 🔵 **unsafe_assumptions**: Negative env var values are accepted silently. `parseInt("-1", 10)` returns `-1` (truthy, so the `|| 16_384` fallback doesn't fire), causing `text.slice(0, -1)` in `truncateForTrace` which silently chops the last character instead of truncating to a limit. Consider `Math.max(1, parseInt(..., 10)) || 16_384` to reject non-positive values. `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -1728,4 +1791,140 @@ describe('runQueryLoop', () => {
       vi.useRealTimers();
     });
   });
+
+  describe('observability — span events for agent output', () => {
```
dimakis, line 1795: 🔵 **missing_tests**: No test exercises the truncation path. All four test cases use short content that fits within TRACE_CONTENT_MAX_CHARS, so the `truncated` flag is never asserted as present (only asserted as absent in the first test). A test that sets a low TRACE_CONTENT_MAX_CHARS (or sends content exceeding it) would cover the `block.truncated` / `tool.input.truncated` / `tool.result.truncated` attributes and the `.slice()` logic. `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -4,6 +4,69 @@ import { ConnectionRegistry } from '../../packages/harness/src/connection-regist
...
+    },
+    updateName(n: string) {
+      span.name = n;
+      return this;
+    },
+    isRecording: () => true,
+    recordException: () => {},
+    spanContext: () => ({ traceId: 'test', spanId: 'test', traceFlags: 1 }),
+    // Expose internals for assertions
+    get _events() { return span.events; },
```
dimakis, line 53: 🔵 **style**: The `_events` and `_name` getters on the test span object are unused dead code — all assertions access `.events` and `.name` directly (which work correctly via the spread reference copy). Consider removing them to keep the test helper minimal. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


### `server/constants.ts`
_Clean, well-structured observability addition. Span event placement is correct (text/thinking on turn span, tool input on tool span, tool result on session span), content/metadata log level split is consistent with existing patterns, and truncation helper is sound. Three minor suggestions: guard against negative env var values, add a truncation-path test, and remove unused test helper getters._

- 🔵 **unsafe_assumptions** (L54): Negative env var values are accepted silently. `parseInt("-1", 10)` returns `-1` (truthy, so the `|| 16_384` fallback doesn't fire), causing `text.slice(0, -1)` in `truncateForTrace` which silently chops the last character instead of truncating to a limit. Consider `Math.max(1, parseInt(..., 10)) || 16_384` to reject non-positive values. `[fixable]`


_Clean, well-structured observability addition. Span event placement is correct (text/thinking on turn span, tool input on tool span, tool result on session span), content/metadata log level split is consistent with existing patterns, and truncation helper is sound. Three minor suggestions: guard against negative env var values, add a truncation-path test, and remove unused test helper getters._

- 🔵 **missing_tests** (L1795): No test exercises the truncation path. All four test cases use short content that fits within TRACE_CONTENT_MAX_CHARS, so the `truncated` flag is never asserted as present (only asserted as absent in the first test). A test that sets a low TRACE_CONTENT_MAX_CHARS (or sends content exceeding it) would cover the `block.truncated` / `tool.input.truncated` / `tool.result.truncated` attributes and the `.slice()` logic. `[fixable]`
- 🔵 **style** (L53): The `_events` and `_name` getters on the test span object are unused dead code — all assertions access `.events` and `.name` directly (which work correctly via the spread reference copy). Consider removing them to keep the test helper minimal. `[fixable]`

PR #304, opened 2026-05-01
feat(sse): event bus + session activity overview

dimakis:
### Phase 1 — SSE Infrastructure
- **SseRegistry** (harness): server-side SSE connection manager with heartbeat (30s), broadcast to all clients, per-client `sendTo` for hydration, and graceful cleanup
- **GET /api/events** route: SSE endpoint with auth, `connected` event on hydration, proper headers (`text/event-stream`, `X-Accel-Buffering: no`)
- **Broadcast wiring**: all 6 server broadcast callbacks (`update_available`, `inbox_updated`, `task_state` x2, `todo_update`, `loop_status`) now also push to SSE
- **EventBus** (client): typed EventSource wrapper with `on(event, listener)` dispatch, `ensureConnected()` for iOS resume recovery, connection state tracking

### Phase 2 — Session Activity Overview
- **SessionOverviewEmitter** (server): derives session state (`working`, `waiting`, `done`, `idle`, `paused`) from registry + permissions + task store. Broadcasts via SSE with 200ms coalescing to prevent re-render storms
- **getPendingCountBySession** (harness): new permissions API to query pending permission count per session
- **Session change callbacks**: wired into chat lifecycle (start/end/turn_end) and query-loop for real-time state updates
- **useSessionOverview** hook (frontend): subscribes to `session_activity` SSE events, sorts by attend tier (waiting → done → working), computes `attendCount`
- **SessionOverview** component: collapsible card section on home page — auto-expands when sessions need attention, shows state icons, repo prefix, progress, elapsed time, red badge for attend count. Tap navigates to session

State is derived lazily at broadcast time — no stored state, no client-side timers. Server owns all timeout logic (Done → Idle after 5min).

Design doc: `mgmt/local_features/session-overview-widget/sse-event-bus.md`

## Test plan

- [x] 15 EventBus unit tests (Phase 1)
- [x] 20 SessionOverviewEmitter unit tests (state derivation, coalescing, timeouts, permissions, ATB progress, repo extraction)
- [x] 6 useSessionOverview hook tests (subscribe/unsubscribe, tier sorting, attend count)
- [x] 8 SessionOverview component tests (render states, badge, auto-expand, navigation, toggle)
- [x] Full test suite: 2351 passed, 2 failed (pre-existing timeout in chat.test.ts)
- [ ] Manual: open home page → see active sessions widget
- [ ] Manual: start a session → card appears with "working" state
- [ ] Manual: trigger permission prompt → card switches to "waiting" with badge
- [ ] Manual: iOS background/foreground → SSE reconnects, widget updates

### DIFF
--- a/frontend/src/client-store.ts
+++ b/frontend/src/client-store.ts
@@ -12,6 +12,7 @@ import { apiFetch, getApiBaseUrl, getWsChatUrl } from './lib/api-fetch';
 import { registerCapacitorLifecycle } from './lib/capacitor';
 import { configureKeyboard } from './lib/keyboard';
 import { initPushNotifications } from './lib/push';
+import { eventBus } from './lib/event-bus-singleton';
 
 export const clientStore = createMitzoStore({
   transport: {
@@ -27,7 +28,10 @@ export const clientStore = createMitzoStore({
 
 // Wire Capacitor app lifecycle → force WS reconnect on resume, send suspend on background
 registerCapacitorLifecycle(
-  () => clientStore.getState().forceReconnect(),
+  () => {
+    clientStore.getState().forceReconnect();
+    eventBus.ensureConnected();
+  },
   () => clientStore.getState().sendSuspend(),
 );
 
--- a/frontend/src/components/SessionOverview.tsx
+++ b/frontend/src/components/SessionOverview.tsx
@@ -28,6 +28,7 @@ function SessionActivityCard({
   onTap: (sessionId: string) => void;
 }) {
   const config = STATE_CONFIG[activity.state];
+  // Elapsed updates on each SSE event, not on a timer — acceptable staleness
   const elapsed = Date.now() - activity.lastEventAt;
   const timeLabel = formatElapsed(elapsed);
 
--- a/frontend/src/components/SessionOverview.tsx
+++ b/frontend/src/components/SessionOverview.tsx
@@ -1,0 +1,133 @@
+import { useState, useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import {
+  useSessionOverview,
+  type SessionActivity,
+  type SessionActivityState,
+} from '../hooks/useSessionOverview';
+import { selectionChanged } from '../lib/haptics';
+
+// ─── State visuals ──────────────────────────────────────────────────────────
+
+const STATE_CONFIG: Record<SessionActivityState, { icon: string; color: string; label: string }> = {
+  init: { icon: '\u25CB', color: '#888', label: 'init' },
+  working: { icon: '\u25CF', color: '#b48cff', label: 'working' },
+  waiting: { icon: '\u26A0', color: '#ff6d6d', label: 'waiting' },
+  done: { icon: '\u2713', color: '#4ade80', label: 'done' },
+  idle: { icon: '\u25CB', color: '#555', label: 'idle' },
+  paused: { icon: '\u23F8', color: '#888', label: 'paused' },
+};
+
+// ─── Card ───────────────────────────────────────────────────────────────────
+
+function SessionActivityCard({
+  activity,
+  onTap,
+}: {
+  activity: SessionActivity;
+  onTap: (sessionId: string) => void;
+}) {
+  const config = STATE_CONFIG[activity.state];
+  const elapsed = Date.now() - activity.lastEventAt;
+  const timeLabel = formatElapsed(elapsed);
+
+  let metaText = config.label;
+  if (activity.waitReason === 'permission') metaText = 'permission';
+  else if (activity.waitReason === 'review') metaText = 'review needed';
+  else if (activity.waitReason === 'blocked') metaText = 'blocked';
+
+  if (activity.progress) {
+    metaText += ` \u00B7 ${activity.progress.done}/${activity.progress.total}`;
+  }
+
+  metaText += ` \u00B7 ${timeLabel}`;
+
+  return (
+    <button
+      className="overview-card"
+      onClick={() => onTap(activity.sessionId)}
+      style={{ '--card-accent': config.color } as React.CSSProperties}
+    >
+      <span className="overview-card-icon" style={{ color: config.color }}>
+        {config.icon}
+      </span>
+      <div className="overview-card-content">
+        <div className="overview-card-title">
+          {activity.repo && <span className="overview-card-repo">{activity.repo}:</span>}{' '}
+          {activity.title}
+        </div>
+        <div className="overview-card-meta">{metaText}</div>
+      </div>
+    </button>
+  );
+}
+
+// ─── Main component ─────────────────────────────────────────────────────────
+
+export function SessionOverview() {
+  const { activities, attendCount } = useSessionOverview();
+  const navigate = useNavigate();
+
+  // Filter out idle/init — only show interesting sessions
+  const visible = activities.filter((a) => a.state !== 'idle' && a.state !== 'init');
+
+  // Auto-expand when tier 1 items exist
+  const hasUrgent = attendCount > 0;
+  const [manualOpen, setManualOpen] = useState<boolean | null>(null);
+  const isOpen = manualOpen ?? hasUrgent;
+
+  const toggleOpen = useCallback(() => {
+    setManualOpen((prev) => !(prev ?? hasUrgent));
+  }, [hasUrgent]);
+
+  const handleTap = useCallback(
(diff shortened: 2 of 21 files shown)


dimakis:
## Centaur Review

### `packages/client/src/event-bus.ts`
_Solid SSE infrastructure with clean separation between client and server. Two issues to fix before merge: a duplicate-dispatch bug in EventBus when listeners are removed and re-added for the same event type, and missing `sseRegistry.destroy()` in the server shutdown path. SseRegistry also needs tests per the project's TDD requirement._

- 🟡 **bugs** (L62): Duplicate dispatch after unsubscribe/resubscribe cycle. When the last listener for an event type is removed (line 64), the Set is deleted from the map, but the EventSource handler registered via `registerEventDispatch` is never removed (anonymous closure, no reference saved). On the next `on()` call for the same event type, a new Set is created and `registerEventDispatch` adds a *second* handler to the EventSource. Both handlers resolve the same Set, so every listener fires twice per event. Each unsub/resub cycle adds another duplicate. Fix: either track the dispatch handler reference and remove it, or don't delete the event key from the map (just empty the Set). `[fixable]`
- 🔵 **style** (L1): Multi-paragraph doc comment (11 lines). Project convention per CLAUDE.md: 'default to writing no comments. Only add one when the WHY is non-obvious' and 'Never write multi-paragraph docstrings or multi-line comment blocks — one short line max.' The class-level block and most method comments exceed this. `[fixable]`


### `server/index.ts`
_Solid SSE infrastructure with clean separation between client and server. Two issues to fix before merge: a duplicate-dispatch bug in EventBus when listeners are removed and re-added for the same event type, and missing `sseRegistry.destroy()` in the server shutdown path. SseRegistry also needs tests per the project's TDD requirement._

- 🟡 **bugs** (L785): `sseRegistry.destroy()` is never called in `shutdown()`. The SseRegistry has a `destroy()` method that closes all client connections and stops the heartbeat interval, but the shutdown function (line 785-795) only cleans up WebSocket clients, the server, and other stores. This leaks the heartbeat timer and leaves SSE connections unclosed on graceful shutdown. `[fixable]`


### `packages/harness/src/sse-registry.ts`
_Solid SSE infrastructure with clean separation between client and server. Two issues to fix before merge: a duplicate-dispatch bug in EventBus when listeners are removed and re-added for the same event type, and missing `sseRegistry.destroy()` in the server shutdown path. SseRegistry also needs tests per the project's TDD requirement._

- 🟡 **missing_tests**: No test file for SseRegistry. The project follows TDD (per CLAUDE.md) and every other registry in `packages/harness/__tests__/` has tests (connection-registry, session-registry). SseRegistry has broadcast, sendTo, heartbeat start/stop, and destroy logic that should be covered. `[fixable]`
- 🔵 **unsafe_assumptions** (L124): Heartbeat silently swallows write failures without cleaning up dead connections (line 130: `// will be cleaned up on next broadcast`). If broadcasts are infrequent, dead connections accumulate and the heartbeat keeps writing to them every 30s. Consider collecting failures during heartbeat and calling `remove()` for them, matching the pattern already used in `broadcast()`. `[fixable]`


### `packages/client/__tests__/event-bus.test.ts`
_Solid SSE infrastructure with clean separation between client and server. Two issues to fix before merge: a duplicate-dispatch bug in EventBus when listeners are removed and re-added for the same event type, and missing `sseRegistry.destroy()` in the server shutdown path. SseRegistry also needs tests per the project's TDD requirement._

- 🔵 **missing_tests**: No test for the duplicate-dispatch scenario (subscribe → unsubscribe → resubscribe for the same event type, then verify the listener fires exactly once). This is the untested path that masks the bug identified above. `[fixable]`

dimakis:
## Centaur Review

### `packages/harness/src/permissions.ts`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🟡 **bugs** (L65): The JSDoc for `denyPendingBySession` (lines 65-68) is orphaned — `getPendingCountBySession` was inserted between it and the function it documents (line 81). The result is two stacked JSDoc comments: one for `denyPendingBySession` and one for `getPendingCountBySession`, where the former now documents nothing. `[fixable]`


### `packages/client/src/event-bus.ts`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🟡 **bugs** (L64): Listener leak on unsub/resub: when the last listener for an event type unsubscribes, the listeners map entry is deleted (line 64), but the `addEventListener` handler on the EventSource is never removed. If a new listener later subscribes to the same event type, `on()` calls `registerEventDispatch()` again, adding a duplicate handler to the EventSource. The old handler is a no-op (its `set` lookup still works) so behavior is correct, but handlers accumulate without bound on the EventSource for subscribe/unsubscribe cycles. `[fixable]`


### `server/index.ts`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🟡 **regressions** (L821): `sseRegistry.destroy()` is never called during shutdown. The `SseRegistry` has a `destroy()` method that ends all client responses and clears the heartbeat interval, but `shutdown()` only calls `overviewEmitter.destroy()` — SSE clients won't get a clean close, and the heartbeat interval timer may keep the process alive past the grace period. `[fixable]`


### `packages/harness/src/sse-registry.ts`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🟡 **missing_tests**: No test file for `SseRegistry`. The class has non-trivial logic: heartbeat start/stop lifecycle, broadcast failure cleanup, `formatEvent` SSE encoding, and `destroy()` cleanup. The `EventBus` client side has thorough tests, but the server-side registry has none.


_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🔵 **unsafe_assumptions** (L252): `checkTaskWaitState()` only checks direct children of the goal (`task.parentId === goalId`). If the task tree is deeper than one level (grandchild tasks), blocked or pending_review tasks at deeper levels won't be detected. This may be intentional for Phase 2 scope, but the comment says 'flat tree — children follow parent' which isn't guaranteed by `getTree()` which returns the full tree. `[fixable]`
- 🔵 **unsafe_assumptions** (L279): `extractRepoName` uses a non-greedy `(.+?)` in the worktree regex, which matches the shortest possible path. For a path like `/Users/foo/a/b/.claude/worktrees/id`, this correctly captures `/Users/foo/a/b`. However, for nested `.claude` directories (unlikely but possible), the non-greedy match would stop at the first `.claude`. A greedy match `(.+)` would be safer. `[fixable]`


### `server/chat.ts`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🔵 **style** (L47): The `setSessionChangeCallback` and `_onSessionChange` callback pattern (module-level mutable state) adds a fifth `set*` callback pattern to `chat.ts`. This module already has `setConnectionRegistry`. Consider whether this could be wired through the existing `QueryLoopOptions.onTurnEnd` pattern instead of adding another module-level setter.


### `frontend/src/components/SessionOverview.tsx`
_Well-structured SSE event bus with good test coverage on client and session-overview. Key issues: orphaned JSDoc in permissions.ts, missing sseRegistry.destroy() in shutdown, no tests for SseRegistry, and a minor EventSource handler leak on unsub/resub cycles._

- 🔵 **bugs** (L34): `formatElapsed` is called with `Date.now() - activity.lastEventAt` on each render, but since the component doesn't re-render on a timer, the elapsed time will be stale. A session showing 'just now' will keep showing 'just now' until the next SSE event triggers a re-render. Consider a periodic re-render or accepting this as a known limitation.

PR #307, opened 2026-05-02
feat(sse): wire consumer hooks to SSE events — Phase 2

dimakis:
## Summary

- **useTodoData**: subscribes to `todo_update` SSE → auto-refetch (replaces stale one-shot fetch)
- **useTaskBoard**: subscribes to `loop_status` → reload status, `task_state` → refresh tasks (supplements existing WS)
- **useTabBadges**: subscribes to `inbox_updated` → reload inbox count, `todo_update` → reload todo count (replaces visibility-change-only updates)
- **useSessionList**: subscribes to `session_activity` → live `isActive`/`isAttached` dots without full refetch (replaces stale session list)

Also addresses Centaur review findings from PR #304:
- Fix EventBus handler leak on unsub/resub cycles
- Add `sseRegistry.destroy()` to server shutdown
- Fix orphaned JSDoc in permissions.ts
- Clean up dead SSE connections during heartbeat
- Add SseRegistry tests (13 tests)
- Add EventBus unsub/resub cycle test
- Greedy regex in `extractRepoName`, clarified `checkTaskWaitState` scope

## Test plan

- [x] All 649 frontend tests pass (71 files)
- [x] All server tests pass
- [x] Event bus mock added to 5 test files that transitively import event-bus-singleton
- [ ] Manual: open home page → modify a Telos item via CLI → Telos view auto-updates
- [ ] Manual: start ATB loop → task board shows live loop status without page refresh
- [ ] Manual: inbox item arrives → tab badge updates immediately
- [ ] Manual: session starts/stops → session list dots update in real-time

### DIFF
--- a/frontend/src/hooks/useSessionList.ts
+++ b/frontend/src/hooks/useSessionList.ts
@@ -2,6 +2,8 @@ import { useState, useEffect, useCallback, useRef } from 'react';
 import type { Session } from '../types/chat';
 import { renameSession as renameSessionApi } from '../lib/rename-session';
 import { apiFetch } from '../lib/api-fetch';
+import { eventBus } from '../lib/event-bus-singleton';
+import type { SessionActivity } from './useSessionOverview';
 
 export interface QuickAction {
   label: string;
@@ -86,8 +88,24 @@ export function useSessionList(): UseSessionListReturn {
     };
     document.addEventListener('visibilitychange', onVisible);
 
+    // Live session dots via SSE — update isActive/isAttached without full refetch
+    const unsubActivity = eventBus.on('session_activity', (data) => {
+      const activities = data as SessionActivity[];
+      const activeIds = new Set(activities.map((a) => a.sessionId));
+      setSessions((prev) =>
+        prev.map((s) => ({
+          ...s,
+          isActive: activeIds.has(s.id),
+          isAttached: activities.some(
+            (a) => a.sessionId === s.id && a.state !== 'idle' && a.state !== 'done',
+          ),
+        })),
+      );
+    });
+
     return () => {
       document.removeEventListener('visibilitychange', onVisible);
+      unsubActivity();
     };
   }, []);
 
--- a/frontend/src/hooks/useSessionOverview.ts
+++ b/frontend/src/hooks/useSessionOverview.ts
@@ -1,24 +1,8 @@
 import { useState, useEffect, useMemo } from 'react';
 import { eventBus } from '../lib/event-bus-singleton';
+import type { SessionActivity } from '@mitzo/protocol';
 
-// ─── Types (mirror server/session-overview.ts) ──────────────────────────────
-
-export type SessionActivityState = 'init' | 'working' | 'waiting' | 'done' | 'idle' | 'paused';
-
-export type WaitReason = 'permission' | 'review' | 'blocked';
-
-export interface SessionActivity {
-  sessionId: string;
-  clientId: string;
-  title: string;
-  repo?: string;
-  state: SessionActivityState;
-  flags: SessionActivityState[];
-  waitReason?: WaitReason;
-  progress?: { done: number; total: number };
-  lastEventAt: number;
-  taskId?: string;
-}
+export type { SessionActivity, SessionActivityState, WaitReason } from '@mitzo/protocol';
 
 // ─── Tier sorting ───────────────────────────────────────────────────────────
 
--- a/packages/client/src/event-bus.ts
+++ b/packages/client/src/event-bus.ts
@@ -1,15 +1,3 @@
-/**
- * EventBus — SSE client for broadcast events.
- *
- * Wraps native EventSource to receive typed server-sent events (session state,
- * task updates, health, etc.). One instance per app, lives for the app's
- * lifetime. Complements the WebSocket (per-session, bidirectional) with a
- * global unidirectional awareness channel.
- *
- * EventSource auto-reconnects natively — no custom reconnect logic needed.
- * On reconnect, the server sends a fresh hydration snapshot.
- */
-
 /** EventSource readyState constants — avoids referencing the global. */
 const ES_OPEN = 1;
 const ES_CLOSED = 2;
@@ -30,23 +18,12 @@ export class EventBus {
       factory ?? ((url: string) => new EventSource(url, { withCredentials: true }));
   }
 
-  /**
-   * Connect to the SSE endpoint. Call once on app mount.
-   */
   connect(url: string): void {
     if (this.source) return;
     this.url = url;
     this.createSource();
   }
 
-  /**
-   * Subscribe to a specific event type. Returns an unsubscribe function.
-   *
-   * Listeners are tracked by the bus itself — they survive EventSource
-   * reconnects because we re-register them on each new source.
-   * The dispatch path goes through the listener set (not through
-   * individual EventSource handlers), so unsubscribe is immediate.
-   */
   on(event: string, listener: EventBusListener): () => void {
     let set = this.listeners.get(event);
     if (!set) {
@@ -64,19 +41,12 @@ export class EventBus {
     };
   }
 
-  /**
-   * Register a callback for connection state changes.
-   */
   onConnectionChange(cb: ConnectionChangeCallback): () => void {
     this.connectionChangeListeners.add(cb);
     return () => this.connectionChangeListeners.delete(cb);
   }
 
-  /**
-   * Ensure the SSE connection is alive. Call on app resume (iOS foreground).
-   * If the EventSource is in CLOSED state (gave up reconnecting), recreates it.
-   * No-op if still connected or reconnecting.
-   */
+  // Recreates EventSource if CLOSED (gave up); no-op if connected/reconnecting.
   ensureConnected(): void {
     if (!this.url) return;
     if (!this.source || this.source.readyState === ES_CLOSED) {
@@ -92,9 +62,6 @@ export class EventBus {
     }
   }
 
-  /**
-   * Disconnect. Call on app unmount.
-   */
   disconnect(): void {
     if (this.source) {
       try {
@@ -106,9 +73,6 @@ export class EventBus {
     }
   }
 
-  /**
-   * Whether the SSE connection is currently open.
-   */
   get connected(): boolean {
     return this.source?.readyState === ES_OPEN;
   }
@@ -129,19 +93,11 @@ export class EventBus {
       this.notifyConnectionChange(false);
     };
 
-    // Register one dispatch handler per event type on the new source.
-    // Each handler fans out to all listeners in the set — no per-listener
-    // handlers on the EventSource, so unsubscribe is just a Set.delete().
     for (const event of this.listeners.keys()) {
       this.registerEventDispatch(event);
     }
   }
 
-  /**
-   * Register a single EventSource handler for an event type that dispatches
-   * to all listeners in the set. This indirection means unsubscribe doesn't
-   * need to touch the EventSource — it just removes from the Set.
-   */
   private registerEventDispatch(event: string): void {
     if (!this.source) return;
     this.source.addEventListener(event, ((e: MessageEvent) => {
--- a/packages/client/src/event-bus.ts
+++ b/packages/client/src/event-bus.ts
@@ -1,4 +1,6 @@
-/** EventSource readyState constants — avoids referencing the global. */
+// SSE client for broadcast events (session state, tasks, health).
+
+// EventSource readyState constants.
 const ES_OPEN = 1;
 const ES_CLOSED = 2;
 
@@ -24,6 +26,7 @@ export class EventBus {
     this.createSource();
   }
 
+  // Listeners survive EventSource reconnects — dispatch goes through the Set.
   on(event: string, listener: EventBusListener): () => void {
     let set = this.listeners.get(event);
     if (!set) {
@@ -46,7 +49,7 @@ export class EventBus {
     return () => this.connectionChangeListeners.delete(cb);
   }
 
-  // Recreates EventSource if CLOSED (gave up); no-op if connected/reconnecting.
+  // Recreate EventSource if CLOSED (gave up). No-op if still alive.
   ensureConnected(): void {
     if (!this.url) return;
     if (!this.source || this.source.readyState === ES_CLOSED) {
--- a/packages/client/src/event-bus.ts
+++ b/packages/client/src/event-bus.ts
@@ -96,8 +96,8 @@ export class EventBus {
       this.notifyConnectionChange(false);
     };
 
-    for (const event of this.listeners.keys()) {
-      this.registerEventDispatch(event);
+    for (const [event, set] of this.listeners) {
+      if (set.size > 0) this.registerEventDispatch(event);
     }
   }
 
--- a/packages/harness/src/sse-registry.ts
+++ b/packages/harness/src/sse-registry.ts
@@ -1,15 +1,3 @@
-/**
- * SseRegistry — manages Server-Sent Events connections for broadcast events.
- *
- * Unlike WebSocket (bidirectional, per-session), SSE is unidirectional
- * (server → client) and global (not tied to a specific session). It delivers
- * broadcast data like session state changes, task updates, todo changes, and
- * health status.
- *
- * Auto-reconnects natively via EventSource (client-side) — no custom reconnect
- * logic needed. Each client gets a fresh snapshot on connect (hydration).
- */
-
 import type { Response } from 'express';
 import { createLogger } from './logger.js';
 
@@ -26,11 +14,6 @@ export class SseRegistry {
   private clients = new Map<string, SseClient>();
   private heartbeatTimer: ReturnType<typeof setInterval> | null = null;
 
-  /**
-   * Register a new SSE client. Starts heartbeat if this is the first client.
-   * The response object must have headers already written
-   * (Content-Type: text/event-stream, etc.) before calling this.
-   */
   add(id: string, res: Response): void {
     this.clients.set(id, { id, res });
     log.info('SSE client connected', { id, total: this.clients.size });
@@ -40,9 +23,6 @@ export class SseRegistry {
     }
   }
 
-  /**
-   * Remove a client (on close/error). Stops heartbeat if this was the last client.
-   */
   remove(id: string): void {
     if (!this.clients.has(id)) return;
 
@@ -54,13 +34,6 @@ export class SseRegistry {
     }
   }
 
-  /**
-   * Send a typed event to all connected clients.
-   * SSE format:
-   *   event: <type>\n
-   *   data: <json>\n
-   *   \n
-   */
   broadcast(event: string, data: unknown): void {
     if (this.clients.size === 0) return;
 
@@ -86,9 +59,6 @@ export class SseRegistry {
     }
   }
 
-  /**
-   * Send a typed event to a specific client (for hydration on connect).
-   */
   sendTo(clientId: string, event: string, data: unknown): boolean {
     const client = this.clients.get(clientId);
     if (!client) return false;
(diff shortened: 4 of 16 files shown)


packages/harness/src/sse-registry.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    log.info('Starting SSE heartbeat', { intervalMs: HEARTBEAT_INTERVAL_MS });
+    this.heartbeatTimer = setInterval(() => {
+      const payload = ':heartbeat\n\n'; // SSE comment line — ignored by EventSource
+      for (const [, client] of this.clients) {
+        try {
+          client.res.write(payload);
+        } catch {
+          // Write failed — clean up dead connection immediately
+          this.remove(client.id);
```
dimakis: 🟡 **bugs**: `this.remove(client.id)` inside `for (const [, client] of this.clients)` mutates the Map during iteration. While JS spec permits Map deletion during for...of, `remove()` also calls `stopHeartbeat()` (clearing the interval from inside its own callback) when the last client is removed. The `broadcast()` method already handles this correctly by collecting failures first and removing after the loop — the heartbeat should use the same pattern for consistency and safety. `[fixable]`

server/index.ts
```diff
@@ -123,6 +129,8 @@ setTaskBroadcast((event) => {
     if (client.readyState === client.OPEN) client.send(msg);
   });
   connRegistry.broadcastAll(event as Record<string, unknown>);
+  sseRegistry.broadcast('task_state', event);
+  overviewEmitter.scheduleBroadcast();
```
dimakis, line 137: 🟡 **unsafe_assumptions**: `overviewEmitter.scheduleBroadcast()` is called inside the `setTaskBroadcast` callback (line 133) and the orchestrator's `broadcastStatus` callback (line 196), but `overviewEmitter` isn't declared until line 222 (`const`). This works at runtime because these closures only fire after module initialization completes, but it's fragile — a refactor that triggers a task broadcast during initialization would hit a TDZ ReferenceError. Consider moving the `overviewEmitter` declaration above the orchestrator construction, or using the same `let`-with-setter pattern used in `app.ts`. `[fixable]`

frontend/src/hooks/useSessionOverview.ts
```diff
@@ -0,0 +1,76 @@
+import { useState, useEffect, useMemo } from 'react';
+import { eventBus } from '../lib/event-bus-singleton';
+
+// ─── Types (mirror server/session-overview.ts) ──────────────────────────────
+
+export type SessionActivityState = 'init' | 'working' | 'waiting' | 'done' | 'idle' | 'paused';
```
dimakis: 🔵 **style**: `SessionActivity`, `SessionActivityState`, and `WaitReason` types are duplicated between `server/session-overview.ts` and `frontend/src/hooks/useSessionOverview.ts` (comment says "mirror server/session-overview.ts"). These shared protocol types should live in `@mitzo/protocol` to prevent drift — this is the established pattern for all other shared types in this codebase. `[fixable]`

packages/client/src/event-bus.ts
```diff
@@ -0,0 +1,170 @@
...
+      set = new Set();
+      this.listeners.set(event, set);
+      // First listener for this event type — register the dispatch handler
+      if (this.source) {
+        this.registerEventDispatch(event);
+      }
+    }
+    set.add(listener);
+
+    return () => {
```
dimakis, line 42: 🔵 **bugs**: When the last listener for an event type unsubscribes, the empty Set stays in `this.listeners` and the EventSource dispatch handler remains registered. On reconnect via `ensureConnected()`, `createSource()` re-registers dispatch handlers for all event types in the Map — including ones with zero listeners. Consider cleaning up the Set and Map entry when the last listener unsubscribes, or at minimum skipping registration for empty sets in `createSource()`. `[fixable]`

server/app.ts
```diff
@@ -487,6 +496,31 @@ app.post('/api/sessions/suspend', (req, res) => {
 
 app.use('/api', authMiddleware);
 
+// --- SSE Event Bus ---
+
+app.get('/api/events', (req, res) => {
```
dimakis, line 501: 🔵 **missing_tests**: The SSE endpoint (`GET /api/events`) has no integration test verifying: (1) auth middleware applies (unauthenticated requests get 401), (2) correct SSE headers are sent, (3) hydration events (`connected`, `session_activity`) are delivered on connect. The SseRegistry unit tests cover the registry logic, but the route wiring is untested. `[fixable]`

server/chat.ts
```diff
@@ -44,6 +44,12 @@ let _connRegistry: ConnectionRegistry | null = null;
 export function setConnectionRegistry(registry: ConnectionRegistry): void {
   _connRegistry = registry;
 }
+
```
dimakis, line 47: 🔵 **style**: The `SessionChangeCallback` type and `setSessionChangeCallback` follow a module-level mutable setter pattern. This is the fourth such setter in this file (`setConnectionRegistry`, `setSessionChangeCallback`, plus the EventStore/prompt-compare imports). Consider consolidating these into a single options/deps object to reduce the scattered mutable state, though this is a broader refactor concern.

dimakis (commented):
## Centaur Review

### `packages/harness/src/sse-registry.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🟡 **bugs** (L131): `this.remove(client.id)` inside `for (const [, client] of this.clients)` mutates the Map during iteration. While JS spec permits Map deletion during for...of, `remove()` also calls `stopHeartbeat()` (clearing the interval from inside its own callback) when the last client is removed. The `broadcast()` method already handles this correctly by collecting failures first and removing after the loop — the heartbeat should use the same pattern for consistency and safety. `[fixable]`


### `server/index.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🟡 **unsafe_assumptions** (L133): `overviewEmitter.scheduleBroadcast()` is called inside the `setTaskBroadcast` callback (line 133) and the orchestrator's `broadcastStatus` callback (line 196), but `overviewEmitter` isn't declared until line 222 (`const`). This works at runtime because these closures only fire after module initialization completes, but it's fragile — a refactor that triggers a task broadcast during initialization would hit a TDZ ReferenceError. Consider moving the `overviewEmitter` declaration above the orchestrator construction, or using the same `let`-with-setter pattern used in `app.ts`. `[fixable]`


### `frontend/src/hooks/useSessionOverview.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🔵 **style** (L6): `SessionActivity`, `SessionActivityState`, and `WaitReason` types are duplicated between `server/session-overview.ts` and `frontend/src/hooks/useSessionOverview.ts` (comment says "mirror server/session-overview.ts"). These shared protocol types should live in `@mitzo/protocol` to prevent drift — this is the established pattern for all other shared types in this codebase. `[fixable]`


### `packages/client/src/event-bus.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🔵 **bugs** (L62): When the last listener for an event type unsubscribes, the empty Set stays in `this.listeners` and the EventSource dispatch handler remains registered. On reconnect via `ensureConnected()`, `createSource()` re-registers dispatch handlers for all event types in the Map — including ones with zero listeners. Consider cleaning up the Set and Map entry when the last listener unsubscribes, or at minimum skipping registration for empty sets in `createSource()`. `[fixable]`


### `server/app.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🔵 **missing_tests** (L501): The SSE endpoint (`GET /api/events`) has no integration test verifying: (1) auth middleware applies (unauthenticated requests get 401), (2) correct SSE headers are sent, (3) hydration events (`connected`, `session_activity`) are delivered on connect. The SseRegistry unit tests cover the registry logic, but the route wiring is untested. `[fixable]`


### `server/chat.ts`
_Well-structured SSE event bus with solid test coverage; main concerns are Map mutation during heartbeat iteration, fragile TDZ reference ordering in index.ts, and duplicated protocol types that should live in @mitzo/protocol._

- 🔵 **style** (L47): The `SessionChangeCallback` type and `setSessionChangeCallback` follow a module-level mutable setter pattern. This is the fourth such setter in this file (`setConnectionRegistry`, `setSessionChangeCallback`, plus the EventStore/prompt-compare imports). Consider consolidating these into a single options/deps object to reduce the scattered mutable state, though this is a broader refactor concern.

packages/harness/src/sse-registry.ts
```diff
@@ -0,0 +1,126 @@
...
+    if (this.heartbeatTimer) return;
+
+    log.info('Starting SSE heartbeat', { intervalMs: HEARTBEAT_INTERVAL_MS });
+    this.heartbeatTimer = setInterval(() => {
+      const payload = ':heartbeat\n\n'; // SSE comment line — ignored by EventSource
+      for (const [, client] of this.clients) {
+        try {
+          client.res.write(payload);
+        } catch {
+          // Write failed — clean up dead connection immediately
```
dimakis: 🟡 **bugs**: Map mutation during iteration: the heartbeat loop calls `this.remove(client.id)` inside `for (const [, client] of this.clients)`, which deletes from the Map being iterated. While V8's Map implementation tolerates this (deleted entries are skipped, no crash), it's fragile and inconsistent with `broadcast()` which correctly collects failures first then cleans up. Apply the same collect-then-remove pattern here. `[fixable]`

packages/client/src/event-bus.ts
```diff
@@ -0,0 +1,129 @@
...
+    source.onerror = () => {
+      this.notifyConnectionChange(false);
+    };
+
+    for (const event of this.listeners.keys()) {
+      this.registerEventDispatch(event);
+    }
+  }
+
+  private registerEventDispatch(event: string): void {
```
dimakis, line 104: 🟡 **bugs**: EventSource listener accumulation: `registerEventDispatch` adds an anonymous handler via `addEventListener` but never removes it. When `ensureConnected()` recreates the EventSource this is fine (new object, old GC'd). But if `on()` is called for the same event type on a source that already had dispatch registered via `createSource()`, the first-listener-per-set guard at line 33-37 prevents duplication. However, if a component unmounts (clearing the Set) and remounts (creating a new Set for the same event), `registerEventDispatch` is called again on the same EventSource, adding a second identical dispatch handler. Both will fire, causing duplicate listener invocations. Consider tracking which events are registered per source instance. `[fixable]`

server/session-overview.ts
```diff
@@ -0,0 +1,283 @@
...
+    loopStatus: LoopStatus,
+    now: number,
+  ): SessionActivity {
+    const sessionId = session.sessionId!;
+    const lastEventAt = this.lastEventTimes.get(session.clientId) ?? now;
+    const elapsed = now - lastEventAt;
+
+    // Collect all applicable states
+    const states: SessionActivityState[] = [];
+    let waitReason: WaitReason | undefined;
```
dimakis, line 132: 🟡 **unsafe_assumptions**: `const sessionId = session.sessionId!` uses non-null assertion. While the `compute()` method filters `s.sessionId` truthy, `deriveActivity` is a separate method — if ever called directly or if the filter changes, this will produce an activity with `undefined` sessionId cast to string. Consider accepting the narrowed type or adding a guard. `[fixable]`

server/session-overview.ts
```diff
@@ -0,0 +1,283 @@
...
+    let progress: { done: number; total: number } | undefined;
+    let taskId: string | undefined;
+
```
dimakis, line 135: 🔵 **unsafe_assumptions**: `lastEventTimes.get(session.clientId) ?? now` means untouched sessions default to 'just now', which makes them appear as 'done' rather than 'idle'. This is intentional per the design (sessions start fresh), but means a server restart loses all timing data and every active-but-not-streaming session briefly shows as 'done' until the 5-minute timeout. Consider persisting touch times or deriving from session start time.

server/app.ts
dimakis, line 501: 🟡 **missing_tests**: The `/api/events` SSE endpoint has no integration test. Key behaviors to test: SSE headers are set correctly, auth middleware is enforced (unauthenticated request rejected), hydration events are sent on connect (`connected` + `session_activity`), and client cleanup on `req.close`. `[fixable]`

frontend/src/hooks/useSessionOverview.ts
```diff
@@ -0,0 +1,76 @@
...
+
```
dimakis: 🔵 **style**: The `SessionActivity` and `SessionActivityState` types are duplicated between `frontend/src/hooks/useSessionOverview.ts` and `server/session-overview.ts`. Consider defining these in `@mitzo/protocol` (the shared types package) to ensure they stay in sync and to follow the project convention of shared types living in protocol. `[fixable]`

frontend/src/hooks/useSessionList.ts
```diff
@@ -86,8 +88,24 @@ export function useSessionList(): UseSessionListReturn {
     };
     document.addEventListener('visibilitychange', onVisible);
 
+    // Live session dots via SSE — update isActive/isAttached without full refetch
+    const unsubActivity = eventBus.on('session_activity', (data) => {
+      const activities = data as SessionActivity[];
+      const activeIds = new Set(activities.map((a) => a.sessionId));
+      setSessions((prev) =>
```
dimakis, line 95: 🔵 **style**: `data as SessionActivity[]` is an unsafe cast on unvalidated SSE data. All other SSE handlers in this PR do similar casts. Since SSE data crosses a trust boundary (network), consider adding a lightweight runtime check or Zod parse — consistent with the project's existing use of Zod schemas in `ws-schemas.ts` and `api-schemas.ts`. `[fixable]`

server/chat.ts
```diff
@@ -660,6 +666,7 @@ async function _startChatInner(
 
   const session = registry.get(clientId)!;
   session.inputQueue = inputQueue as { push: (msg: unknown) => void; close: () => void };
+  _onSessionChange?.(clientId, 'start');
```
dimakis, line 670: 🔵 **regressions**: The `_onSessionChange?.(clientId, 'start')` call fires before the `query()` loop begins (line 670, before the async query invocation). The `'end'` event fires in the `finally` block (line 808). This means a session that fails during startup (e.g., SDK init error) will emit both 'start' and 'end' — correct. But note that the removed comment 'Phase 2e: worktrees survive until explicit close or stale GC' was replaced by the callback. Verify that this comment removal doesn't signal an intentional behavior change that was lost.

dimakis (commented):
## Centaur Review

### `packages/harness/src/sse-registry.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🟡 **bugs** (L97): Map mutation during iteration: the heartbeat loop calls `this.remove(client.id)` inside `for (const [, client] of this.clients)`, which deletes from the Map being iterated. While V8's Map implementation tolerates this (deleted entries are skipped, no crash), it's fragile and inconsistent with `broadcast()` which correctly collects failures first then cleans up. Apply the same collect-then-remove pattern here. `[fixable]`


### `packages/client/src/event-bus.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🟡 **bugs** (L104): EventSource listener accumulation: `registerEventDispatch` adds an anonymous handler via `addEventListener` but never removes it. When `ensureConnected()` recreates the EventSource this is fine (new object, old GC'd). But if `on()` is called for the same event type on a source that already had dispatch registered via `createSource()`, the first-listener-per-set guard at line 33-37 prevents duplication. However, if a component unmounts (clearing the Set) and remounts (creating a new Set for the same event), `registerEventDispatch` is called again on the same EventSource, adding a second identical dispatch handler. Both will fire, causing duplicate listener invocations. Consider tracking which events are registered per source instance. `[fixable]`


_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🟡 **unsafe_assumptions** (L148): `const sessionId = session.sessionId!` uses non-null assertion. While the `compute()` method filters `s.sessionId` truthy, `deriveActivity` is a separate method — if ever called directly or if the filter changes, this will produce an activity with `undefined` sessionId cast to string. Consider accepting the narrowed type or adding a guard. `[fixable]`
- 🔵 **unsafe_assumptions** (L151): `lastEventTimes.get(session.clientId) ?? now` means untouched sessions default to 'just now', which makes them appear as 'done' rather than 'idle'. This is intentional per the design (sessions start fresh), but means a server restart loses all timing data and every active-but-not-streaming session briefly shows as 'done' until the 5-minute timeout. Consider persisting touch times or deriving from session start time.


### `server/app.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🟡 **missing_tests** (L501): The `/api/events` SSE endpoint has no integration test. Key behaviors to test: SSE headers are set correctly, auth middleware is enforced (unauthenticated request rejected), hydration events are sent on connect (`connected` + `session_activity`), and client cleanup on `req.close`. `[fixable]`


### `frontend/src/hooks/useSessionOverview.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🔵 **style** (L7): The `SessionActivity` and `SessionActivityState` types are duplicated between `frontend/src/hooks/useSessionOverview.ts` and `server/session-overview.ts`. Consider defining these in `@mitzo/protocol` (the shared types package) to ensure they stay in sync and to follow the project convention of shared types living in protocol. `[fixable]`


### `frontend/src/hooks/useSessionList.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🔵 **style** (L95): `data as SessionActivity[]` is an unsafe cast on unvalidated SSE data. All other SSE handlers in this PR do similar casts. Since SSE data crosses a trust boundary (network), consider adding a lightweight runtime check or Zod parse — consistent with the project's existing use of Zod schemas in `ws-schemas.ts` and `api-schemas.ts`. `[fixable]`


### `server/chat.ts`
_Solid SSE infrastructure with good test coverage for new components. Main concerns: heartbeat Map mutation during iteration (inconsistent with broadcast's safer pattern), potential EventBus listener duplication on component remount, and duplicated types between server/frontend that should live in @mitzo/protocol._

- 🔵 **regressions** (L670): The `_onSessionChange?.(clientId, 'start')` call fires before the `query()` loop begins (line 670, before the async query invocation). The `'end'` event fires in the `finally` block (line 808). This means a session that fails during startup (e.g., SDK init error) will emit both 'start' and 'end' — correct. But note that the removed comment 'Phase 2e: worktrees survive until explicit close or stale GC' was replaced by the callback. Verify that this comment removal doesn't signal an intentional behavior change that was lost.

PR #308, opened 2026-05-02
feat(sse): health consolidation — Phase 3

dimakis:
## Summary

- Server-side `HealthMonitor` polls Yapper + ContexGin every 30s, broadcasts `health` SSE event only on status change (not every tick)
- New `useServiceHealth` hook replaces independent 30s polling intervals in `useVoice` and `useDocumentReader` with a single SSE subscription
- `ServiceStatus` component shows service health dots on the home page
- `ServiceHealthPayload` / `ServiceHealthStatus` types added to `@mitzo/protocol`
- Removed `YAPPER_HEALTH_POLL_MS` constant (no longer needed)
- Tests updated: health assertions now use mock `useServiceHealth` instead of mock fetch

## Test plan

- [ ] Build passes (`tsc --noEmit`)
- [ ] All tests pass (`vitest run` — 2378/2378, excluding pre-existing chat.test.ts timeouts)
- [ ] Yapper/ContexGin dots appear on home page when services are running
- [ ] Dots update without page refresh when a service goes down/up
- [ ] Voice recording and TTS still work correctly
- [ ] Document reader still works correctly

### DIFF
--- a/frontend/src/components/ServiceStatus.tsx
+++ b/frontend/src/components/ServiceStatus.tsx
@@ -1,0 +1,29 @@
+// Service health indicator — shows Yapper and ContexGin status on the home page.
+
+import { useServiceHealth } from '../hooks/useServiceHealth';
+import type { ServiceHealthStatus } from '@mitzo/protocol';
+
+function ServiceDot({ service }: { service: ServiceHealthStatus | null }) {
+  if (!service) return null;
+  const color = service.ok ? '#4ade80' : '#ff6d6d';
+  return (
+    <span className="service-dot" style={{ color }}>
+      <span className="service-dot-indicator" style={{ background: color }} />
+      {service.name}
+    </span>
+  );
+}
+
+export function ServiceStatus() {
+  const { yapper, contexgin, checkedAt } = useServiceHealth();
+
+  // Don't render until first health check arrives
+  if (checkedAt === 0) return null;
+
+  return (
+    <div className="service-status">
+      <ServiceDot service={yapper} />
+      <ServiceDot service={contexgin} />
+    </div>
+  );
+}
--- a/frontend/src/components/ServiceStatus.tsx
+++ b/frontend/src/components/ServiceStatus.tsx
@@ -17,8 +17,8 @@ function ServiceDot({ service }: { service: ServiceHealthStatus | null }) {
 export function ServiceStatus() {
   const { yapper, contexgin, checkedAt } = useServiceHealth();
 
-  // Don't render until first health check arrives
-  if (checkedAt === 0) return null;
+  // Don't render until first health check arrives, or if no services are present
+  if (checkedAt === 0 || (!yapper && !contexgin)) return null;
 
   return (
     <div className="service-status">
--- a/frontend/src/hooks/useServiceHealth.ts
+++ b/frontend/src/hooks/useServiceHealth.ts
@@ -1,0 +1,33 @@
+// SSE-driven service health — replaces per-hook polling for Yapper/ContexGin.
+
+import { useState, useEffect, useMemo } from 'react';
+import { eventBus } from '../lib/event-bus-singleton';
+import type { ServiceHealthPayload, ServiceHealthStatus } from '@mitzo/protocol';
+
+export interface UseServiceHealthReturn {
+  services: ServiceHealthStatus[];
+  yapper: ServiceHealthStatus | null;
+  contexgin: ServiceHealthStatus | null;
+  checkedAt: number;
+}
+
+export function useServiceHealth(): UseServiceHealthReturn {
+  const [payload, setPayload] = useState<ServiceHealthPayload>({ services: [], checkedAt: 0 });
+
+  useEffect(() => {
+    return eventBus.on('health', (data) => {
+      setPayload(data as ServiceHealthPayload);
+    });
+  }, []);
+
+  const yapper = useMemo(
+    () => payload.services.find((s) => s.name === 'yapper') ?? null,
+    [payload],
+  );
+  const contexgin = useMemo(
+    () => payload.services.find((s) => s.name === 'contexgin') ?? null,
+    [payload],
+  );
+
+  return { services: payload.services, yapper, contexgin, checkedAt: payload.checkedAt };
+}
--- a/frontend/src/hooks/useVoice.ts
+++ b/frontend/src/hooks/useVoice.ts
@@ -1,13 +1,8 @@
 // Voice integration hook — Yapper health, mic capture, streaming + batch transcription, TTS playback.
 
 import { useState, useEffect, useRef, useCallback } from 'react';
-import {
-  YAPPER_URL,
-  YAPPER_HEALTH_POLL_MS,
-  TTS_ENABLED_KEY,
-  TTS_VOICE_KEY,
-  DEFAULT_TTS_VOICE,
-} from '../lib/constants';
+import { YAPPER_URL, TTS_ENABLED_KEY, TTS_VOICE_KEY, DEFAULT_TTS_VOICE } from '../lib/constants';
+import { useServiceHealth } from './useServiceHealth';
 import {
   negotiateMimeType,
   createRecorder,
@@ -25,11 +20,6 @@ import {
   unlockAudioContext,
   closeAudioContext,
 } from '../lib/tts';
-
-interface YapperHealth {
-  status: string;
-  models?: { stt?: boolean; tts?: boolean };
-}
 
 export interface Voice {
   id: string;
@@ -76,15 +66,16 @@ function mimeToFormat(mime: string): string {
 }
 
 export function useVoice(): UseVoiceReturn {
+  // --- Service health (SSE-driven) ---
+  const { yapper } = useServiceHealth();
+  const available = yapper?.ok === true && yapper.detail?.stt !== false;
+  const ttsAvailable = yapper?.ok === true && yapper.detail?.tts !== false;
+
   // --- STT state ---
-  const [available, setAvailable] = useState(false);
   const [recording, setRecording] = useState(false);
   const [transcribing, setTranscribing] = useState(false);
   const [micBlocked, setMicBlocked] = useState(false);
   const [error, setError] = useState<string | null>(null);
-
-  // --- TTS state ---
-  const [ttsAvailable, setTtsAvailable] = useState(false);
   const [ttsEnabled, setTtsEnabledState] = useState(
     () => localStorage.getItem(TTS_ENABLED_KEY) === 'true',
   );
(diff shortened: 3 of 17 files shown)


frontend/src/hooks/useVoice.ts
```diff
@@ -76,15 +66,16 @@ function mimeToFormat(mime: string): string {
 }
 
 export function useVoice(): UseVoiceReturn {
+  // --- Service health (SSE-driven) ---
+  const { yapper } = useServiceHealth();
+  const available = yapper?.ok === true && yapper.detail?.stt !== false;
```
dimakis, line 74: 🟡 **bugs**: Behavioral change: the old code checked `data.status === 'ready' || data.status === 'ok'` AND `data.models.stt === true`. The new code uses `yapper?.detail?.stt !== false`, which means when Yapper responds without a `models` field (detail is `undefined`), `undefined !== false` evaluates to `true` — so STT is assumed available. The server-side `parseDetail` returns `undefined` when `models` is absent (line 25 of health-monitor.ts), so `detail` will be missing entirely. The old frontend code handled this case by treating a missing `models` field as 'assume both capabilities when status is ready', so the behavior is preserved. However, when Yapper is `ok: true` but the health endpoint returns a non-ready status (e.g. `status: 'loading'`), the server sets `ok: false` so this path doesn't fire. The logic is safe but the `!== false` pattern is subtle — a brief inline comment noting the intentional fallthrough would help future readers. `[fixable]`

server/health-monitor.ts
```diff
@@ -0,0 +1,110 @@
...
+const POLL_INTERVAL_MS = 30_000;
+const CHECK_TIMEOUT_MS = 3_000;
+
+interface ServiceCheck {
+  name: string;
+  url: string;
+  parseDetail?: (data: unknown) => Record<string, unknown> | undefined;
+}
+
+const SERVICES: ServiceCheck[] = [
```
dimakis, line 19: 🔵 **unsafe_assumptions**: `SERVICES` is evaluated at module load time, reading `process.env.YAPPER_PROXY_TARGET` and `process.env.CONTEXGIN_URL` once. If these env vars are set after module import (unlikely but possible in test scenarios), the URLs will use defaults. This is consistent with how other parts of the codebase handle env vars, so it's fine for production — just worth noting for testability.

server/app.ts
```diff
@@ -518,6 +523,11 @@ app.get('/api/events', (req, res) => {
     sseRegistry.sendTo(clientId, 'session_activity', overviewEmitter.getSnapshot());
   }
 
+  if (healthMonitor) {
+    const snapshot = healthMonitor.getSnapshot();
+    if (snapshot) sseRegistry.sendTo(clientId, 'health', snapshot);
```
dimakis: 🔵 **style**: `getSnapshot()` always returns a truthy object (never null/undefined — it falls back to `{ services: [], checkedAt: 0 }`), so `if (snapshot)` is always true. This isn't a bug but the guard is misleading — it suggests `getSnapshot()` can return a falsy value. Consider removing the `if` or adjusting the type signature to match the intent. Contrast with the `if (overviewEmitter)` guard above it, which checks the nullable module-level variable. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useVoice.ts`
_Clean consolidation that correctly replaces per-hook polling with centralized SSE health events. Main gap is missing unit tests for the three new modules (HealthMonitor, useServiceHealth, ServiceStatus)._

- 🟡 **bugs** (L71): Behavioral change: the old code checked `data.status === 'ready' || data.status === 'ok'` AND `data.models.stt === true`. The new code uses `yapper?.detail?.stt !== false`, which means when Yapper responds without a `models` field (detail is `undefined`), `undefined !== false` evaluates to `true` — so STT is assumed available. The server-side `parseDetail` returns `undefined` when `models` is absent (line 25 of health-monitor.ts), so `detail` will be missing entirely. The old frontend code handled this case by treating a missing `models` field as 'assume both capabilities when status is ready', so the behavior is preserved. However, when Yapper is `ok: true` but the health endpoint returns a non-ready status (e.g. `status: 'loading'`), the server sets `ok: false` so this path doesn't fire. The logic is safe but the `!== false` pattern is subtle — a brief inline comment noting the intentional fallthrough would help future readers. `[fixable]`


### `server/health-monitor.ts`
_Clean consolidation that correctly replaces per-hook polling with centralized SSE health events. Main gap is missing unit tests for the three new modules (HealthMonitor, useServiceHealth, ServiceStatus)._

- 🟡 **missing_tests**: No test file exists for `HealthMonitor`. The class has meaningful logic worth testing: `hasChanged()` change detection, `checkService()` parsing (especially the Yapper `parseDetail` branch for missing `models`), the initial-check-on-start behavior, and the broadcast-only-on-change contract. This is a server-side component with clear seams for unit testing (inject a mock `SseRegistry`, stub `fetch`). `[fixable]`
- 🔵 **unsafe_assumptions** (L19): `SERVICES` is evaluated at module load time, reading `process.env.YAPPER_PROXY_TARGET` and `process.env.CONTEXGIN_URL` once. If these env vars are set after module import (unlikely but possible in test scenarios), the URLs will use defaults. This is consistent with how other parts of the codebase handle env vars, so it's fine for production — just worth noting for testability.


### `frontend/src/hooks/useServiceHealth.ts`
_Clean consolidation that correctly replaces per-hook polling with centralized SSE health events. Main gap is missing unit tests for the three new modules (HealthMonitor, useServiceHealth, ServiceStatus)._

- 🟡 **missing_tests**: No dedicated test file for `useServiceHealth`. The hook is indirectly tested through the `useVoice` and `useDocumentReader` test suites (which mock it), but there are no tests verifying the hook's own behavior: subscribing to the event bus, unsubscribing on unmount, the `useMemo` lookups by service name, or handling payloads with unknown service names. `[fixable]`


### `frontend/src/components/ServiceStatus.tsx`
_Clean consolidation that correctly replaces per-hook polling with centralized SSE health events. Main gap is missing unit tests for the three new modules (HealthMonitor, useServiceHealth, ServiceStatus)._

- 🔵 **missing_tests**: No test file for the `ServiceStatus` component. It has conditional rendering logic (`checkedAt === 0` → null, color selection based on `ok`) that would benefit from a basic render test. `[fixable]`


### `server/app.ts`
_Clean consolidation that correctly replaces per-hook polling with centralized SSE health events. Main gap is missing unit tests for the three new modules (HealthMonitor, useServiceHealth, ServiceStatus)._

- 🔵 **style** (L528): `getSnapshot()` always returns a truthy object (never null/undefined — it falls back to `{ services: [], checkedAt: 0 }`), so `if (snapshot)` is always true. This isn't a bug but the guard is misleading — it suggests `getSnapshot()` can return a falsy value. Consider removing the `if` or adjusting the type signature to match the intent. Contrast with the `if (overviewEmitter)` guard above it, which checks the nullable module-level variable. `[fixable]`

frontend/src/components/ServiceStatus.tsx
```diff
@@ -0,0 +1,29 @@
...
+  );
+}
+
+export function ServiceStatus() {
+  const { yapper, contexgin, checkedAt } = useServiceHealth();
+
+  // Don't render until first health check arrives
+  if (checkedAt === 0) return null;
+
+  return (
```
dimakis, line 23: 🔵 **bugs**: When `checkedAt > 0` but both `yapper` and `contexgin` are null (e.g., neither service is configured in the health payload), the component renders an empty `<div className="service-status">` with vertical padding (`var(--space-2) 0`), creating a visible gap in the layout with no content. Consider returning null when both services are null, similar to the `checkedAt === 0` guard. `[fixable]`

server/health-monitor.ts
dimakis, line 19: 🟡 **unsafe_assumptions**: The `SERVICES` array reads `process.env.YAPPER_PROXY_TARGET` and `process.env.CONTEXGIN_URL` at module load time (top-level const). This means env vars set after import are ignored, and the URLs cannot be overridden in tests without mocking the module. This works today because these env vars are set in `.env` before startup, but it's worth noting the coupling. Not blocking since the same pattern is used elsewhere in the codebase. `[fixable]`

server/health-monitor.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  constructor(sseRegistry: SseRegistry) {
+    this.sseRegistry = sseRegistry;
+  }
+
+  start(): void {
+    if (this.timer) return;
+    log.info('Starting health monitor', { intervalMs: POLL_INTERVAL_MS });
+    // Initial check immediately
+    this.check();
```
dimakis, line 47: 🔵 **unsafe_assumptions**: `start()` calls `this.check()` (fire-and-forget async) then immediately sets the interval. If the first check takes longer than `POLL_INTERVAL_MS` (unlikely at 30s, but possible under extreme network conditions), two checks could overlap. The current `setInterval` doesn't wait for the previous check to complete. Consider `setTimeout` chaining if overlapping checks become a concern, though at 30s intervals with 3s timeouts this is very unlikely. `[fixable]`

frontend/src/hooks/useServiceHealth.ts
```diff
@@ -0,0 +1,33 @@
...
+  contexgin: ServiceHealthStatus | null;
+  checkedAt: number;
+}
+
+export function useServiceHealth(): UseServiceHealthReturn {
+  const [payload, setPayload] = useState<ServiceHealthPayload>({ services: [], checkedAt: 0 });
+
+  useEffect(() => {
+    return eventBus.on('health', (data) => {
+      setPayload(data as ServiceHealthPayload);
```
dimakis, line 19: 🔵 **style**: The `data as ServiceHealthPayload` cast bypasses type safety. The eventBus `on` callback receives `unknown` data. While the server always sends the correct shape, a Zod parse (or at minimum a runtime shape check) would be more defensive. Other SSE consumers in the codebase may use similar casts, so this is consistent — but worth flagging. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/components/ServiceStatus.tsx`
_Clean consolidation of per-hook health polling into SSE-driven events. Well-tested, follows existing codebase patterns. Minor suggestions around empty-state rendering and edge-case test coverage._

- 🔵 **bugs** (L23): When `checkedAt > 0` but both `yapper` and `contexgin` are null (e.g., neither service is configured in the health payload), the component renders an empty `<div className="service-status">` with vertical padding (`var(--space-2) 0`), creating a visible gap in the layout with no content. Consider returning null when both services are null, similar to the `checkedAt === 0` guard. `[fixable]`


### `server/health-monitor.ts`
_Clean consolidation of per-hook health polling into SSE-driven events. Well-tested, follows existing codebase patterns. Minor suggestions around empty-state rendering and edge-case test coverage._

- 🟡 **unsafe_assumptions** (L19): The `SERVICES` array reads `process.env.YAPPER_PROXY_TARGET` and `process.env.CONTEXGIN_URL` at module load time (top-level const). This means env vars set after import are ignored, and the URLs cannot be overridden in tests without mocking the module. This works today because these env vars are set in `.env` before startup, but it's worth noting the coupling. Not blocking since the same pattern is used elsewhere in the codebase. `[fixable]`
- 🔵 **unsafe_assumptions** (L47): `start()` calls `this.check()` (fire-and-forget async) then immediately sets the interval. If the first check takes longer than `POLL_INTERVAL_MS` (unlikely at 30s, but possible under extreme network conditions), two checks could overlap. The current `setInterval` doesn't wait for the previous check to complete. Consider `setTimeout` chaining if overlapping checks become a concern, though at 30s intervals with 3s timeouts this is very unlikely. `[fixable]`


### `frontend/src/hooks/useServiceHealth.ts`
_Clean consolidation of per-hook health polling into SSE-driven events. Well-tested, follows existing codebase patterns. Minor suggestions around empty-state rendering and edge-case test coverage._

- 🔵 **style** (L19): The `data as ServiceHealthPayload` cast bypasses type safety. The eventBus `on` callback receives `unknown` data. While the server always sends the correct shape, a Zod parse (or at minimum a runtime shape check) would be more defensive. Other SSE consumers in the codebase may use similar casts, so this is consistent — but worth flagging. `[fixable]`


### `server/__tests__/health-monitor.test.ts`
_Clean consolidation of per-hook health polling into SSE-driven events. Well-tested, follows existing codebase patterns. Minor suggestions around empty-state rendering and edge-case test coverage._

- 🔵 **missing_tests**: No test covers the `AbortSignal.timeout(CHECK_TIMEOUT_MS)` behavior — i.e., what happens when a health check times out (hangs for >3s). A test with a never-resolving fetch promise and advancing timers past 3s would verify the timeout path returns `ok: false`. `[fixable]`
- 🔵 **missing_tests**: No test verifies that calling `start()` twice is idempotent (doesn't create duplicate intervals). The guard `if (this.timer) return` exists at line 44 but isn't tested. `[fixable]`

PR #309, opened 2026-05-03
feat: subagent visibility - nested agent execution UI

dimakis:
## Summary

Full implementation of subagent visibility — makes nested agent execution (Agent tool spawning subagents) visible in the Mitzo UI instead of treating them as black-box tool calls.

### Phase 1: Protocol Types
- Added `SubagentUsage`, `StreamingSubagentState`, `FinishedSubagentState` types
- Extended `StreamingBlock` and `FinishedBlock` with optional `subagent` field
- Added 6 new WS message types: `subagent_start`, `subagent_block_{start,delta,end}`, `subagent_tool_result`, `subagent_end`
- Tests: `packages/protocol/__tests__/subagent-types.test.ts` (5 passing)

### Phase 2: Server Detection & Event Emission
- Track `activeSubagents` Map keyed by `parent_tool_use_id` in query-loop
- Detect when SDK events have `parent_tool_use_id` set (agent spawned a subagent)
- Emit `subagent_*` wrapper events instead of normal block events for subagent turns
- Persistent `toolId → blockId` map for parent block lookup
- Structured logging for subagent lifecycle (started, tool call, completed)
- Tests: `server/__tests__/subagent-events.test.ts` (3 passing)

### Phase 3: Client State & Protocol Parser
- Added 6 new `MessagesAction` types for subagent lifecycle
- Reducer: `SUBAGENT_START` initializes nested subagent state on parent tool block
- Reducer: `SUBAGENT_BLOCK_{START,DELTA,END}` manage streaming blocks within subagent
- Reducer: `SUBAGENT_TOOL_RESULT` patches tool results into subagent blocks
- Reducer: `SUBAGENT_END` converts streaming → finished state with summary + usage
- Protocol parser routes 6 new WS messages to reducer actions
- Tests: `packages/client/__tests__/subagent-reducer.test.ts` (6 passing)

### Phase 4: SubagentCard UI Component
- `SubagentCard` component renders nested subagent activity inside parent Agent pill
- Collapsible header: summary/status, token usage (input↓ output↑), chevron
- Pulsing dot when running, checkmark when complete
- Expanded view shows nested blocks: `ThinkingBlock`, `ToolPill`, text content
- Indented card with accent border-left (`--success`), subtle background
- Integrated into `ToolPill`: renders after result section when `block.subagent` exists
- Tests: `frontend/src/components/__tests__/SubagentCard.test.tsx` (5 passing)

## What's Visible Now

When Claude spawns a subagent via the Agent tool:
- **Header**: "Working..." → final summary, token usage, expandable
- **Expanded**: Full subagent thinking, tool calls, results nested inside parent Agent pill
- **Progress**: Pulsing indicator while running, checkmark when done

## Design Doc

Based on `local_features/subagent-visibility/spec.md` (PR #87 in mgmt)

## Test Coverage

**19 new tests**, all passing:
- Protocol types: 5 tests
- Server event emission: 3 tests
- Client reducer: 6 tests
- UI component: 5 tests

## Next Steps (Future PRs)

- Deep OTel spans (nested `subagent` → `subagent.tool` spans in Jaeger)
- Recursive nesting (subagents spawning subagents)
- Token rollup (parent + subagent cumulative usage)

### DIFF
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -14,6 +14,7 @@
   --accent-hover: #5a52e0;
   --danger: #f44336;
   --success: #4caf50;
+  --success-rgb: 76, 175, 80;
   --code-bg: #0e0e10;
   --code-font: 'SF Mono', 'Fira Code', 'Cascadia Code', 'Menlo', monospace;
 
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -5237,3 +5237,87 @@ textarea:focus {
 .isolation-toggle:hover {
   opacity: 0.8;
 }
+
+/* ━━━ Subagent Card ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ */
+
+.subagent-card {
+  margin: 8px 0 0 0;
+  padding-left: 12px;
+  border-left: 2px solid var(--success);
+  background: rgba(var(--success-rgb), 0.05);
+  border-radius: 4px;
+  overflow: hidden;
+}
+
+.subagent-header {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  padding: 8px 12px 8px 0;
+  width: 100%;
+  cursor: pointer;
+  background: transparent;
+  border: none;
+  -webkit-tap-highlight-color: transparent;
+  touch-action: manipulation;
+  font-family: inherit;
+  font-size: var(--text-sm);
+  color: var(--text);
+}
+
+.subagent-header:active {
+  background: rgba(255, 255, 255, 0.03);
+}
+
+.subagent-dot {
+  width: 6px;
+  height: 6px;
+  border-radius: 50%;
+  flex-shrink: 0;
+}
+
+.subagent-dot--running {
+  background: var(--success);
+  animation: pulse-dot 1.2s ease-in-out infinite;
+}
+
+.subagent-dot--done {
+  background: var(--success);
+}
+
+.subagent-summary {
+  flex: 1;
+  text-align: left;
+  font-weight: 500;
+  color: var(--text);
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.subagent-tokens {
+  font-size: var(--text-xs);
+  color: var(--text-dim);
+  font-variant-numeric: tabular-nums;
+  white-space: nowrap;
+  flex-shrink: 0;
+}
+
+.subagent-chevron {
+  color: var(--text-dim);
+  font-size: var(--text-xs);
+  flex-shrink: 0;
+}
+
+.subagent-detail {
+  padding: 0 12px 8px 0;
+}
+
+.subagent-text {
+  padding: 8px 0;
+  color: var(--text-dim);
+  font-size: var(--text-sm);
+  line-height: 1.5;
+  white-space: pre-wrap;
+  word-break: break-word;
+}
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -152,6 +152,7 @@ export function finishCurrent(current: StreamingMessage): FinishedMessage {
       rawInput: b.rawInput,
       toolResult: b.toolResult,
       toolError: b.toolError,
+      subagent: b.subagent,
     };
   });
   return { messageId: current.messageId, role: 'assistant', blocks, timestamp: Date.now() };
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -77,6 +77,43 @@ export type MessagesAction =
     }
   | { type: 'MESSAGE_END'; messageId: string; sessionId?: string }
   | { type: 'SESSION_END'; sessionId?: string }
+  // Subagent events
+  | { type: 'SUBAGENT_START'; parentBlockId: string; subagentMessageId: string }
+  | {
+      type: 'SUBAGENT_BLOCK_START';
+      parentBlockId: string;
+      blockId: string;
+      blockType: BlockType;
+      toolName?: string;
+    }
+  | { type: 'SUBAGENT_BLOCK_DELTA'; parentBlockId: string; blockId: string; delta: string }
+  | {
+      type: 'SUBAGENT_BLOCK_END';
+      parentBlockId: string;
+      blockId: string;
+      toolName?: string;
+      toolId?: string;
+      input?: string;
+      rawInput?: RawToolInput;
+    }
+  | {
+      type: 'SUBAGENT_TOOL_RESULT';
+      parentBlockId: string;
+      toolId: string;
+      result: string;
+      isError: boolean;
+    }
+  | {
+      type: 'SUBAGENT_END';
+      parentBlockId: string;
+      summary?: string;
+      usage?: {
+        inputTokens: number;
+        outputTokens: number;
+        cacheReadTokens: number;
+        cacheCreationTokens: number;
+      };
+    }
   // Reattach snapshot
   | { type: 'MESSAGE_SNAPSHOT'; messageId: string; blocks: FinishedBlock[] }
   // Session / UI lifecycle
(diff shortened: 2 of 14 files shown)


server/query-loop.ts
```diff
@@ -693,6 +873,58 @@ async function _runQueryLoopInner(
...
+                  toolId: toolEntry.id,
+                });
+              } else if (blockId) {
+                // Text or thinking block
+                emit(
+                  v2('subagent_block_end', {
+                    parentBlockId: subagent.parentBlockId,
+                    subagentMessageId: subagent.subagentMessageId,
+                    blockId,
+                    blockType: subagent.subagentBlockIdByIndex.get(index) ? 'text' : 'thinking',
```
dimakis: 🔴 **bugs**: blockType is always 'text', never 'thinking'. The condition `subagent.subagentBlockIdByIndex.get(index)` is always truthy here because `blockId` was obtained from that same map on line 882, and we're inside `else if (blockId)`. The block type should be tracked separately (e.g., store the blockType in `subagentBlockIdByIndex` or a parallel map) instead of re-deriving it from map presence. `[fixable]`

server/query-loop.ts
```diff
@@ -510,6 +555,54 @@ async function _runQueryLoopInner(
           log.debug('stream event', { clientId, evtType: evt?.type });
 
           if (evt?.type === 'message_start') {
+            const parentToolUseId = msg.parent_tool_use_id as string | undefined;
+
+            // Subagent message_start — emit subagent_start instead of normal message_start
+            if (parentToolUseId) {
+              const parentBlockId = toolIdToBlockId.get(parentToolUseId);
+              const apiMsg = evt.message as Record<string, unknown> | undefined;
```
dimakis, line 563: 🟡 **unsafe_assumptions**: `parentToolName` is hardcoded to `'Agent'` with the comment 'We don't track this persistently, assume Agent'. If other tools spawn subagents in the future (or the tool name changes), this will be wrong. Consider looking up the tool name from `toolIdToBlockId` → block data, or dropping the field if it's unused. `[fixable]`

frontend/src/styles/global.css
```diff
@@ -5237,3 +5237,87 @@ textarea:focus {
...
   opacity: 0.8;
 }
+
+/* ━━━ Subagent Card ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ */
+
+.subagent-card {
+  margin: 8px 0 0 0;
+  padding-left: 12px;
+  border-left: 2px solid var(--success);
+  background: rgba(var(--success-rgb), 0.05);
```
dimakis, line 5248: 🟡 **bugs**: CSS variable `--success-rgb` is used but never defined. Only `--success: #4caf50` exists. The `rgba(var(--success-rgb), 0.05)` declaration will silently fail, making the `.subagent-card` background transparent instead of a subtle green tint. Either define `--success-rgb: 76, 175, 80;` in `:root` or use a different approach like `color-mix()`. `[fixable]`

packages/client/src/slices/messages.ts
```diff
@@ -466,6 +503,177 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
...
+    }
+
+    case 'SUBAGENT_END': {
+      if (!state.current) return state;
+      const parentBlock = state.current.blocks.get(action.parentBlockId);
+      if (!parentBlock?.subagent) return state;
+
+      // Convert streaming subagent state to finished state
+      const finishedBlocks: FinishedBlock[] = parentBlock.subagent.blockOrder.map((blockId) => {
+        const b = parentBlock.subagent!.blocks.get(blockId)!;
```
dimakis, line 650: 🟡 **bugs**: Non-null assertion `parentBlock.subagent!.blocks.get(blockId)!` will throw if `blockOrder` contains a blockId that isn't in the blocks Map (e.g., due to a missed SUBAGENT_BLOCK_START). Since this is a reducer (called on every render), a crash here would break the entire UI. Add a guard: skip missing blocks or filter them out. `[fixable]`

dimakis (commented):
## Centaur Review

Found **8** issue(s) (2 critical) (5 warning).


### `server/query-loop.ts`
_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🔴 **bugs** (L921): blockType is always 'text', never 'thinking'. The condition `subagent.subagentBlockIdByIndex.get(index)` is always truthy here because `blockId` was obtained from that same map on line 882, and we're inside `else if (blockId)`. The block type should be tracked separately (e.g., store the blockType in `subagentBlockIdByIndex` or a parallel map) instead of re-deriving it from map presence. `[fixable]`
- 🟡 **regressions**: `forceFlushPendingMessage()` doesn't emit `subagent_end` for entries in `activeSubagents`. If a session disconnects or aborts mid-subagent, the client will see a perpetually 'running' subagent with no end event. Consider emitting `subagent_end` (without usage) for all remaining `activeSubagents` entries in `forceFlushPendingMessage` or at session teardown. `[fixable]`
- 🟡 **unsafe_assumptions** (L563): `parentToolName` is hardcoded to `'Agent'` with the comment 'We don't track this persistently, assume Agent'. If other tools spawn subagents in the future (or the tool name changes), this will be wrong. Consider looking up the tool name from `toolIdToBlockId` → block data, or dropping the field if it's unused. `[fixable]`


### `packages/protocol/src/index.ts`
_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🔴 **bugs**: Missing exports: `SubagentCard.tsx` imports `FinishedSubagentState` and `StreamingSubagentState` from `@mitzo/protocol`, but these types are not re-exported from `packages/protocol/src/index.ts`. The `SubagentUsage` and `SubagentState` types are also not exported. This will cause a compile error. `[fixable]`


_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🟡 **bugs** (L5247): CSS variable `--success-rgb` is used but never defined. Only `--success: #4caf50` exists. The `rgba(var(--success-rgb), 0.05)` declaration will silently fail, making the `.subagent-card` background transparent instead of a subtle green tint. Either define `--success-rgb: 76, 175, 80;` in `:root` or use a different approach like `color-mix()`. `[fixable]`


_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🟡 **bugs** (L649): Non-null assertion `parentBlock.subagent!.blocks.get(blockId)!` will throw if `blockOrder` contains a blockId that isn't in the blocks Map (e.g., due to a missed SUBAGENT_BLOCK_START). Since this is a reducer (called on every render), a crash here would break the entire UI. Add a guard: skip missing blocks or filter them out. `[fixable]`


_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🟡 **missing_tests**: No tests for `parseServerMessage` handling of the six new subagent message types. The reducer tests verify state transitions, but the protocol parser's mapping from server messages to reducer actions is untested — type casting errors or incorrect field mappings (e.g., `msg.parentBlockId as string`) won't be caught. `[fixable]`


_Two critical bugs (wrong blockType in subagent_block_end always emitting 'text', and missing protocol type exports that will fail compilation), plus a missing CSS variable and no cleanup of orphaned subagents on disconnect._

- 🔵 **missing_tests**: Tests don't cover edge cases: (1) subagent with no blocks (empty turn), (2) multiple concurrent subagents from different parent tools, (3) subagent interrupted mid-stream (abort signal). The happy-path coverage is good but these edge cases are where the Map-based tracking is most likely to break. `[fixable]`

dimakis:
## R2 Review: Subagent Visibility UI

This PR adds nested agent execution visibility to the Mitzo UI — subagent events flow from query-loop through the protocol to the frontend, rendering as collapsible cards under tool blocks. The protocol design is clean and the reducer logic is thorough, but there's a critical blockType bug and a missing type export that will break compilation.

### Critical Issues

🔴 **blockType always 'text' in subagent_block_end** (query-loop.ts:1758): The ternary `subagent.subagentBlockIdByIndex.get(index) ? 'text' : 'thinking'` is always truthy because we're inside `else if (blockId)` where `blockId = subagent.subagentBlockIdByIndex.get(index)`. The actual blockType from `content_block_start` is never stored — only the blockId is. This means all subagent blocks (text, thinking, tool_use) will emit `blockType: 'text'` in the `_end` event, breaking the UI rendering (ThinkingBlock won't render for thinking blocks). Fix: store `{blockId, blockType}` in `subagentBlockIdByIndex` (make it a Map<number, {id: string, type: string}>), then look up the type in `content_block_stop`. `[must fix]`

🔴 **Missing type exports in @mitzo/protocol** (packages/protocol/src/index.ts): `SubagentCard.tsx` imports `FinishedSubagentState` and `StreamingSubagentState` from `@mitzo/protocol`, but these types are not re-exported from `packages/protocol/src/index.ts`. TypeScript compilation will fail with "cannot find name". Add to exports: `export type { SubagentUsage, StreamingSubagentState, FinishedSubagentState, SubagentState } from './types.js';` `[must fix]`

### Warnings

🟡 **No cleanup of orphaned subagents on disconnect** (query-loop.ts:2060): `activeSubagents` is a Map keyed by `parentToolUseId`. If a session disconnects or is aborted mid-subagent (e.g., iOS background kill), the entry is never removed from the map and `subagent_end` is never emitted. The client will see a perpetually "running" subagent. Fix: in the query loop's `finally` block (or in `forceFlushPendingMessage`), emit `subagent_end` (without usage) for all remaining `activeSubagents.keys()`. `[should fix]`

🟡 **CSS variable `--success-rgb` undefined** (global.css:231): `.subagent-card` background uses `rgba(var(--success-rgb), 0.05)` but `--success-rgb` is never defined in `:root`. Only `--success: #4caf50` exists. The rgba will silently fail (transparent background instead of subtle green). Add `--success-rgb: 76, 175, 80;` to `:root` in global.css. `[should fix]`

🟡 **Unsafe non-null assertion in SUBAGENT_END reducer** (messages.ts:991): `parentBlock.subagent!.blocks.get(blockId)!` double non-null assertion assumes `blockOrder` always contains valid blockIds. If a `SUBAGENT_BLOCK_START` was missed (e.g., due to a network race), `blockOrder` will reference a missing block and the reducer will throw, crashing the UI. Add a guard: `const b = parentBlock.subagent!.blocks.get(blockId); if (!b) continue;` `[should fix]`

### Nits

🔵 **parentToolName hardcoded to 'Agent'** (query-loop.ts:1561): Comment says "We don't track this persistently, assume Agent". If other tools (e.g., a future `Planner` tool) spawn subagents, this will be wrong. Not urgent, but consider looking up the tool name from `toolIdToBlockId` → parent block's `toolName`. `[nit]`

🔵 **Protocol parser uses bare 'as' casts** (protocol-parser.ts:732): `msg.parentBlockId as string` assumes the server always sends the field. If missing, the reducer will silently no-op (fine) but the type system won't catch it. Consider adding a runtime guard or `?? ''` fallback for critical fields like `parentBlockId`. `[nit]`

🔵 **No test for subagent tool_result relay path**: Tests cover `subagent_start`, block lifecycle, and `subagent_end`, but the `msg.type === 'user'` path that emits `subagent_tool_result` (query-loop.ts:1783) is untested. Add a test case with a tool_result block inside a subagent turn. `[test gap]`

### Verdict

**REQUEST CHANGES**

The blockType bug will break the UI for thinking/tool_use blocks in subagents, and the missing type export will fail compilation. Fix those two, and this is solid. The orphaned subagent cleanup and CSS variable are important but can be follow-ups if needed.

dimakis:
## Centaur Review

Found **7** issue(s) (1 critical) (3 warning).


_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🔴 **bugs** (L155): Type mismatch: `finishCurrent()` copies `b.subagent` (typed `StreamingSubagentState`) directly into a `FinishedBlock.subagent` field (typed `FinishedSubagentState`). These types are structurally incompatible — `StreamingSubagentState.blocks` is a `Map<string, StreamingBlock>` while `FinishedSubagentState.blocks` is `FinishedBlock[]`, and `running: true` vs `running?: never`. This should fail to compile under strict mode. At runtime, if `finishCurrent()` is called before SUBAGENT_END (e.g. during SESSION_END force-finalization), the subagent will have Map blocks that downstream code (SubagentCard) would need to handle. The fix: convert the subagent state in `finishCurrent()` the same way SUBAGENT_END does — map the streaming blocks to finished blocks and drop the `running` flag. `[fixable]`


### `server/query-loop.ts`
_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🟡 **bugs** (L923): Narrow type cast `blockEntry!.blockType as 'text' | 'thinking'` excludes `'redacted_thinking'`, which is a valid block type handled at content_block_start (line 680). The runtime value will be correct, but the TypeScript type is wrong. Should be `as BlockType` or `as 'text' | 'thinking' | 'redacted_thinking'`. `[fixable]`
- 🟡 **regressions** (L613): `toolIdToBlockId` is never cleared between turns. At message_start (line 613), `blockIdByIndex.clear()` and `toolInputBuffers.clear()` are called, but `toolIdToBlockId` grows unbounded across the session. While tool IDs are globally unique so collisions are unlikely, this is inconsistent with how sibling maps are managed and constitutes a minor memory leak over long sessions. `[fixable]`


_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🟡 **missing_tests**: No test covers the case where `finishCurrent()` is called while a subagent is still in streaming state (i.e., SESSION_END arrives before SUBAGENT_END). This is the exact scenario where the type mismatch bug on line 155 would manifest. The existing `finishCurrent` tests only check finished and absent subagent states. `[fixable]`


_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🔵 **missing_tests**: No test covers error/abort cleanup of `activeSubagents`. If the query loop errors mid-subagent, entries in `activeSubagents` are never cleaned up. The finally block should call `activeSubagents.clear()` and a test should verify this. `[fixable]`


_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🔵 **style** (L19): `subagent.summary` is accessed without type narrowing — when `isRunning` is false, TypeScript hasn't narrowed the union type, so `subagent.summary` on a `StreamingSubagentState` accesses a `never` property. Use a type guard: `const summary = 'summary' in subagent && subagent.summary ? subagent.summary : isRunning ? 'Working...' : 'Complete'`. `[fixable]`


_The subagent visibility feature has solid architecture and good test coverage for happy paths, but has a critical type mismatch in `finishCurrent()` where streaming subagent state is assigned to a finished subagent field without conversion, plus a few edge cases around error cleanup and type narrowing._

- 🔵 **style**: The `@keyframes pulse-dot` animation referenced by `.subagent-dot--running` is not defined in this diff. If it doesn't already exist in global.css, the pulsing indicator won't animate. `[fixable]`

PR #310, opened 2026-05-03
fix(connection-registry): add delivery guarantee with cursors and periodic sync

dimakis:
## Problem

Events could be lost when `broadcast()` failed during:
- iOS backgrounding killing WebSocket mid-stream  
- Network timing races (WS closes before send completes)
- Connection drops during event delivery

The bug manifested as stuck "Running..." UI — subagents completed but clients never received the `session_end` event.

**Reproduction**: From production logs (session `42fda507`, 2026-05-03 ~21:22):
1. User sends message "Add me some cards for dreame in HA"
2. Agent spawns subagent ("Explore HA Dreame entities")  
3. iOS suspends app → WS dies mid-stream
4. Server-side subagent completes, query loop ends with `doneSent: false`
5. Client reconnects, replays events from EventStore
6. Query loop already ended → no more events to replay  
7. UI shows "Running..." forever

## Root Cause

Event delivery was fire-and-forget. `broadcast()` sent events via WS but had no retry logic. If `send()` failed, the event was lost. EventStore replay only worked if the client reconnected **before** the query loop ended — events lost in that timing window were gone forever.

## Solution

Implemented **connection-level event cursors with periodic sync**:

### 1. Cursor Tracking
Each connection maintains `last-delivered seq` per session. Updated on successful `broadcast()`.

### 2. Periodic Sync  
Background task (5s interval) retries missed events by comparing cursor vs EventStore. Bounded to 50 events/round to avoid overwhelming slow connections.

### 3. Reconnect Reset
Cursor resets to client's `lastSeq` on reconnect to prevent duplicate replay (EventStore handles the gap).

### 4. Out-of-Order Safety
Cursor only advances forward, handles reordered delivery gracefully.

### `ConnectionRegistry` (`packages/harness/src/connection-registry.ts`)
- Add `cursors` map (per-connection per-session tracking)
- Add `EventStoreAdapter` interface (injected to avoid circular deps)  
- Update `broadcast()` to track delivery success and update cursors
- Add `resetCursor()` for reconnect cursor sync
- Add `startPeriodicSync()` / `stopPeriodicSync()` for background retry
- Add `setEventStore()` for EventStore injection
- Add `dispose()` for graceful cleanup

### `ws-handler-v2.ts` (`server/ws-handler-v2.ts`)
- Call `resetCursor()` in `handleReconnect()` after EventStore replay

### `server/index.ts`
- Wire `EventStore` to `ConnectionRegistry` via `setEventStore()`
- Start periodic sync in `server.listen()` callback
- Stop periodic sync in `shutdown()` function

### Tests (`packages/harness/__tests__/connection-registry.test.ts`)
Added 14 new tests covering:
- Cursor initialization and cleanup
- Cursor update on successful broadcast  
- Cursor preservation on failed send
- Out-of-order delivery handling
- Reconnect cursor reset
- Periodic sync retry logic
- Batch size limiting (50 events/round)
- EventStore fetch error handling
- Closed connection skipping
- Sync start/stop lifecycle

## Edge Cases Handled

| Scenario | Behavior |
|----------|----------|
| Broadcast fails mid-delivery | Cursor stays behind, periodic sync retries within 5s |
| EventStore append + broadcast race | Sync catches missed event within 5s |
| Slow client (backpressure) | Sync rate-limited to 50 events/round |
| Multi-device sessions | Each connection has independent cursor |
| Observer connections | Same cursor mechanism applies |
| Session ends while suspended | Sync delivers missed events on resume |
| Reconnect with stale cursor | Reset to client's `lastSeq` prevents dupes |
| Out-of-order delivery | Cursor only advances forward (max seen) |
| EventStore fetch fails | Logged, next sync round retries |
| Sync already running | Warning logged, no-op |

## Testing

All tests pass:
- ✅ `packages/harness/__tests__/connection-registry.test.ts` — 35 tests (14 new)
- ✅ `server/__tests__/ws-handler-v2.test.ts` — 99 tests (no regressions)  
- ✅ Full test suite — 2437 tests pass

## Deployment Notes

- **Zero config required** — periodic sync starts automatically on server boot
- **No schema changes** — uses existing EventStore
- **Backward compatible** — clients without `seq` support work fine (cursors stay at 0)
- **Performance impact** — minimal (5s interval, 50 events/round cap)

Fixes the stalled Agent tool bug identified in production (follow-up to PR #239).

### DIFF
--- a/frontend/src/components/SubagentCard.tsx
+++ b/frontend/src/components/SubagentCard.tsx
@@ -1,0 +1,64 @@
+import { useState } from 'react';
+import type { FinishedSubagentState, StreamingSubagentState } from '@mitzo/protocol';
+import { ThinkingBlock } from './ThinkingBlock';
+import { ToolPill } from './ToolPill';
+
+interface SubagentCardProps {
+  subagent: FinishedSubagentState | StreamingSubagentState;
+}
+
+function formatTokens(usage?: { inputTokens: number; outputTokens: number }): string {
+  if (!usage) return '';
+  return `${usage.inputTokens}↓ ${usage.outputTokens}↑`;
+}
+
+export function SubagentCard({ subagent }: SubagentCardProps) {
+  const [expanded, setExpanded] = useState(false);
+
+  const isRunning = 'running' in subagent && subagent.running;
+  const summary = isRunning ? 'Working...' : subagent.summary || 'Complete';
+  const usage = 'usage' in subagent ? subagent.usage : undefined;
+
+  // Convert blocks to array for rendering
+  const blocks = Array.isArray(subagent.blocks)
+    ? subagent.blocks
+    : Array.from(subagent.blocks.values());
+
+  return (
+    <div className="subagent-card">
+      <button
+        className="subagent-header"
+        onClick={() => setExpanded((e) => !e)}
+        aria-expanded={expanded}
+      >
+        <span
+          className={`subagent-dot ${isRunning ? 'subagent-dot--running' : 'subagent-dot--done'}`}
+        />
+        <span className="subagent-summary">{summary}</span>
+        {usage && <span className="subagent-tokens">{formatTokens(usage)}</span>}
+        <span className="subagent-chevron">{expanded ? '▾' : '▸'}</span>
+      </button>
+
+      {expanded && (
+        <div className="subagent-detail">
+          {blocks.map((block) => {
+            if (block.blockType === 'thinking' || block.blockType === 'redacted_thinking') {
+              return <ThinkingBlock key={block.blockId} block={block} />;
+            }
+            if (block.blockType === 'tool_use') {
+              return <ToolPill key={block.blockId} block={block} />;
+            }
+            if (block.blockType === 'text') {
+              return (
+                <div key={block.blockId} className="subagent-text">
+                  {block.content}
+                </div>
+              );
+            }
+            return null;
+          })}
+        </div>
+      )}
+    </div>
+  );
+}
--- a/frontend/src/components/ToolPill.tsx
+++ b/frontend/src/components/ToolPill.tsx
@@ -1,5 +1,6 @@
 import { useState } from 'react';
 import type { StreamingBlock, FinishedBlock, RawToolInput } from '../types/chat';
+import { SubagentCard } from './SubagentCard';
 
 interface Props {
   block: StreamingBlock | FinishedBlock;
@@ -71,6 +72,7 @@ export function ToolPill({ block }: Props) {
           )}
         </div>
       )}
+      {block.subagent && <SubagentCard subagent={block.subagent} />}
     </div>
   );
 }
--- a/frontend/src/components/__tests__/SubagentCard.test.tsx
+++ b/frontend/src/components/__tests__/SubagentCard.test.tsx
@@ -1,0 +1,120 @@
+// @vitest-environment jsdom
+import { describe, it, expect, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup } from '@testing-library/react';
+import { SubagentCard } from '../SubagentCard';
+import type { FinishedBlock } from '../../types/chat';
+
+afterEach(() => cleanup());
+
+describe('SubagentCard', () => {
+  it('renders collapsed by default with summary', () => {
+    const subagent = {
+      messageId: 'msg-sub-1',
+      blocks: [
+        {
+          blockId: 'b1',
+          blockType: 'text' as const,
+          content: 'Subagent output',
+        },
+      ],
+      summary: 'Search complete',
+      usage: {
+        inputTokens: 100,
+        outputTokens: 50,
+        cacheReadTokens: 0,
+        cacheCreationTokens: 0,
+      },
+    };
+
+    render(<SubagentCard subagent={subagent} />);
+
+    expect(screen.getByText('Search complete')).toBeTruthy();
+    expect(screen.getByText(/100.*50/)).toBeTruthy(); // Token display
+  });
+
+  it('renders "Working..." when subagent is still running', () => {
+    const subagent = {
+      messageId: 'msg-sub-1',
+      blocks: new Map([
+        [
+          'b1',
+          {
+            blockId: 'b1',
+            blockType: 'thinking' as const,
+            content: 'Analyzing...',
+            done: false,
+          },
+        ],
+      ]),
+      blockOrder: ['b1'],
+      running: true as const,
+    };
+
+    render(<SubagentCard subagent={subagent} />);
+
+    expect(screen.getByText('Working...')).toBeTruthy();
+  });
+
+  it('expands to show nested blocks when clicked', () => {
+    const blocks: FinishedBlock[] = [
+      {
+        blockId: 'b1',
+        blockType: 'thinking',
+        content: 'Let me search for that',
+      },
+      {
+        blockId: 'b2',
+        blockType: 'text',
+        content: 'Search results here',
+      },
+    ];
+
+    const subagent = {
+      messageId: 'msg-sub-1',
+      blocks,
+      summary: 'Done',
+    };
+
+    const { container } = render(<SubagentCard subagent={subagent} />);
+
+    // Initially collapsed - detail not visible
+    expect(container.querySelector('.subagent-detail')).not.toBeTruthy();
+
+    // Click to expand
+    const header = container.querySelector('.subagent-header');
+    if (header) {
+      fireEvent.click(header);
+    }
+
+    // Now detail section is visible
+    expect(container.querySelector('.subagent-detail')).toBeTruthy();
+    expect(screen.getByText('Search results here')).toBeTruthy();
+  });
+
+  it('shows pulsing indicator when running', () => {
+    const subagent = {
+      messageId: 'msg-sub-1',
+      blocks: new Map(),
+      blockOrder: [],
+      running: true as const,
+    };
+
+    const { container } = render(<SubagentCard subagent={subagent} />);
+
+    const dot = container.querySelector('.subagent-dot--running');
+    expect(dot).toBeTruthy();
+  });
+
+  it('shows checkmark when complete', () => {
+    const subagent = {
+      messageId: 'msg-sub-1',
+      blocks: [],
+      summary: 'Complete',
+    };
+
+    const { container } = render(<SubagentCard subagent={subagent} />);
+
+    const dot = container.querySelector('.subagent-dot--done');
+    expect(dot).toBeTruthy();
+  });
+});
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -14,6 +14,7 @@
   --accent-hover: #5a52e0;
   --danger: #f44336;
   --success: #4caf50;
+  --success-rgb: 76, 175, 80;
   --code-bg: #0e0e10;
   --code-font: 'SF Mono', 'Fira Code', 'Cascadia Code', 'Menlo', monospace;
 
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -5237,3 +5237,87 @@ textarea:focus {
 .isolation-toggle:hover {
   opacity: 0.8;
 }
+
+/* ━━━ Subagent Card ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ */
+
+.subagent-card {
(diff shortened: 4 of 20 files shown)


dimakis:
## Centaur Review

Found **8** issue(s) (2 critical) (3 warning).


### `server/query-loop.ts`
_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🔴 **bugs** (L921): blockType ternary is always 'text'. The condition `subagent.subagentBlockIdByIndex.get(index)` is guaranteed truthy because we're in the `else if (blockId)` branch where `blockId = subagent.subagentBlockIdByIndex.get(index)`. The block type from content_block_start is not stored — only the blockId is. The fix is to store {blockId, blockType} in subagentBlockIdByIndex so content_block_stop can look up the original type. `[fixable]`


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🔴 **bugs** (L145): finishCurrent() does not preserve the `subagent` field when converting StreamingBlock → FinishedBlock. When MESSAGE_END, MESSAGE_START (finalizing orphan), or SESSION_END triggers finishCurrent(), all subagent state is silently dropped. Add `subagent: b.subagent` (with appropriate streaming→finished conversion) to the returned object. `[fixable]`
- 🟡 **regressions** (L503): SUBAGENT_END converts blocks from Map to array but the RESTORE action (which validates FinishedMessage shape) doesn't account for subagent state on blocks. If a restored session has blocks with subagent data, the data will be preserved (it's on the FinishedBlock), but only if finishCurrent() is also fixed (see related finding). Verify RESTORE round-trips subagent state correctly.


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🟡 **bugs** (L5247): CSS variable `--success-rgb` is used in `rgba(var(--success-rgb), 0.05)` but never defined in :root. `--success` is `#4caf50` — add `--success-rgb: 76, 175, 80;` to the :root block. Without it, the background of .subagent-card is transparent (silently fails). `[fixable]`


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🟡 **unsafe_assumptions** (L23): Calling `Array.from(subagent.blocks.values())` assumes blocks is always a Map when not an array. However, the SubagentState union type allows blocks to be either `Map<string, StreamingBlock>` or `FinishedBlock[]`. If a future code path passes a plain object, this will throw. The type narrowing is correct for the current union, but adding an explicit type guard (e.g. `blocks instanceof Map`) would be safer than duck-typing on `Array.isArray`. `[fixable]`


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🔵 **style** (L443): Subagent event parsing uses bare `as` casts (e.g., `msg.parentBlockId as string`) instead of runtime validation. The existing parser follows this pattern, so it's consistent, but these fields are critical for reducer correctness — a missing parentBlockId would silently produce an action that no-ops in the reducer. Consider adding a guard or at least a `?? ''` fallback. `[fixable]`


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🔵 **missing_tests**: No test covers subagent tool_result emission (the `msg.type === 'user'` path with `parent_tool_use_id`). The test file covers subagent_start, block lifecycle, and subagent_end, but the tool result relay through the 'user' event type is untested. This is the path at query-loop.ts:1050-1075. `[fixable]`


_Two critical bugs (blockType ternary always evaluates to 'text' in subagent block_stop, and finishCurrent() drops subagent state), one missing CSS variable, and a few test gaps — otherwise solid architecture with clean separation between connection-level delivery guarantees and subagent visibility._

- 🔵 **missing_tests**: No test for edge case: SUBAGENT_START on a non-existent parentBlockId (should be a no-op). Also no test for SUBAGENT_BLOCK_DELTA/END when subagent or subBlock is missing. The reducer handles these with early returns, but the behavior is unverified. `[fixable]`

dimakis:
## R2 Review: Connection Registry Delivery Guarantee

This PR adds cursor tracking and periodic sync to ConnectionRegistry, implementing a delivery guarantee layer for the v2 WebSocket protocol. The architecture is sound — cursors track what was delivered, periodic sync retries gaps — but there are several critical bugs and edge cases.

### Critical Issues

🔴 **Cursor update race in reconnect flow** (ws-handler-v2.ts:371): After replaying buffered events, the cursor is reset to `newCursor = events.length > 0 ? events[events.length - 1].seq : entry.lastSeq`. But if periodic sync fires between the replay and the cursor reset, it will see the old cursor (0 or client's lastSeq) and re-deliver the same events that were just replayed. The cursor reset should happen **before** replay starts, not after. Move `resetCursor()` to line 295 (right after `entry.lastSeq` is validated). `[must fix]`

🔴 **Periodic sync ignores session activity state** (connection-registry.ts:1465): The sync loop fetches events for all `watchedSessions` without checking if the session's query loop is still active. If a session ends but the connection is still open (e.g., user switches sessions), periodic sync will keep retrying events for the ended session forever. Check `eventStore.isSessionActive(sessionId)` before fetching (or add a `stoppedSessions` set and skip those). `[must fix]`

### Warnings

🟡 **No backpressure on slow clients** (connection-registry.ts:1469): `SYNC_BATCH_LIMIT = 50` bounds a single sync round, but if a client is consistently slow (e.g., mobile on bad network), the cursor will never advance past the first 50 events. The next sync round will retry the same 50, and the gap grows unbounded. Either: (a) track consecutive failures and exponentially back off, or (b) emit a `cursor_stuck` warning after N failed rounds. `[should fix]`

🟡 **EventStore fetch errors are swallowed silently** (connection-registry.ts:1470): `try/catch` around `getEventsAfter()` logs and continues. If the EventStore is persistently broken (e.g., SQLite locked), periodic sync will silently fail every 5 seconds with no user-visible signal. Consider emitting a `sync_error` event to the client or incrementing a Prometheus metric. `[should fix]`

🟡 **Test coverage gap for interleaved broadcast + sync**: The tests verify periodic sync in isolation and broadcast in isolation, but don't cover the case where broadcast() and periodic sync race on the same session. Specifically: broadcast() updates cursor on success, but if send() throws mid-batch, the cursor is left at the last successful event. Periodic sync will retry from there. If the broadcast batch was `[seq: 10, 11, 12]` and `send(seq: 11)` failed, cursor is at 10. But the client may have received 10 successfully and is waiting for 11. Periodic sync will re-send 11 (correct), but if the original broadcast continues after sync, the client gets 11 twice. Add a test for this interleaving. `[test gap]`

### Nits

🔵 **Magic number for SYNC_INTERVAL_MS**: 5 seconds is reasonable, but it's not tunable. Consider making it an env var (`MITZO_SYNC_INTERVAL_MS`) for testing or for deployments with different network characteristics. `[nit]`

🔵 **Cursor map never shrinks**: When a connection unwatches a session (`unwatch()`), the cursor entry for that session is never removed from `connCursors`. Over time, this leaks memory for long-lived connections that watch many transient sessions. Add cleanup in `unwatch()`. `[nit]`

### Verdict

**REQUEST CHANGES**

The two critical bugs (reconnect cursor race and sync ignoring session activity) will cause duplicate delivery and unbounded retries. Fix those, add the interleaving test, and this is merge-ready. The backpressure and error visibility warnings are important but can be addressed in a follow-up if needed.

dimakis:
## Centaur Review

Found **10** issue(s) (5 warning).


_Solid delivery-guarantee and subagent-visibility features with good test coverage, but MESSAGE_SNAPSHOT drops subagent state on reattach and activeSubagents are not cleaned up on abnormal query-loop exit — both will cause visible UI issues._

- 🟡 **bugs** (L289): MESSAGE_SNAPSHOT does not copy the `subagent` field when reconstructing StreamingBlocks from FinishedBlocks. Any snapshot-based reattach (iOS recovery) will silently drop subagent state from tool_use blocks. Add `subagent: b.subagent` alongside the other optional fields. `[fixable]`
- 🔵 **missing_tests** (L289): No test covers MESSAGE_SNAPSHOT with subagent state. Given the bug above (subagent field dropped), a test confirming round-trip preservation through MESSAGE_SNAPSHOT would catch this regression. `[fixable]`


### `server/query-loop.ts`
_Solid delivery-guarantee and subagent-visibility features with good test coverage, but MESSAGE_SNAPSHOT drops subagent state on reattach and activeSubagents are not cleaned up on abnormal query-loop exit — both will cause visible UI issues._

- 🟡 **bugs** (L376): activeSubagents are only cleaned up when a matching `assistant` event with `parent_tool_use_id` arrives. If the query loop terminates abnormally (abort, error) while a subagent is running, no `subagent_end` event is emitted. The client will see the subagent perpetually in 'Working...' state. The finally block should iterate over `activeSubagents` and emit synthetic `subagent_end` events before `session_end`. `[fixable]`
- 🟡 **bugs** (L617): When a parent `message_start` fires, `blockCounter` is reset to 0 and `blockIdByIndex` is cleared, but `toolIdToBlockId` is never cleared. This is intentional (subagents need to look up parent tool blocks across turns), but the map grows unboundedly for the lifetime of the query loop. For long-running sessions with many tool calls, this is a memory leak. Consider pruning entries for tools that have completed (received tool_result). `[fixable]`
- 🔵 **style** (L569): `parentToolName: 'Agent'` is hardcoded with a comment 'We don't track this persistently, assume Agent'. The actual tool name is available from `toolIdToBlockId` lookup context — or better yet, from the parent block's snapshot. If future tools (e.g., a hypothetical 'SubAgent' or 'Delegate' tool) also spawn subagents, this assumption would silently produce incorrect metadata. `[fixable]`


### `server/ws-handler-v2.ts`
_Solid delivery-guarantee and subagent-visibility features with good test coverage, but MESSAGE_SNAPSHOT drops subagent state on reattach and activeSubagents are not cleaned up on abnormal query-loop exit — both will cause visible UI issues._

- 🟡 **bugs** (L136): Reconnect replay sends events directly via `transport.send()` bypassing `ConnectionRegistry.broadcast()`, so the cursor is not updated per-event during replay. While this is safe in single-threaded Node.js (setInterval can't fire mid-synchronous loop), if `getEventsAfter` or any step becomes async in the future, periodic sync could fire mid-replay and re-send events. The final `resetCursor` at line 145 mitigates this, but adding a comment documenting the single-threaded assumption would improve safety. `[fixable]`


_Solid delivery-guarantee and subagent-visibility features with good test coverage, but MESSAGE_SNAPSHOT drops subagent state on reattach and activeSubagents are not cleaned up on abnormal query-loop exit — both will cause visible UI issues._

- 🟡 **unsafe_assumptions** (L145): broadcast() casts `data.seq` as `number | undefined` without validation. If a caller passes a non-numeric `seq` (e.g., a string), the cursor comparison `seq > current` would silently produce wrong results via JS type coercion. Consider adding a `typeof seq === 'number'` guard. `[fixable]`
- 🔵 **style** (L213): The `startPeriodicSync()` method logs an error ('cannot start periodic sync: EventStore not set') but returns silently. The caller in `server/index.ts` has no way to know sync failed to start. Consider throwing, or at least returning a boolean. `[fixable]`
- 🔵 **unsafe_assumptions** (L233): Periodic sync runs synchronously within setInterval — iterating all connections, fetching from EventStore, and retrying sends. With many connections * many sessions * 50-event batches, this could block the event loop. Consider limiting total work per tick (not just per-connection) or making the sync function async. `[fixable]`


_Solid delivery-guarantee and subagent-visibility features with good test coverage, but MESSAGE_SNAPSHOT drops subagent state on reattach and activeSubagents are not cleaned up on abnormal query-loop exit — both will cause visible UI issues._

- 🔵 **missing_tests** (L441): No test covers the new subagent_* cases in `parseServerMessage()`. The existing parser tests should have corresponding cases for each new message type to verify the field mapping (especially the `as` casts which bypass type checking). `[fixable]`

PR #312, opened 2026-05-04
fix(protocol,client): resolve subagent type mismatches breaking CI

dimakis:
## Summary

- Fixes the CI build failure on `main` caused by subagent type mismatches in `packages/client` and `frontend`
- `StreamingBlock.subagent` was typed as `StreamingSubagentState` only, but the `SUBAGENT_END` reducer replaces it with `FinishedSubagentState` (subagent finishes while parent message is still streaming)
- `finishCurrent()` copied `b.subagent` without converting from streaming to finished form
- `SubagentCard.tsx` imported `FinishedSubagentState`/`StreamingSubagentState` which weren't exported from `@mitzo/protocol`

## Changes

- **`packages/protocol/src/types.ts`**: Widen `StreamingBlock.subagent` to `StreamingSubagentState | FinishedSubagentState`
- **`packages/protocol/src/index.ts`**: Export `StreamingSubagentState`, `FinishedSubagentState`, `SubagentState`, `SubagentUsage`
- **`packages/client/src/slices/messages.ts`**: Add `finishSubagent()` helper (handles both Map→array conversion and finished passthrough), cast to `StreamingSubagentState` in streaming reducer cases

## Test plan

- [x] `packages/client/__tests__/messages-slice.test.ts` — 64 tests pass
- [x] `packages/harness/__tests__/connection-registry.test.ts` — 37 tests pass
- [x] `npx tsc -b` in frontend — clean
- [x] `npm run build` — full Vite build succeeds

### DIFF
--- a/packages/client/__tests__/messages-slice.test.ts
+++ b/packages/client/__tests__/messages-slice.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { messagesReducer, finishCurrent, INITIAL_MESSAGES_STATE } from '../src/slices/messages.js';
 import type { MessagesState } from '../src/slices/messages.js';
-import type { FinishedBlock, StreamingMessage } from '@mitzo/protocol';
+import type { FinishedBlock, FinishedSubagentState, StreamingMessage } from '@mitzo/protocol';
 
 const INITIAL = INITIAL_MESSAGES_STATE;
 
@@ -1180,8 +1180,13 @@ describe('finishCurrent', () => {
     };
 
     const finished = finishCurrent(current);
-    expect(finished.blocks[0].subagent).toBeDefined();
-    expect(finished.blocks[0].subagent!.messageId).toBe('sub-msg-1');
+    const sub = finished.blocks[0].subagent as FinishedSubagentState;
+    expect(sub).toBeDefined();
+    expect(sub.messageId).toBe('sub-msg-1');
+    // Verify streaming Map was converted to finished array
+    expect(Array.isArray(sub.blocks)).toBe(true);
+    expect(sub.blocks).toHaveLength(1);
+    expect(sub.blocks[0].content).toBe('subagent output');
   });
 
   it('works normally when subagent field is absent', () => {
--- a/packages/client/__tests__/subagent-reducer.test.ts
+++ b/packages/client/__tests__/subagent-reducer.test.ts
@@ -406,4 +406,107 @@ describe('Subagent Reducer Actions', () => {
     expect(sub.blocks[0].content).toBe('Partial');
     expect(sub.summary).toBeUndefined();
   });
+
+  it('finishSubagent skips stale blockOrder entries not in the Map', () => {
+    const state = {
+      ...INITIAL_MESSAGES_STATE,
+      current: {
+        messageId: 'msg-1',
+        blocks: new Map<string, StreamingBlock>([
+          [
+            'b1',
+            {
+              blockId: 'b1',
+              blockType: 'tool_use',
+              content: '',
+              done: true,
+              toolName: 'Agent',
+              toolId: 't1',
+              subagent: {
+                messageId: 'msg-sub-1',
+                // blockOrder references 'ghost' which is NOT in the Map
+                blocks: new Map([
+                  [
+                    'b-sub-1',
+                    {
+                      blockId: 'b-sub-1',
+                      blockType: 'text',
+                      content: 'Real',
+                      done: true,
+                    },
+                  ],
+                ]),
+                blockOrder: ['b-sub-1', 'ghost'],
+                running: true as const,
+              },
+            },
+          ],
+        ]),
+        blockOrder: ['b1'],
+      },
+    };
+
+    // Should not throw — stale entry is filtered out
+    const finished = finishCurrent(state.current!);
+    const sub = finished.blocks[0].subagent as FinishedSubagentState;
+    expect(sub.blocks).toHaveLength(1);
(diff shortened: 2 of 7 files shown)


dimakis:
## Centaur Review

_Type-safe fix for the subagent union type, but the `as StreamingSubagentState` casts bypass the new union without narrowing, and the existing test assertions may produce tsc errors under the widened type._

- 🟡 **bugs** (L560): Unsafe `as StreamingSubagentState` cast (repeated in 5 reducer cases: lines 560, 590, 614, 642, 671). If a SUBAGENT_BLOCK_* or SUBAGENT_TOOL_RESULT event arrives after SUBAGENT_END has already converted the subagent to `FinishedSubagentState`, the cast silently lies — `sub.blocks` is a `FinishedBlock[]` not a `Map`, and `sub.blockOrder` doesn't exist. This would throw at runtime. A narrowing guard (`if (!('blockOrder' in parentBlock.subagent)) return state;` or checking `blocks instanceof Map`) would make this safe without relying on event ordering guarantees. `[fixable]`
- 🔵 **missing_tests** (L148): The `finishSubagent` helper's early-return branch (`Array.isArray(sub.blocks)` — subagent already finished) is untested. This path is exercised when `finishCurrent` is called on a message whose subagent already received SUBAGENT_END. A test that runs the full SUBAGENT_START → SUBAGENT_END → MESSAGE_END sequence would cover this. `[fixable]`


_Type-safe fix for the subagent union type, but the `as StreamingSubagentState` casts bypass the new union without narrowing, and the existing test assertions may produce tsc errors under the widened type._

- 🟡 **regressions** (L39): Existing tests access `subagent?.blocks.size`, `subagent?.blocks.get(...)`, and `subagent?.blockOrder` — all Map/StreamingSubagentState-specific APIs. With the type of `StreamingBlock.subagent` widened to `StreamingSubagentState | FinishedSubagentState`, these expressions are type errors (`FinishedBlock[]` has no `.size` or `.get()`). The tests pass at runtime (vitest uses esbuild which strips types) but will fail any `tsc --noEmit` check that includes test files — potentially re-breaking the CI this PR aims to fix. `[fixable]`

dimakis:
## Centaur Review

_Solid fix that properly types the streaming→finished subagent transition and adds the missing `finishSubagent` converter. The guard pattern (`'blockOrder' in parentBlock.subagent`) is correct but repetitive — consider extracting a helper. The non-null assertions in `finishSubagent` mirror existing patterns but add another crash site if protocol invariants break._

- 🟡 **bugs** (L559): The `'blockOrder' in parentBlock.subagent` guard correctly rejects `FinishedSubagentState` (which lacks `blockOrder`), but it also means if a second `SUBAGENT_BLOCK_START` arrives after `SUBAGENT_END` has already fired (e.g., out-of-order WebSocket delivery), the event is silently dropped with no diagnostic. This is probably the right behavior, but worth a brief comment since the same guard is repeated verbatim in five case arms (SUBAGENT_BLOCK_START/DELTA/END, SUBAGENT_TOOL_RESULT, SUBAGENT_END). Consider extracting a tiny helper like `getStreamingSubagent(parentBlock)` to DRY it up and centralize the invariant. `[fixable]`
- 🟡 **unsafe_assumptions** (L155): The non-null assertion `streaming.blocks.get(blockId)!` in `finishSubagent` mirrors the same pattern in `finishCurrent` (line 173) and `SUBAGENT_END` (line 680). If `blockOrder` ever contains a stale blockId not in the Map (e.g., due to a protocol bug or dropped event), this will throw at runtime. A defensive fallback (`.get(blockId) ?? { blockId, blockType: 'text', content: '' }`) or a `filter(Boolean)` would be more resilient, matching the general pattern where the project guards against protocol edge cases (e.g., `groupBlocks()` guards against undefined input per CLAUDE.md). `[fixable]`


_Solid fix that properly types the streaming→finished subagent transition and adds the missing `finishSubagent` converter. The guard pattern (`'blockOrder' in parentBlock.subagent`) is correct but repetitive — consider extracting a helper. The non-null assertions in `finishSubagent` mirror existing patterns but add another crash site if protocol invariants break._

- 🔵 **missing_tests**: The new `finishCurrent` tests cover the happy path (already-finished passthrough) and the edge case (still-streaming conversion), but there is no test for a block with `subagent: undefined` flowing through `finishSubagent` — this is implicitly covered by the ternary in `finishCurrent` (line 184), but an explicit test would document the contract. Additionally, there's no test for `finishSubagent` when `blockOrder` references a blockId not present in the Map (the `!` assertion on line 155 would throw). This is an unlikely edge case but the `!` is a potential null dereference. `[fixable]`
- 🔵 **style** (L40): Several test assertions use `as StreamingSubagentState` or `as FinishedSubagentState` type casts. While this is a pragmatic fix for the CI type errors, these casts bypass the type checker and could mask regressions if the underlying type changes. Consider using a type-narrowing helper (e.g., an `isStreamingSubagent` function that checks `sub.blocks instanceof Map`) to keep the tests type-safe. `[fixable]`


_Solid fix that properly types the streaming→finished subagent transition and adds the missing `finishSubagent` converter. The guard pattern (`'blockOrder' in parentBlock.subagent`) is correct but repetitive — consider extracting a helper. The non-null assertions in `finishSubagent` mirror existing patterns but add another crash site if protocol invariants break._

- 🔵 **regressions** (L1183): The existing `finishCurrent` test at line 1146 sets up a streaming subagent with `running: true` but doesn't verify the output shape — it only checks `messageId` is preserved. After this PR, `finishSubagent` converts the Map to an array, so the existing test should also verify that `finished.blocks[0].subagent!.blocks` is now an array (not a Map), to confirm the new conversion logic works in the pre-existing test fixture. `[fixable]`

PR #315, opened 2026-05-06
feat(ui): desktop page navigation + New Chat fix

dimakis:
## Summary
- Adds `DesktopNav` component to the left sidebar so desktop users can navigate to Calendar, Inbox, Telos, Tasks, and Files (previously only reachable by typing URLs)
- Non-chat pages now render inside `DesktopShell` on desktop via `PageRoute` wrapper — nav sidebar on the left, no right panel
- Fixes the New Chat button which was broken because `history.replaceState` desynchronized React Router — `handleNewChat` now calls `storeNewSession()` directly

## Test plan
- [x] DesktopShell tests updated and passing (7/7)
- [x] TypeScript clean (no frontend errors)
- [ ] Manual: verify nav items render in left sidebar on desktop
- [ ] Manual: verify clicking nav items navigates to correct pages
- [ ] Manual: verify active state highlighting
- [ ] Manual: verify New Chat button creates fresh session
- [ ] Manual: verify mobile layout unchanged (TabBar still works)

### DIFF
--- a/frontend/src/components/DesktopNav.tsx
+++ b/frontend/src/components/DesktopNav.tsx
@@ -1,0 +1,48 @@
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useTabBadges } from '../hooks/useTabBadges';
+
+interface NavItem {
+  label: string;
+  path: string;
+  match: (pathname: string) => boolean;
+  badge?: number;
+}
+
+export function DesktopNav() {
+  const location = useLocation();
+  const navigate = useNavigate();
+  const { inboxCount, todoCount } = useTabBadges();
+
+  const items: NavItem[] = [
+    {
+      label: 'Chat',
+      path: '/',
+      match: (p) => p === '/' || p === '/chat' || p.startsWith('/chat/'),
+    },
+    { label: 'Calendar', path: '/calendar', match: (p) => p.startsWith('/calendar') },
+    { label: 'Inbox', path: '/inbox', match: (p) => p === '/inbox', badge: inboxCount },
+    {
+      label: 'Telos',
+      path: '/todos',
+      match: (p) => p === '/todos' || p.startsWith('/todos/'),
+      badge: todoCount,
+    },
+    { label: 'Tasks', path: '/tasks', match: (p) => p.startsWith('/tasks') },
+    { label: 'Files', path: '/files', match: (p) => p.startsWith('/files') },
+  ];
+
+  return (
+    <nav className="desktop-nav">
+      {items.map((item) => (
+        <button
+          key={item.label}
+          className={`desktop-nav-item${item.match(location.pathname) ? ' desktop-nav-item--active' : ''}`}
+          onClick={() => navigate(item.path)}
+        >
+          <span>{item.label}</span>
+          {item.badge ? <span className="desktop-nav-badge">{item.badge}</span> : null}
+        </button>
+      ))}
+    </nav>
+  );
+}
--- a/frontend/src/components/DesktopShell.tsx
+++ b/frontend/src/components/DesktopShell.tsx
@@ -1,9 +1,10 @@
 import { useState, useCallback, type ReactNode } from 'react';
+import { DesktopNav } from './DesktopNav';
 
 export interface DesktopShellProps {
-  left: ReactNode;
+  left?: ReactNode;
   center: ReactNode;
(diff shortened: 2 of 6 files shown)


frontend/src/components/DesktopNav.tsx
```diff
@@ -0,0 +1,48 @@
...
+  match: (pathname: string) => boolean;
+  badge?: number;
+}
+
+export function DesktopNav() {
+  const location = useLocation();
+  const navigate = useNavigate();
+  const { inboxCount, todoCount } = useTabBadges();
+
+  const items: NavItem[] = [
```
dimakis, line 16: 🔵 **style**: The `items` array (with its closures) is re-allocated on every render. For 6 items this is negligible, but wrapping it in `useMemo(() => [...], [inboxCount, todoCount])` would make dependencies explicit and avoid the allocation when only `location` changes. `[fixable]`

frontend/src/components/DesktopShell.tsx
```diff
@@ -55,25 +56,32 @@ export function DesktopShell({ left, center, right, statusBar }: DesktopShellPro
           <button
             className="desktop-collapse-btn"
             onClick={toggleLeft}
-            title={leftCollapsed ? 'Show sessions' : 'Hide sessions'}
+            title={leftCollapsed ? 'Show sidebar' : 'Hide sidebar'}
           >
             {leftCollapsed ? '\u25B6' : '\u25C0'}
           </button>
-          {!leftCollapsed && left}
+          {!leftCollapsed && (
```
dimakis, line 63: 🟡 **bugs**: When `left` is undefined (the PageRoute case), the left sidebar still renders DesktopNav and the collapse toggle — which is the desired behavior. However, when the sidebar is collapsed and then expanded, the fragment renders `<DesktopNav />` followed by `{undefined}`. This is harmless in React (undefined children are ignored), but worth noting: there's no visual separator or empty-state handling between the nav and a missing session panel. If the left sidebar should be entirely hidden when no `left` content exists, wrap the whole sidebar in a conditional too.

dimakis (commented):
## Centaur Review

### `frontend/src/components/DesktopNav.tsx`
_Solid feature PR — the New Chat fix is correct (`storeNewSession` clears store state before navigation), and DesktopShell's optional props cleanly support the new PageRoute pattern. Main gap is missing tests for the new DesktopNav component and the center-only DesktopShell rendering path._

- 🟡 **missing_tests**: New component has no unit tests. It contains routing logic (match functions), badge rendering, and navigation — all worth covering. The project follows TDD, so tests should exist alongside the implementation. `[fixable]`
- 🔵 **style** (L16): The `items` array (with its closures) is re-allocated on every render. For 6 items this is negligible, but wrapping it in `useMemo(() => [...], [inboxCount, todoCount])` would make dependencies explicit and avoid the allocation when only `location` changes. `[fixable]`


### `frontend/src/components/__tests__/DesktopShell.test.tsx`
_Solid feature PR — the New Chat fix is correct (`storeNewSession` clears store state before navigation), and DesktopShell's optional props cleanly support the new PageRoute pattern. Main gap is missing tests for the new DesktopNav component and the center-only DesktopShell rendering path._

- 🟡 **missing_tests**: No test covers the new center-only rendering path (right omitted). The PR made `right` optional and wrapped it in a conditional (`{right && ...}`), but every existing test still passes `right`. A test rendering `<DesktopShell center={...} />` without left/right would verify the PageRoute usage path and confirm the right sidebar collapse button is correctly absent. `[fixable]`


### `frontend/src/components/DesktopShell.tsx`
_Solid feature PR — the New Chat fix is correct (`storeNewSession` clears store state before navigation), and DesktopShell's optional props cleanly support the new PageRoute pattern. Main gap is missing tests for the new DesktopNav component and the center-only DesktopShell rendering path._

- 🟡 **bugs** (L63): When `left` is undefined (the PageRoute case), the left sidebar still renders DesktopNav and the collapse toggle — which is the desired behavior. However, when the sidebar is collapsed and then expanded, the fragment renders `<DesktopNav />` followed by `{undefined}`. This is harmless in React (undefined children are ignored), but worth noting: there's no visual separator or empty-state handling between the nav and a missing session panel. If the left sidebar should be entirely hidden when no `left` content exists, wrap the whole sidebar in a conditional too.

PR #318, opened 2026-05-13
feat(tracing): nested OTel spans for subagent hierarchy

dimakis:
## Summary

- Adds nested OTel span hierarchy for subagent execution
- Jaeger now shows: `session → turn → tool.Agent → subagent → subagent.turn → subagent.tool.*`
- Span attributes: `subagent.parent_tool_id`, `subagent.message_id`, `subagent.duration_ms`, `subagent.total_tokens`
- Cleanup in finally block ensures spans are always closed on abnormal exit

Completes Phase 4 of the subagent visibility spec (dimakis/mgmt#87).
Phases 1-3 (server events, client state, frontend rendering) shipped in PR #309.

## Test plan

- [ ] Start session with `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318`
- [ ] Spawn subagent via Agent tool
- [ ] Open Jaeger, verify span hierarchy: session → turn → tool.Agent → subagent → subagent.tool.*
- [ ] Verify `subagent.duration_ms` and `subagent.total_tokens` attributes populated
- [ ] Verify spans close cleanly on session abort (ctrl+c)

### DIFF
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -196,6 +196,11 @@ async function _runQueryLoopInner(
         cacheReadTokens: number;
         cacheCreationTokens: number;
       } | null;
+      // OTel span hierarchy for subagent
+      span: Span | null;
+      turnSpan: Span | null;
+      toolSpans: Map<string, Span>; // blockId → tool span
+      startedAt: number;
     }
   >();
 
@@ -378,10 +383,40 @@ async function _runQueryLoopInner(
         if (msg.type === 'assistant') {
           const parentToolUseId = msg.parent_tool_use_id as string | undefined;
 
-          // Subagent turn complete — emit subagent_end
+          // Subagent turn complete — emit subagent_end + close spans
           if (parentToolUseId) {
             const subagent = activeSubagents.get(parentToolUseId);
             if (subagent) {
+              const durationMs = Date.now() - subagent.startedAt;
+
+              // End any open subagent tool spans
+              for (const [, ts] of subagent.toolSpans) {
+                ts.setStatus({ code: SpanStatusCode.OK });
+                ts.end();
+              }
+              subagent.toolSpans.clear();
+
+              // End subagent turn span
+              if (subagent.turnSpan) {
+                subagent.turnSpan.setStatus({ code: SpanStatusCode.OK });
+                subagent.turnSpan.end();
+              }
+
+              // End subagent span with attributes
+              if (subagent.span) {
+                subagent.span.setAttribute('subagent.duration_ms', durationMs);
+                if (subagent.usage) {
+                  const totalTokens =
+                    subagent.usage.inputTokens +
+                    subagent.usage.outputTokens +
+                    subagent.usage.cacheReadTokens +
+                    subagent.usage.cacheCreationTokens;
+                  subagent.span.setAttribute('subagent.total_tokens', totalTokens);
+                }
+                subagent.span.setStatus({ code: SpanStatusCode.OK });
+                subagent.span.end();
+              }
+
               emit(
                 v2('subagent_end', {
                   parentBlockId: subagent.parentBlockId,
@@ -394,6 +429,7 @@ async function _runQueryLoopInner(
                 parentToolId: parentToolUseId,
                 subagentMessageId: subagent.subagentMessageId,
                 usage: subagent.usage,
+                durationMs,
               });
               activeSubagents.delete(parentToolUseId);
             }
@@ -588,14 +624,32 @@ async function _runQueryLoopInner(
               const subagentMessageId = (apiMsg?.id as string | undefined) ?? `msg-${Date.now()}`;
 
               if (parentBlockId) {
+                // Create OTel subagent span under the parent tool span
+                const parentToolSpan = toolSpans.get(parentBlockId);
+                const subagentParentCtx = parentToolSpan
+                  ? trace.setSpan(context.active(), parentToolSpan)
+                  : context.active();
+                const subagentSpan = tracer.startSpan('subagent', {}, subagentParentCtx);
+                subagentSpan.setAttribute('subagent.parent_tool_id', parentToolUseId);
+                subagentSpan.setAttribute('subagent.message_id', subagentMessageId);
+
+                // Start first turn span under the subagent span
+                const subTurnCtx = trace.setSpan(context.active(), subagentSpan);
+                const subTurnSpan = tracer.startSpan('subagent.turn', {}, subTurnCtx);
+                subTurnSpan.setAttribute('subagent.turn.index', 0);
+
                 activeSubagents.set(parentToolUseId, {
                   parentBlockId,
-                  parentToolName: 'Agent', // We don't track this persistently, assume Agent
+                  parentToolName: 'Agent',
                   taskId: null,
                   subagentMessageId,
                   subagentBlockIdByIndex: new Map(),
                   subagentToolInputBuffers: new Map(),
                   usage: null,
+                  span: subagentSpan,
+                  turnSpan: subTurnSpan,
+                  toolSpans: new Map(),
+                  startedAt: Date.now(),
                 });
 
                 // Extract usage from message_start
@@ -722,6 +776,20 @@ async function _runQueryLoopInner(
                   id: toolId,
                   inputBuf: '',
                 });
+
+                // Create OTel span for subagent tool under its turn span
+                const subToolParent = subagent.turnSpan
+                  ? trace.setSpan(context.active(), subagent.turnSpan)
+                  : context.active();
+                const subToolSpan = tracer.startSpan(
+                  `subagent.tool.${toolName}`,
+                  {},
+                  subToolParent,
+                );
+                subToolSpan.setAttribute('tool.name', toolName);
+                subToolSpan.setAttribute('tool.id', toolId);
+                subagent.toolSpans.set(blockId, subToolSpan);
+
                 emit(
                   v2('subagent_block_start', {
                     parentBlockId: subagent.parentBlockId,
@@ -935,6 +1003,14 @@ async function _runQueryLoopInner(
                     ...(rawInput ? { rawInput } : {}),
                   }),
                 );
+                // End subagent tool span
+                const subToolSpan = subagent.toolSpans.get(blockId);
+                if (subToolSpan) {
+                  subToolSpan.setStatus({ code: SpanStatusCode.OK });
+                  subToolSpan.end();
+                  subagent.toolSpans.delete(blockId);
+                }
+
                 log.info('subagent tool call', {
                   clientId,
                   parentToolId: parentToolUseId,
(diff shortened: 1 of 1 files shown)


dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid instrumentation that correctly mirrors the existing span patterns. Main concerns: no test coverage for the new span hierarchy, cleanup spans always report OK status even on error paths, and duplicated cleanup logic._

- 🟡 **bugs** (L1263): Subagent spans cleaned up in the `finally` block always get `SpanStatusCode.OK`, even when the query loop exits via the `catch` block (error/abort). This is consistent with the existing pattern for parent tool spans (line 1278), but means error-path cleanup produces misleading OK-status spans in Jaeger. Consider checking whether the catch block was entered (e.g., a `caughtError` flag) and using `SpanStatusCode.ERROR` for cleanup spans in that case. `[fixable]`
- 🟡 **bugs** (L638): The subagent turn span index is hardcoded to 0. For multi-turn subagents (where the SDK emits multiple message_start/assistant cycles with the same `parent_tool_use_id`), each cycle creates a new subagent entry that resets the index to 0. All subagent turn spans will have `subagent.turn.index = 0`, making the attribute meaningless in traces. If multi-turn subagents are expected, the turn index should either increment across the lifecycle or the attribute should be removed. `[fixable]`
- 🔵 **style** (L392): The span cleanup pattern (`setStatus(OK) → end()` in a loop, then null-check turnSpan, then null-check span) is duplicated between the normal subagent_end path (lines 392-417) and the finally-block cleanup (lines 1261-1275). Consider extracting a helper like `endSubagentSpans(sub: ActiveSubagent)` to reduce duplication and ensure consistent cleanup order. `[fixable]`
- 🔵 **unsafe_assumptions** (L631): When the parent tool span is not found in `toolSpans` (line 627), the subagent span falls back to `context.active()`, which is the session-level context. This silently creates the subagent span under the session rather than the tool, breaking the expected hierarchy. A `log.warn` for this fallback would help diagnose tracing gaps — the parent tool span should always exist at this point unless events arrive out of order. `[fixable]`


_Solid instrumentation that correctly mirrors the existing span patterns. Main concerns: no test coverage for the new span hierarchy, cleanup spans always report OK status even on error paths, and duplicated cleanup logic._

- 🟡 **missing_tests**: No tests verify the OTel span hierarchy for subagents. The existing `subagent-events.test.ts` tests v2 event emission but doesn't assert span creation, parent-child relationships, or attribute values. The project has a precedent for testing span hierarchy in `with-span.test.ts` (using `InMemorySpanExporter`). At minimum, a test should verify: (1) subagent span parents to the tool span, (2) subagent.turn parents to the subagent span, (3) subagent.tool.* parents to the turn span, (4) all spans are ended on cleanup. `[fixable]`

server/query-loop.ts
```diff
@@ -1181,6 +1257,23 @@ async function _runQueryLoopInner(
       if (store && resolvedSessionId) {
         store.markSessionInactive(resolvedSessionId);
       }
+      // Clean up any open subagent spans
+      for (const [, sub] of activeSubagents) {
+        for (const [, ts] of sub.toolSpans) {
+          ts.setStatus({ code: SpanStatusCode.OK });
```
dimakis: 🟡 **bugs**: Subagent spans cleaned up in the `finally` block always get `SpanStatusCode.OK`, even when the query loop exits via the `catch` block (error/abort). This is consistent with the existing pattern for parent tool spans (line 1278), but means error-path cleanup produces misleading OK-status spans in Jaeger. Consider checking whether the catch block was entered (e.g., a `caughtError` flag) and using `SpanStatusCode.ERROR` for cleanup spans in that case. `[fixable]`

server/query-loop.ts
```diff
@@ -587,13 +623,31 @@ async function _runQueryLoopInner(
...
+                  ? trace.setSpan(context.active(), parentToolSpan)
+                  : context.active();
+                const subagentSpan = tracer.startSpan('subagent', {}, subagentParentCtx);
+                subagentSpan.setAttribute('subagent.parent_tool_id', parentToolUseId);
+                subagentSpan.setAttribute('subagent.message_id', subagentMessageId);
+
+                // Start first turn span under the subagent span
+                const subTurnCtx = trace.setSpan(context.active(), subagentSpan);
+                const subTurnSpan = tracer.startSpan('subagent.turn', {}, subTurnCtx);
+                subTurnSpan.setAttribute('subagent.turn.index', 0);
```
dimakis: 🟡 **bugs**: The subagent turn span index is hardcoded to 0. For multi-turn subagents (where the SDK emits multiple message_start/assistant cycles with the same `parent_tool_use_id`), each cycle creates a new subagent entry that resets the index to 0. All subagent turn spans will have `subagent.turn.index = 0`, making the attribute meaningless in traces. If multi-turn subagents are expected, the turn index should either increment across the lifecycle or the attribute should be removed. `[fixable]`

server/query-loop.ts
```diff
@@ -377,10 +382,40 @@ async function _runQueryLoopInner(
...
 
-          // Subagent turn complete — emit subagent_end
+          // Subagent turn complete — emit subagent_end + close spans
           if (parentToolUseId) {
             const subagent = activeSubagents.get(parentToolUseId);
             if (subagent) {
+              const durationMs = Date.now() - subagent.startedAt;
+
+              // End any open subagent tool spans
+              for (const [, ts] of subagent.toolSpans) {
```
dimakis: 🔵 **style**: The span cleanup pattern (`setStatus(OK) → end()` in a loop, then null-check turnSpan, then null-check span) is duplicated between the normal subagent_end path (lines 392-417) and the finally-block cleanup (lines 1261-1275). Consider extracting a helper like `endSubagentSpans(sub: ActiveSubagent)` to reduce duplication and ensure consistent cleanup order. `[fixable]`

server/query-loop.ts
```diff
@@ -587,13 +623,31 @@ async function _runQueryLoopInner(
               const subagentMessageId = (apiMsg?.id as string | undefined) ?? `msg-${Date.now()}`;
 
               if (parentBlockId) {
+                // Create OTel subagent span under the parent tool span
+                const parentToolSpan = toolSpans.get(parentBlockId);
+                const subagentParentCtx = parentToolSpan
+                  ? trace.setSpan(context.active(), parentToolSpan)
+                  : context.active();
+                const subagentSpan = tracer.startSpan('subagent', {}, subagentParentCtx);
```
dimakis: 🔵 **unsafe_assumptions**: When the parent tool span is not found in `toolSpans` (line 627), the subagent span falls back to `context.active()`, which is the session-level context. This silently creates the subagent span under the session rather than the tool, breaking the expected hierarchy. A `log.warn` for this fallback would help diagnose tracing gaps — the parent tool span should always exist at this point unless events arrive out of order. `[fixable]`

server/query-loop.ts
```diff
@@ -587,14 +644,59 @@ async function _runQueryLoopInner(
...
-                  parentBlockId,
-                  parentToolName: 'Agent', // We don't track this persistently, assume Agent
-                  subagentMessageId,
-                  subagentBlockIdByIndex: new Map(),
-                  subagentToolInputBuffers: new Map(),
-                  usage: null,
-                });
+                const existing = activeSubagents.get(parentToolUseId);
+
+                if (existing) {
```
dimakis, line 650: 🟡 **bugs**: Multi-turn subagent path does not end open tool spans from the previous turn before ending the turn span. If any subagent tool spans from turn N are still open when turn N+1 starts, they become children of an already-ended turn span. Add `for (const [, ts] of existing.toolSpans) { ts.setStatus({ code: SpanStatusCode.OK }); ts.end(); } existing.toolSpans.clear();` before ending the turn span at line 651. `[fixable]`

server/query-loop.ts
```diff
@@ -377,10 +430,13 @@ async function _runQueryLoopInner(
         if (msg.type === 'assistant') {
           const parentToolUseId = msg.parent_tool_use_id as string | undefined;
 
-          // Subagent turn complete — emit subagent_end
+          // Subagent turn complete — emit subagent_end + close spans
           if (parentToolUseId) {
             const subagent = activeSubagents.get(parentToolUseId);
             if (subagent) {
+              const durationMs = Date.now() - subagent.startedAt;
```
dimakis, line 438: 🟡 **bugs**: `durationMs` is calculated here via `Date.now() - subagent.startedAt`, then `endSubagentSpans` recalculates it independently at line 62. The two values will differ slightly. The span gets the second (later) value while the log at line 452 gets the first. Either pass `durationMs` into `endSubagentSpans` or compute it only once inside the function and return it. `[fixable]`

server/query-loop.ts
```diff
@@ -28,6 +28,52 @@ function truncateForTrace(text: string): { text: string; truncated?: true } {
   return { text: text.slice(0, TRACE_CONTENT_MAX_CHARS), truncated: true };
 }
 
+/** End all open OTel spans for a subagent entry. */
+function endSubagentSpans(
```
dimakis, line 32: 🔵 **style**: The inline type literal for `endSubagentSpans`'s `sub` parameter duplicates the shape already defined in the `activeSubagents` Map generic. Extract the subagent entry type (e.g., `type SubagentEntry = { ... }`) and reference it in both the Map and the function signature to avoid drift. `[fixable]`

server/query-loop.ts
```diff
@@ -587,14 +644,59 @@ async function _runQueryLoopInner(
...
+                      parentBlockId,
+                      parentToolUseId,
+                    });
+                  }
+                  const subagentParentCtx = parentToolSpan
+                    ? trace.setSpan(context.active(), parentToolSpan)
+                    : context.active();
+                  const subagentSpan = tracer.startSpan('subagent', {}, subagentParentCtx);
+                  subagentSpan.setAttribute('subagent.parent_tool_id', parentToolUseId);
+                  subagentSpan.setAttribute('subagent.message_id', subagentMessageId);
```
dimakis, line 680: 🔵 **style**: `subagent.message_id` span attribute is set only on first turn (line 679). On multi-turn subagents, `subagentMessageId` is updated (line 656) but the span attribute is not. Consider adding the latest message ID as an event or updating the attribute so traces reflect the full lifecycle. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Solid span hierarchy with proper cleanup ordering, but multi-turn path has a tool-span leak when previous-turn tools are still open, and duration is double-calculated with slight divergence between logs and traces._

- 🟡 **bugs** (L649): Multi-turn subagent path does not end open tool spans from the previous turn before ending the turn span. If any subagent tool spans from turn N are still open when turn N+1 starts, they become children of an already-ended turn span. Add `for (const [, ts] of existing.toolSpans) { ts.setStatus({ code: SpanStatusCode.OK }); ts.end(); } existing.toolSpans.clear();` before ending the turn span at line 651. `[fixable]`
- 🟡 **bugs** (L437): `durationMs` is calculated here via `Date.now() - subagent.startedAt`, then `endSubagentSpans` recalculates it independently at line 62. The two values will differ slightly. The span gets the second (later) value while the log at line 452 gets the first. Either pass `durationMs` into `endSubagentSpans` or compute it only once inside the function and return it. `[fixable]`
- 🔵 **style** (L32): The inline type literal for `endSubagentSpans`'s `sub` parameter duplicates the shape already defined in the `activeSubagents` Map generic. Extract the subagent entry type (e.g., `type SubagentEntry = { ... }`) and reference it in both the Map and the function signature to avoid drift. `[fixable]`
- 🔵 **style** (L679): `subagent.message_id` span attribute is set only on first turn (line 679). On multi-turn subagents, `subagentMessageId` is updated (line 656) but the span attribute is not. Consider adding the latest message ID as an event or updating the attribute so traces reflect the full lifecycle. `[fixable]`


_Solid span hierarchy with proper cleanup ordering, but multi-turn path has a tool-span leak when previous-turn tools are still open, and duration is double-calculated with slight divergence between logs and traces._

- 🟡 **missing_tests**: No tests cover `endSubagentSpans`, subagent span lifecycle (creation, multi-turn rotation, cleanup), or the error-path status propagation (`caughtError` flag). The existing 2100+ tests cover message/event flow but not span management. At minimum, test: (1) spans are ended with OK on normal completion, (2) spans are ended with ERROR when the query loop throws, (3) multi-turn subagents get new turn spans while the parent span persists. `[fixable]`

PR #320, opened 2026-05-14
fix(client,server): prevent duplicate messages on iOS foreground recovery

dimakis:
## Summary

- **Bug:** Messages (user prompt + assistant reply) render twice in the UI after iOS foreground recovery. Self-heals on next session re-entry.
- **Root cause:** `_foreground` (REST RESTORE) and WS reconnect replay both fire simultaneously, each inserting messages without deduplicating against the other.
- **Fix:** Three surgical changes:
  - `MESSAGE_START`/`MESSAGE_END`: skip if messageId already exists in `messages[]`
  - `replayEventsToMessages`: use stored messageId instead of `Date.now()` so REST and WS agree
  - `RESTORE`: clear `current` when its messageId conflicts with the restored set

## Test plan

- [x] Unit tests: MESSAGE_START/END dedup guards (positive + negative)
- [x] Unit tests: RESTORE clears stale `current` (normal + interrupted paths)
- [x] Unit tests: deterministic initial prompt messageId (reuse stored + fallback)
- [x] Integration test: full foreground recovery sequence (RESTORE → WS replay → no duplication)
- [x] Integration test: multi-turn dedup (only replayed message suppressed, earlier ones kept)
- [ ] Manual: iOS foreground recovery — send message, background, return, verify no duplicates

### DIFF
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -481,14 +481,14 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
           merged.splice(insertAfter + 1, 0, opt);
         }
         merged.push(notice);
-        const currentStale = state.current &&
-          merged.some((m) => m.messageId === state.current!.messageId);
+        const currentStale =
+          state.current && merged.some((m) => m.messageId === state.current!.messageId);
         return { ...state, messages: merged, current: currentStale ? null : state.current };
       }
       // Clear current if the restored set already contains it (prevents
       // MESSAGE_END from re-inserting a message that RESTORE already has).
-      const currentStale = state.current &&
-        valid.some((m) => m.messageId === state.current!.messageId);
+      const currentStale =
+        state.current && valid.some((m) => m.messageId === state.current!.messageId);
       return { ...state, messages: valid, current: currentStale ? null : state.current };
     }
 
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -237,6 +237,10 @@ export function patchToolResult(
 export function messagesReducer(state: MessagesState, action: MessagesAction): MessagesState {
   switch (action.type) {
     case 'MESSAGE_START': {
+      // Dedup: skip if this message was already restored (e.g. WS replay after RESTORE)
+      if (state.messages.some((m) => m.messageId === action.messageId)) {
+        return state;
+      }
       const base = state.current
         ? { ...state, messages: [...state.messages, finishCurrent(state.current)] }
         : state;
@@ -309,6 +313,10 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
 
     case 'MESSAGE_END': {
       if (!state.current) return { ...state };
+      // Dedup: if this message was already restored, discard the streaming copy
+      if (state.messages.some((m) => m.messageId === state.current!.messageId)) {
+        return { ...state, current: null };
+      }
       const finished = finishCurrent(state.current);
       return { ...state, messages: [...state.messages, finished], current: null };
     }
@@ -473,9 +481,15 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
           merged.splice(insertAfter + 1, 0, opt);
         }
         merged.push(notice);
-        return { ...state, messages: merged };
+        const currentStaleI = state.current &&
+          merged.some((m) => m.messageId === state.current!.messageId);
+        return { ...state, messages: merged, ...(currentStaleI ? { current: null } : {}) };
       }
-      return { ...state, messages: valid };
+      // Clear current if the restored set already contains it (prevents
+      // MESSAGE_END from re-inserting a message that RESTORE already has).
+      const currentStale = state.current &&
+        valid.some((m) => m.messageId === state.current!.messageId);
+      return { ...state, messages: valid, ...(currentStale ? { current: null } : {}) };
     }
 
     case 'USER_MESSAGE_RECEIVED': {
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -312,7 +312,7 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
     }
 
     case 'MESSAGE_END': {
-      if (!state.current) return { ...state };
+      if (!state.current) return state;
       // Dedup: if this message was already restored, discard the streaming copy
       if (state.messages.some((m) => m.messageId === state.current!.messageId)) {
         return { ...state, current: null };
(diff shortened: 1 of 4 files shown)


server/chat.ts
```diff
@@ -1455,9 +1455,14 @@ export function replayEventsToMessages(
   // Inject the initial prompt as the first message.
   // Priority: initialPrompt param (from session metadata) > legacy out-of-order event
   if (initialPrompt) {
+    // Reuse the stored messageId so REST recovery and WS replay agree on IDs
+    const matchingEvt = events.find(
+      (e) => e.type === 'user_message' && e.payload.text === initialPrompt,
```
dimakis, line 1548: 🟡 **unsafe_assumptions**: Matching initial prompt by text content (`e.payload.text === initialPrompt`) is fragile. If the user sends the same text as a follow-up message later, `events.find()` returns the first match (which is correct for the initial prompt), but if the initial prompt was mutated (e.g., skill expansion prepending to `fullPrompt`), no match is found and the fallback fires. This is safe today because the initial prompt stored in session metadata and the event payload both use `fullPrompt` after expansion, but worth noting as a coupling assumption.

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid fix for a real iOS foreground recovery race condition. The dedup logic in MESSAGE_START/MESSAGE_END/RESTORE is correct and well-tested. The `finishSubagent` extraction and subagent type-guard additions are clean refactors. The deterministic message ID change in `replayEventsToMessages` correctly addresses the root cause. Minor style nits and a couple missing edge-case tests, but no critical issues._

- 🟡 **unsafe_assumptions** (L1460): Matching initial prompt by text content (`e.payload.text === initialPrompt`) is fragile. If the user sends the same text as a follow-up message later, `events.find()` returns the first match (which is correct for the initial prompt), but if the initial prompt was mutated (e.g., skill expansion prepending to `fullPrompt`), no match is found and the fallback fires. This is safe today because the initial prompt stored in session metadata and the event payload both use `fullPrompt` after expansion, but worth noting as a coupling assumption.


_Solid fix for a real iOS foreground recovery race condition. The dedup logic in MESSAGE_START/MESSAGE_END/RESTORE is correct and well-tested. The `finishSubagent` extraction and subagent type-guard additions are clean refactors. The deterministic message ID change in `replayEventsToMessages` correctly addresses the root cause. Minor style nits and a couple missing edge-case tests, but no critical issues._

- 🔵 **style** (L457): The variable name `currentStaleI` is unclear — the trailing `I` apparently means 'interrupted path' (vs `currentStale` on line 463 for the normal path). Consider renaming to `currentStaleInterrupted` or just reusing the same name `currentStale` since both branches have the same logic. `[fixable]`
- 🟡 **bugs** (L291): MESSAGE_END returns `{ ...state }` (a new object) when `current` is null, but MESSAGE_START dedup returns `state` (same reference). The inconsistency means MESSAGE_END without a current always triggers a React re-render even though nothing changed. Pre-existing issue, but the dedup path at line 294 correctly returns `{ ...state, current: null }` which is also a new object even if current was already null. Consider `return state` for the `!state.current` early return to avoid unnecessary re-renders. `[fixable]`
- 🔵 **style** (L457): The conditional spread `...(currentStaleI ? { current: null } : {})` is used in two places (lines 459 and 465). A simple ternary for the `current` field would be clearer: `current: currentStale ? null : state.current`. `[fixable]`


_Solid fix for a real iOS foreground recovery race condition. The dedup logic in MESSAGE_START/MESSAGE_END/RESTORE is correct and well-tested. The `finishSubagent` extraction and subagent type-guard additions are clean refactors. The deterministic message ID change in `replayEventsToMessages` correctly addresses the root cause. Minor style nits and a couple missing edge-case tests, but no critical issues._

- 🔵 **missing_tests**: No test covers the RESTORE interrupted path's new `currentStaleI` logic (line 457-459). The existing 'nullifies current in interrupted RESTORE too' test at line 1341 exercises the non-interrupted RESTORE path's `currentStale` check. A test where RESTORE fires with `interrupted: true`, optimistic user messages exist, AND current overlaps with restored messages would cover the `currentStaleI` branch. `[fixable]`
- 🔵 **missing_tests**: No test covers the `finishSubagent` helper directly or the scenario where `finishCurrent` is called on a message with subagent blocks — e.g., the MESSAGE_START dedup path at line 220-221 calls `finishCurrent(state.current)` when a new message arrives while a subagent-containing message is streaming. A test for `finishCurrent` with subagent state would validate the new `finishSubagent` extraction. `[fixable]`

packages/client/src/slices/messages.ts
```diff
@@ -424,9 +454,15 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
           merged.splice(insertAfter + 1, 0, opt);
         }
         merged.push(notice);
-        return { ...state, messages: merged };
+        const currentStaleI = state.current &&
```
dimakis: 🔵 **style**: The variable name `currentStaleI` is unclear — the trailing `I` apparently means 'interrupted path' (vs `currentStale` on line 463 for the normal path). Consider renaming to `currentStaleInterrupted` or just reusing the same name `currentStale` since both branches have the same logic. `[fixable]`

packages/client/src/slices/messages.ts
```diff
@@ -263,6 +289,10 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
 
     case 'MESSAGE_END': {
       if (!state.current) return { ...state };
```
dimakis: 🟡 **bugs**: MESSAGE_END returns `{ ...state }` (a new object) when `current` is null, but MESSAGE_START dedup returns `state` (same reference). The inconsistency means MESSAGE_END without a current always triggers a React re-render even though nothing changed. Pre-existing issue, but the dedup path at line 294 correctly returns `{ ...state, current: null }` which is also a new object even if current was already null. Consider `return state` for the `!state.current` early return to avoid unnecessary re-renders. `[fixable]`

packages/client/src/slices/messages.ts
dimakis: 🔵 **style**: The conditional spread `...(currentStaleI ? { current: null } : {})` is used in two places (lines 459 and 465). A simple ternary for the `current` field would be clearer: `current: currentStale ? null : state.current`. `[fixable]`

dimakis:
## Centaur Review

_Clean, well-tested fix for a real iOS foreground-recovery race condition. Two minor defense-in-depth suggestions — the core dedup logic is correct._

- 🟡 **bugs** (L324): SESSION_END also finalizes `current` into `messages` (line 326-330) without a dedup check. The RESTORE and MESSAGE_START fixes make this unreachable in practice — RESTORE now clears stale `current`, and MESSAGE_START blocks re-creation — but SESSION_END sits outside that transitive protection. If a future change breaks the invariant (e.g. MESSAGE_SNAPSHOT re-creates `current` with a stale ID), SESSION_END would silently duplicate. Adding the same one-line guard (`if (state.messages.some(…)) return { ...state, running: false, current: null }`) makes the invariant explicit. `[fixable]`


### `server/chat.ts`
_Clean, well-tested fix for a real iOS foreground-recovery race condition. Two minor defense-in-depth suggestions — the core dedup logic is correct._

- 🔵 **unsafe_assumptions** (L1539): `matchingEvt.payload.messageId as string` — `payload` is `Record<string, unknown>`, so `messageId` could be `undefined` for corrupt event data. The `as string` cast silently produces `undefined`, which would pass through as the messageId and break downstream dedup checks (they compare strings). A safe fallback would be: `const messageId = (matchingEvt?.payload.messageId as string | undefined) ?? 'umsg-initial'` — collapsing the ternary into one expression with nullish coalescing. `[fixable]`

PR #321, opened 2026-05-14
fix(harness): cascade stopTask to subagents on interrupt

dimakis:
## Summary

- **Bug:** `interruptChat()` only called `interrupt()` on the parent query, which was blocked waiting for a running subagent (Agent tool call) — causing the session to hang until the subagent completed naturally
- **Fix:** Track `task_id` from SDK `task_started` events on the session, call `stopTask()` for each active subagent before interrupting the parent
- **Frontend:** `subagent_cancelled` event renders as a finished block with "Cancelled" summary

## Details

The SDK provides `query.stopTask(taskId)` and emits `task_started`/`task_notification` system events, but Mitzo was silently ignoring them. This PR:

1. Expands `ManagedSession` with `activeTaskIds` map and `stopTask` on `queryInstance`
2. Handles `task_started` and `task_notification` system events in the query loop
3. Calls `stopTask()` for all active subagents in `interruptChat()` before `interrupt()`
4. Maps `subagent_cancelled` to `SUBAGENT_END` in the protocol parser for clean UI transition

## Risk

`task_started` may only fire for background agents (`run_in_background: true`). If foreground agents don't emit it, `activeTaskIds` will be empty and `stopTask` is a no-op — graceful degradation to old behavior. Check server logs for `subagent task started` to verify.

## Test plan

- [x] `tsc -b` passes
- [x] `npm test` passes (6/6 in send-to-chat, pre-existing failures unrelated)
- [ ] Manual: start session, trigger Explore agent, send interrupt message — verify subagent stops and UI shows "Cancelled"
- [ ] Check logs for `subagent task started` / `subagent task finished` entries
- [ ] Test with multiple concurrent subagents
- [ ] Test with `run_in_background: true` agents

### DIFF
--- a/frontend/src/types/ws-messages.ts
+++ b/frontend/src/types/ws-messages.ts
@@ -239,7 +239,8 @@ export type ServerMessage =
   | SubagentBlockDeltaMsg
   | SubagentBlockEndMsg
   | SubagentToolResultMsg
-  | SubagentEndMsg;
+  | SubagentEndMsg
+  | SubagentCancelledMsg;
 
 export interface ProgressStartMsg {
   type: 'progress_start';
--- a/frontend/src/types/ws-messages.ts
+++ b/frontend/src/types/ws-messages.ts
@@ -368,3 +368,13 @@ export interface SubagentEndMsg {
     cacheCreationTokens: number;
   };
 }
+
+export interface SubagentCancelledMsg {
+  type: 'subagent_cancelled';
+  v: 2;
+  ts: number;
+  sessionId: string;
+  parentBlockId: string;
+  subagentMessageId: string;
+  taskId: string;
+}
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -504,6 +504,14 @@ export function parseServerMessage(
           | undefined,
       });
       break;
+
+    case 'subagent_cancelled':
+      result.messagesActions.push({
+        type: 'SUBAGENT_END',
+        parentBlockId: msg.parentBlockId as string,
+        summary: 'Cancelled',
+      });
+      break;
   }
 
   return result;
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -194,6 +194,23 @@ export function parseServerMessage(
       });
       break;
 
+    case 'boot_context': {
+      const source = msg.source === 'contexgin' ? 'contexgin' : 'local-fallback';
+      const sourceCount = typeof msg.sourceCount === 'number' ? msg.sourceCount : 0;
+      const tokenCount = typeof msg.tokenCount === 'number' ? msg.tokenCount : 0;
+      const trimmedCount = typeof msg.trimmedCount === 'number' ? msg.trimmedCount : 0;
+      // Validate each element is a string — filter out non-string entries
+      const rawSources = Array.isArray(msg.sources) ? msg.sources : [];
+      const sources: string[] = rawSources.filter(
+        (s: unknown): s is string => typeof s === 'string',
+      );
+      result.messagesActions.push({
+        type: 'SET_BOOT_CONTEXT',
+        bootContext: { source, sourceCount, tokenCount, trimmedCount, sources },
+      });
+      break;
+    }
+
     case 'worktree_opened':
       result.messagesActions.push({
         type: 'WORKTREE_OPENED',
--- a/packages/client/src/slices/messages.ts
+++ b/packages/client/src/slices/messages.ts
@@ -13,6 +13,8 @@ import type {
   PermissionRequest,
   RawToolInput,
   BlockType,
+  StreamingSubagentState,
+  FinishedSubagentState,
 } from '@mitzo/protocol';
 
 // ─── State ───────────────────────────────────────────────────────────────────
@@ -139,6 +141,26 @@ export type MessagesAction =
 
 // ─── Helpers ─────────────────────────────────────────────────────────────────
 
+function finishSubagent(sa: StreamingSubagentState | FinishedSubagentState): FinishedSubagentState {
+  if (Array.isArray(sa.blocks)) return sa as FinishedSubagentState;
+  const streaming = sa as StreamingSubagentState;
+  const blocks: FinishedBlock[] = streaming.blockOrder.map((id) => {
+    const sb = streaming.blocks.get(id)!;
+    return {
+      blockId: sb.blockId,
+      blockType: sb.blockType,
+      content: sb.content,
+      toolName: sb.toolName,
+      toolId: sb.toolId,
+      toolInput: sb.toolInput,
+      rawInput: sb.rawInput,
+      toolResult: sb.toolResult,
+      toolError: sb.toolError,
+    };
+  });
+  return { messageId: streaming.messageId, blocks };
+}
+
 export function finishCurrent(current: StreamingMessage): FinishedMessage {
   const blocks: FinishedBlock[] = current.blockOrder.map((blockId) => {
     const b = current.blocks.get(blockId)!;
@@ -152,7 +174,7 @@ export function finishCurrent(current: StreamingMessage): FinishedMessage {
       rawInput: b.rawInput,
       toolResult: b.toolResult,
       toolError: b.toolError,
-      subagent: b.subagent,
+      subagent: b.subagent ? finishSubagent(b.subagent) : undefined,
     };
   });
   return { messageId: current.messageId, role: 'assistant', blocks, timestamp: Date.now() };
@@ -527,7 +549,8 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
     case 'SUBAGENT_BLOCK_START': {
       if (!state.current) return state;
       const parentBlock = state.current.blocks.get(action.parentBlockId);
-      if (!parentBlock?.subagent) return state;
+      if (!parentBlock?.subagent || !('blockOrder' in parentBlock.subagent)) return state;
+      const sa = parentBlock.subagent;
 
       const newBlock: StreamingBlock = {
         blockId: action.blockId,
@@ -537,17 +560,13 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
         ...(action.toolName ? { toolName: action.toolName } : {}),
       };
 
-      const newSubBlocks = new Map(parentBlock.subagent.blocks);
+      const newSubBlocks = new Map(sa.blocks);
       newSubBlocks.set(action.blockId, newBlock);
 
       const newBlocks = new Map(state.current.blocks);
       newBlocks.set(action.parentBlockId, {
         ...parentBlock,
-        subagent: {
-          ...parentBlock.subagent,
-          blocks: newSubBlocks,
-          blockOrder: [...parentBlock.subagent.blockOrder, action.blockId],
-        },
+        subagent: { ...sa, blocks: newSubBlocks, blockOrder: [...sa.blockOrder, action.blockId] },
       });
 
       return { ...state, current: { ...state.current, blocks: newBlocks } };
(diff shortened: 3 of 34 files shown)


frontend/src/types/ws-messages.ts
```diff
@@ -368,3 +368,13 @@ export interface SubagentEndMsg {
     cacheCreationTokens: number;
   };
 }
+
+export interface SubagentCancelledMsg {
```
dimakis, line 373: 🔴 **bugs**: `SubagentCancelledMsg` is defined but not added to the `ServerMessage` union type (line 242). The union ends at `SubagentEndMsg`. While the `@mitzo/client` protocol parser uses a loose `{type: string; [key: string]: unknown}` type so it still works at runtime, any frontend code that pattern-matches on `ServerMessage` won't include this variant. Add `| SubagentCancelledMsg` to the union. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -504,6 +504,14 @@ export function parseServerMessage(
           | undefined,
       });
       break;
+
+    case 'subagent_cancelled':
```
dimakis, line 525: 🟡 **missing_tests**: The `subagent_cancelled` → `SUBAGENT_END` mapping in the protocol parser has no test coverage. `protocol-parser.test.ts` has no subagent tests at all. A test should verify that a `subagent_cancelled` message produces a `SUBAGENT_END` action with `summary: 'Cancelled'` and correct `parentBlockId`. `[fixable]`

server/query-loop.ts
```diff
@@ -1057,13 +1062,47 @@ async function _runQueryLoopInner(
...
+
+          // Track compaction events from SDK system status messages
           const compactResult = (msg as Record<string, unknown>).compact_result;
           if (subtype === 'status' && compactResult === 'success') {
             numCompactions++;
             log.info('compaction completed', { clientId, numCompactions });
           }
+
+          // Track subagent task lifecycle for interrupt cancellation
+          if (subtype === 'task_started') {
```
dimakis, line 1075: 🟡 **missing_tests**: No tests for the new `task_started` and `task_notification` system message handling in the query loop. The existing `query-loop.test.ts` covers subagent block events but not the task lifecycle tracking or `subagent_cancelled` emission. Key cases to cover: task_started populates `activeTaskIds`, task_notification with `status=stopped` emits `subagent_cancelled`, and task_notification with other statuses does not. `[fixable]`

server/query-loop.ts
```diff
@@ -1057,13 +1062,47 @@ async function _runQueryLoopInner(
...
+            const taskId = (msg as Record<string, unknown>).task_id as string | undefined;
```
dimakis, line 1076: 🔵 **unsafe_assumptions**: The `task_id`, `tool_use_id`, and `status` fields are extracted from system messages via `as string | undefined` casts with no runtime validation. While these are SDK-internal messages and unlikely to be malformed, a missing field silently no-ops (no log). Consider adding a debug-level log when `task_started` has a missing `task_id` or `tool_use_id` to aid troubleshooting. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid feature implementation with correct race-condition handling and idempotent guards, but `SubagentCancelledMsg` is missing from the frontend `ServerMessage` union type, and the new server-side and client-side logic lacks test coverage._

- 🔴 **bugs** (L372): `SubagentCancelledMsg` is defined but not added to the `ServerMessage` union type (line 242). The union ends at `SubagentEndMsg`. While the `@mitzo/client` protocol parser uses a loose `{type: string; [key: string]: unknown}` type so it still works at runtime, any frontend code that pattern-matches on `ServerMessage` won't include this variant. Add `| SubagentCancelledMsg` to the union. `[fixable]`


_Solid feature implementation with correct race-condition handling and idempotent guards, but `SubagentCancelledMsg` is missing from the frontend `ServerMessage` union type, and the new server-side and client-side logic lacks test coverage._

- 🟡 **missing_tests** (L508): The `subagent_cancelled` → `SUBAGENT_END` mapping in the protocol parser has no test coverage. `protocol-parser.test.ts` has no subagent tests at all. A test should verify that a `subagent_cancelled` message produces a `SUBAGENT_END` action with `summary: 'Cancelled'` and correct `parentBlockId`. `[fixable]`


### `server/query-loop.ts`
_Solid feature implementation with correct race-condition handling and idempotent guards, but `SubagentCancelledMsg` is missing from the frontend `ServerMessage` union type, and the new server-side and client-side logic lacks test coverage._

- 🟡 **missing_tests** (L1075): No tests for the new `task_started` and `task_notification` system message handling in the query loop. The existing `query-loop.test.ts` covers subagent block events but not the task lifecycle tracking or `subagent_cancelled` emission. Key cases to cover: task_started populates `activeTaskIds`, task_notification with `status=stopped` emits `subagent_cancelled`, and task_notification with other statuses does not. `[fixable]`
- 🔵 **unsafe_assumptions** (L1076): The `task_id`, `tool_use_id`, and `status` fields are extracted from system messages via `as string | undefined` casts with no runtime validation. While these are SDK-internal messages and unlikely to be malformed, a missing field silently no-ops (no log). Consider adding a debug-level log when `task_started` has a missing `task_id` or `tool_use_id` to aid troubleshooting. `[fixable]`


### `server/chat.ts`
_Solid feature implementation with correct race-condition handling and idempotent guards, but `SubagentCancelledMsg` is missing from the frontend `ServerMessage` union type, and the new server-side and client-side logic lacks test coverage._

- 🟡 **missing_tests** (L935): The `interruptChat` stopTask cascade logic (iterating `activeTaskIds` and calling `stopTask`) has no test. The existing `send-to-chat.test.ts` tests for `interruptChat` only update the mock to include `stopTask` in `queryInstance` but don't exercise a scenario where `activeTaskIds` is populated and `stopTask` is actually called. `[fixable]`


_Solid feature implementation with correct race-condition handling and idempotent guards, but `SubagentCancelledMsg` is missing from the frontend `ServerMessage` union type, and the new server-side and client-side logic lacks test coverage._

- 🔵 **style** (L145): `finishSubagent` uses `Array.isArray(sa.blocks)` as a runtime discriminator between `StreamingSubagentState` (Map) and `FinishedSubagentState` (array). This works but is fragile — it relies on a structural coincidence rather than the `running` field that the types were designed with. Consider `if ('running' in sa && sa.running === true)` or `if (sa.blocks instanceof Map)` for a discriminator that's more obviously tied to the type contract. `[fixable]`

server/chat.ts
```diff
@@ -929,6 +929,17 @@ export async function interruptChat(
     const echo = { type: 'user_message', messageId, text: fullPrompt };
     send(session.transport, echo);
     broadcastToObservers(session.observers, echo);
+    // Stop all active subagent tasks before interrupting the parent query.
+    // Without this, interrupt() only halts the parent — which is blocked
+    // waiting for the subagent, so the session hangs.
+    if (session.activeTaskIds.size > 0) {
```
dimakis, line 1012: 🟡 **missing_tests**: The `interruptChat` stopTask cascade logic (iterating `activeTaskIds` and calling `stopTask`) has no test. The existing `send-to-chat.test.ts` tests for `interruptChat` only update the mock to include `stopTask` in `queryInstance` but don't exercise a scenario where `activeTaskIds` is populated and `stopTask` is actually called. `[fixable]`

packages/client/src/slices/messages.ts
```diff
@@ -139,6 +141,26 @@ export type MessagesAction =
 
 // ─── Helpers ─────────────────────────────────────────────────────────────────
 
+function finishSubagent(sa: StreamingSubagentState | FinishedSubagentState): FinishedSubagentState {
+  if (Array.isArray(sa.blocks)) return sa as FinishedSubagentState;
```
dimakis: 🔵 **style**: `finishSubagent` uses `Array.isArray(sa.blocks)` as a runtime discriminator between `StreamingSubagentState` (Map) and `FinishedSubagentState` (array). This works but is fragile — it relies on a structural coincidence rather than the `running` field that the types were designed with. Consider `if ('running' in sa && sa.running === true)` or `if (sa.blocks instanceof Map)` for a discriminator that's more obviously tied to the type contract. `[fixable]`

server/query-loop.ts
```diff
@@ -1060,13 +1062,49 @@ async function _runQueryLoopInner(
...
+              currentSession?.activeTaskIds.delete(taskId);
+              log.info('subagent task finished', { clientId, taskId, status });
+            }
+            if (status === 'stopped' && toolUseId) {
+              const subagent = activeSubagents.get(toolUseId);
+              if (subagent) {
+                emit(
+                  v2('subagent_cancelled', {
+                    parentBlockId: subagent.parentBlockId,
+                    subagentMessageId: subagent.subagentMessageId,
```
dimakis, line 1100: 🟡 **bugs**: subagent.subagentMessageId is typed `string | null` in the activeSubagents map, but SubagentCancelledMsg.subagentMessageId is typed `string` (non-nullable) in ws-messages.ts. In practice, subagentMessageId is always initialized to a string (fallback `msg-${Date.now()}`), so this won't produce a runtime error — but the type mismatch is a latent bug. If future code ever sets subagentMessageId to null, it would silently emit a null where the client expects a string. `[fixable]`

server/query-loop.ts
```diff
@@ -1060,13 +1062,49 @@ async function _runQueryLoopInner(
...
+
+          // Track compaction events from SDK system status messages
           const compactResult = (msg as Record<string, unknown>).compact_result;
           if (subtype === 'status' && compactResult === 'success') {
             numCompactions++;
             log.info('compaction completed', { clientId, numCompactions });
           }
+
+          // Track subagent task lifecycle for interrupt cancellation
+          if (subtype === 'task_started') {
```
dimakis, line 1075: 🔵 **style**: The task_started handler uses optional chaining on currentSession (currentSession?.activeTaskIds.set) but doesn't log when currentSession is null. If task_started arrives without a current session, the tracking silently fails. The existing debug log only fires when taskId/toolUseId are missing, not when currentSession is null. `[fixable]`

server/__tests__/send-to-chat.test.ts
```diff
@@ -176,4 +184,36 @@ describe('interruptChat emits user_message via transport', () => {
...
+      interrupt: interruptSpy,
+      close: vi.fn(),
+      stopTask: stopTaskSpy,
+    };
+    session.activeTaskIds.set('task-abc', 'tool-1');
+    session.activeTaskIds.set('task-def', 'tool-2');
+
+    await interruptChat(CLIENT_ID, 'Stop everything');
+
+    expect(stopTaskSpy).toHaveBeenCalledTimes(2);
```
dimakis, line 214: 🔵 **missing_tests**: Missing test for stopTask failure resilience. The implementation uses `.catch()` + `Promise.allSettled()` to ensure interrupt() is still called even if stopTask throws, but there's no test verifying that behavior. A test where stopTaskSpy rejects would confirm interrupt is still called. `[fixable]`

server/chat.ts
```diff
@@ -1006,6 +1006,17 @@ export async function interruptChat(
...
```
dimakis, line 1012: 🔵 **unsafe_assumptions**: interruptChat does not clear activeTaskIds after calling stopTask. It relies on task_notification events arriving in the query loop to clean them up. If interrupt() terminates the query loop before those notifications arrive, stale entries could persist in activeTaskIds until the session is removed. Consider clearing the map after Promise.allSettled completes, since the tasks have been stopped regardless of whether the notifications arrive. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid interrupt-cascade implementation with good test coverage for the interruptChat path and client parser, but the core server-side event translation logic in query-loop.ts (task_started/task_notification handling) lacks test coverage, and there's a minor type mismatch on subagentMessageId nullability._

- 🟡 **missing_tests**: No query-loop tests for the new task_started/task_notification system message handling. The task lifecycle tracking (activeTaskIds.set/delete, subagent_cancelled emission) in query-loop.ts is untested. The protocol-parser test covers the client-side mapping, and send-to-chat.test.ts covers the interruptChat side, but the server-side event translation — the core logic of this PR — has no test coverage. `[fixable]`


### `server/query-loop.ts`
_Solid interrupt-cascade implementation with good test coverage for the interruptChat path and client parser, but the core server-side event translation logic in query-loop.ts (task_started/task_notification handling) lacks test coverage, and there's a minor type mismatch on subagentMessageId nullability._

- 🟡 **bugs** (L1100): subagent.subagentMessageId is typed `string | null` in the activeSubagents map, but SubagentCancelledMsg.subagentMessageId is typed `string` (non-nullable) in ws-messages.ts. In practice, subagentMessageId is always initialized to a string (fallback `msg-${Date.now()}`), so this won't produce a runtime error — but the type mismatch is a latent bug. If future code ever sets subagentMessageId to null, it would silently emit a null where the client expects a string. `[fixable]`
- 🔵 **style** (L1075): The task_started handler uses optional chaining on currentSession (currentSession?.activeTaskIds.set) but doesn't log when currentSession is null. If task_started arrives without a current session, the tracking silently fails. The existing debug log only fires when taskId/toolUseId are missing, not when currentSession is null. `[fixable]`


_Solid interrupt-cascade implementation with good test coverage for the interruptChat path and client parser, but the core server-side event translation logic in query-loop.ts (task_started/task_notification handling) lacks test coverage, and there's a minor type mismatch on subagentMessageId nullability._

- 🔵 **missing_tests** (L214): Missing test for stopTask failure resilience. The implementation uses `.catch()` + `Promise.allSettled()` to ensure interrupt() is still called even if stopTask throws, but there's no test verifying that behavior. A test where stopTaskSpy rejects would confirm interrupt is still called. `[fixable]`


### `server/chat.ts`
_Solid interrupt-cascade implementation with good test coverage for the interruptChat path and client parser, but the core server-side event translation logic in query-loop.ts (task_started/task_notification handling) lacks test coverage, and there's a minor type mismatch on subagentMessageId nullability._

- 🔵 **unsafe_assumptions** (L1012): interruptChat does not clear activeTaskIds after calling stopTask. It relies on task_notification events arriving in the query loop to clean them up. If interrupt() terminates the query loop before those notifications arrive, stale entries could persist in activeTaskIds until the session is removed. Consider clearing the map after Promise.allSettled completes, since the tasks have been stopped regardless of whether the notifications arrive. `[fixable]`

PR #322, opened 2026-05-14
fix(worktree): branch from main to prevent cross-session contamination

dimakis:
## Summary

- `createWorktree()` and `createWorktreeAsync()` called `git worktree add -b <branch> <path>` without a start-point, causing new sessions to inherit whatever HEAD was in the base repo — often another session branch with unmerged commits
- Fix: add `startPoint` option (default `'main'`) as the final argument to `git worktree add`
- When falling back to an existing branch, reset it to `startPoint` before attaching

## Context

Discovered while investigating mgmt PR #105, where a design doc PR inherited 7 unrelated `jira_process` commits from a prior session. The root cause: worktree creation branched from the repo's current HEAD (a stale session branch) instead of `main`.

## Test plan

- [x] New test: worktree branches from main, not HEAD (verifies the bug fix)
- [x] New test: custom `startPoint` option respected
- [x] New test: existing branch reset to `startPoint` on fallback
- [x] All 35 worktree tests pass
- [x] Lint + format pass (pre-commit hooks)

### DIFF
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -35,11 +35,11 @@ const log = createLogger('worktree');
  */
 export function detectDefaultBranch(repoPath: string): string {
   try {
-    const ref = execFileSync(
-      'git',
-      ['-C', repoPath, 'symbolic-ref', 'refs/remotes/origin/HEAD'],
-      { encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe'], timeout: WORKTREE_GIT_TIMEOUT_MS },
-    ).trim();
+    const ref = execFileSync('git', ['-C', repoPath, 'symbolic-ref', 'refs/remotes/origin/HEAD'], {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: WORKTREE_GIT_TIMEOUT_MS,
+    }).trim();
     // ref looks like "refs/remotes/origin/main" — extract the last segment
     const branch = ref.split('/').pop();
     if (branch) return branch;
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -31,7 +31,10 @@ const log = createLogger('worktree');
 /**
  * Detect the default branch of a repo (e.g. 'main' or 'master').
  * Uses `git symbolic-ref refs/remotes/origin/HEAD` when an origin exists,
- * otherwise falls back to 'main'.
+ * otherwise falls back to 'main' (verified) or HEAD.
+ *
+ * NOTE: Uses execFileSync intentionally — called during session setup (not hot path).
+ * Worktree creation is a one-time initialization per session, so blocking is acceptable.
  */
 export function detectDefaultBranch(repoPath: string): string {
   try {
@@ -40,13 +43,25 @@ export function detectDefaultBranch(repoPath: string): string {
       stdio: ['pipe', 'pipe', 'pipe'],
       timeout: WORKTREE_GIT_TIMEOUT_MS,
     }).trim();
-    // ref looks like "refs/remotes/origin/main" — extract the last segment
-    const branch = ref.split('/').pop();
+    // ref looks like "refs/remotes/origin/main" or "refs/remotes/origin/release/stable"
+    const branch = ref.replace('refs/remotes/origin/', '');
     if (branch) return branch;
   } catch {
     // No origin or symbolic-ref not set — fall back
   }
-  return 'main';
+
+  // Verify 'main' exists before using it as fallback
+  try {
+    execFileSync('git', ['-C', repoPath, 'rev-parse', '--verify', 'main'], {
+      stdio: 'pipe',
+      timeout: WORKTREE_GIT_TIMEOUT_MS,
+    });
+    return 'main';
+  } catch {
+    // 'main' doesn't exist — fall back to HEAD
+  }
+
+  return 'HEAD';
 }
 
 /** Worktrees live inside each repo at .claude/worktrees/<sessionId>. */
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -56,6 +56,8 @@ export interface CreateWorktreeOptions {
   dir?: string;
   branch?: string;
   prefix?: '.claude' | '.cursor';
+  /** Git ref to branch from. Defaults to 'main' to avoid inheriting unmerged session commits. */
+  startPoint?: string;
 }
 
 export function createWorktree(
@@ -69,6 +71,7 @@ export function createWorktree(
 
   const worktreePath = join(dir, sessionId);
   const branch = opts?.branch ?? `${WORKTREE_BRANCH_PREFIX}${sessionId}`;
+  const startPoint = opts?.startPoint ?? 'main';
 
   // If the worktree path already exists (stale from a previous session),
   // check whether it's a valid worktree we can reuse or a stale directory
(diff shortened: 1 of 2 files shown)


dimakis:
## Centaur Review

### `server/worktree.ts`
_Good fix for a real cross-session contamination bug. The hardcoded `'main'` default is the main concern — it breaks portability for repos not using `main` as default branch. Tests need `git init -b main` to be reliable, and the sync `createWorktree` variant lacks test coverage for the new behavior._

- 🟡 **unsafe_assumptions** (L74): Hardcoded `'main'` as default startPoint. Repos using `master` or another default branch will fail with `fatal: not a valid object name: 'main'`. Secondary repos configured in `.mitzo.json` `repos` are particularly at risk since they're external. Consider detecting the default branch dynamically (e.g. `git symbolic-ref refs/remotes/origin/HEAD`) or reading from `.mitzo.json` per-repo config. `[fixable]`
- 🔵 **regressions** (L79): The 'reuse existing worktree' early return (lines 79-87 sync, 145-151 async) does not reset the branch to `startPoint`. If a worktree from a prior run still exists and is valid but points to contaminated commits, it will be reused as-is, leaving the contamination bug unfixed for that path. This is likely acceptable since session IDs include random hex making collisions improbable, but worth documenting.


_Good fix for a real cross-session contamination bug. The hardcoded `'main'` default is the main concern — it breaks portability for repos not using `main` as default branch. Tests need `git init -b main` to be reliable, and the sync `createWorktree` variant lacks test coverage for the new behavior._

- 🟡 **missing_tests**: The sync `createWorktree` function received the same startPoint changes as `createWorktreeAsync`, but has no corresponding tests. The sync version is the one used in production by `createSessionWorktrees` (worktree.ts:272) and `app.ts:375`. The async variant's tests don't cover the sync code path. `[fixable]`
- 🟡 **bugs** (L127): Test `beforeEach` uses `git init` without `-b main`, so the default branch name depends on the system's `init.defaultBranch` git config. On systems where it defaults to `master`, all three new tests will fail because the `startPoint` default resolves `'main'` which won't exist. Use `git init -b main` to make tests self-contained. `[fixable]`

server/worktree.ts
```diff
@@ -28,6 +28,27 @@ import { createLogger } from './logger.js';
...
+ */
+export function detectDefaultBranch(repoPath: string): string {
+  try {
+    const ref = execFileSync(
+      'git',
+      ['-C', repoPath, 'symbolic-ref', 'refs/remotes/origin/HEAD'],
+      { encoding: 'utf-8', stdio: ['pipe', 'pipe', 'pipe'], timeout: WORKTREE_GIT_TIMEOUT_MS },
+    ).trim();
+    // ref looks like "refs/remotes/origin/main" — extract the last segment
+    const branch = ref.split('/').pop();
```
dimakis: 🟡 **bugs**: `ref.split('/').pop()` incorrectly extracts branch names that contain slashes. If the default branch is `release/stable`, the symbolic-ref is `refs/remotes/origin/release/stable` and `.pop()` returns `stable` instead of `release/stable`. Use `ref.replace('refs/remotes/origin/', '')` instead. `[fixable]`

server/worktree.ts
```diff
@@ -28,6 +28,27 @@ import { createLogger } from './logger.js';
...
+    if (branch) return branch;
+  } catch {
+    // No origin or symbolic-ref not set — fall back
+  }
+  return 'main';
```
dimakis: 🟡 **regressions**: Hardcoded fallback to `'main'` regresses repos that use `master` (or another name) as default branch and have no origin configured. Previously, omitting `startPoint` caused `git worktree add` to branch from HEAD, which always works. Now it passes `'main'` as startPoint, which fails with a fatal error if the ref doesn't exist. Consider verifying the fallback ref exists (e.g. `git rev-parse --verify main`) and falling back to `HEAD` if it doesn't. `[fixable]`

server/worktree.ts
```diff
@@ -28,6 +28,27 @@ import { createLogger } from './logger.js';
...
 const log = createLogger('worktree');
 
+/**
+ * Detect the default branch of a repo (e.g. 'main' or 'master').
+ * Uses `git symbolic-ref refs/remotes/origin/HEAD` when an origin exists,
+ * otherwise falls back to 'main'.
+ */
+export function detectDefaultBranch(repoPath: string): string {
+  try {
+    const ref = execFileSync(
```
dimakis: 🔵 **unsafe_assumptions**: `detectDefaultBranch` uses synchronous `execFileSync` but is called from the async `createWorktreeAsync` path (line 164), blocking the event loop. Consider an async variant or accepting that the git call is fast enough (~5ms) to not matter in practice. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_Solid fix for a real cross-session contamination bug with thorough test coverage. Two issues worth addressing: `split('/').pop()` mishandles slashed branch names, and the hardcoded `'main'` fallback can break repos using `master` without an origin._

- 🟡 **bugs** (L44): `ref.split('/').pop()` incorrectly extracts branch names that contain slashes. If the default branch is `release/stable`, the symbolic-ref is `refs/remotes/origin/release/stable` and `.pop()` returns `stable` instead of `release/stable`. Use `ref.replace('refs/remotes/origin/', '')` instead. `[fixable]`
- 🟡 **regressions** (L49): Hardcoded fallback to `'main'` regresses repos that use `master` (or another name) as default branch and have no origin configured. Previously, omitting `startPoint` caused `git worktree add` to branch from HEAD, which always works. Now it passes `'main'` as startPoint, which fails with a fatal error if the ref doesn't exist. Consider verifying the fallback ref exists (e.g. `git rev-parse --verify main`) and falling back to `HEAD` if it doesn't. `[fixable]`
- 🔵 **unsafe_assumptions** (L38): `detectDefaultBranch` uses synchronous `execFileSync` but is called from the async `createWorktreeAsync` path (line 164), blocking the event loop. Consider an async variant or accepting that the git call is fast enough (~5ms) to not matter in practice. `[fixable]`

dimakis (commented):
## Codex Review

### `server/worktree.ts`
_PR fixes the HEAD contamination bug and adds strong coverage, but default-branch detection may regress repos whose local base branch intentionally differs from origin/HEAD._

- 🟡 **regressions** (L39): `detectDefaultBranch()` now prefers `origin/HEAD` over the local repo state, which changes behavior for repos whose local checked-out default branch intentionally differs from the remote default. In that case new session worktrees will suddenly branch from the remote default branch instead of the local base branch, potentially creating sessions from the wrong code line. Consider preferring an explicit local branch/ref when present (e.g. current repo default branch or configured branch) and only using `origin/HEAD` as a fallback. `[fixable]`

server/worktree.ts
```diff
@@ -28,6 +28,42 @@ import { createLogger } from './logger.js';
...
 
+/**
+ * Detect the default branch of a repo (e.g. 'main' or 'master').
+ * Uses `git symbolic-ref refs/remotes/origin/HEAD` when an origin exists,
+ * otherwise falls back to 'main' (verified) or HEAD.
+ *
+ * NOTE: Uses execFileSync intentionally — called during session setup (not hot path).
+ * Worktree creation is a one-time initialization per session, so blocking is acceptable.
+ */
+export function detectDefaultBranch(repoPath: string): string {
```
dimakis, line 40: 🟡 **regressions**: `detectDefaultBranch()` now prefers `origin/HEAD` over the local repo state, which changes behavior for repos whose local checked-out default branch intentionally differs from the remote default. In that case new session worktrees will suddenly branch from the remote default branch instead of the local base branch, potentially creating sessions from the wrong code line. Consider preferring an explicit local branch/ref when present (e.g. current repo default branch or configured branch) and only using `origin/HEAD` as a fallback. `[fixable]`

server/worktree.ts
```diff
@@ -96,13 +136,19 @@ export function createWorktree(
...
+    execFileSync(
+      'git',
+      ['-C', baseRepo, 'worktree', 'add', '-b', branch, worktreePath, startPoint],
+      { stdio: 'pipe', timeout: WORKTREE_GIT_TIMEOUT_MS },
+    );
   } catch (err: unknown) {
     const message = err instanceof Error ? err.message : String(err);
     if (message.includes('already exists')) {
+      // Branch exists but no worktree — reset it to startPoint before attaching
+      execFileSync('git', ['-C', baseRepo, 'branch', '-f', branch, startPoint], {
```
dimakis, line 148: 🟡 **bugs**: The fallback path force-resets an existing session branch with `git branch -f` before attaching it. If that branch is currently checked out in another existing worktree, Git rejects force-updating it (`cannot force update the branch ... checked out at ...`), so `createWorktree`/`createWorktreeAsync` will now fail in a case that previously succeeded. This can happen if a stale branch remains attached elsewhere or concurrent session setup races on the same session id. `[fixable]`

dimakis (commented):
## Codex Review

### `server/worktree.ts`
_PR fixes the HEAD contamination bug and adds strong coverage, but the new branch-reset fallback can fail when the branch is checked out in another worktree._

- 🟡 **bugs** (L148): The fallback path force-resets an existing session branch with `git branch -f` before attaching it. If that branch is currently checked out in another existing worktree, Git rejects force-updating it (`cannot force update the branch ... checked out at ...`), so `createWorktree`/`createWorktreeAsync` will now fail in a case that previously succeeded. This can happen if a stale branch remains attached elsewhere or concurrent session setup races on the same session id. `[fixable]`

PR #319, opened 2026-05-13
feat(ui): inline markdown preview cards in chat

dimakis:
## Summary
- Markdown file paths (`.md`/`.mdx`) in chat messages now render as collapsible preview cards instead of plain links
- Cards lazy-load file content on first expand via `/api/files/read` and render with ReactMarkdown + remark-gfm
- Solo-paragraph `.md` links become cards; inline references within text stay as clickable links
- "Open" button on card header navigates to FileViewer for full reading/editing
- Follows existing ToolPill/ThinkingBlock expand/collapse pattern

## Files Changed
- **New:** `frontend/src/components/MarkdownPreviewCard.tsx` — standalone card component
- **Modified:** `frontend/src/components/MessageBubble.tsx` — `p` renderer promotes solo `.md` links to cards, `a` renderer tags `.md` paths via `data-md-preview`
- **Modified:** `frontend/src/styles/global.css` — `.md-preview-card-*` styles with scoped markdown rendering

## Test plan
- [ ] Send a message that references a `.md` file path on its own line — should render as a collapsed card
- [ ] Expand card — content fetched and rendered as markdown
- [ ] Click "Open" — navigates to FileViewer with back-navigation preserved
- [ ] Reference a `.md` file inline with other text — should stay as a regular link
- [ ] Reference a non-`.md` file — should stay as a regular link
- [ ] Reference a missing `.md` file — card should show error on expand
- [ ] Verify no regressions in existing file path link behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #323, opened 2026-05-14
fix(ui): markdown preview card promotion broken in ReactMarkdown v10

dimakis:
## Summary

- **Bug:** `.md` file paths in assistant messages rendered as plain links instead of expandable preview cards (introduced in PR #319)
- **Root cause:** Two compounding issues in ReactMarkdown v10:
  1. v10 sanitizes unknown URL schemes — `file-path://` was stripped to `""`, so the `a` handler never matched
  2. The `p` handler checked `data-file-path` on children, but in v10 children are unrendered component instances (the `a` handler hasn't executed yet), so the prop never existed
- **Fix:** Add `urlTransform` to preserve all URLs, and check `href` prop (which ReactMarkdown passes pre-render) instead of `data-file-path`

## Test plan

- [x] All 27 MessageBubble tests pass (updated 3 promotion tests to match new approach)
- [ ] Verify standalone `.md` path in chat renders as collapsible preview card
- [ ] Verify non-`.md` file paths still render as clickable links
- [ ] Verify inline `.md` paths (mixed with text) stay as links, not cards

### DIFF
--- a/frontend/src/components/MessageBubble.tsx
+++ b/frontend/src/components/MessageBubble.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect, useRef } from 'react';
-import ReactMarkdown from 'react-markdown';
+import ReactMarkdown, { defaultUrlTransform } from 'react-markdown';
 import remarkGfm from 'remark-gfm';
 import { useNavigate, useLocation } from 'react-router-dom';
 import type { FinishedMessage } from '../types/chat';
@@ -106,7 +106,7 @@ export function TextBubble({ content, streaming = false, timestamp, readAloud }:
       <div className="msg-bubble-markdown" ref={contentRef}>
         <ReactMarkdown
           remarkPlugins={[remarkGfm]}
-          urlTransform={(url) => url}
+          urlTransform={(url) => (url.startsWith(FILE_SCHEME) ? url : defaultUrlTransform(url))}
           components={{
             table: ({ children, ...props }) => (
               <div className="table-scroll-wrapper">
--- a/frontend/src/components/MessageBubble.tsx
+++ b/frontend/src/components/MessageBubble.tsx
@@ -106,6 +106,7 @@ export function TextBubble({ content, streaming = false, timestamp, readAloud }:
       <div className="msg-bubble-markdown" ref={contentRef}>
         <ReactMarkdown
           remarkPlugins={[remarkGfm]}
+          urlTransform={(url) => url}
           components={{
             table: ({ children, ...props }) => (
               <div className="table-scroll-wrapper">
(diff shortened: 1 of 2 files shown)


frontend/src/components/MessageBubble.tsx
```diff
@@ -106,6 +106,7 @@ export function TextBubble({ content, streaming = false, timestamp, readAloud }:
       <div className="msg-bubble-markdown" ref={contentRef}>
         <ReactMarkdown
           remarkPlugins={[remarkGfm]}
+          urlTransform={(url) => url}
```
dimakis: 🟡 **unsafe_assumptions**: `urlTransform={(url) => url}` disables ReactMarkdown's built-in URL sanitization, which blocks `javascript:`, `vbscript:`, and `data:` schemes. While assistant content is less attacker-controlled than user input, a prompt-injection scenario could produce a clickable `javascript:` link. A targeted transform would be safer: `urlTransform={(url) => url.startsWith(FILE_SCHEME) ? url : defaultUrlTransform(url)}` (import `defaultUrlTransform` from `react-markdown`). `[fixable]`

dimakis (commented):
## Centaur Review

_Correct fix for the v10 render-order change, but the identity `urlTransform` disables built-in URL sanitization — use a targeted transform that only passes `file-path://` through._

- 🟡 **unsafe_assumptions** (L109): `urlTransform={(url) => url}` disables ReactMarkdown's built-in URL sanitization, which blocks `javascript:`, `vbscript:`, and `data:` schemes. While assistant content is less attacker-controlled than user input, a prompt-injection scenario could produce a clickable `javascript:` link. A targeted transform would be safer: `urlTransform={(url) => url.startsWith(FILE_SCHEME) ? url : defaultUrlTransform(url)}` (import `defaultUrlTransform` from `react-markdown`). `[fixable]`


_Correct fix for the v10 render-order change, but the identity `urlTransform` disables built-in URL sanitization — use a targeted transform that only passes `file-path://` through._

- 🔵 **missing_tests**: The regex `/\.mdx?$/i` also matches `.mdx` files, but all promotion tests only use `.md` extensions. Adding a test with a `.mdx` link would verify both branches of the quantifier. `[fixable]`
- 🔵 **missing_tests**: No test covers the case where a paragraph contains a single link element whose `href` does NOT start with `FILE_SCHEME` (e.g., a regular `https://` URL). This would verify the `p` handler falls through to rendering a plain `<p>` for non-file-path links. `[fixable]`

dimakis:
## Centaur Review

Found **2** issue(s).


_Clean, well-scoped fix for the ReactMarkdown v10 breaking change. Two new tests fill real gaps (.mdx promotion, regular-URL rejection). Minor gap: the test mock doesn't export `defaultUrlTransform` and no test covers the `urlTransform` callback itself._

- 🔵 **missing_tests** (L7): The `react-markdown` mock only exports `default` but the component now imports the named export `defaultUrlTransform`. It will be `undefined` in tests. Currently harmless because the mock ReactMarkdown never invokes the `urlTransform` callback, but if a future test exercises that path it will throw a TypeError. Add `defaultUrlTransform: (url: string) => url` to the mock factory. `[fixable]`
- 🔵 **missing_tests**: No test verifies the `urlTransform` prop itself — i.e., that `file-path://` URLs are preserved while other URLs pass through `defaultUrlTransform`. This is the mechanism that prevents ReactMarkdown v10 from sanitizing the custom scheme. A targeted unit test for the callback would guard against future regressions if the `urlTransform` signature changes. `[fixable]`

PR #324, opened 2026-05-15
fix(ui): add Load More button to desktop session panel

dimakis:
## Summary
- Desktop `SessionPanel` was missing pagination — only the first 20 sessions were visible with no way to load more
- The `useSessionList` hook already exposes `hasMore`/`loadMore`/`loadingMore`; the mobile `SessionList` uses them but desktop never wired them up
- Added the same "Load More" button (reusing existing `.session-load-more` CSS class) to the desktop "All" session list

## Test plan
- [ ] Open desktop UI, switch to "All" tab in session sidebar
- [ ] Verify "Load More" appears at bottom when >20 sessions exist
- [ ] Click it — older sessions append, button disappears when no more pages
- [ ] Verify button shows "Loading..." while fetching

### DIFF
--- a/frontend/src/components/SessionPanel.tsx
+++ b/frontend/src/components/SessionPanel.tsx
@@ -23,7 +23,7 @@ function readViewMode(): ViewMode {
 }
 
 export function SessionPanel({ activeSessionId, onSelectSession, onNewChat }: SessionPanelProps) {
-  const { sessions, loading, dismissSession } = useSessionList();
+  const { sessions, loading, loadingMore, hasMore, loadMore, dismissSession } = useSessionList();
   const { attendCount } = useSessionOverview();
   const [viewMode, setViewMode] = useState<ViewMode>(readViewMode);
 
@@ -116,6 +116,11 @@ export function SessionPanel({ activeSessionId, onSelectSession, onNewChat }: Se
                   </button>
                 </div>
               ))}
+              {hasMore && (
+                <button className="session-load-more" onClick={loadMore} disabled={loadingMore}>
+                  {loadingMore ? 'Loading...' : 'Load More'}
+                </button>
+              )}
             </div>
           )}
         </>
--- a/frontend/src/components/__tests__/SessionPanel.test.tsx
+++ b/frontend/src/components/__tests__/SessionPanel.test.tsx
@@ -7,6 +7,15 @@ vi.mock('../../hooks/useSessionList', () => ({
   useSessionList: vi.fn(),
 }));
 
+// Mock useSessionOverview to avoid EventSource dependency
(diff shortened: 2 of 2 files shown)


frontend/src/components/SessionPanel.tsx
```diff
@@ -99,6 +99,11 @@ export function SessionPanel({ activeSessionId, onSelectSession, onNewChat }: Se
                   </button>
                 </div>
               ))}
+              {hasMore && (
```
dimakis, line 119: 🔵 **missing_tests**: The existing SessionPanel test file (components/__tests__/SessionPanel.test.tsx) mocks `hasMore: false` but has no test cases exercising the Load More button — rendering when `hasMore` is true, the disabled state when `loadingMore` is true, or that `loadMore` is called on click. The mobile SessionList page uses the identical pattern and also lacks dedicated tests for it, so this is a pre-existing gap, but worth noting. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `frontend/src/components/SessionPanel.tsx`
_Clean, minimal change that mirrors the existing mobile Load More pattern exactly — hook contract, CSS class, and conditional rendering all verified. Only gap is missing test coverage for the new button._

- 🔵 **missing_tests** (L102): The existing SessionPanel test file (components/__tests__/SessionPanel.test.tsx) mocks `hasMore: false` but has no test cases exercising the Load More button — rendering when `hasMore` is true, the disabled state when `loadingMore` is true, or that `loadMore` is called on click. The mobile SessionList page uses the identical pattern and also lacks dedicated tests for it, so this is a pre-existing gap, but worth noting. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `frontend/src/components/__tests__/SessionPanel.test.tsx`
_Clean, well-tested PR that brings desktop parity with the mobile Load More button. Implementation is correct, mocks match real interfaces, CSS exists, and the only gap is a minor missing negative test case._

- 🔵 **missing_tests**: No test verifies that the Load More button is hidden when hasMore is false (the default). A negative case like `it('does not render Load More when hasMore is false')` with a `queryByText('Load More')` returning null would make the conditional rendering contract explicit. `[fixable]`

dimakis:
All Centaur review items have been addressed:

✅ **Load More button test coverage** — 4 test cases added:
1. Button hidden when `hasMore` is false (negative case)
2. Button renders when `hasMore` is true
3. Button disabled when `loadingMore` is true
4. `loadMore` callback called on click

All tests pass (13/13). Ready for merge.

Commits:
- `5c7028b` — initial test coverage
- `2da5d86` — negative test case

PR #325, opened 2026-05-15
fix(client): recover SSE on browser visibilitychange

dimakis:
## Summary
- iOS Safari kills EventSource connections when the tab is backgrounded without firing an error event
- The Capacitor lifecycle hook already handled this, but browser clients (e.g. phone via Safari PWA) had no recovery
- This caused the health SSE stream to die permanently after a single background cycle, hiding the voice mic button

## Fix
Add a `visibilitychange` listener in `event-bus-singleton.ts` that calls `ensureConnected()` when the page becomes visible.

## Test plan
- [x] New test: `event-bus-singleton.test.ts` — verifies `ensureConnected` called on visible, not called on hidden
- [x] Full test suite passes (5 pre-existing failures unrelated to this change)
- [ ] Manual: open Mitzo in mobile Safari, background the tab, return — mic button should persist

### DIFF
--- a/frontend/src/lib/__tests__/event-bus-singleton.test.ts
+++ b/frontend/src/lib/__tests__/event-bus-singleton.test.ts
@@ -1,0 +1,57 @@
+// @vitest-environment jsdom
+// Tests for the global SSE EventBus singleton's visibilitychange recovery.
+
+import { describe, it, expect, vi } from 'vitest';
+
+// Mock EventSource (jsdom doesn't provide it)
+class MockEventSource {
+  static CONNECTING = 0;
+  static OPEN = 1;
+  static CLOSED = 2;
+  readyState = MockEventSource.CONNECTING;
+  onopen: ((ev: unknown) => void) | null = null;
+  onerror: ((ev: unknown) => void) | null = null;
+  close = vi.fn();
+  addEventListener = vi.fn();
+  removeEventListener = vi.fn();
+}
+
+// Must be set before module loads — static imports are hoisted above beforeAll
+global.EventSource = MockEventSource as unknown as typeof EventSource;
+
+// Dynamic import so the module-level side effects run after EventSource is defined
+const { eventBus } = await import('../event-bus-singleton');
+
+describe('event-bus-singleton visibilitychange recovery', () => {
+  it('calls ensureConnected when page becomes visible', () => {
+    const ensureConnectedSpy = vi.spyOn(eventBus, 'ensureConnected');
+
+    // Simulate page becoming visible
+    Object.defineProperty(document, 'visibilityState', {
+      value: 'visible',
+      writable: true,
+      configurable: true,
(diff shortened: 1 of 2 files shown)


frontend/src/lib/__tests__/event-bus-singleton.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+// Dynamic import so the module-level side effects run after EventSource is defined
+const { eventBus } = await import('../event-bus-singleton');
+
+describe('event-bus-singleton visibilitychange recovery', () => {
+  it('calls ensureConnected when page becomes visible', () => {
+    const ensureConnectedSpy = vi.spyOn(eventBus, 'ensureConnected');
+
+    // Simulate page becoming visible
+    Object.defineProperty(document, 'visibilityState', {
```
dimakis, line 30: 🟡 **unsafe_assumptions**: The test mutates `document.visibilityState` via `Object.defineProperty` without restoring it in an `afterEach`. Since both tests modify the same global, a test ordering dependency exists — the second test ('hidden') could inadvertently pass if it relies on state left by the first test. Use `afterEach` to restore the original descriptor. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/lib/event-bus-singleton.ts`
_The PR ships tests for a visibilitychange recovery feature whose implementation is missing from the diff — the singleton module has no listener, so both tests will fail._

- 🔴 **bugs**: The PR adds tests for a visibilitychange listener that calls `eventBus.ensureConnected()`, but the listener itself was never added to `event-bus-singleton.ts`. The module still only exports the EventBus and calls `connect()` — there is no `document.addEventListener('visibilitychange', ...)`. Both tests will fail because dispatching the event does nothing. The commit message says 'Add a visibilitychange listener in the EventBus singleton' but the implementation is missing from the diff. `[fixable]`


### `frontend/src/lib/__tests__/event-bus-singleton.test.ts`
_The PR ships tests for a visibilitychange recovery feature whose implementation is missing from the diff — the singleton module has no listener, so both tests will fail._

- 🔵 **missing_tests**: No test verifies the actual recovery behavior — that `ensureConnected()` recreates a CLOSED EventSource. The tests only check that `ensureConnected` is called, but a test confirming the EventSource is recreated when readyState is CLOSED (the real bug scenario from the commit message) would add meaningful coverage beyond the spy check. `[fixable]`
- 🟡 **unsafe_assumptions** (L30): The test mutates `document.visibilityState` via `Object.defineProperty` without restoring it in an `afterEach`. Since both tests modify the same global, a test ordering dependency exists — the second test ('hidden') could inadvertently pass if it relies on state left by the first test. Use `afterEach` to restore the original descriptor. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #326, opened 2026-05-16
Fix session list timestamp reset and missing new sessions

dimakis:
## Summary

Fixes two bugs in the session list:

1. **All sessions show same timestamp after server restart** — when sessions are resumed (e.g., after a server restart), their `updated_at` timestamps were being reset to "now" instead of being preserved. This caused all recently-resumed sessions to show identical relative times like "18m ago".

2. **Newly created sessions disappear from list** — sessions with zero turns and zero prompts that aren't currently active were being filtered out, hiding newly created sessions that hadn't completed their first turn yet.

## Changes

**`server/chat.ts:687-697`** — Resume timestamp preservation
- Read existing session metadata via `eventStore.getSession()` before calling `upsertSession()` on resume
- Pass existing `updatedAt` explicitly to prevent EventStore from defaulting to current time
- Only applies when session already exists (resume path); new sessions get natural timestamp

**`server/chat.ts:1354-1358`** — Grace period for new sessions
- Keep zero-turn, zero-prompt, inactive sessions visible for 1 hour after creation
- After 1 hour with no activity, apply the filter as before (hides automated/discovered sessions)
- Handles race conditions where session is created but query loop hasn't finished first turn

## Test Plan

- [ ] Verify session list shows correct relative timestamps (not all identical)
- [ ] Verify newly created sessions appear in list immediately
- [ ] Verify sessions created >1h ago with no turns are still filtered out
- [ ] Restart server and confirm session timestamps don't reset

### DIFF
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -689,7 +689,9 @@ async function _startChatInner(
 
   // Pre-register resumed sessions in EventStore so they're discoverable
   // even if the query loop dies before the first assistant event.
+  // Preserve existing updatedAt so server restarts don't reset all timestamps.
   if (options.resume) {
+    const existingMeta = eventStore.getSession(options.resume);
     eventStore.upsertSession({
       sessionId: options.resume,
       cwd,
@@ -698,6 +700,7 @@ async function _startChatInner(
       isActive: true,
       ...(worktreePath ? { wtId } : {}),
       ...(options.telosTaskId ? { telosTaskId: options.telosTaskId } : {}),
+      ...(existingMeta ? { updatedAt: existingMeta.updatedAt } : {}),
     });
   }
 
@@ -1509,10 +1512,12 @@ export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
   const all = eventStore.listSessions().filter((m) => {
     // Hide sessions that were never used through Mitzo (e.g. automated
     // code review sessions discovered from filesystem).  Active sessions
-    // always show regardless of turn count.
-    // Note: new sessions are created with is_active=1 by default (see EventStore
-    // schema), so a brand-new session will never be filtered out here.
-    if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) return false;
+    // always show regardless of turn count.  Recently created sessions
+    // (< 1 hour) are kept even with no turns — they may still be starting.
+    if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) {
+      const ONE_HOUR = 60 * 60 * 1000;
+      return Date.now() - m.createdAt < ONE_HOUR;
+    }
     return true;
   });
   const page = all.slice(offset, offset + limit);
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -40,6 +40,9 @@ import { INTERNAL_TOKEN } from './internal-token.js';
 import { buildTaskSystemPrompt } from './task-context.js';
 import type { TaskStore } from './task-store.js';
 
+/** Grace period for zero-turn inactive sessions — recently created ones stay visible. */
+const ZERO_TURN_GRACE_MS = 60 * 60 * 1000; // 1 hour
+
 let _taskStore: TaskStore | null = null;
 export function setTaskStore(store: TaskStore): void {
   _taskStore = store;
@@ -1515,8 +1518,7 @@ export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
     // always show regardless of turn count.  Recently created sessions
     // (< 1 hour) are kept even with no turns — they may still be starting.
     if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) {
-      const ONE_HOUR = 60 * 60 * 1000;
-      return Date.now() - m.createdAt < ONE_HOUR;
+      return Date.now() - m.createdAt < ZERO_TURN_GRACE_MS;
     }
     return true;
   });
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -35,13 +35,11 @@ import {
   SESSION_PAGE_SIZE,
   SESSION_MESSAGES_LIMIT,
   USER_CLOSEOUT_TIMEOUT_MS,
+  ZERO_TURN_GRACE_MS,
 } from './constants.js';
 import { INTERNAL_TOKEN } from './internal-token.js';
 import { buildTaskSystemPrompt } from './task-context.js';
 import type { TaskStore } from './task-store.js';
-
-/** Grace period for zero-turn inactive sessions — recently created ones stay visible. */
-const ZERO_TURN_GRACE_MS = 60 * 60 * 1000; // 1 hour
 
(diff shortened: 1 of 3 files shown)


server/chat.ts
```diff
@@ -1348,10 +1351,12 @@ export function getSessionsCached(offset = 0, limit = SESSION_PAGE_SIZE) {
   const all = eventStore.listSessions().filter((m) => {
     // Hide sessions that were never used through Mitzo (e.g. automated
     // code review sessions discovered from filesystem).  Active sessions
-    // always show regardless of turn count.
-    // Note: new sessions are created with is_active=1 by default (see EventStore
-    // schema), so a brand-new session will never be filtered out here.
-    if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) return false;
+    // always show regardless of turn count.  Recently created sessions
+    // (< 1 hour) are kept even with no turns — they may still be starting.
+    if (m.numTurns === 0 && m.promptCount === 0 && !m.isActive) {
```
dimakis, line 1519: 🟡 **missing_tests**: The 1-hour grace period logic is not covered by existing tests in server/__tests__/session-cache.test.ts. The test suite verifies that zero-turn inactive sessions are hidden and that active sessions are shown, but no test creates a recent zero-turn inactive session to verify it's kept, nor an old one to verify it's hidden. Add two tests: one with createdAt < 1 hour ago (should show) and one with createdAt > 1 hour ago (should hide). `[fixable]`

dimakis, line 1519: 🟡 **regressions**: The old filter unconditionally hid zero-turn, zero-prompt, inactive sessions. The new filter keeps them visible for up to 1 hour. This means sessions discovered from the filesystem (e.g. automated code review sessions) that were created within the last hour will now appear in the session list, where previously they were always hidden. This is likely intentional but is a behavioral change — the PR title mentions 'missing new sessions', so this is presumably the fix for that. Worth a note in the PR description.

server/chat.ts
```diff
@@ -684,14 +684,17 @@ async function _startChatInner(
 
   // Pre-register resumed sessions in EventStore so they're discoverable
   // even if the query loop dies before the first assistant event.
+  // Preserve existing updatedAt so server restarts don't reset all timestamps.
   if (options.resume) {
+    const existingMeta = eventStore.getSession(options.resume);
```
dimakis, line 695: 🔵 **missing_tests**: The timestamp-preservation path (passing existing updatedAt on resume) has no direct test at the chat.ts integration level. The underlying EventStore.upsertSession respecting an explicit updatedAt is well-tested in protocol/__tests__/event-store.test.ts, so the risk is low, but a test in session-cache.test.ts or a chat-level test verifying that resuming a session doesn't change its updatedAt would confirm the end-to-end behavior. `[fixable]`

server/chat.ts
dimakis, line 1519: 🔵 **style**: The constant `ONE_HOUR` is defined inline inside the filter callback and re-created on every call to getSessionsCached (and for every session in the list). Consider hoisting it to module scope or into constants.ts alongside other server-wide constants for reuse and discoverability. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Correct and well-scoped fix for two related session-list bugs; main gap is missing test coverage for the new 1-hour grace period filter logic._

- 🟡 **missing_tests** (L1356): The 1-hour grace period logic is not covered by existing tests in server/__tests__/session-cache.test.ts. The test suite verifies that zero-turn inactive sessions are hidden and that active sessions are shown, but no test creates a recent zero-turn inactive session to verify it's kept, nor an old one to verify it's hidden. Add two tests: one with createdAt < 1 hour ago (should show) and one with createdAt > 1 hour ago (should hide). `[fixable]`
- 🟡 **regressions** (L1356): The old filter unconditionally hid zero-turn, zero-prompt, inactive sessions. The new filter keeps them visible for up to 1 hour. This means sessions discovered from the filesystem (e.g. automated code review sessions) that were created within the last hour will now appear in the session list, where previously they were always hidden. This is likely intentional but is a behavioral change — the PR title mentions 'missing new sessions', so this is presumably the fix for that. Worth a note in the PR description.
- 🔵 **missing_tests** (L689): The timestamp-preservation path (passing existing updatedAt on resume) has no direct test at the chat.ts integration level. The underlying EventStore.upsertSession respecting an explicit updatedAt is well-tested in protocol/__tests__/event-store.test.ts, so the risk is low, but a test in session-cache.test.ts or a chat-level test verifying that resuming a session doesn't change its updatedAt would confirm the end-to-end behavior. `[fixable]`
- 🔵 **style** (L1356): The constant `ONE_HOUR` is defined inline inside the filter callback and re-created on every call to getSessionsCached (and for every session in the list). Consider hoisting it to module scope or into constants.ts alongside other server-wide constants for reuse and discoverability. `[fixable]`

dimakis:
## Centaur Review

### `server/__tests__/session-cache.test.ts`
_Solid fix for timestamp preservation and grace period filtering. Main concern is the existing test at line 79 now relies on NaN arithmetic to pass — it should explicitly set `createdAt` to an old timestamp to properly exercise the new code path._

- 🟡 **bugs** (L79): Existing test 'hides sessions with numTurns=0...' no longer matches the implementation's logic path. After this PR, the filter accesses `m.createdAt`, but the mock at line 79 doesn't include `createdAt`. The test passes accidentally because `Date.now() - undefined` yields `NaN`, and `NaN < ZERO_TURN_GRACE_MS` is `false`. This is fragile — if createdAt handling changes (e.g., a default fallback), the test could silently break. Add `createdAt: 1000` (or a similarly old timestamp) to the `sess-unused` mock to explicitly test the 'past grace period' path. `[fixable]`
- 🟡 **missing_tests** (L392): The 'resume timestamp preservation' tests re-implement the upsert logic from `chat.ts` inline rather than calling the actual function (`_startChatInner` or a testable wrapper). They verify that the mock pattern works, not that the real code does the right thing. If the implementation in `chat.ts` drifts from this pattern (e.g., someone changes the spread order or condition), these tests will still pass while the real behavior is broken. Consider testing through the actual exported function if possible, or at minimum add a comment acknowledging this is a pattern test.


### `server/chat.ts`
_Solid fix for timestamp preservation and grace period filtering. Main concern is the existing test at line 79 now relies on NaN arithmetic to pass — it should explicitly set `createdAt` to an old timestamp to properly exercise the new code path._

- 🔵 **unsafe_assumptions** (L1357): `Date.now()` is called on every filter iteration inside `getSessionsCached()`. For large session lists, this creates slightly different timestamps across items. Not a practical bug, but extracting `const now = Date.now()` before the `.filter()` call would be more precise and consistent — and avoids repeated syscalls. `[fixable]`
- 🔵 **style** (L38): The constant `ZERO_TURN_GRACE_MS` is well-named but lives at module top-level in `chat.ts`, which is already a large file. Consider colocating it with `SESSION_PAGE_SIZE` in `constants.ts` since it's also a session-list behavioral constant. `[fixable]`

dimakis (commented):
## Codex Review

LGTM — no issues found.

PR #327, opened 2026-05-16
fix(watch): trust Tailscale certs + shared Keychain

dimakis:
## Summary
- Add Embed Watch Content build phase to properly bundle MitzoWatch.app into iOS app
- Implement lazy apiClient/wsClient creation in WatchRelay with fallback token mechanism
- Switch from MagicDNS (.tail) to Tailscale IP (100.91.50.57) for URLSession hostname resolution
- Add slimMessages() to handle WCSession 64KB payload limit (strips tool blocks, truncates)
- Add retry logic for session loading (once after 3s delay on failure)
- Add error display in SessionListView when relay connection fails
- Add token notification chain so relay can bootstrap clients when watch makes first request
- Trust Tailscale IPs (100.*) in TailscaleTrust delegate
- Enable shared Keychain access group (com.mitzo.app) across iOS and watchOS

## Context
The watch can't connect directly to Tailscale VPN addresses due to NECP policy blocks. This PR implements a WatchConnectivity relay where the iPhone acts as a bridge - the watch sends requests via WCSession messages, the iPhone relay forwards them to the backend over Tailscale, and returns responses back to the watch.

The relay handles lazy client creation, token fallback (watch passes its token in relay messages), message slimming for WCSession limits, and proper error handling with retries.

## Test plan
- [ ] Build and run watchOS app on physical device
- [ ] Verify session list loads via relay (check retry on first failure)
- [ ] Send message from watch and verify relay to backend works
- [ ] Verify error display appears on connection failure
- [ ] Check logs for relay message forwarding and size warnings


dimakis:
## Centaur Review

### `frontend/ios/MitzoShared/Sources/MitzoShared/Auth/AuthManager.swift`
_Sound approach — switches all Mitzo networking to a Tailscale-trusting URLSession and wires shared Keychain for watch. Main concerns: hardcoded Team ID hurts portability, and YapperClient was missed in the URLSession migration._

- 🟡 **unsafe_assumptions** (L17): Team ID `Y4QGXHYSY3` is hardcoded. The entitlements files use the build-time variable `$(AppIdentifierPrefix)` which adapts automatically, but the Swift code will break for anyone with a different Apple Developer account. For an open-source project that aims for portability, consider reading the prefix from a build-injected Info.plist key or a shared config. `[fixable]`


### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/YapperClient.swift`
_Sound approach — switches all Mitzo networking to a Tailscale-trusting URLSession and wires shared Keychain for watch. Main concerns: hardcoded Team ID hurts portability, and YapperClient was missed in the URLSession migration._

- 🟡 **bugs** (L30): YapperClient still uses `URLSession.shared` (health check, line 30) and a bare `URLSession(configuration: .default)` (WebSocket, line 60). If Yapper is reachable over Tailscale, these calls will fail with a TLS error since they don't trust the Tailscale cert. The same pattern fixed in MitzoAPIClient, MitzoWSClient, and AuthManager was not applied here. `[fixable]`


### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/TailscaleTrust.swift`
_Sound approach — switches all Mitzo networking to a Tailscale-trusting URLSession and wires shared Keychain for watch. Main concerns: hardcoded Team ID hurts portability, and YapperClient was missed in the URLSession migration._

- 🔵 **unsafe_assumptions** (L22): `.tail` is accepted as a trusted suffix but Capacitor's `allowNavigation` only lists `*.ts.net`. These are different layers (URLSession vs WKWebView), so they don't conflict functionally, but the mismatch is easy to forget. A comment noting why `.tail` is needed (MagicDNS short names) would help future readers. `[fixable]`
- 🔵 **unsafe_assumptions** (L23): The delegate accepts the server trust credential without calling `SecTrustEvaluateWithError` — any certificate (self-signed, expired, etc.) from a matching host is trusted. This is intentional for Tailscale's own PKI, but it means a compromised DNS entry for a `.ts.net` domain would be trusted silently. Acceptable tradeoff for a Tailscale-only deployment, but worth a one-line comment stating the deliberate trust model.
- 🔵 **missing_tests**: No tests for TailscaleTrustDelegate's host-matching logic. The suffix checks (`.ts.net`, `.tail`, `localhost`) are simple but a unit test would catch regressions if the allowed-host list changes. Existing tests in MitzoSharedTests only cover message encoding. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

dimakis:
Closing — stale session branch with 51 accumulated commits. Unique watch/plist changes will be cherry-picked onto a clean branch if needed.

PR #328, opened 2026-05-16
fix(ios): forward APNs device token to Capacitor

dimakis:
## Summary
- AppDelegate.swift was missing `didRegisterForRemoteNotificationsWithDeviceToken` and `didFailToRegisterForRemoteNotificationsWithError`
- iOS delivered the APNs token but it was never forwarded to Capacitor via NotificationCenter, so no device token ever reached the server
- Adds the two required methods per Capacitor PushNotifications plugin docs

## Test plan
- [ ] Build in Xcode, run on physical iOS device
- [ ] Check that notification permission prompt appears (or is already granted)
- [ ] Verify `~/redhat/mgmt/.mitzo/device-tokens.json` is created with a token
- [ ] Background the app, send a message — confirm native push notification arrives

### DIFF
--- a/frontend/ios/App/App/App.entitlements
+++ b/frontend/ios/App/App/App.entitlements
@@ -4,5 +4,9 @@
 <dict>
 	<key>aps-environment</key>
 	<string>development</string>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.mitzo.app</string>
+	</array>
 </dict>
 </plist>
--- a/frontend/ios/App/App/AppDelegate.swift
+++ b/frontend/ios/App/App/AppDelegate.swift
@@ -46,4 +46,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return ApplicationDelegateProxy.shared.application(application, continue: userActivity, restorationHandler: restorationHandler)
     }
 
+    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
+        NotificationCenter.default.post(name: .capacitorDidRegisterForRemoteNotifications, object: deviceToken)
+    }
+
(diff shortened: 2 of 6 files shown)


dimakis:
## Centaur Review

Found **3** issue(s).


### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/YapperClient.swift`
_Clean, well-scoped fix: APNs token forwarding follows standard Capacitor guidance, and the shared TailscaleTrustDelegate correctly centralizes certificate trust across all networking clients. One pre-existing inconsistency in YapperClient and a minor doc-comment mismatch._

- 🔵 **unsafe_assumptions** (L30): Pre-existing: YapperClient still uses URLSession.shared for the health check (line 30) and creates a plain URLSession for WebSocket (line 60-61). If Yapper is served over Tailscale HTTPS, both will fail with certificate errors — the same issue this PR fixes for the other clients. Consider using tailscaleURLSession here too. `[fixable]`


### `frontend/ios/MitzoShared/Sources/MitzoShared/Networking/TailscaleTrust.swift`
_Clean, well-scoped fix: APNs token forwarding follows standard Capacitor guidance, and the shared TailscaleTrustDelegate correctly centralizes certificate trust across all networking clients. One pre-existing inconsistency in YapperClient and a minor doc-comment mismatch._

- 🔵 **style** (L8): Doc comment says 'matching Capacitor's allowNavigation config' but the trust list (.ts.net, localhost, .tail) is broader than allowNavigation which only lists *.ts.net. The comment is misleading — this intentionally trusts a superset (adding .tail for watchOS MagicDNS and localhost for dev). `[fixable]`
- 🔵 **missing_tests**: No unit tests for TailscaleTrustDelegate host-matching logic. The suffix checks (.ts.net, .tail, localhost) are simple but untested — a typo or future edit could silently break trust or over-trust. Testing URLSession delegates is hard, but the host-matching logic could be extracted into a pure function and tested. `[fixable]`

PR #331, opened 2026-05-16
feat(infra): persist Jaeger traces with badger storage

dimakis:
## Summary

- Switches Jaeger from in-memory (ephemeral, ~37h retention) to badger persistent storage
- 14-day TTL via `BADGER_SPAN_STORE_TTL=336h`
- Data persists at `.jaeger-data/` (gitignored)
- Enables the Trace Dreamer agent (dimakis/mgmt#116) to extract operational insights from historical traces

## Telemetry fixes

- **Incomplete sessions**: `session.total_tokens`, `session.num_turns`, and `session.id` now set as fallback span attributes when the SDK `result` event never fires (abort, timeout, connection loss)
- **Cost fallback**: `session.cost_usd` set from `cumulativeCostUsd` for multi-query sessions that had at least one `result` event before abort

## Test plan

- [x] Jaeger persists traces across container restarts
- [x] 50/50 query-loop tests pass
- [x] TypeScript type check passes
- [x] Prettier formatting check passes

Refs: #333

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,13 @@ services:
       - '16686:16686' # Jaeger UI
     environment:
       COLLECTOR_OTLP_ENABLED: 'true'
+      SPAN_STORAGE_TYPE: badger
+      BADGER_EPHEMERAL: 'false'
+      BADGER_DIRECTORY_VALUE: /badger/data
+      BADGER_DIRECTORY_KEY: /badger/key
+      BADGER_SPAN_STORE_TTL: 336h  # 14 days retention
+    volumes:
+      - ./.jaeger-data:/badger
 
   loki:
     image: grafana/loki:3.4.2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       BADGER_EPHEMERAL: 'false'
       BADGER_DIRECTORY_VALUE: /badger/data
       BADGER_DIRECTORY_KEY: /badger/key
-      BADGER_SPAN_STORE_TTL: 336h  # 14 days retention
+      BADGER_SPAN_STORE_TTL: 336h # 14 days retention
     volumes:
       - ./.jaeger-data:/badger
 
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -1368,6 +1368,18 @@ async function _runQueryLoopInner(
         currentTurnSpan = null;
       }
 
+      // Ensure span attributes are set even if the SDK result event never arrived.
+      // The result handler (msg.type === 'result') sets these, but sessions that
+      // abort, timeout, or lose connection skip that path. Use live counters as fallback.
+      if (!doneSent) {
+        span.setAttribute('session.total_tokens', liveSessionTokens);
+        span.setAttribute('session.num_turns', turnIndex);
+      }
+      // Always set session.id if we resolved it (idempotent with line 472)
+      if (resolvedSessionId) {
+        span.setAttribute('session.id', resolvedSessionId);
+      }
+
       span.setStatus({ code: SpanStatusCode.OK });
       log.info('query loop ended', { clientId, doneSent });
     }
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -1380,7 +1380,7 @@ async function _runQueryLoopInner(
           span.setAttribute('session.cost_usd', finalSession.cumulativeCostUsd);
         }
       }
-      // Always set session.id if we resolved it (idempotent with line 472)
+      // Always set session.id if we resolved it (idempotent with first-event handler)
       if (resolvedSessionId) {
         span.setAttribute('session.id', resolvedSessionId);
       }
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -1374,6 +1374,11 @@ async function _runQueryLoopInner(
       if (!doneSent) {
         span.setAttribute('session.total_tokens', liveSessionTokens);
         span.setAttribute('session.num_turns', turnIndex);
+        // Cost is only available from the SDK result event. For multi-query sessions
+        // (compaction), earlier result events may have set cumulativeCostUsd before abort.
+        if (finalSession && finalSession.cumulativeCostUsd > 0) {
+          span.setAttribute('session.cost_usd', finalSession.cumulativeCostUsd);
+        }
       }
       // Always set session.id if we resolved it (idempotent with line 472)
       if (resolvedSessionId) {
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -1323,6 +1323,10 @@ async function _runQueryLoopInner(
       }
     } finally {
       clearTimeout(firstEventTimer);
+      // NOTE: finalSession is read before registry.remove() below. After remove(),
+      // the object reference remains valid (Map.delete doesn't mutate the value).
+      // cumulativeCostUsd is read from finalSession after remove — safe due to
+      // reference semantics, but keep the ordering if refactoring.
       const finalSession = registry.get(clientId);
       if (finalSession) {
         finalSession.currentSnapshot = null;
@@ -1385,8 +1389,8 @@ async function _runQueryLoopInner(
         span.setAttribute('session.id', resolvedSessionId);
       }
 
-      span.setStatus({ code: SpanStatusCode.OK });
-      log.info('query loop ended', { clientId, doneSent });
+      span.setStatus({ code: caughtError ? SpanStatusCode.ERROR : SpanStatusCode.OK });
+      log.info('query loop ended', { clientId, doneSent, caughtError });
     }
   } finally {
     span.end();
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ frontend/ios/local.xcconfig
 mcp-server/build/
 certs/
 logs/
+.jaeger-data/
 .DS_Store
 .claude/worktrees/
 .cursor/worktrees/


docker-compose.yml
```diff
@@ -13,6 +13,13 @@ services:
       - '16686:16686' # Jaeger UI
     environment:
       COLLECTOR_OTLP_ENABLED: 'true'
+      SPAN_STORAGE_TYPE: badger
+      BADGER_EPHEMERAL: 'false'
+      BADGER_DIRECTORY_VALUE: /badger/data
+      BADGER_DIRECTORY_KEY: /badger/key
+      BADGER_SPAN_STORE_TTL: 336h  # 14 days retention
+    volumes:
+      - ./.jaeger-data:/badger
```
dimakis, line 22: 🔵 **unsafe_assumptions**: The bind mount `./.jaeger-data:/badger` will be created as a root-owned directory by Docker if it doesn't exist. The Jaeger container process may fail to write if there's a UID mismatch. Consider documenting `mkdir -p .jaeger-data` or adding an init container / healthcheck, though in practice jaegertracing/jaeger images typically run as root so this is low risk.

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `docker-compose.yml`
_Clean infrastructure change — adds Badger persistent storage to Jaeger with a 14-day TTL and properly gitignores the data directory. No bugs or regressions._

- 🔵 **unsafe_assumptions** (L22): The bind mount `./.jaeger-data:/badger` will be created as a root-owned directory by Docker if it doesn't exist. The Jaeger container process may fail to write if there's a UID mismatch. Consider documenting `mkdir -p .jaeger-data` or adding an init container / healthcheck, though in practice jaegertracing/jaeger images typically run as root so this is low risk.

server/query-loop.ts
```diff
@@ -1367,6 +1367,18 @@ async function _runQueryLoopInner(
...
+      if (!doneSent) {
+        span.setAttribute('session.total_tokens', liveSessionTokens);
+        span.setAttribute('session.num_turns', turnIndex);
+      }
+      // Always set session.id if we resolved it (idempotent with line 472)
+      if (resolvedSessionId) {
+        span.setAttribute('session.id', resolvedSessionId);
+      }
+
       span.setStatus({ code: SpanStatusCode.OK });
```
dimakis: 🟡 **bugs**: When `caughtError` is true (the catch block was entered), `span.setStatus` is still set to `SpanStatusCode.OK` on line 1382. The span status should reflect errors for aborted/failed sessions. However, this is pre-existing behavior not introduced by this PR — the new code only adds attributes above this line. No regression from the diff itself. `[fixable]`

server/query-loop.ts
```diff
@@ -1367,6 +1367,18 @@ async function _runQueryLoopInner(
         currentTurnSpan = null;
       }
 
+      // Ensure span attributes are set even if the SDK result event never arrived.
+      // The result handler (msg.type === 'result') sets these, but sessions that
+      // abort, timeout, or lose connection skip that path. Use live counters as fallback.
+      if (!doneSent) {
```
dimakis, line 1378: 🔵 **missing_tests**: The fallback span attribute logic (setting `session.total_tokens` and `session.num_turns` when `!doneSent`) has no test coverage. The existing `query-loop.test.ts` doesn't assert on span attributes at all. Given that this is observability-only code with no functional impact, this is low priority, but a test verifying the fallback fires on abort/timeout would prevent silent regressions.

server/query-loop.ts
```diff
@@ -1367,6 +1367,18 @@ async function _runQueryLoopInner(
...
+        span.setAttribute('session.total_tokens', liveSessionTokens);
+        span.setAttribute('session.num_turns', turnIndex);
+      }
+      // Always set session.id if we resolved it (idempotent with line 472)
```
dimakis: 🔵 **style**: Comment says "idempotent with line 472" but line numbers drift as the file changes. Consider referencing the code location by description (e.g., "idempotent with the first-event handler") instead of a line number. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Clean infrastructure change — badger persistence is correctly configured, gitignore is updated, and the span attribute fallback for incomplete sessions is logically sound with no regressions._

- 🟡 **bugs** (L1382): When `caughtError` is true (the catch block was entered), `span.setStatus` is still set to `SpanStatusCode.OK` on line 1382. The span status should reflect errors for aborted/failed sessions. However, this is pre-existing behavior not introduced by this PR — the new code only adds attributes above this line. No regression from the diff itself. `[fixable]`
- 🔵 **missing_tests** (L1373): The fallback span attribute logic (setting `session.total_tokens` and `session.num_turns` when `!doneSent`) has no test coverage. The existing `query-loop.test.ts` doesn't assert on span attributes at all. Given that this is observability-only code with no functional impact, this is low priority, but a test verifying the fallback fires on abort/timeout would prevent silent regressions.
- 🔵 **style** (L1377): Comment says "idempotent with line 472" but line numbers drift as the file changes. Consider referencing the code location by description (e.g., "idempotent with the first-event handler") instead of a line number. `[fixable]`

server/query-loop.ts
```diff
@@ -1367,6 +1367,23 @@ async function _runQueryLoopInner(
...
+        if (finalSession && finalSession.cumulativeCostUsd > 0) {
+          span.setAttribute('session.cost_usd', finalSession.cumulativeCostUsd);
+        }
+      }
+      // Always set session.id if we resolved it (idempotent with first-event handler)
+      if (resolvedSessionId) {
+        span.setAttribute('session.id', resolvedSessionId);
+      }
+
       span.setStatus({ code: SpanStatusCode.OK });
```
dimakis: 🟡 **bugs**: Pre-existing bug amplified by this PR: `span.setStatus({ code: SpanStatusCode.OK })` at line 1387 runs unconditionally in the `finally` block, overwriting the `SpanStatusCode.ERROR` set in the `catch` block (line 1306). The new fallback attributes (tokens, turns, cost) will thus always appear on an OK span, even for errored/aborted sessions. The subagent cleanup at line 1351 already uses `caughtError` to pick the right status — this line should do the same: `span.setStatus({ code: caughtError ? SpanStatusCode.ERROR : SpanStatusCode.OK })`. `[fixable]`

server/query-loop.ts
```diff
@@ -1367,6 +1367,23 @@ async function _runQueryLoopInner(
         currentTurnSpan = null;
       }
 
+      // Ensure span attributes are set even if the SDK result event never arrived.
+      // The result handler (msg.type === 'result') sets these, but sessions that
+      // abort, timeout, or lose connection skip that path. Use live counters as fallback.
+      if (!doneSent) {
```
dimakis, line 1378: 🟡 **missing_tests**: No tests cover the new fallback span-attribute path (`!doneSent` branch in the finally block). A test that simulates an aborted session (no `result` event emitted) should verify that `session.total_tokens`, `session.num_turns`, and `session.cost_usd` are set on the span from live counters.

server/query-loop.ts
```diff
@@ -1367,6 +1367,23 @@ async function _runQueryLoopInner(
...
+        span.setAttribute('session.total_tokens', liveSessionTokens);
+        span.setAttribute('session.num_turns', turnIndex);
+        // Cost is only available from the SDK result event. For multi-query sessions
+        // (compaction), earlier result events may have set cumulativeCostUsd before abort.
+        if (finalSession && finalSession.cumulativeCostUsd > 0) {
```
dimakis, line 1383: 🔵 **unsafe_assumptions**: `finalSession` is retrieved via `registry.get(clientId)` at line 1325, but `registry.remove(clientId)` is called at line 1344 — before `finalSession.cumulativeCostUsd` is accessed at line 1378. If `registry.remove` invalidates or mutates the returned object, this would read stale/undefined data. Currently safe because `registry.get` returns the object reference and `remove` only deletes the map entry, but this ordering is fragile. Consider reading `cumulativeCostUsd` before `registry.remove`, or add a comment noting the dependency on reference semantics. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_The tracing fallback logic is sound in intent, but an existing bug at line 1387 unconditionally overwrites the span status to OK — negating error reporting for aborted sessions and making the new fallback attributes misleading. That should be fixed in this PR since the new code deepens the dependency on correct span status._

- 🟡 **bugs** (L1387): Pre-existing bug amplified by this PR: `span.setStatus({ code: SpanStatusCode.OK })` at line 1387 runs unconditionally in the `finally` block, overwriting the `SpanStatusCode.ERROR` set in the `catch` block (line 1306). The new fallback attributes (tokens, turns, cost) will thus always appear on an OK span, even for errored/aborted sessions. The subagent cleanup at line 1351 already uses `caughtError` to pick the right status — this line should do the same: `span.setStatus({ code: caughtError ? SpanStatusCode.ERROR : SpanStatusCode.OK })`. `[fixable]`
- 🟡 **missing_tests** (L1373): No tests cover the new fallback span-attribute path (`!doneSent` branch in the finally block). A test that simulates an aborted session (no `result` event emitted) should verify that `session.total_tokens`, `session.num_turns`, and `session.cost_usd` are set on the span from live counters.
- 🔵 **unsafe_assumptions** (L1378): `finalSession` is retrieved via `registry.get(clientId)` at line 1325, but `registry.remove(clientId)` is called at line 1344 — before `finalSession.cumulativeCostUsd` is accessed at line 1378. If `registry.remove` invalidates or mutates the returned object, this would read stale/undefined data. Currently safe because `registry.get` returns the object reference and `remove` only deletes the map entry, but this ordering is fragile. Consider reading `cumulativeCostUsd` before `registry.remove`, or add a comment noting the dependency on reference semantics. `[fixable]`


### `docker-compose.yml`
_The tracing fallback logic is sound in intent, but an existing bug at line 1387 unconditionally overwrites the span status to OK — negating error reporting for aborted sessions and making the new fallback attributes misleading. That should be fixed in this PR since the new code deepens the dependency on correct span status._

- 🔵 **style**: The Jaeger container uses a bind-mount (`./.jaeger-data:/badger`) instead of a named Docker volume. This is fine for a local dev stack, but bind mounts can have permission issues on Linux (the container process may not own the host directory). A named volume would be more portable. Minor concern since this is explicitly a local dev tool. `[fixable]`

PR #335, opened 2026-05-16
feat(ui): fix What's Next to prioritize sessions over Telos

dimakis:
## Summary

- **Sessions awaiting reply** (agent spoke last, user hasn't responded) now appear as Tier 1 items above Telos
- **Uncommitted worktree work** surfaces as Tier 1 attention items with cached git status checks
- **Sub-priority sorting** within tiers ensures session items (0-2) always sort above Telos items (3)
- Previously all 14 manual Telos items (urgency=1.0, starred) filled all 5 slots, making What's Next useless

## Changes

| Layer | File | What |
|-------|------|------|
| Protocol | `types.ts` | Add `awaitingReply`, `uncommittedWork`, `idleMinutes` to `SessionActivity`; `lastSpeaker`/`lastSpeakerAt` to `SessionMeta` |
| Protocol | `event-store.ts` | `last_speaker` column + migration, `updateLastSpeaker()`, `getAttentionSessions()` |
| Server | `session-overview.ts` | Wire eventStore, persistent attention sessions, uncommitted work cache, awaitingReply derivation |
| Server | `chat.ts` | Mark `last_speaker = 'user'` on user messages |
| Server | `index.ts` | Mark `last_speaker = 'assistant'` on turn end, wire eventStore into emitter |
| Server | `worktree.ts` | Export `hasUncommittedWork` |
| Frontend | `useAttentionFeed.ts` | Sub-priority sorting, handle `awaitingReply` + `uncommittedWork` session states |

## Test plan

- [x] 27 session-overview tests pass (7 new for awaitingReply, uncommittedWork, persistent sessions)
- [x] 38 event-store tests pass (no regressions)
- [x] 1184/1187 server tests pass (3 failures are pre-existing on main)
- [ ] Deploy to dev, verify home screen shows sessions above Telos
- [ ] Verify: when no sessions need attention, Telos items still fill all slots

### DIFF
--- a/frontend/src/hooks/useAttentionFeed.ts
+++ b/frontend/src/hooks/useAttentionFeed.ts
@@ -60,7 +60,7 @@ function telosToAttention(items: TodoItem[]): AttentionItem[] {
         tier: 1,
         subPriority: SUB_TELOS,
         title: item.summary,
-        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} \u00B7 ${item.profile}`,
+        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} · ${item.profile}`,
         accentColor: item.urgency >= 0.8 ? COLOR_RED : COLOR_AMBER,
         icon: '\u2605', // ★
         navigateTo: `/todos/${item.id}`,
@@ -75,7 +75,7 @@ function telosToAttention(items: TodoItem[]): AttentionItem[] {
         tier: 2,
         subPriority: SUB_TELOS,
         title: item.summary,
-        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} \u00B7 ${item.profile}`,
+        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} · ${item.profile}`,
         accentColor: item.starred ? COLOR_AMBER : COLOR_PURPLE,
         icon: item.starred ? '\u2605' : '\u25CF', // ★ or ●
         navigateTo: `/todos/${item.id}`,
@@ -191,7 +191,7 @@ function sessionsToAttention(activities: SessionActivity[]): AttentionItem[] {
         id: `session-${a.sessionId}`,
         source: 'session',
         tier: 2,
-        subPriority: SUB_AWAITING_REPLY,
+        subPriority: SUB_TELOS,
         title: a.title,
         meta: 'done',
         accentColor: COLOR_GREEN,
--- a/frontend/src/hooks/useAttentionFeed.ts
+++ b/frontend/src/hooks/useAttentionFeed.ts
@@ -15,6 +15,8 @@ export interface AttentionItem {
   id: string;
   source: AttentionSource;
   tier: AttentionTier;
+  /** Within same tier, lower = higher priority. Sessions sort above Telos. */
+  subPriority: number;
   title: string;
   meta: string;
   accentColor: string;
@@ -32,6 +34,17 @@ const COLOR_RED = '#ff6d6d';
 const COLOR_PURPLE = '#b48cff';
 const COLOR_GREEN = '#4ade80';
 
+// ─── Sub-priority constants ────────────────────────────────────────────────
+
+/** Sessions awaiting user reply — highest priority within any tier */
+const SUB_AWAITING_REPLY = 0;
+/** Sessions/ATB waiting for input (permission, review, blocked) */
+const SUB_WAITING_INPUT = 1;
+/** Sessions with uncommitted worktree work */
+const SUB_UNCOMMITTED = 2;
+/** Telos items — fill remaining slots */
+const SUB_TELOS = 3;
+
 // ─── Derive attention items from Telos ─────────────────────────────────────
 
 function telosToAttention(items: TodoItem[]): AttentionItem[] {
@@ -45,8 +58,9 @@ function telosToAttention(items: TodoItem[]): AttentionItem[] {
         id: `telos-${item.id}`,
         source: 'telos',
         tier: 1,
+        subPriority: SUB_TELOS,
         title: item.summary,
-        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} · ${item.profile}`,
+        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} \u00B7 ${item.profile}`,
         accentColor: item.urgency >= 0.8 ? COLOR_RED : COLOR_AMBER,
         icon: '\u2605', // ★
         navigateTo: `/todos/${item.id}`,
@@ -59,8 +73,9 @@ function telosToAttention(items: TodoItem[]): AttentionItem[] {
         id: `telos-${item.id}`,
         source: 'telos',
         tier: 2,
+        subPriority: SUB_TELOS,
         title: item.summary,
-        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} · ${item.profile}`,
+        meta: `${item.ageDays === 0 ? 'new' : `${item.ageDays}d`} \u00B7 ${item.profile}`,
         accentColor: item.starred ? COLOR_AMBER : COLOR_PURPLE,
         icon: item.starred ? '\u2605' : '\u25CF', // ★ or ●
         navigateTo: `/todos/${item.id}`,
@@ -92,6 +107,7 @@ function atbToAttention(tasks: Task[]): AttentionItem[] {
       id: `atb-${t.id}`,
       source: 'atb' as AttentionSource,
       tier: 1 as AttentionTier,
+      subPriority: SUB_WAITING_INPUT,
       title: t.title,
       meta:
         t.status === 'pending_review'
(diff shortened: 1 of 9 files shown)


dimakis (commented):
## Centaur Review

### `frontend/src/hooks/useAttentionFeed.ts`
_Solid feature with good test coverage for the new session-overview logic. Main concerns: synchronous `execFileSync` blocking the event loop during broadcasts, git error sentinels being misinterpreted as uncommitted work in the attention feed, and missing EventStore unit tests for the new SQL methods._

- 🟡 **bugs** (L194): Tier 2 done sessions use `subPriority: SUB_AWAITING_REPLY` (value 0), which is semantically incorrect — these are plain done sessions, not awaiting-reply. While this doesn't cause a functional bug (tier separation prevents cross-tier ordering issues), it will sort done sessions above any future tier-2 items with higher subPriority values. Consider introducing a `SUB_DONE = 0` or reusing `SUB_TELOS` to place done sessions below telos items within tier 2. `[fixable]`
- 🔵 **regressions** (L62): The `·` separator in telos meta strings was changed from a literal middle-dot character to `\u00B7`. This is functionally identical but diverges from the rest of the codebase's convention. If the original literal was intentional for readability, this change adds inconsistency. `[fixable]`


_Solid feature with good test coverage for the new session-overview logic. Main concerns: synchronous `execFileSync` blocking the event loop during broadcasts, git error sentinels being misinterpreted as uncommitted work in the attention feed, and missing EventStore unit tests for the new SQL methods._

- 🟡 **unsafe_assumptions** (L301): `hasUncommittedWork()` uses `execFileSync` with a 30-second timeout, blocking the Node.js event loop during `compute()`. When the 5-minute cache expires, all sessions with a `cwd` are checked synchronously in sequence. With N sessions pointing to different cwds, this can block the event loop for up to N × 30 seconds in the worst case. Consider using an async approach (`execFile` with promises) or running the checks outside the broadcast path. `[fixable]`
- 🟡 **bugs** (L301): `hasUncommittedWork()` returns a non-null sentinel string (`'[git status failed: ...]'`) when git fails (e.g., worktree deleted, disk issue). `checkUncommittedCached` treats any non-null result as dirty (`!== null`), so git errors are cached as 'uncommitted work' for 5 minutes. This is the intended behavior for worktree cleanup (conservative), but in the attention feed context it causes a phantom 'uncommitted work' card. Consider distinguishing error sentinels from genuine dirty output. `[fixable]`
- 🔵 **style** (L304): The `catch` block in `checkUncommittedCached` is unreachable: `hasUncommittedWork()` catches all exceptions internally and returns a sentinel string. The try/catch here silently swallows a code path that can never execute. `[fixable]`


_Solid feature with good test coverage for the new session-overview logic. Main concerns: synchronous `execFileSync` blocking the event loop during broadcasts, git error sentinels being misinterpreted as uncommitted work in the attention feed, and missing EventStore unit tests for the new SQL methods._

- 🟡 **missing_tests**: The new `updateLastSpeaker()` and `getAttentionSessions()` methods on `EventStore` have no unit tests. The `migrateAttentionTracking()` migration is also untested. The session-overview tests mock the EventStore, so these SQL statements are never exercised against a real database. Key cases to test: speaker toggling, the `LIMIT 10` boundary, filtering by `is_active`/`is_hidden`/`last_speaker`, and `rowToSession` mapping of the new fields. `[fixable]`


_Solid feature with good test coverage for the new session-overview logic. Main concerns: synchronous `execFileSync` blocking the event loop during broadcasts, git error sentinels being misinterpreted as uncommitted work in the attention feed, and missing EventStore unit tests for the new SQL methods._

- 🔵 **missing_tests**: No tests verify the cache TTL behavior of `checkUncommittedCached`. A test should confirm that a second call within 5 minutes returns the cached value without calling `hasUncommittedWork()` again, and that a call after TTL expiration re-checks. `[fixable]`

frontend/src/hooks/useAttentionFeed.ts
```diff
@@ -114,32 +130,86 @@ function atbToAttention(tasks: Task[]): AttentionItem[] {
...
+      });
+      continue;
+    }
+    // Tier 2: done sessions (recently finished)
+    if (a.state === 'done') {
+      items.push({
+        id: `session-${a.sessionId}`,
+        source: 'session',
+        tier: 2,
+        subPriority: SUB_AWAITING_REPLY,
```
dimakis: 🟡 **bugs**: Tier 2 done sessions use `subPriority: SUB_AWAITING_REPLY` (value 0), which is semantically incorrect — these are plain done sessions, not awaiting-reply. While this doesn't cause a functional bug (tier separation prevents cross-tier ordering issues), it will sort done sessions above any future tier-2 items with higher subPriority values. Consider introducing a `SUB_DONE = 0` or reusing `SUB_TELOS` to place done sessions below telos items within tier 2. `[fixable]`

frontend/src/hooks/useAttentionFeed.ts
```diff
@@ -45,8 +58,9 @@ function telosToAttention(items: TodoItem[]): AttentionItem[] {
         id: `telos-${item.id}`,
         source: 'telos',
         tier: 1,
+        subPriority: SUB_TELOS,
         title: item.summary,
```
dimakis, line 62: 🔵 **regressions**: The `·` separator in telos meta strings was changed from a literal middle-dot character to `\u00B7`. This is functionally identical but diverges from the rest of the codebase's convention. If the original literal was intentional for readability, this change adds inconsistency. `[fixable]`

server/session-overview.ts
```diff
@@ -221,9 +255,57 @@ export class SessionOverviewEmitter {
...
+  /**
+   * Check uncommitted work with a TTL cache to avoid expensive git calls.
+   */
+  private checkUncommittedCached(cwd: string, now: number): boolean {
+    const cached = this.uncommittedCache.get(cwd);
+    if (cached && now - cached.checkedAt < UNCOMMITTED_CACHE_TTL_MS) {
+      return cached.dirty;
+    }
+    try {
+      const dirty = hasUncommittedWork(cwd) !== null;
```
dimakis: 🟡 **unsafe_assumptions**: `hasUncommittedWork()` uses `execFileSync` with a 30-second timeout, blocking the Node.js event loop during `compute()`. When the 5-minute cache expires, all sessions with a `cwd` are checked synchronously in sequence. With N sessions pointing to different cwds, this can block the event loop for up to N × 30 seconds in the worst case. Consider using an async approach (`execFile` with promises) or running the checks outside the broadcast path. `[fixable]`

dimakis: 🟡 **bugs**: `hasUncommittedWork()` returns a non-null sentinel string (`'[git status failed: ...]'`) when git fails (e.g., worktree deleted, disk issue). `checkUncommittedCached` treats any non-null result as dirty (`!== null`), so git errors are cached as 'uncommitted work' for 5 minutes. This is the intended behavior for worktree cleanup (conservative), but in the attention feed context it causes a phantom 'uncommitted work' card. Consider distinguishing error sentinels from genuine dirty output. `[fixable]`

server/session-overview.ts
```diff
@@ -221,9 +255,57 @@ export class SessionOverviewEmitter {
...
+      this.uncommittedCache.set(cwd, { dirty, checkedAt: now });
+      return dirty;
+    } catch {
```
dimakis: 🔵 **style**: The `catch` block in `checkUncommittedCached` is unreachable: `hasUncommittedWork()` catches all exceptions internally and returns a sentinel string. The try/catch here silently swallows a code path that can never execute. `[fixable]`

dimakis:
## Centaur Review

### `server/chat.ts`
_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🟡 **bugs** (L852): `updateLastSpeaker(options.resume, 'user')` is called for resumed sessions and `sendToChat`/`interruptChat`, but the initial user prompt in a brand-new session (non-resume path through `_startChatInner`) never calls `updateLastSpeaker('user')`. This means after a new session starts and the agent responds, `lastSpeaker` goes from `null` directly to `'assistant'` on `turn_end`, making the session appear as "awaiting reply" even though it's the first turn and the user's initial prompt was the last real user action. The first `turn_end` fires before the user has a chance to send a follow-up, so every new session will briefly show as awaiting reply in the attention feed. `[fixable]`


_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🟡 **unsafe_assumptions** (L303): `hasUncommittedWork()` calls `execFileSync('git', ...)` with a 30-second timeout (`WORKTREE_GIT_TIMEOUT_MS`). This is invoked synchronously inside `compute()`, which runs on `getSnapshot()` (SSE hydration) and `broadcast()` (coalesced timer callback). With multiple active sessions, each having a cwd, the event loop can be blocked for up to 30s × N sessions on a cold cache. The 5-minute TTL cache helps on repeat calls, but the first compute after server restart or cache expiry can stall all SSE clients. Consider making this async or running the git checks in a worker/background interval rather than inline with compute. `[fixable]`


### `frontend/src/hooks/__tests__/useAttentionFeed.test.ts`
_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🔵 **missing_tests**: The existing `useAttentionFeed` tests don't cover any of the new session-derived attention items (awaitingReply, uncommittedWork, waiting) or the sub-priority sort order. All 7 test cases are telos-only. Adding at least one test that mixes session and telos items and verifies sub-priority ordering (sessions sort above telos within the same tier) would catch regressions in the core behavioral change this PR introduces. `[fixable]`


### `server/worktree.ts`
_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🔵 **style** (L316): Duplicate JSDoc comment block — the short one-liner at line 316–318 is immediately followed by the full doc block at lines 319–323. The short one should be removed. `[fixable]`


### `frontend/src/hooks/useAttentionFeed.ts`
_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🔵 **bugs** (L194): Done sessions that have `awaitingReply=true` are captured by the first branch (tier 1, SUB_AWAITING_REPLY) and `continue`, so they never reach the tier 2 'done' branch — which is correct. But done sessions that fell through to tier 2 are assigned `subPriority: SUB_TELOS` (value 3), making them sort identically to telos items. A distinct sub-priority constant (e.g., `SUB_DONE_SESSION = 2`) would let completed sessions sort above telos within tier 2 if that's the intended behavior. `[fixable]`


_Solid feature addition with good test coverage on the backend. Main concern: missing `updateLastSpeaker('user')` for new (non-resume) sessions causes every fresh session to appear as 'awaiting reply' after the first assistant turn. The synchronous `execFileSync` in the compute path is a latency risk worth addressing._

- 🔵 **missing_tests**: `updateLastSpeaker` silently succeeds (0 rows affected) when called with a non-existent session ID. A test documenting this behavior (no-op or no error) would prevent future contributors from assuming it throws. `[fixable]`

dimakis (commented):
## Codex Review

### `server/index.ts`
_PR mostly achieves the prioritization goal, but attention tracking can misclassify sessions and surface false awaiting-reply items._

- 🟡 **bugs** (L256): `lastSpeaker` is updated to `assistant` on every `turn_end`, even when the turn produced no assistant reply (e.g. user interrupts/cancels before output, or a turn ends with only internal/tool activity). That can incorrectly mark sessions as `awaitingReply` and surface them in What's Next despite the assistant never having spoken. The speaker update should happen when an assistant message is actually emitted/persisted, not unconditionally at turn end. `[fixable]`


_PR mostly achieves the prioritization goal, but attention tracking can misclassify sessions and surface false awaiting-reply items._

- 🟡 **regressions** (L250): `awaitingReply` is derived solely from `lastSpeaker === 'assistant' && !session.hasSnapshot`, so live sessions in `waiting` state (permission/review/blocked) can also be marked as awaiting reply if the assistant spoke last. In `useAttentionFeed`, awaiting-reply items sort ahead of waiting-input items, which reverses the intended priority for sessions that still require explicit user action. Exclude `state === 'waiting'` from awaiting-reply derivation. `[fixable]`

server/index.ts
```diff
@@ -255,6 +256,11 @@ setSessionChangeCallback((clientId, event) => {
     overviewEmitter.forget(clientId);
```
dimakis, line 256: 🟡 **bugs**: `lastSpeaker` is updated to `assistant` on every `turn_end`, even when the turn produced no assistant reply (e.g. user interrupts/cancels before output, or a turn ends with only internal/tool activity). That can incorrectly mark sessions as `awaitingReply` and surface them in What's Next despite the assistant never having spoken. The speaker update should happen when an assistant message is actually emitted/persisted, not unconditionally at turn end. `[fixable]`

server/session-overview.ts
```diff
@@ -210,6 +243,15 @@ export class SessionOverviewEmitter {
     // Derive repo from cwd
     const repo = session.cwd ? extractRepoName(session.cwd) : undefined;
 
+    // Check uncommitted work for live sessions (from background cache)
+    const uncommittedWork = session.cwd ? this.checkUncommittedCached(session.cwd) : false;
+
+    // Check if awaiting user reply (assistant spoke last, not streaming)
+    const meta = this.deps.eventStore.getSession(sessionId);
```
dimakis, line 251: 🟡 **regressions**: `awaitingReply` is derived solely from `lastSpeaker === 'assistant' && !session.hasSnapshot`, so live sessions in `waiting` state (permission/review/blocked) can also be marked as awaiting reply if the assistant spoke last. In `useAttentionFeed`, awaiting-reply items sort ahead of waiting-input items, which reverses the intended priority for sessions that still require explicit user action. Exclude `state === 'waiting'` from awaiting-reply derivation. `[fixable]`

server/index.ts
dimakis, line 256: 🟡 **bugs**: `turn_end` now unconditionally marks the assistant as the last speaker. This also fires after turns where the assistant produced no user-visible reply (e.g. interrupted/cancelled/error/tool-only/no-op turns), which can incorrectly surface sessions as `awaitingReply` and persist them via `getAttentionSessions()`. The last-speaker update should happen only when an assistant message was actually emitted/completed, not on every turn boundary. `[fixable]`

server/session-overview.ts
```diff
@@ -210,6 +243,17 @@ export class SessionOverviewEmitter {
     // Derive repo from cwd
```
dimakis, line 243: 🟡 **regressions**: `awaitingReply` is derived solely from `lastSpeaker === 'assistant'`, `!hasSnapshot`, and `primaryState !== 'waiting'`. That means actively running sessions that are not currently streaming (`state: 'running'`) can now be flagged as awaiting reply if the assistant spoke last between tool calls/turn phases, causing the UI to prioritize in-progress sessions as if they were waiting on the user. This should likely be limited to idle/done sessions, matching the frontend prioritization logic. `[fixable]`

dimakis, line 243: 🔵 **missing_tests**: There is good coverage for `awaitingReply` on idle/waiting/streaming sessions, but no test that a non-streaming `running` session does NOT get `awaitingReply`. Given the new persistence/sorting behavior, that edge case should be locked down. `[fixable]`

dimakis (commented):
## Codex Review

### `server/index.ts`
_PR is close, but last-speaker/awaiting-reply tracking appears too eager and can incorrectly prioritize sessions that are not actually waiting on the user._

- 🟡 **bugs** (L256): `turn_end` now unconditionally marks the assistant as the last speaker. This also fires after turns where the assistant produced no user-visible reply (e.g. interrupted/cancelled/error/tool-only/no-op turns), which can incorrectly surface sessions as `awaitingReply` and persist them via `getAttentionSessions()`. The last-speaker update should happen only when an assistant message was actually emitted/completed, not on every turn boundary. `[fixable]`


_PR is close, but last-speaker/awaiting-reply tracking appears too eager and can incorrectly prioritize sessions that are not actually waiting on the user._

- 🟡 **regressions** (L243): `awaitingReply` is derived solely from `lastSpeaker === 'assistant'`, `!hasSnapshot`, and `primaryState !== 'waiting'`. That means actively running sessions that are not currently streaming (`state: 'running'`) can now be flagged as awaiting reply if the assistant spoke last between tool calls/turn phases, causing the UI to prioritize in-progress sessions as if they were waiting on the user. This should likely be limited to idle/done sessions, matching the frontend prioritization logic. `[fixable]`
- 🔵 **missing_tests** (L243): There is good coverage for `awaitingReply` on idle/waiting/streaming sessions, but no test that a non-streaming `running` session does NOT get `awaitingReply`. Given the new persistence/sorting behavior, that edge case should be locked down. `[fixable]`

PR #337, opened 2026-05-16
feat(workload): complete Telos → Task Board bridge

dimakis:
## Summary

- Add promote button to TodoDetailView — creates a root task from a Telos workload item and navigates to the task board
- Add "↗ Task Board" badge when an item is already promoted (clickable, navigates to linked task)
- Wire WebSocket sync for `workload_item_created`, `workload_item_updated`, `workload_batch_updated` events (new workload slice + protocol parser + store integration)
- Call `orchestrator.onTaskCompleted()` from both HTTP endpoints (POST complete + PATCH status=done) to eliminate ~10s completion delay

## Test plan

- [x] 7 frontend tests: promote button visibility, badge, navigation, API call, error display, loading state
- [x] 2 backend route tests: broadcast verification, goalId link verification
- [x] 4 integration tests: full lifecycle (ingest→promote→orchestrate→complete), failure path, single root goal, idempotency
- [ ] Manual: promote a Telos item, verify task board navigation, complete task, verify Telos item auto-completes

### DIFF
--- a/frontend/src/components/__tests__/TodoCard.test.tsx
+++ b/frontend/src/components/__tests__/TodoCard.test.tsx
@@ -25,6 +25,7 @@ const mockItem: TodoItem = {
       snippet: 'Some description',
     },
   ],
+  goalId: null,
   contextHints: {
     repos: ['dimakis/mitzo'],
     paths: [],
--- a/frontend/src/hooks/__tests__/useAttentionFeed.test.ts
+++ b/frontend/src/hooks/__tests__/useAttentionFeed.test.ts
@@ -44,6 +44,7 @@ function makeTodo(overrides: Partial<TodoItem> = {}): TodoItem {
     children: [],
     childCount: 0,
     completedChildCount: 0,
+    goalId: null,
     sources: [],
     contextHints: {
       repos: [],
--- a/frontend/src/lib/__tests__/todo-utils.test.ts
+++ b/frontend/src/lib/__tests__/todo-utils.test.ts
@@ -37,6 +37,7 @@ const fullItem: TodoItem = {
   children: [],
   childCount: 0,
   completedChildCount: 0,
+  goalId: null,
   sources: [
     {
       type: 'github',
--- a/frontend/src/pages/TodoDetailView.tsx
+++ b/frontend/src/pages/TodoDetailView.tsx
@@ -28,6 +28,8 @@ export function TodoDetailView() {
   const stateItem = (location.state as { item?: TodoItem } | null)?.item;
   const [fetchedItem, setFetchedItem] = useState<TodoItem | null>(null);
   const [fetchFailed, setFetchFailed] = useState(false);
+  const [promoting, setPromoting] = useState(false);
+  const [promoteError, setPromoteError] = useState<string | null>(null);
 
   useEffect(() => {
     if (stateItem || !id) return;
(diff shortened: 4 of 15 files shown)


dimakis:
## Centaur Review

_The `workload_item_created` store handler silently drops new items (uses replace-only `updateWorkloadItem` instead of `upsertWorkloadItem`), and the PATCH task endpoint only notifies the orchestrator on `done` but not `failed`, which can leave the loop stuck._

- 🔴 **bugs** (L770): `workload_item_created` uses `updateWorkloadItem()` which only replaces existing items via `.map()`. A genuinely new item (not already in the array) will be silently dropped. Should use `upsertWorkloadItem()` — but it's not even imported. The `workload_item_updated` case sharing the same handler masks this: updates work, but creations are lost. `[fixable]`
- 🟡 **missing_tests** (L770): No test covers the `workload_item_created` store handler path. A test would have caught the `updateWorkloadItem` vs `upsertWorkloadItem` bug — a new item arriving via WS should appear in `state.workload.items`. `[fixable]`


### `server/app.ts`
_The `workload_item_created` store handler silently drops new items (uses replace-only `updateWorkloadItem` instead of `upsertWorkloadItem`), and the PATCH task endpoint only notifies the orchestrator on `done` but not `failed`, which can leave the loop stuck._

- 🟡 **bugs** (L581): PATCH `/api/tasks/:id` only calls `orchestrator.onTaskCompleted()` when `status === 'done'`, but the orchestrator's `tick()` also handles `failed` as a terminal state (line 274 of task-orchestrator.ts). Manually failing a task via REST won't notify the orchestrator, leaving it stuck in `running` state. The guard should include `failed` (and arguably `skipped`). `[fixable]`
- 🔵 **missing_tests** (L636): The new `orchestrator.onTaskCompleted()` call in `/api/internal/task-tools/complete` is not tested through the HTTP layer. The integration test in `workload-task-bridge.test.ts` calls `orchestrator.onTaskCompleted()` directly, bypassing the route. A supertest-level test would verify the wiring. `[fixable]`


_The `workload_item_created` store handler silently drops new items (uses replace-only `updateWorkloadItem` instead of `upsertWorkloadItem`), and the PATCH task endpoint only notifies the orchestrator on `done` but not `failed`, which can leave the loop stuck._

- 🟡 **bugs** (L282): When the orchestrator calls `workloadStore.completeByGoal()` after goal completion, there is no WebSocket broadcast of the workload item update. The `onWorkloadBroadcast` callback is only wired in the HTTP route handlers (app.ts), not in the orchestrator. Clients won't see the workload item transition to `completed` until a full page refresh. `[fixable]`


_The `workload_item_created` store handler silently drops new items (uses replace-only `updateWorkloadItem` instead of `upsertWorkloadItem`), and the PATCH task endpoint only notifies the orchestrator on `done` but not `failed`, which can leave the loop stuck._

- 🔵 **unsafe_assumptions** (L468): `msg.item as WorkloadItem` and `msg.items as WorkloadItem[]` are unchecked type assertions on data arriving over the wire. The other protocol paths (tasks, tokens) similarly use casts, so this is consistent with the codebase — but it's a wider pattern worth noting for future hardening. `[fixable]`

dimakis:
## Centaur Review

_Critical bug: `workload_item_created` WS events silently drop new items because `updateWorkloadItem` (map-only) is used instead of `upsertWorkloadItem` (insert-or-update). Also missing double-promote guard on the server API._

- 🔴 **bugs** (L774): `workload_item_created` uses `updateWorkloadItem()` which only does a map/replace — if the item doesn't already exist in the array, it is silently dropped. Should use `upsertWorkloadItem()` (which is defined in the slice but never imported in store.ts). The `workload_item_updated` case can correctly use `updateWorkloadItem`, but `workload_item_created` must use `upsertWorkloadItem` to add new items. `[fixable]`


### `server/app.ts`
_Critical bug: `workload_item_created` WS events silently drop new items because `updateWorkloadItem` (map-only) is used instead of `upsertWorkloadItem` (insert-or-update). Also missing double-promote guard on the server API._

- 🟡 **bugs** (L1687): The `/promote` endpoint has no guard against double promotion. If the item already has a `goalId`, calling promote again creates a new orphaned task and silently overwrites the link. The frontend guards this with `!item.goalId`, but the API itself is unprotected — direct API calls or race conditions can create orphan tasks. Add `if (item.goalId) return res.status(409).json({ error: 'Already promoted' });` before task creation. `[fixable]`
- 🟡 **bugs** (L581): `orchestrator.onTaskCompleted` is called for any PATCH setting `status: 'done'`, even for tasks not part of an active orchestration loop. `onTaskCompleted` guards with `if (this.state !== 'running') return` so this is safe, but it's misleading — a status of `'failed'` or `'blocked'` set via PATCH won't trigger orchestrator tick, meaning external REST mutations for non-done statuses are silently ignored by the orchestrator.


_Critical bug: `workload_item_created` WS events silently drop new items because `updateWorkloadItem` (map-only) is used instead of `upsertWorkloadItem` (insert-or-update). Also missing double-promote guard on the server API._

- 🔵 **missing_tests** (L517): No test for double-promote (calling POST `/promote` on an item that already has a `goalId`). Whether the intended behavior is 409 rejection or silent overwrite, it should be documented by a test. `[fixable]`


_Critical bug: `workload_item_created` WS events silently drop new items because `updateWorkloadItem` (map-only) is used instead of `upsertWorkloadItem` (insert-or-update). Also missing double-promote guard on the server API._

- 🟡 **unsafe_assumptions** (L467): Workload message fields are cast with `as WorkloadItem` / `as WorkloadItem[]` / `as number` from the generic `msg` object without any runtime validation. If the server sends a malformed message, the client will silently store corrupted state. Other slices (tasks, tokens) also use casts, but workload items have more complex nested structures (`contextHints`, `sources[]`) where a shape mismatch is more likely to cause downstream errors. `[fixable]`


### `packages/client/src/slices/workload.ts`
_Critical bug: `workload_item_created` WS events silently drop new items because `updateWorkloadItem` (map-only) is used instead of `upsertWorkloadItem` (insert-or-update). Also missing double-promote guard on the server API._

- 🔵 **style** (L1): The comment `// Mirror of backend TodoItem type` is misleading — these are `WorkloadItem` types, not `TodoItem`. The workload and todo concepts appear to be related but distinct (different stores, different status enums). The comment should be removed or corrected. `[fixable]`

PR #338, opened 2026-05-17
feat: enriched iOS notifications with session context

dimakis:
## Summary

- **Session title in notification title** — "Mitzo: Debug Auth Flow" instead of "Mitzo: Agent replied", with "Mitzo" fallback when no title exists
- **Thread grouping** — notifications grouped by session via APNs `threadId`, so multiple sessions don't pile up in one stack
- **Action buttons** — Reply (text input, foreground), View (foreground), Later (background dismiss) registered as `SESSION_UPDATE` category
- **Better auto-rename** — Haiku prompt tuned from generic "generate a title" to intent-focused "what is the user trying to accomplish?"
- **All notification channels** — APNs, ntfy, and Pushover all receive session title

## Follow-up needed

- Server-side handler for Reply action (inject text into session)
- Server-side handler for Later action (snooze/re-queue)
- Frontend deep-link handling for View action (open correct session)
- Capacitor plugin to bridge notification action responses to the web layer

## Test plan

- [x] APNs tests pass (14/14) — new tests for session title, fallback, category constant
- [x] ntfy tests pass (6/6) — new tests for title with/without session title
- [x] pushover tests pass (5/5)
- [x] notification-helpers tests pass (6/6)
- [ ] Deploy and verify notification appearance on iOS device
- [ ] Verify thread grouping in notification center with multiple sessions
- [ ] Verify Reply/View/Later action buttons appear on long-press

### DIFF
--- a/frontend/src/lib/push.ts
+++ b/frontend/src/lib/push.ts
@@ -35,9 +35,36 @@ export async function initPushNotifications(): Promise<void> {
   });
 
   await PushNotifications.addListener('pushNotificationActionPerformed', (action) => {
-    const data = (action as { notification: { data: Record<string, string> } }).notification.data;
-    if (data?.sessionId) {
-      window.location.href = `/chat/${data.sessionId}`;
+    const typed = action as {
+      actionId: string;
+      inputValue?: string;
+      notification: { data: Record<string, string> };
+    };
+    const { actionId, inputValue } = typed;
+    const data = typed.notification.data;
+    const sessionId = data?.sessionId;
+
+    if (sessionId && actionId === 'REPLY_ACTION' && inputValue) {
+      // Send reply text to the session, then navigate
+      apiFetch('/api/push/notification-action', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ sessionId, actionId, userText: inputValue }),
+      }).finally(() => {
+        window.location.href = `/chat/${sessionId}`;
+      });
+    } else if (sessionId) {
+      // VIEW_ACTION, LATER_ACTION, or default tap — navigate to session
+      if (actionId && actionId !== 'VIEW_ACTION') {
+        apiFetch('/api/push/notification-action', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ sessionId, actionId }),
+        });
+      }
+      if (actionId !== 'LATER_ACTION') {
+        window.location.href = `/chat/${sessionId}`;
+      }
     }
   });
 
--- a/frontend/src/lib/push.ts
+++ b/frontend/src/lib/push.ts
@@ -1,7 +1,7 @@
 // Push notification integration for Capacitor iOS. No-op in browser.
 
 import { Capacitor } from '@capacitor/core';
-import { PushNotifications } from '@capacitor/push-notifications';
+import { PushNotifications, type ActionPerformed } from '@capacitor/push-notifications';
 import { apiFetch } from './api-fetch';
 
 let initialized = false;
@@ -34,14 +34,9 @@ export async function initPushNotifications(): Promise<void> {
     // Foreground — WS handles live updates, no action needed
   });
 
-  await PushNotifications.addListener('pushNotificationActionPerformed', (action) => {
-    const typed = action as {
-      actionId: string;
-      inputValue?: string;
-      notification: { data: Record<string, string> };
-    };
-    const { actionId, inputValue } = typed;
-    const data = typed.notification.data;
+  await PushNotifications.addListener('pushNotificationActionPerformed', (action: ActionPerformed) => {
+    const { actionId, inputValue } = action;
+    const data = action.notification.data as Record<string, string> | undefined;
     const sessionId = data?.sessionId;
 
     if (sessionId && actionId === 'REPLY_ACTION' && inputValue) {
--- a/frontend/src/lib/push.ts
+++ b/frontend/src/lib/push.ts
@@ -45,19 +45,21 @@ export async function initPushNotifications(): Promise<void> {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify({ sessionId, actionId, userText: inputValue }),
-      }).finally(() => {
+      }).then(() => {
+        window.location.href = `/chat/${sessionId}`;
+      }).catch(() => {
+        // Still navigate — user can resend from chat view
         window.location.href = `/chat/${sessionId}`;
       });
     } else if (sessionId) {
-      // VIEW_ACTION, LATER_ACTION, or default tap — navigate to session
-      if (actionId && actionId !== 'VIEW_ACTION') {
+      if (actionId === 'LATER_ACTION') {
         apiFetch('/api/push/notification-action', {
           method: 'POST',
           headers: { 'Content-Type': 'application/json' },
           body: JSON.stringify({ sessionId, actionId }),
         });
-      }
-      if (actionId !== 'LATER_ACTION') {
+      } else {
+        // VIEW_ACTION or default tap — just navigate
         window.location.href = `/chat/${sessionId}`;
       }
     }
--- a/frontend/src/lib/push.ts
+++ b/frontend/src/lib/push.ts
@@ -34,36 +34,41 @@ export async function initPushNotifications(): Promise<void> {
     // Foreground — WS handles live updates, no action needed
   });
 
-  await PushNotifications.addListener('pushNotificationActionPerformed', (action: ActionPerformed) => {
-    const { actionId, inputValue } = action;
-    const data = action.notification.data as Record<string, string> | undefined;
-    const sessionId = data?.sessionId;
+  await PushNotifications.addListener(
+    'pushNotificationActionPerformed',
+    (action: ActionPerformed) => {
+      const { actionId, inputValue } = action;
+      const data = action.notification.data as Record<string, string> | undefined;
+      const sessionId = data?.sessionId;
 
-    if (sessionId && actionId === 'REPLY_ACTION' && inputValue) {
-      // Send reply text to the session, then navigate
-      apiFetch('/api/push/notification-action', {
-        method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({ sessionId, actionId, userText: inputValue }),
-      }).then(() => {
-        window.location.href = `/chat/${sessionId}`;
-      }).catch(() => {
-        // Still navigate — user can resend from chat view
-        window.location.href = `/chat/${sessionId}`;
-      });
-    } else if (sessionId) {
-      if (actionId === 'LATER_ACTION') {
+      if (sessionId && actionId === 'REPLY_ACTION' && inputValue) {
+        // Send reply text to the session, then navigate
         apiFetch('/api/push/notification-action', {
           method: 'POST',
           headers: { 'Content-Type': 'application/json' },
-          body: JSON.stringify({ sessionId, actionId }),
-        });
-      } else {
-        // VIEW_ACTION or default tap — just navigate
-        window.location.href = `/chat/${sessionId}`;
+          body: JSON.stringify({ sessionId, actionId, userText: inputValue }),
+        })
+          .then(() => {
+            window.location.href = `/chat/${sessionId}`;
+          })
+          .catch(() => {
+            // Still navigate — user can resend from chat view
+            window.location.href = `/chat/${sessionId}`;
+          });
+      } else if (sessionId) {
+        if (actionId === 'LATER_ACTION') {
+          apiFetch('/api/push/notification-action', {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({ sessionId, actionId }),
+          });
+        } else {
+          // VIEW_ACTION or default tap — just navigate
+          window.location.href = `/chat/${sessionId}`;
+        }
       }
-    }
-  });
+    },
+  );
 
   await PushNotifications.register();
 }
--- a/server/__tests__/apns.test.ts
+++ b/server/__tests__/apns.test.ts
@@ -113,3 +113,37 @@ describe('sendPush', () => {
     await expect(sendPush('Test', 'Body')).resolves.toBeUndefined();
   });
 });
+
+describe('sendTurnCompleteNotification', () => {
+  it('uses session title in notification title when provided', async () => {
+    vi.resetModules();
+    delete process.env.APNS_KEY_PATH;
+    const { sendTurnCompleteNotification } = await import('../apns.js');
+    // Not configured, so sendPush is a no-op — we test the interface contract
+    await expect(
+      sendTurnCompleteNotification('sess-1', 'Some snippet', 'Jira Sprint Analysis'),
+    ).resolves.toBeUndefined();
+  });
+
+  it('falls back to "Mitzo" when no session title', async () => {
+    vi.resetModules();
+    delete process.env.APNS_KEY_PATH;
+    const { sendTurnCompleteNotification } = await import('../apns.js');
+    await expect(sendTurnCompleteNotification('sess-1', 'Snippet')).resolves.toBeUndefined();
+  });
+
+  it('accepts undefined snippet and sessionId', async () => {
+    vi.resetModules();
+    delete process.env.APNS_KEY_PATH;
+    const { sendTurnCompleteNotification } = await import('../apns.js');
+    await expect(sendTurnCompleteNotification()).resolves.toBeUndefined();
+  });
+});
+
+describe('APNS_CATEGORY', () => {
+  it('exports the notification category constant', async () => {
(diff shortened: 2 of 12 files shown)


packages/client/src/store.ts
```diff
@@ -757,6 +765,26 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       }
     }
 
+    if (result.workloadUpdate) {
+      switch (result.workloadUpdate.type) {
+        case 'workload_item_created':
+        case 'workload_item_updated': {
+          const item = result.workloadUpdate.item;
+          store.setState((s) => ({
+            workload: { ...s.workload, items: updateWorkloadItem(s.workload.items, item) },
```
dimakis, line 774: 🔴 **bugs**: `workload_item_created` uses `updateWorkloadItem()` which only maps over existing items — new items are silently dropped. Should use `upsertWorkloadItem()` (already defined in the workload slice but not imported). The `batchUpdateWorkloadItems` path correctly uses upsert internally, making this inconsistency easy to miss. `[fixable]`

server/app.ts
```diff
@@ -578,6 +579,9 @@ app.patch('/api/tasks/:id', (req, res) => {
     res.status(404).json({ error: 'Task not found' });
     return;
   }
+  if (body.data.status === 'done' && orchestrator) {
+    orchestrator.onTaskCompleted(req.params.id);
```
dimakis, line 638: 🟡 **bugs**: PATCH `/api/tasks/:id` calls `orchestrator.onTaskCompleted()` without first calling `taskStore.cascadeStatus()`. The internal endpoint (POST `/api/internal/task-tools/complete`) gets cascade via `handleTaskComplete()`. Without cascade, parent task statuses in nested trees won't propagate, so `deriveParentStatus(goalId)` in the orchestrator's `tick()` may compute stale results and fail to detect goal completion. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -35,9 +35,36 @@ export async function initPushNotifications(): Promise<void> {
...
+      apiFetch('/api/push/notification-action', {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({ sessionId, actionId, userText: inputValue }),
+      }).finally(() => {
+        window.location.href = `/chat/${sessionId}`;
+      });
+    } else if (sessionId) {
+      // VIEW_ACTION, LATER_ACTION, or default tap — navigate to session
+      if (actionId && actionId !== 'VIEW_ACTION') {
```
dimakis: 🟡 **bugs**: Default notification tap (Capacitor actionId = `'tap'`) is truthy and not `'VIEW_ACTION'`, so it falls through to the `apiFetch` call at line 59 with actionId `'tap'`. The server returns 400 `Unknown actionId: tap`. The request is fire-and-forget so no user-visible failure, but it produces noise in server logs. Guard with an explicit allowlist: `if (actionId === 'LATER_ACTION')`. `[fixable]`

server/__tests__/apns.test.ts
```diff
@@ -113,3 +113,37 @@ describe('sendPush', () => {
     await expect(sendPush('Test', 'Body')).resolves.toBeUndefined();
   });
 });
+
+describe('sendTurnCompleteNotification', () => {
```
dimakis, line 117: 🔵 **missing_tests**: The `sendTurnCompleteNotification` tests only verify the function resolves when APNS is not configured (no-op path). They don't verify the title is set to `Mitzo: ${sessionTitle}` or that `threadId`/`category` are passed through. The ntfy tests (notify.test.ts) do check headers, so the APNS tests are weaker by comparison. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -457,6 +464,22 @@ export function parseServerMessage(
       };
       break;
 
+    case 'workload_item_created':
```
dimakis, line 467: 🔵 **unsafe_assumptions**: `msg.item as WorkloadItem` is an unsafe cast. Since `ServerMessage` is `{ type: string; [key: string]: unknown }`, there's no compile-time guarantee that `msg.item` exists or has the right shape. A malformed message would produce `undefined` cast to `WorkloadItem`, causing downstream crashes. Consider a runtime check or adding the workload message types to the `WsMsg` discriminated union. `[fixable]`

dimakis (commented):
## Centaur Review

_Critical bug: new workload items silently dropped in store due to `updateWorkloadItem` vs `upsertWorkloadItem` mismatch; missing `cascadeStatus` in PATCH handler can break orchestrator goal detection for nested task trees._

- 🔴 **bugs** (L774): `workload_item_created` uses `updateWorkloadItem()` which only maps over existing items — new items are silently dropped. Should use `upsertWorkloadItem()` (already defined in the workload slice but not imported). The `batchUpdateWorkloadItems` path correctly uses upsert internally, making this inconsistency easy to miss. `[fixable]`


### `server/app.ts`
_Critical bug: new workload items silently dropped in store due to `updateWorkloadItem` vs `upsertWorkloadItem` mismatch; missing `cascadeStatus` in PATCH handler can break orchestrator goal detection for nested task trees._

- 🟡 **bugs** (L583): PATCH `/api/tasks/:id` calls `orchestrator.onTaskCompleted()` without first calling `taskStore.cascadeStatus()`. The internal endpoint (POST `/api/internal/task-tools/complete`) gets cascade via `handleTaskComplete()`. Without cascade, parent task statuses in nested trees won't propagate, so `deriveParentStatus(goalId)` in the orchestrator's `tick()` may compute stale results and fail to detect goal completion. `[fixable]`


### `frontend/src/lib/push.ts`
_Critical bug: new workload items silently dropped in store due to `updateWorkloadItem` vs `upsertWorkloadItem` mismatch; missing `cascadeStatus` in PATCH handler can break orchestrator goal detection for nested task trees._

- 🟡 **bugs** (L58): Default notification tap (Capacitor actionId = `'tap'`) is truthy and not `'VIEW_ACTION'`, so it falls through to the `apiFetch` call at line 59 with actionId `'tap'`. The server returns 400 `Unknown actionId: tap`. The request is fire-and-forget so no user-visible failure, but it produces noise in server logs. Guard with an explicit allowlist: `if (actionId === 'LATER_ACTION')`. `[fixable]`


### `server/__tests__/apns.test.ts`
_Critical bug: new workload items silently dropped in store due to `updateWorkloadItem` vs `upsertWorkloadItem` mismatch; missing `cascadeStatus` in PATCH handler can break orchestrator goal detection for nested task trees._

- 🔵 **missing_tests** (L117): The `sendTurnCompleteNotification` tests only verify the function resolves when APNS is not configured (no-op path). They don't verify the title is set to `Mitzo: ${sessionTitle}` or that `threadId`/`category` are passed through. The ntfy tests (notify.test.ts) do check headers, so the APNS tests are weaker by comparison. `[fixable]`


_Critical bug: new workload items silently dropped in store due to `updateWorkloadItem` vs `upsertWorkloadItem` mismatch; missing `cascadeStatus` in PATCH handler can break orchestrator goal detection for nested task trees._

- 🔵 **unsafe_assumptions** (L467): `msg.item as WorkloadItem` is an unsafe cast. Since `ServerMessage` is `{ type: string; [key: string]: unknown }`, there's no compile-time guarantee that `msg.item` exists or has the right shape. A malformed message would produce `undefined` cast to `WorkloadItem`, causing downstream crashes. Consider a runtime check or adding the workload message types to the `WsMsg` discriminated union. `[fixable]`

packages/client/src/store.ts
dimakis, line 774: 🔴 **bugs**: `workload_item_created` uses `updateWorkloadItem` which only does `.map()` — it replaces existing items but never adds new ones. New workload items received via WS will be silently dropped. Should use `upsertWorkloadItem` (already exported from the slice but not imported here). `[fixable]`

server/app.ts
```diff
@@ -578,6 +579,9 @@ app.patch('/api/tasks/:id', (req, res) => {
     res.status(404).json({ error: 'Task not found' });
     return;
   }
+  if (body.data.status === 'done' && orchestrator) {
```
dimakis, line 637: 🟡 **bugs**: PATCH `/api/tasks/:id` calls `orchestrator.onTaskCompleted()` without first calling `taskStore.cascadeStatus()`. The internal `/api/internal/task-tools/complete` endpoint calls `handleTaskComplete` which does cascade. Missing cascade means parent task statuses won't be persisted when a child is marked done via PATCH, leading to stale parent status in the DB and UI until something else triggers a re-derive. `[fixable]`

server/query-loop.ts
```diff
@@ -627,9 +627,10 @@ async function _runQueryLoopInner(
           if (!registry.isAttached(clientId)) {
             const snippet = extractSnippet(snapshotBlocks, NOTIFY_SNIPPET_MAX_CHARS);
             const sid = (msg.session_id as string) || currentSession.sessionId;
-            ntfyTurnComplete(sid, snippet).catch(() => {});
-            pushoverTurnComplete(sid, snippet).catch(() => {});
-            apnsTurnComplete(sid, snippet).catch(() => {});
+            const sessionTitle = store?.getSession(sid)?.summary ?? undefined;
```
dimakis: 🟡 **unsafe_assumptions**: `store?.getSession(sid)?.summary` assumes `summary` is populated. Sessions only get a summary after auto-rename fires (every N prompts). For the first few turns, `summary` will be `null`/`undefined`, falling back to the generic 'Mitzo' title. Not a bug per se, but on first turn completion the notification title carries no session context — exactly when the user most needs to know which session is asking.

server/__tests__/apns.test.ts
```diff
@@ -113,3 +113,37 @@ describe('sendPush', () => {
     await expect(sendPush('Test', 'Body')).resolves.toBeUndefined();
   });
 });
+
```
dimakis, line 116: 🔵 **missing_tests**: The APNS `sendTurnCompleteNotification` tests only verify the function doesn't throw — they never check that `threadId` or `category` are actually passed to the push notification. The tests run with APNS unconfigured so `sendPush` is a no-op, making them interface-contract-only. Consider adding a test with a mocked provider to verify `notification.threadId` and `notification.category` are set. `[fixable]`

frontend/src/lib/push.ts
```diff
@@ -34,10 +34,32 @@ export async function initPushNotifications(): Promise<void> {
...
-    if (data?.sessionId) {
-      window.location.href = `/chat/${data.sessionId}`;
+  await PushNotifications.addListener('pushNotificationActionPerformed', (action: ActionPerformed) => {
+    const { actionId, inputValue } = action;
+    const data = action.notification.data as Record<string, string> | undefined;
+    const sessionId = data?.sessionId;
+
+    if (sessionId && actionId === 'REPLY_ACTION' && inputValue) {
+      // Send reply text to the session, then navigate
+      apiFetch('/api/push/notification-action', {
```
dimakis: 🔵 **unsafe_assumptions**: The REPLY_ACTION handler navigates via `window.location.href` inside `.finally()` — this fires even if the `apiFetch` fails (network error, 500, etc.). The reply text is lost and the user lands on the chat page with no indication the reply wasn't delivered. Consider showing an error or retrying instead of always navigating. `[fixable]`

packages/client/src/protocol-parser.ts
dimakis, line 467: 🔵 **style**: The three workload message cases use `as WorkloadItem` / `as WorkloadItem[]` / `as number` casts on untyped `msg` fields. The rest of the parser uses Zod schemas or typed message interfaces. Consider adding these message types to the WS schema for consistent type safety. `[fixable]`

dimakis (commented):
## Centaur Review

_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🔴 **bugs** (L774): `workload_item_created` uses `updateWorkloadItem` which only does `.map()` — it replaces existing items but never adds new ones. New workload items received via WS will be silently dropped. Should use `upsertWorkloadItem` (already exported from the slice but not imported here). `[fixable]`


### `server/app.ts`
_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🟡 **bugs** (L582): PATCH `/api/tasks/:id` calls `orchestrator.onTaskCompleted()` without first calling `taskStore.cascadeStatus()`. The internal `/api/internal/task-tools/complete` endpoint calls `handleTaskComplete` which does cascade. Missing cascade means parent task statuses won't be persisted when a child is marked done via PATCH, leading to stale parent status in the DB and UI until something else triggers a re-derive. `[fixable]`


### `server/query-loop.ts`
_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🟡 **unsafe_assumptions** (L630): `store?.getSession(sid)?.summary` assumes `summary` is populated. Sessions only get a summary after auto-rename fires (every N prompts). For the first few turns, `summary` will be `null`/`undefined`, falling back to the generic 'Mitzo' title. Not a bug per se, but on first turn completion the notification title carries no session context — exactly when the user most needs to know which session is asking.


### `server/__tests__/apns.test.ts`
_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🔵 **missing_tests** (L116): The APNS `sendTurnCompleteNotification` tests only verify the function doesn't throw — they never check that `threadId` or `category` are actually passed to the push notification. The tests run with APNS unconfigured so `sendPush` is a no-op, making them interface-contract-only. Consider adding a test with a mocked provider to verify `notification.threadId` and `notification.category` are set. `[fixable]`


### `frontend/src/lib/push.ts`
_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🔵 **unsafe_assumptions** (L44): The REPLY_ACTION handler navigates via `window.location.href` inside `.finally()` — this fires even if the `apiFetch` fails (network error, 500, etc.). The reply text is lost and the user lands on the chat page with no indication the reply wasn't delivered. Consider showing an error or retrying instead of always navigating. `[fixable]`


_The workload_item_created handler silently drops new items (uses map-only update instead of upsert) — this is the most impactful bug. The missing cascadeStatus in the PATCH task endpoint is also a real issue. Notification enrichment and iOS action handlers look solid otherwise._

- 🔵 **style** (L467): The three workload message cases use `as WorkloadItem` / `as WorkloadItem[]` / `as number` casts on untyped `msg` fields. The rest of the parser uses Zod schemas or typed message interfaces. Consider adding these message types to the WS schema for consistent type safety. `[fixable]`

dimakis (commented):
## Codex Review

_PR is close, but the new workload WS handling drops created items on the client due to using update-only logic._

- 🟡 **bugs** (L771): `workload_item_created` is handled with `updateWorkloadItem(...)`, which only replaces existing entries and never inserts new ones. Any newly created workload item delivered over WS will be dropped from client state unless it was already present. Use `upsertWorkloadItem(...)` for created events (and likely updated events too for resilience). `[fixable]`

packages/client/src/store.ts
```diff
@@ -757,6 +765,26 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       }
     }
 
+    if (result.workloadUpdate) {
+      switch (result.workloadUpdate.type) {
+        case 'workload_item_created':
+        case 'workload_item_updated': {
```
dimakis, line 771: 🟡 **bugs**: `workload_item_created` is handled with `updateWorkloadItem(...)`, which only replaces existing entries and never inserts new ones. Any newly created workload item delivered over WS will be dropped from client state unless it was already present. Use `upsertWorkloadItem(...)` for created events (and likely updated events too for resilience). `[fixable]`

dimakis, line 771: 🟡 **bugs**: `workload_item_created` is handled with `updateWorkloadItem(...)`, which only replaces existing entries and never inserts new ones. As a result, newly created workload items delivered over WS are silently dropped from client state unless a separate full refresh happens. Use `upsertWorkloadItem(...)` for created events (and likely updated events too for resilience). `[fixable]`

dimakis (commented):
## Codex Review

_PR is close, but the client currently drops newly created workload items from websocket updates due to using update-only logic._

- 🟡 **bugs** (L771): `workload_item_created` is handled with `updateWorkloadItem(...)`, which only replaces existing entries and never inserts new ones. As a result, newly created workload items delivered over WS are silently dropped from client state unless a separate full refresh happens. Use `upsertWorkloadItem(...)` for created events (and likely updated events too for resilience). `[fixable]`

PR #340, opened 2026-05-18
docs(design): session state machine for lifecycle bug fix

dimakis:
## Summary

- Design doc for fixing the session reattach hanging bug via an explicit session lifecycle state machine
- Root cause: two sources of truth (SessionRegistry in-memory + EventStore SQLite `isActive` flag) diverge on reconnect, causing the stale check to kill running queries
- Solution: explicit 7-state lifecycle (CREATED → STARTING → ACTIVE → DETACHED/SUSPENDED → CLOSING → ENDED) with validated transitions
- 4-phase rollout: write state (Phase 1) → observe (Phase 2) → enforce routing (Phase 3) → validate transitions (Phase 4)

## Context

Traced the bug live: on iOS reattach after backgrounding, `ws-handler-v2.ts:388` sees `isActive: false` in EventStore while session exists in registry → removes session → query loop dies → 0-token `session_end` → user sees "thinking..." with no response.

## Test plan

- [ ] Review design doc for completeness and correctness
- [ ] Validate state machine transitions cover all lifecycle paths
- [ ] Confirm Phase 1 is safe (write-only, no behavior change)
- [ ] Check that crash recovery strategy is sound

### DIFF
--- a/docs/design/session-state-machine.md
+++ b/docs/design/session-state-machine.md
@@ -1,0 +1,996 @@
+# Session State Machine
+
+**Status:** Proposed  
+**Author:** Claude Sonnet 4.5 + Dimitri Saridakis  
+**Created:** 2026-05-18  
+**Related PRs:** TBD
+
+## Problem Statement
+
+### Symptom
+
+Users experience hanging sessions on mobile reattach. Pattern:
+
+1. User backgrounds app for >30 minutes
+2. Returns, sends message
+3. UI shows "thinking..." but no response
+4. Repeated messages eventually trigger a response
+5. Event log shows `session_end` with 0 tokens before recovery
+
+### Root Cause
+
+**Two sources of truth for session lifecycle with no explicit state machine:**
+
+**SessionRegistry** (in-memory):
+
+- `has(clientId)` → query loop is running
+- `!has(clientId)` → no query loop
+
+**EventStore** (SQLite):
+
+- `isActive: true` → query was running last time we checked
+- `isActive: false` → query ended last time we checked
+
+These diverge because:
+
+- `isActive` updated only on query **start** (first SDK event) and **end**
+- Detach/reattach/suspend only touch SessionRegistry
+- Rapid state changes create windows where in-memory and durable state disagree
+
+### The Bug
+
+`ws-handler-v2.ts:388-394`:
+
+```typescript
+const storeMeta = ctx.eventStore.getSession(sessionId);
+const staleInMemory = storeMeta && !storeMeta.isActive;
+
+if (staleInMemory) {
+  log.info('removing stale session from registry (send)');
+  ctx.sessionRegistry.remove(found.clientId); // ← kills query loop
+}
+```
+
+On reattach after long idle:
+
+1. Registry has session (from reconnect or lingering state)
+2. EventStore shows `isActive: false` (from previous session end)
+3. Code removes session from registry
+4. Query loop sees `!currentSession` on next iteration → breaks → emits 0-token `session_end`
+5. Fallthrough starts fresh query with `resume: sessionId` → works
+
+This is one symptom. The architectural issue causes:
+
+- Race conditions on rapid reconnect
+- Impossible to distinguish legitimate stale vs. just-started vs. detached
+- Crash recovery can't determine session state
+- No metrics on state transitions or dwell times
+- Debugging requires correlating logs across two systems
+
+## Solution: Explicit State Machine
+
+### States
+
+```
+CREATED    → session allocated in registry, no SDK call yet
+STARTING   → query() called, waiting for first SDK event
+ACTIVE     → SDK yielded events, query running, transport attached
+DETACHED   → transport disconnected, query still running
+SUSPENDED  → iOS backgrounded, buffering events
+CLOSING    → closeout phase (10min timer)
+ENDED      → query finished, cleaned up
+```
+
+### Transitions
+
+```
+CREATED → STARTING        startChat() calls query()
+STARTING → ACTIVE         first SDK event received
+STARTING → ENDED          query fails before first event (timeout/error)
+
+ACTIVE → DETACHED         transport.close() without abort
+ACTIVE → SUSPENDED        iOS background + suspend buffer active
+ACTIVE → CLOSING          closeout handler fires
+ACTIVE → ENDED            query completes normally
+
+DETACHED → ACTIVE         reattach()
+DETACHED → SUSPENDED      iOS background while detached
+DETACHED → CLOSING        detach TTL expires, closeout fires
+DETACHED → ENDED          detach TTL expires without closeout, or query aborts
+
+SUSPENDED → ACTIVE        resume from background, reattach
+SUSPENDED → DETACHED      resume but stay detached (shouldn't happen)
+SUSPENDED → ENDED         suspend timeout expires, query aborts
+
+CLOSING → ENDED           closeout completes or timeout expires
+
+ENDED → STARTING          explicit resume with startChat(resume: sessionId)
+```
+
+### Invalid Transitions (bugs if observed)
+
+- Any state → CREATED (sessions don't un-create)
+- ENDED → any except STARTING (can't revive without explicit resume)
+- CLOSING → any except ENDED (closeout is terminal)
+- SUSPENDED → CLOSING (suspend timeout goes to ENDED, not CLOSING)
+
+### Invariants
+
+1. **Registry ⇔ State consistency:**
+   - `registry.has(clientId)` iff state ∈ {CREATED, STARTING, ACTIVE, DETACHED, SUSPENDED, CLOSING}
+   - `!registry.has(clientId)` iff state = ENDED
+
+2. **Query running:**
+   - state ∈ {STARTING, ACTIVE, DETACHED, SUSPENDED, CLOSING} → query loop is running
+   - state ∈ {CREATED, ENDED} → no query loop
+
+3. **Transport attached:**
+   - state ∈ {ACTIVE, CLOSING} → `registry.isAttached(clientId) === true`
+   - state ∈ {DETACHED, SUSPENDED} → `registry.isAttached(clientId) === false`
+   - state ∈ {CREATED, STARTING, ENDED} → transport may be attached (STARTING) or not exist (ENDED)
+
+4. **Monotonic timestamp:**
+   - `last_state_change` increases on every transition
+
+## Implementation Plan
+
+### Phase 1: Add State Column (No Behavior Change)
+
+**Goal:** Write state everywhere, don't read it yet. Validate transitions in production.
+
+#### Schema Migration
+
+```sql
+-- server/migrations/003_session_state.sql
+ALTER TABLE sessions ADD COLUMN state TEXT DEFAULT 'ENDED';
+ALTER TABLE sessions ADD COLUMN last_state_change INTEGER;
+
+CREATE INDEX idx_sessions_state ON sessions(session_id, state);
+```
+
+#### EventStore API
+
+```typescript
+// server/event-store.ts
+
+export type SessionState =
+  | 'CREATED'
+  | 'STARTING'
+  | 'ACTIVE'
+  | 'DETACHED'
+  | 'SUSPENDED'
+  | 'CLOSING'
+  | 'ENDED';
+
+interface SessionStateTransition {
+  fromState: SessionState | null;
+  toState: SessionState;
+  timestamp: number;
+  clientId: string;
+  reason?: string;
+}
+
+class EventStore {
+  setState(
+    sessionId: string,
+    newState: SessionState,
+    opts?: { clientId?: string; reason?: string },
+  ): void {
+    const current = this.getSession(sessionId);
+    const fromState = current?.state ?? null;
+    const now = Date.now();
+
+    this.db.run(
+      `UPDATE sessions 
+       SET state = ?, last_state_change = ?
+       WHERE session_id = ?`,
+      [newState, now, sessionId],
+    );
+
+    // Emit state transition for tracing/metrics
+    this.appendStateTransition(sessionId, {
+      fromState,
+      toState: newState,
+      timestamp: now,
+      clientId: opts?.clientId ?? 'unknown',
+      reason: opts?.reason,
+    });
+
+    log.info('session state transition', {
+      sessionId,
+      fromState,
+      toState: newState,
+      clientId: opts?.clientId,
+      reason: opts?.reason,
+    });
+  }
+
+  getSessionState(sessionId: string): SessionState | null {
+    const row = this.db.get('SELECT state FROM sessions WHERE session_id = ?', [sessionId]);
+    return row?.state ?? null;
+  }
+}
+```
+
+#### Write State in All Lifecycle Operations
+
+**startChat (chat.ts):**
+
+```typescript
+export function startChat(...) {
+  registry.register(clientId, { ... });
+  if (sessionId) {
+    eventStore.setState(sessionId, 'CREATED', { clientId });
+  }
+
+  const q = query(...);
+  if (sessionId) {
+    eventStore.setState(sessionId, 'STARTING', { clientId });
+  }
+
+  runQueryLoop(q, clientId, ...);
+}
+```
+
+**First SDK event (query-loop.ts):**
+
+```typescript
+if (!firstEventReceived) {
+  firstEventReceived = true;
+  if (resolvedSessionId) {
+    store?.setState(resolvedSessionId, 'ACTIVE', {
+      clientId,
+      reason: 'first_sdk_event',
+    });
+  }
+}
+```
+
+**detach (session-registry.ts):**
+
+```typescript
+detach(clientId: string, store?: EventStore): void {
+  const session = this.sessions.get(clientId);
+  if (!session) return;
+
+  this.attached.delete(clientId);
+  if (session.sessionId && store) {
+    store.setState(session.sessionId, 'DETACHED', {
+      clientId,
+      reason: 'transport_close'
+    });
+  }
+  // ... start TTL timer
+}
+```
+
+**reattach (session-registry.ts):**
+
+```typescript
+reattach(clientId: string, transport: SessionTransport, store?: EventStore): boolean {
+  const session = this.sessions.get(clientId);
+  if (!session) return false;
+
+  session.transport = transport;
+  this.attached.add(clientId);
+
+  if (session.sessionId && store) {
+    store.setState(session.sessionId, 'ACTIVE', {
+      clientId,
+      reason: 'reattach'
+    });
+  }
+  // ... clear timers
+  return true;
+}
+```
+
+**suspend (session-registry.ts):**
+
+```typescript
+suspend(clientId: string, store?: EventStore): boolean {
+  const session = this.sessions.get(clientId);
+  if (!session) return false;
+
+  this.suspended.add(clientId);
+  this.attached.delete(clientId);
+
+  if (session.sessionId && store) {
+    store.setState(session.sessionId, 'SUSPENDED', {
+      clientId,
+      reason: 'ios_background'
+    });
+  }
+  // ... start suspend timer
+  return true;
+}
+```
+
+**closeout (chat.ts):**
+
+```typescript
+function closeoutSession(clientId: string): void {
+  const session = registry.get(clientId);
+  if (session?.sessionId) {
+    eventStore.setState(session.sessionId, 'CLOSING', {
+      clientId,
+      reason: 'detach_ttl_closeout',
+    });
+  }
+  // ... inject closeout prompt
+}
+```
+
+**query end (query-loop.ts finally block):**
+
+```typescript
+finally {
+  const sid = finalSession.sessionId ?? resolvedSessionId;
+  if (store && sid) {
+    store.setState(sid, 'ENDED', {
+      clientId,
+      reason: caughtError ? 'error' : 'completed'
+    });
+  }
+  registry.remove(clientId);
+}
+```
+
+#### Validation (No Enforcement Yet)
+
+Add telemetry to catch impossible transitions:
+
+```typescript
+// event-store.ts
+setState(sessionId: string, newState: SessionState, opts?: ...): void {
+  const current = this.getSession(sessionId);
+  const fromState = current?.state ?? null;
+
+  // Warn on invalid transitions but don't block
+  if (fromState && !isValidTransition(fromState, newState)) {
+    log.warn('invalid session state transition', {
+      sessionId,
+      fromState,
+      toState: newState,
+      clientId: opts?.clientId,
+      stack: new Error().stack,
+    });
+    // Record metric for monitoring
+    invalidTransitionCounter.inc({ from: fromState, to: newState });
+  }
+
+  // Always proceed with state change in Phase 1
+  this.db.run(...);
+}
+```
+
+#### Acceptance Criteria (Phase 1)
+
+- [ ] All lifecycle operations set state
+- [ ] State transitions appear in logs
+- [ ] No invalid transitions observed in production for 1 week
+- [ ] Existing behavior unchanged (stale checks still run)
+- [ ] Tests verify state is written on all paths
+
+### Phase 2: Observability (Read State, Don't Enforce)
+
+**Goal:** Surface state in debugging, detect mismatches, don't change routing logic yet.
+
+#### Add State to All Relevant Logs
+
+```typescript
+// ws-handler-v2.ts handleSendV2
+const found = ctx.sessionRegistry.findBySessionId(sessionId);
+const state = ctx.eventStore.getSessionState(sessionId);
+
+log.info('send routing decision', {
+  sessionId,
+  registryHas: !!found,
+  storeState: state,
+  isOwner,
+  isDetached,
+});
+```
+
+#### Detect State Mismatches
+
+```typescript
+// Add mismatch detector that runs on every send/interrupt
+function detectStateMismatch(
+  sessionId: string,
+  registry: SessionRegistry,
+  store: EventStore,
+): { mismatch: boolean; details?: string } {
+  const found = registry.findBySessionId(sessionId);
+  const state = store.getSessionState(sessionId);
+
+  const registryHas = !!found;
+  const shouldHave = state && state !== 'ENDED';
+
+  if (registryHas && !shouldHave) {
+    return {
+      mismatch: true,
+      details: `registry has session but state=${state}`,
+    };
+  }
+
+  if (!registryHas && shouldHave) {
+    return {
+      mismatch: true,
+      details: `registry missing session but state=${state}`,
+    };
+  }
+
+  if (found && state) {
+    const attached = registry.isAttached(found.clientId);
+    const shouldBeAttached = state === 'ACTIVE' || state === 'CLOSING';
+    const shouldBeDetached = state === 'DETACHED' || state === 'SUSPENDED';
+
+    if (attached && shouldBeDetached) {
+      return {
+        mismatch: true,
+        details: `attached but state=${state}`,
+      };
+    }
+    if (!attached && shouldBeAttached) {
+      return {
+        mismatch: true,
+        details: `detached but state=${state}`,
+      };
+    }
+  }
+
+  return { mismatch: false };
+}
+
+// ws-handler-v2.ts handleSendV2
+const mismatch = detectStateMismatch(sessionId, ctx.sessionRegistry, ctx.eventStore);
+if (mismatch.mismatch) {
+  log.error('session state mismatch detected', {
+    sessionId,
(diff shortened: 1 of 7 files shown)


server/session-overview.ts
```diff
@@ -299,6 +301,27 @@ export class SessionOverviewEmitter {
     };
```
dimakis, line 307: 🟡 **bugs**: Speaker cache is never updated for 'user' messages. `overviewEmitter.updateSpeaker(sid, 'user')` is not called from `sendToChat`, `interruptChat`, or `startChat(resume)` — only the EventStore is updated. Between a user sending a follow-up and the next `turn_end` event, the cached speaker remains 'assistant', causing `awaitingReply` to be incorrectly `true` in broadcasts during that window. The same gap exists for the new `store.updateLastSpeaker(resolvedSessionId, 'user')` call in query-loop.ts — it updates SQLite but not the emitter cache. `[fixable]`

server/session-overview.ts
```diff
@@ -248,11 +250,11 @@ export class SessionOverviewEmitter {
...
     // Exclude 'waiting' state — sessions needing permission/review should sort as waiting, not awaiting reply
-    const meta = this.deps.eventStore.getSession(sessionId);
+    const speaker = this.getCachedSpeaker(sessionId);
     const awaitingReply =
-      meta?.lastSpeaker === 'assistant' && !session.hasSnapshot && primaryState !== 'waiting';
-    const speakerAt = meta?.lastSpeakerAt ?? lastEventAt;
-    const idleMinutes = Math.round((now - speakerAt) / 60_000);
+      speaker.speaker === 'assistant' && !session.hasSnapshot && primaryState !== 'waiting';
+    const speakerAt = speaker.at ?? lastEventAt;
+    const idleMinutes = Math.max(0, Math.round((now - speakerAt) / 60_000));
```
dimakis, line 263: 🔵 **style**: The `Math.max(0, ...)` guard on `idleMinutes` is a good defensive fix, but it appears in two places (lines 257 and 284). If `speakerAt` or `lastEventAt` could be in the future (e.g., clock skew), this silently clamps to 0 rather than surfacing the issue. Fine as-is, but worth noting the implicit assumption.

server/__tests__/session-overview.test.ts
```diff
@@ -65,6 +65,12 @@ vi.mock('@mitzo/harness', async (importOriginal) => {
   };
 });
 
+// ─── Mock the worktree module ─────────────────────────────────────────────────
+
+vi.mock('../worktree.js', () => ({
```
dimakis: 🔵 **style**: Duplicate `vi.mock('../worktree.js')` — the PR adds a mock at line 70 (exporting sync `hasUncommittedWork`) while an existing mock at line 79 (exporting async `hasUncommittedWorkAsync`) already covers the module. Vitest hoists and deduplicates mocks, so only the last one takes effect. The new mock at line 70 is dead code. Either combine them into one mock or remove the new one. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -155,8 +155,7 @@ export class EventStore {
       ),
       getAttentionSessions: db.prepare(
         `SELECT * FROM sessions
```
dimakis, line 157: 🔵 **bugs**: Removing `is_active = 1` from the `getAttentionSessions` query is intentional (new test confirms it), but the query now returns ALL non-hidden sessions where assistant spoke last, including very old completed sessions. The `LIMIT 10` caps the count, but if many old sessions exist with `last_speaker = 'assistant'`, recent relevant sessions could be displaced by stale ones. Consider adding a time-bound filter (e.g., `AND last_speaker_at > unixepoch() - 86400*7`) or an `ORDER BY` that favors recent sessions (already present via `last_speaker_at DESC`, so this is minor). `[fixable]`

dimakis (commented):
## Centaur Review

_Solid caching optimization and query fix with good test coverage. Main concern: the speaker cache is updated for 'assistant' turns but not for 'user' messages, creating a brief window where `awaitingReply` is stale after a user sends a follow-up. The duplicate worktree mock is harmless but should be cleaned up._

- 🟡 **bugs** (L301): Speaker cache is never updated for 'user' messages. `overviewEmitter.updateSpeaker(sid, 'user')` is not called from `sendToChat`, `interruptChat`, or `startChat(resume)` — only the EventStore is updated. Between a user sending a follow-up and the next `turn_end` event, the cached speaker remains 'assistant', causing `awaitingReply` to be incorrectly `true` in broadcasts during that window. The same gap exists for the new `store.updateLastSpeaker(resolvedSessionId, 'user')` call in query-loop.ts — it updates SQLite but not the emitter cache. `[fixable]`
- 🔵 **style** (L257): The `Math.max(0, ...)` guard on `idleMinutes` is a good defensive fix, but it appears in two places (lines 257 and 284). If `speakerAt` or `lastEventAt` could be in the future (e.g., clock skew), this silently clamps to 0 rather than surfacing the issue. Fine as-is, but worth noting the implicit assumption.


_Solid caching optimization and query fix with good test coverage. Main concern: the speaker cache is updated for 'assistant' turns but not for 'user' messages, creating a brief window where `awaitingReply` is stale after a user sends a follow-up. The duplicate worktree mock is harmless but should be cleaned up._

- 🔵 **style** (L70): Duplicate `vi.mock('../worktree.js')` — the PR adds a mock at line 70 (exporting sync `hasUncommittedWork`) while an existing mock at line 79 (exporting async `hasUncommittedWorkAsync`) already covers the module. Vitest hoists and deduplicates mocks, so only the last one takes effect. The new mock at line 70 is dead code. Either combine them into one mock or remove the new one. `[fixable]`
- 🔵 **missing_tests**: No test verifies that `getCachedSpeaker` properly falls back when the EventStore returns a session with `null` lastSpeaker (the `meta?.lastSpeaker ?? null` path). The existing tests pre-populate the cache via `updateSpeaker()` or mock `getSession` with a speaker value, but don't cover the null-speaker lazy-load branch. `[fixable]`


_Solid caching optimization and query fix with good test coverage. Main concern: the speaker cache is updated for 'assistant' turns but not for 'user' messages, creating a brief window where `awaitingReply` is stale after a user sends a follow-up. The duplicate worktree mock is harmless but should be cleaned up._

- 🔵 **bugs** (L157): Removing `is_active = 1` from the `getAttentionSessions` query is intentional (new test confirms it), but the query now returns ALL non-hidden sessions where assistant spoke last, including very old completed sessions. The `LIMIT 10` caps the count, but if many old sessions exist with `last_speaker = 'assistant'`, recent relevant sessions could be displaced by stale ones. Consider adding a time-bound filter (e.g., `AND last_speaker_at > unixepoch() - 86400*7`) or an `ORDER BY` that favors recent sessions (already present via `last_speaker_at DESC`, so this is minor). `[fixable]`

server/index.ts
```diff
@@ -260,6 +260,7 @@ setSessionChangeCallback((clientId, event) => {
     const session = registry.get(clientId);
     if (session?.sessionId) {
       eventStore.updateLastSpeaker(session.sessionId, 'assistant');
+      overviewEmitter.updateSpeaker(session.sessionId, 'assistant');
```
dimakis, line 268: 🟡 **bugs**: Speaker cache is only updated for 'assistant' here on turn_end. The corresponding 'user' updates in sendToChat (chat.ts:1040) and interruptChat (chat.ts:1074) write to EventStore via updateLastSpeaker but never call overviewEmitter.updateSpeaker(sessionId, 'user'). After a turn_end caches speaker='assistant', a follow-up user message leaves the cache stale — getCachedSpeaker still returns 'assistant', so awaitingReply remains true until the next turn_end. If the assistant errors out before responding, the stale cache persists indefinitely. `[fixable]`

server/query-loop.ts
```diff
@@ -504,6 +504,7 @@ async function _runQueryLoopInner(
                   messageId: `umsg-${now}-init`,
                   text: initialPrompt,
                 });
+                store.updateLastSpeaker(resolvedSessionId, 'user');
```
dimakis, line 507: 🟡 **bugs**: store.updateLastSpeaker(resolvedSessionId, 'user') correctly updates EventStore, but there is no corresponding overviewEmitter.updateSpeaker(resolvedSessionId, 'user') call. For initial prompts this is benign (cache is cold, so getCachedSpeaker falls through to EventStore), but the asymmetry between the 'user' path here (EventStore only) and the 'assistant' path in index.ts (EventStore + cache) is a latent inconsistency that will bite if the cache is ever pre-warmed before the initial prompt. `[fixable]`

server/__tests__/session-overview.test.ts
dimakis: 🔵 **style**: Duplicate vi.mock('../worktree.js') — this mock (exporting hasUncommittedWork) is immediately overridden by the existing mock at line 79 (exporting hasUncommittedWorkAsync). Vitest hoists all vi.mock calls and the last registration wins, so lines 70-72 are dead code. Remove the duplicate to avoid confusion. `[fixable]`

server/__tests__/session-overview.test.ts
```diff
@@ -658,4 +664,52 @@ describe('SessionOverviewEmitter', () => {
...
+    // First snapshot — should call getSession to populate cache
+    emitter.getSnapshot();
+    expect(getSessionMock).toHaveBeenCalledTimes(1);
+
+    // Second snapshot — should use cached value
+    emitter.getSnapshot();
+    expect(getSessionMock).toHaveBeenCalledTimes(1);
+  });
+
+  it('updateSpeaker bypasses eventStore lookup', () => {
```
dimakis, line 688: 🔵 **missing_tests**: Tests verify that updateSpeaker('assistant') sets awaitingReply=true and that the cache avoids repeated DB lookups, but no test verifies that updateSpeaker(sessionId, 'user') clears awaitingReply back to false. This is the exact scenario for the cache staleness bug — a test would have caught the missing 'user' cache update in sendToChat/interruptChat. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Speaker caching optimization introduces a cache-vs-store divergence: user messages update EventStore but not the speaker cache, leaving awaitingReply stale until the next turn_end. The design doc and EventStore query changes are sound._

- 🟡 **bugs** (L263): Speaker cache is only updated for 'assistant' here on turn_end. The corresponding 'user' updates in sendToChat (chat.ts:1040) and interruptChat (chat.ts:1074) write to EventStore via updateLastSpeaker but never call overviewEmitter.updateSpeaker(sessionId, 'user'). After a turn_end caches speaker='assistant', a follow-up user message leaves the cache stale — getCachedSpeaker still returns 'assistant', so awaitingReply remains true until the next turn_end. If the assistant errors out before responding, the stale cache persists indefinitely. `[fixable]`


### `server/query-loop.ts`
_Speaker caching optimization introduces a cache-vs-store divergence: user messages update EventStore but not the speaker cache, leaving awaitingReply stale until the next turn_end. The design doc and EventStore query changes are sound._

- 🟡 **bugs** (L507): store.updateLastSpeaker(resolvedSessionId, 'user') correctly updates EventStore, but there is no corresponding overviewEmitter.updateSpeaker(resolvedSessionId, 'user') call. For initial prompts this is benign (cache is cold, so getCachedSpeaker falls through to EventStore), but the asymmetry between the 'user' path here (EventStore only) and the 'assistant' path in index.ts (EventStore + cache) is a latent inconsistency that will bite if the cache is ever pre-warmed before the initial prompt. `[fixable]`


_Speaker caching optimization introduces a cache-vs-store divergence: user messages update EventStore but not the speaker cache, leaving awaitingReply stale until the next turn_end. The design doc and EventStore query changes are sound._

- 🔵 **style** (L70): Duplicate vi.mock('../worktree.js') — this mock (exporting hasUncommittedWork) is immediately overridden by the existing mock at line 79 (exporting hasUncommittedWorkAsync). Vitest hoists all vi.mock calls and the last registration wins, so lines 70-72 are dead code. Remove the duplicate to avoid confusion. `[fixable]`
- 🔵 **missing_tests** (L693): Tests verify that updateSpeaker('assistant') sets awaitingReply=true and that the cache avoids repeated DB lookups, but no test verifies that updateSpeaker(sessionId, 'user') clears awaitingReply back to false. This is the exact scenario for the cache staleness bug — a test would have caught the missing 'user' cache update in sendToChat/interruptChat. `[fixable]`

dimakis:
## Centaur Review

_Speaker cache is only updated for assistant turns, not user turns — creating a staleness window where sessions incorrectly show as 'awaiting reply'. The `getAttentionSessions` filter change and test additions are solid, but the duplicate `vi.mock` in the test file is dead code._

- 🟡 **bugs** (L308): Speaker cache is never invalidated for user messages. `overviewEmitter.updateSpeaker(sessionId, 'user')` is never called — only the `'assistant'` path in `index.ts:263` updates the cache. The DB is updated in `chat.ts:919,1040,1074` and `query-loop.ts:507` but the cache retains the stale `'assistant'` entry from the previous `turn_end`. Between a user sending a message and the next `turn_end`, `getCachedSpeaker()` returns the stale cached `'assistant'` value (it won't fall through to DB because a cached entry already exists), causing `awaitingReply` to be incorrectly `true`. The `hasSnapshot` guard masks this in most streaming cases, but it's exposed during resume and edge cases where the agent hasn't started streaming yet. `[fixable]`
- 🟡 **bugs** (L99): `forget(clientId)` cleans up `lastEventTimes` but does not remove the corresponding entry from `speakerCache` (keyed by sessionId). Over the lifetime of the server, speaker cache entries accumulate for every session that was ever active. This is a minor memory leak — consider clearing `speakerCache` in `forget()` (would need a clientId→sessionId lookup, or accept the leak as bounded). `[fixable]`


_Speaker cache is only updated for assistant turns, not user turns — creating a staleness window where sessions incorrectly show as 'awaiting reply'. The `getAttentionSessions` filter change and test additions are solid, but the duplicate `vi.mock` in the test file is dead code._

- 🟡 **style** (L70): Duplicate `vi.mock('../worktree.js')` — lines 70-72 mock `hasUncommittedWork` (sync), but lines 79-81 re-mock the same module with only `hasUncommittedWorkAsync`. In Vitest, the last `vi.mock` for a given module wins, so the mock at lines 70-72 is dead code with no effect. Either merge both exports into a single `vi.mock` call, or remove the first one since `session-overview.ts` only imports `hasUncommittedWorkAsync`. `[fixable]`
- 🔵 **missing_tests** (L714): Speaker cache tests verify caching and `updateSpeaker('assistant')` bypass, but don't test the stale-cache scenario: when a cached entry exists as `'assistant'` and the user sends a message without calling `updateSpeaker('user')`, the cache returns stale data. A test asserting the expected behavior (or documenting the known gap) would clarify intent. `[fixable]`


_Speaker cache is only updated for assistant turns, not user turns — creating a staleness window where sessions incorrectly show as 'awaiting reply'. The `getAttentionSessions` filter change and test additions are solid, but the duplicate `vi.mock` in the test file is dead code._

- 🔵 **missing_tests** (L462): `getAttentionSessions` test for inactive sessions was updated to match the removed `is_active = 1` filter, but there's no test verifying that hidden sessions with `lastSpeaker = 'assistant'` are still excluded. The existing `excludes hidden sessions` test covers this, but adding a test that combines `inactive + hidden` would strengthen coverage of the filter change. `[fixable]`


_Speaker cache is only updated for assistant turns, not user turns — creating a staleness window where sessions incorrectly show as 'awaiting reply'. The `getAttentionSessions` filter change and test additions are solid, but the duplicate `vi.mock` in the test file is dead code._

- 🔵 **style** (L222): The added JSDoc comment `/** Scaffolding — column + mapping wired up, callers added when session lifecycle overhaul lands. */` references future work that may or may not happen. Per project conventions, comments should explain the WHY of current code, not reference future plans. This will rot as the codebase evolves. Consider removing it or replacing with a concrete note about why the method exists without callers. `[fixable]`

server/index.ts
```diff
@@ -253,13 +253,20 @@ setSessionChangeCallback((clientId, event) => {
   if (event === 'start') {
     overviewEmitter.touch(clientId);
   } else if (event === 'end') {
-    overviewEmitter.forget(clientId);
+    const session = registry.get(clientId);
```
dimakis: 🔴 **bugs**: Memory leak: `registry.get(clientId)` always returns `undefined` when the `'end'` event fires, because the session has already been removed from the registry. In the normal flow, `registry.remove(clientId)` runs at query-loop.ts:1350 before `_startChatInner`'s `finally` fires the callback. In the error flow, `registry.abort(clientId)` runs at chat.ts:964 in the `catch` block before `finally`. Since `session?.sessionId` is always `undefined`, `speakerCache.delete()` is never called and entries accumulate indefinitely. Fix: pass `sessionId` as a parameter in the callback itself (e.g., change the callback signature to include sessionId), or capture the sessionId before `registry.remove`/`abort` and pass it to `_onSessionChange`. `[fixable]`

server/session-overview.ts
```diff
@@ -299,6 +304,27 @@ export class SessionOverviewEmitter {
     };
   }
 
+  /**
+   * Get cached speaker info for a session. Lazily populated from EventStore
+   * on first access, then kept up-to-date via updateSpeaker().
+   */
```
dimakis, line 313: 🔵 **style**: The `getCachedSpeaker` return type `{ speaker: string | null; at: number | null }` uses `string` for speaker, but `updateSpeaker` accepts `'user' | 'assistant'`. Consider narrowing the cached type to `'user' | 'assistant' | null` for type safety, matching the domain model. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Speaker caching optimization has a critical bug: `registry.get(clientId)` returns undefined in the 'end' callback because the session is already removed from the registry, so speakerCache entries are never cleaned up (memory leak). The design doc is well-structured._

- 🔴 **bugs** (L256): Memory leak: `registry.get(clientId)` always returns `undefined` when the `'end'` event fires, because the session has already been removed from the registry. In the normal flow, `registry.remove(clientId)` runs at query-loop.ts:1350 before `_startChatInner`'s `finally` fires the callback. In the error flow, `registry.abort(clientId)` runs at chat.ts:964 in the `catch` block before `finally`. Since `session?.sessionId` is always `undefined`, `speakerCache.delete()` is never called and entries accumulate indefinitely. Fix: pass `sessionId` as a parameter in the callback itself (e.g., change the callback signature to include sessionId), or capture the sessionId before `registry.remove`/`abort` and pass it to `_onSessionChange`. `[fixable]`


_Speaker caching optimization has a critical bug: `registry.get(clientId)` returns undefined in the 'end' callback because the session is already removed from the registry, so speakerCache entries are never cleaned up (memory leak). The design doc is well-structured._

- 🟡 **missing_tests**: The new speaker cache tests are thorough for the `SessionOverviewEmitter` in isolation, but no integration test verifies that `overviewEmitter.forget(clientId, sessionId)` actually receives a non-null `sessionId` from the callback in `index.ts`. Given the bug above (registry already cleared when `'end'` fires), a test covering the full `setSessionChangeCallback` → `forget` flow would have caught this. `[fixable]`


_Speaker caching optimization has a critical bug: `registry.get(clientId)` returns undefined in the 'end' callback because the session is already removed from the registry, so speakerCache entries are never cleaned up (memory leak). The design doc is well-structured._

- 🔵 **style** (L310): The `getCachedSpeaker` return type `{ speaker: string | null; at: number | null }` uses `string` for speaker, but `updateSpeaker` accepts `'user' | 'assistant'`. Consider narrowing the cached type to `'user' | 'assistant' | null` for type safety, matching the domain model. `[fixable]`

docs/design/session-state-machine.md
```diff
@@ -0,0 +1,1050 @@
...
+      sessionId,
+      state,
+      clientId: found.clientId,
+    });
+    span.setAttribute('routing.decision', 'deduplicated');
+    // Don't start a second query — the first one will transition to ACTIVE
+    // and pick up this message from the conversation history on next turn.
+    transport.send({
+      type: 'system',
+      message: 'Session is starting, please wait...',
```
dimakis, line 553: 🟡 **bugs**: VALID_TRANSITIONS defines `ENDED: ['STARTING']`, but the Phase 1 startChat code sets CREATED before STARTING for all sessions including resumes (`eventStore.setState(sessionId, 'CREATED')` then `setState(sessionId, 'STARTING')`). A resumed session would attempt ENDED → CREATED, which Phase 4 enforcement would reject. Either add CREATED to ENDED's valid targets or skip CREATED for resume flows (go directly ENDED → STARTING). `[fixable]`

docs/design/session-state-machine.md
```diff
@@ -0,0 +1,1050 @@
...
+    });
+    applySkillPolicy(sessionClientId);
+    return;
+  }
+
+  // Guard: duplicate startChat while session is still booting
+  if (found && (state === 'CREATED' || state === 'STARTING')) {
+    log.warn('message received while session still starting, ignoring duplicate', {
+      sessionId,
+      state,
```
dimakis, line 545: 🟡 **bugs**: VALID_TRANSITIONS defines `CREATED: ['STARTING']` with no path to ENDED. If `query()` throws after CREATED is set but before STARTING (lines 197-203 of the Phase 1 code), the session is stuck in CREATED with no valid transition out. The crash recovery in Phase 4 handles this on restart (`force: true`), but during normal operation the session is in a dead state until server restart. Consider adding ENDED to CREATED's valid transitions: `CREATED: ['STARTING', 'ENDED']`. `[fixable]`

docs/design/session-state-machine.md
```diff
@@ -0,0 +1,1050 @@
...
+#### Add State to All Relevant Logs
+
+```typescript
+// ws-handler-v2.ts handleSendV2
+const found = ctx.sessionRegistry.findBySessionId(sessionId);
+const state = ctx.eventStore.getSessionState(sessionId);
+
+log.info('send routing decision', {
+  sessionId,
+  registryHas: !!found,
```
dimakis, line 399: 🟡 **bugs**: Phase 3 routing code passes `{ force: true }` to `setState()` (line 399: `ctx.eventStore.setState(sessionId, 'ENDED', { reason: 'registry_lost_session', force: true })`), but the `opts` type defined in Phase 1 is `{ clientId?: string; reason?: string }` — `force` isn't added until Phase 4. Implementing Phase 3 before Phase 4 would produce a TypeScript error. Either add `force` to the Phase 1 type definition or restructure so Phase 3 doesn't need it. `[fixable]`

docs/design/session-state-machine.md
```diff
@@ -0,0 +1,1050 @@
...
+DETACHED → ENDED          detach TTL expires without closeout, or query aborts
+
+SUSPENDED → ACTIVE        resume from background, reattach
+SUSPENDED → ENDED         suspend timeout expires, query aborts
+
+CLOSING → ENDED           closeout completes or timeout expires
+
+ENDED → STARTING          explicit resume with startChat(resume: sessionId)
+```
+
```
dimakis, line 108: 🔵 **style**: The invalid transitions section states 'SUSPENDED → DETACHED (if background resumes without transport, go straight to ENDED)' — this parenthetical describes the intended behavior but reads like it contradicts the transition table which already shows SUSPENDED → ACTIVE and SUSPENDED → ENDED as the only valid exits. Consider rephrasing to 'SUSPENDED → DETACHED is invalid — if transport is absent on resume, transition directly to ENDED instead' for clarity. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/session-state-machine.md`
_Code changes (speaker caching + user_message callback) are clean and well-tested. The design doc has two transition-table inconsistencies (ENDED→CREATED on resume, CREATED with no path to ENDED) that would cause validation failures when Phase 4 enforcement is implemented — fix those before using the doc as an implementation spec._

- 🟡 **bugs** (L553): VALID_TRANSITIONS defines `ENDED: ['STARTING']`, but the Phase 1 startChat code sets CREATED before STARTING for all sessions including resumes (`eventStore.setState(sessionId, 'CREATED')` then `setState(sessionId, 'STARTING')`). A resumed session would attempt ENDED → CREATED, which Phase 4 enforcement would reject. Either add CREATED to ENDED's valid targets or skip CREATED for resume flows (go directly ENDED → STARTING). `[fixable]`
- 🟡 **bugs** (L545): VALID_TRANSITIONS defines `CREATED: ['STARTING']` with no path to ENDED. If `query()` throws after CREATED is set but before STARTING (lines 197-203 of the Phase 1 code), the session is stuck in CREATED with no valid transition out. The crash recovery in Phase 4 handles this on restart (`force: true`), but during normal operation the session is in a dead state until server restart. Consider adding ENDED to CREATED's valid transitions: `CREATED: ['STARTING', 'ENDED']`. `[fixable]`
- 🟡 **bugs** (L399): Phase 3 routing code passes `{ force: true }` to `setState()` (line 399: `ctx.eventStore.setState(sessionId, 'ENDED', { reason: 'registry_lost_session', force: true })`), but the `opts` type defined in Phase 1 is `{ clientId?: string; reason?: string }` — `force` isn't added until Phase 4. Implementing Phase 3 before Phase 4 would produce a TypeScript error. Either add `force` to the Phase 1 type definition or restructure so Phase 3 doesn't need it. `[fixable]`
- 🔵 **style** (L108): The invalid transitions section states 'SUSPENDED → DETACHED (if background resumes without transport, go straight to ENDED)' — this parenthetical describes the intended behavior but reads like it contradicts the transition table which already shows SUSPENDED → ACTIVE and SUSPENDED → ENDED as the only valid exits. Consider rephrasing to 'SUSPENDED → DETACHED is invalid — if transport is absent on resume, transition directly to ENDED instead' for clarity. `[fixable]`


_Code changes (speaker caching + user_message callback) are clean and well-tested. The design doc has two transition-table inconsistencies (ENDED→CREATED on resume, CREATED with no path to ENDED) that would cause validation failures when Phase 4 enforcement is implemented — fix those before using the doc as an implementation spec._

- 🔵 **missing_tests**: The new tests cover cache hit/miss, updateSpeaker, and forget-with-sessionId, but there's no test verifying that the `user_message` callback event in chat.ts actually fires and flows through to `updateSpeaker('user')` on the emitter. An integration-style test (or at least a test that `_onSessionChange` is called with `'user_message'` from `sendToChat`/`interruptChat`) would catch wiring regressions. `[fixable]`

server/index.ts
```diff
@@ -249,17 +249,23 @@ setHealthMonitor(healthMonitor);
 healthMonitor.start();
 
 // Hook session lifecycle events into the overview emitter
-setSessionChangeCallback((clientId, event) => {
+setSessionChangeCallback((clientId, event, sessionId) => {
   if (event === 'start') {
     overviewEmitter.touch(clientId);
   } else if (event === 'end') {
```
dimakis, line 255: 🟡 **bugs**: The new `user_message` callback path only updates `speakerCache` when `registry.get(clientId)` still returns a live session. In `_startChatInner()` resume mode, `_onSessionChange?.(clientId, 'user_message')` is fired before the resumed session is registered in-memory, so this branch is skipped and the cache can retain a stale `'assistant'` value from the previous turn. Until the next SQLite refresh or cache eviction, `getSnapshot()` may incorrectly report `awaitingReply: true` immediately after the user sends a follow-up on a resumed session. `[fixable]`

server/chat.ts
```diff
@@ -921,6 +925,7 @@ async function _startChatInner(
         text: fullPrompt,
       });
       eventStore.updateLastSpeaker(options.resume, 'user');
+      _onSessionChange?.(clientId, 'user_message');
```
dimakis, line 928: 🟡 **missing_tests**: There is no test covering the resume-path ordering introduced by the new `user_message` callback. The added session-overview tests only exercise `updateSpeaker()` directly, but not the real flow where `_startChatInner(..., { resume })` emits `user_message` before `registry.register(...)`. A regression test should verify that sending a follow-up on a resumed session clears `awaitingReply` immediately. `[fixable]`

dimakis (commented):
## Codex Review

### `server/index.ts`
_PR improves session overview caching, but the new user-message cache update misses resumed sessions because the callback fires before the session is present in the registry._

- 🟡 **bugs** (L255): The new `user_message` callback path only updates `speakerCache` when `registry.get(clientId)` still returns a live session. In `_startChatInner()` resume mode, `_onSessionChange?.(clientId, 'user_message')` is fired before the resumed session is registered in-memory, so this branch is skipped and the cache can retain a stale `'assistant'` value from the previous turn. Until the next SQLite refresh or cache eviction, `getSnapshot()` may incorrectly report `awaitingReply: true` immediately after the user sends a follow-up on a resumed session. `[fixable]`


### `server/chat.ts`
_PR improves session overview caching, but the new user-message cache update misses resumed sessions because the callback fires before the session is present in the registry._

- 🟡 **missing_tests** (L928): There is no test covering the resume-path ordering introduced by the new `user_message` callback. The added session-overview tests only exercise `updateSpeaker()` directly, but not the real flow where `_startChatInner(..., { resume })` emits `user_message` before `registry.register(...)`. A regression test should verify that sending a follow-up on a resumed session clears `awaitingReply` immediately. `[fixable]`

PR #332, opened 2026-05-16
feat(ui): rich boot context pill with content passthrough

dimakis:
## Summary

- Replace flat file-list boot context pill with a diagnostic tool showing exactly what ContexGin injects
- Each source shows a color-coded kind badge (constitution, profile, memory, service, reference)
- Included sections are expandable — tap to see actual injected content
- Trimmed sections shown with token counts, also expandable
- Budget indicator: "8.0k / 8.0k" showing usage vs budget
- Token budget now read from `.agents/mitzo-conversational.yaml` recipe instead of hardcoded 8000
- Backwards compatible protocol parser (handles both rich objects and plain string sources)

**Depends on:** dimakis/contexgin PR (adds `included` field to `CompiledContext`)

## Files changed

| File | Change |
|------|--------|
| `packages/client/src/slices/messages.ts` | New `BootSourceMeta`, `SectionMeta` types; enriched `BootContextMeta` |
| `packages/client/src/index.ts` | Export new types |
| `packages/client/src/protocol-parser.ts` | Parse rich source/section metadata |
| `packages/client/__tests__/protocol-parser.test.ts` | Updated test expectations |
| `server/chat.ts` | Extract rich metadata from ContexGin; recipe-driven budget |
| `frontend/src/components/BootContextPill.tsx` | Expandable sources + sections + trimmed |
| `frontend/src/styles/global.css` | Kind badges, section rows, content preview |

## Test plan

- [ ] Protocol parser: 54/54 tests pass
- [ ] Server: no tsc errors in chat.ts
- [ ] Visual: open a new Mitzo session, expand the context pill, verify sources show kind badges
- [ ] Visual: tap an included section, verify content preview expands
- [ ] Visual: tap "N sections trimmed", verify trimmed list with token counts
- [ ] Verify budget reads from recipe (change value in yaml, restart, check pill header)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #341, opened 2026-05-18
feat(ui): boot context pill sticky + modal view

dimakis:
## Summary

Fixes two UX issues with the boot context pill reported in Telos ae582634f14580e5:

1. ✅ **Pill disappears as messages scroll** → Now sticky at top of chat
2. ✅ **No way to view full markdown** → New ⎘ button opens modal overlay

### 1. Sticky Positioning

The pill now uses `position: sticky; top: 0; z-index: 10;` to stay visible as messages scroll. Maintains context visibility throughout long conversations.

**Before:** Pill scrolls out of view after 3-4 agent responses  
**After:** Pill stays pinned at top, always accessible

### 2. Modal Pop-Out

New ⎘ button in pill header opens full-screen modal showing complete boot context markdown sent to the model.

**Features:**
- Click outside or ✕ button to close
- Full markdown with proper formatting
- Scrollable for long contexts
- Touch-optimized overlay with blur backdrop

### Protocol Extension

- Added `fullMarkdown?: string` to `BootContextMeta` type
- Server extracts `result` field from ContexGin compilation
- Protocol parser handles optional field (backward compatible)
- Falls back gracefully if ContexGin doesn't provide full markdown

## Files Changed

- `frontend/src/components/BootContextPill.tsx` — modal state + UI
- `frontend/src/styles/global.css` — sticky + modal styles  
- `packages/client/src/slices/messages.ts` — BootContextMeta type
- `packages/client/src/protocol-parser.ts` — extract fullMarkdown
- `server/chat.ts` — send fullMarkdown from ContexGin
- `docs/design/boot-context-pill-ux-fixes.md` — design doc

## Test Plan

- [ ] Start new Mitzo session, verify pill stays at top as messages scroll
- [ ] Tap ⎘ button, verify modal opens with full markdown
- [ ] Click outside modal or ✕ button, verify it closes
- [ ] Verify pill still expands inline to show sources/sections
- [ ] Test on mobile (iOS) for touch interactions
- [ ] Verify backward compat if ContexGin doesn't send fullMarkdown

## Related

- **Telos:** ae582634f14580e5 (parent task)
- **Design:** `docs/design/boot-context-pill-ux-fixes.md`
- **PR #332:** Prior work adding rich pill metadata

### DIFF
--- a/frontend/src/components/BootContextPill.tsx
+++ b/frontend/src/components/BootContextPill.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 import type { BootContextMeta, SectionMeta } from '@mitzo/client';
 
 interface Props {
@@ -54,71 +54,84 @@ export function BootContextPill({ context }: Props) {
       : String(context.tokenBudget);
   const label = `${context.sourceCount} sources \u00b7 ${tokenLabel} / ${budgetLabel}`;
 
+  // Escape key handler for modal
+  useEffect(() => {
+    if (!showModal) return;
+    const handleEscape = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') setShowModal(false);
+    };
+    window.addEventListener('keydown', handleEscape);
+    return () => window.removeEventListener('keydown', handleEscape);
+  }, [showModal]);
+
   return (
-    <div className="boot-context-pill">
-      <button className="boot-context-pill-header" onClick={() => setExpanded((e) => !e)}>
-        <span className={`boot-context-pill-dot ${dotClass}`} />
-        <span className="boot-context-pill-label">{label}</span>
-        <span className="boot-context-pill-engine">{isContexgin ? 'ContexGin' : 'Fallback'}</span>
-        <span className="boot-context-pill-chevron">{expanded ? '\u25BE' : '\u25B8'}</span>
-        {context.fullMarkdown && (
-          <button
-            className="boot-context-pill-view-full"
-            onClick={(e) => {
-              e.stopPropagation();
-              setShowModal(true);
-            }}
-            title="View full markdown"
-          >
-            ⧉
+    <>
+      <div className="boot-context-pill">
+        <div className="boot-context-pill-header-wrapper">
+          <button className="boot-context-pill-header" onClick={() => setExpanded((e) => !e)}>
+            <span className={`boot-context-pill-dot ${dotClass}`} />
+            <span className="boot-context-pill-label">{label}</span>
+            <span className="boot-context-pill-engine">
+              {isContexgin ? 'ContexGin' : 'Fallback'}
+            </span>
+            <span className="boot-context-pill-chevron">{expanded ? '\u25BE' : '\u25B8'}</span>
           </button>
-        )}
-      </button>
-      {expanded && (
-        <div className="boot-context-pill-content">
-          <div className="boot-context-pill-section-label">Sources</div>
-          {context.sources.map((src, idx) => (
-            <div key={idx} className="boot-context-pill-source-row">
-              <span className={`boot-context-pill-kind boot-context-pill-kind--${src.kind}`}>
-                {KIND_LABELS[src.kind] ?? src.kind}
-              </span>
-              <span className="boot-context-pill-source-path">{src.path}</span>
-            </div>
-          ))}
-
-          {context.included.length > 0 && (
-            <>
-              <div className="boot-context-pill-section-label">
-                Included ({context.included.length})
-              </div>
-              {context.included.map((section, idx) => (
-                <SectionRow key={idx} section={section} />
-              ))}
-            </>
-          )}
-
-          {context.trimmed.length > 0 && (
-            <>
-              <button
-                className="boot-context-pill-trimmed-toggle"
-                onClick={(e) => {
-                  e.stopPropagation();
-                  setShowTrimmed((t) => !t);
-                }}
-              >
-                {context.trimmed.length} section{context.trimmed.length !== 1 ? 's' : ''} trimmed
-                <span className="boot-context-pill-chevron-inline">
-                  {showTrimmed ? '\u25BE' : '\u25B8'}
-                </span>
-              </button>
-              {showTrimmed &&
-                context.trimmed.map((section, idx) => (
-                  <SectionRow key={idx} section={section} dimmed />
-                ))}
-            </>
+          {context.fullMarkdown && (
+            <button
+              className="boot-context-pill-view-full"
+              onClick={() => setShowModal(true)}
+              title="View full markdown"
+            >
+              ⧉
+            </button>
           )}
         </div>
-      )}
+        {expanded && (
+          <div className="boot-context-pill-content">
+            <div className="boot-context-pill-section-label">Sources</div>
+            {context.sources.map((src, idx) => (
+              <div key={idx} className="boot-context-pill-source-row">
+                <span className={`boot-context-pill-kind boot-context-pill-kind--${src.kind}`}>
+                  {KIND_LABELS[src.kind] ?? src.kind}
+                </span>
+                <span className="boot-context-pill-source-path">{src.path}</span>
+              </div>
+            ))}
+
+            {context.included.length > 0 && (
+              <>
+                <div className="boot-context-pill-section-label">
+                  Included ({context.included.length})
+                </div>
+                {context.included.map((section, idx) => (
+                  <SectionRow key={idx} section={section} />
+                ))}
+              </>
+            )}
+
+            {context.trimmed.length > 0 && (
+              <>
+                <button
+                  className="boot-context-pill-trimmed-toggle"
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    setShowTrimmed((t) => !t);
+                  }}
+                >
+                  {context.trimmed.length} section{context.trimmed.length !== 1 ? 's' : ''} trimmed
+                  <span className="boot-context-pill-chevron-inline">
+                    {showTrimmed ? '\u25BE' : '\u25B8'}
+                  </span>
+                </button>
+                {showTrimmed &&
+                  context.trimmed.map((section, idx) => (
+                    <SectionRow key={idx} section={section} dimmed />
+                  ))}
+              </>
+            )}
+          </div>
+        )}
+      </div>
 
       {showModal && context.fullMarkdown && (
         <div
@@ -126,7 +139,13 @@ export function BootContextPill({ context }: Props) {
           onClick={() => setShowModal(false)}
           onTouchStart={(e) => e.stopPropagation()}
         >
-          <div className="boot-context-modal" onClick={(e) => e.stopPropagation()}>
+          <div
+            className="boot-context-modal"
+            onClick={(e) => e.stopPropagation()}
+            role="dialog"
+            aria-modal="true"
+            aria-label="Boot context full markdown"
+          >
             <div className="boot-context-modal-header">
               <h3>Boot Context (Full Markdown)</h3>
               <button onClick={() => setShowModal(false)} className="boot-context-modal-close">
@@ -137,6 +156,6 @@ export function BootContextPill({ context }: Props) {
           </div>
         </div>
       )}
-    </div>
+    </>
   );
 }
--- a/frontend/src/components/BootContextPill.tsx
+++ b/frontend/src/components/BootContextPill.tsx
@@ -70,7 +70,7 @@ export function BootContextPill({ context }: Props) {
             }}
             title="View full markdown"
           >
-            ⎘
+            ⧉
           </button>
         )}
       </button>
--- a/frontend/src/components/BootContextPill.tsx
+++ b/frontend/src/components/BootContextPill.tsx
@@ -40,6 +40,7 @@ function SectionRow({ section, dimmed }: { section: SectionMeta; dimmed?: boolea
 export function BootContextPill({ context }: Props) {
   const [expanded, setExpanded] = useState(false);
   const [showTrimmed, setShowTrimmed] = useState(false);
+  const [showModal, setShowModal] = useState(false);
 
   const isContexgin = context.source === 'contexgin';
   const dotClass = isContexgin ? 'boot-context-pill-dot--ok' : 'boot-context-pill-dot--warn';
@@ -60,6 +61,18 @@ export function BootContextPill({ context }: Props) {
         <span className="boot-context-pill-label">{label}</span>
         <span className="boot-context-pill-engine">{isContexgin ? 'ContexGin' : 'Fallback'}</span>
         <span className="boot-context-pill-chevron">{expanded ? '\u25BE' : '\u25B8'}</span>
+        {context.fullMarkdown && (
+          <button
+            className="boot-context-pill-view-full"
+            onClick={(e) => {
+              e.stopPropagation();
+              setShowModal(true);
+            }}
+            title="View full markdown"
+          >
+            ⎘
+          </button>
+        )}
       </button>
       {expanded && (
         <div className="boot-context-pill-content">
@@ -104,6 +117,24 @@ export function BootContextPill({ context }: Props) {
                 ))}
             </>
           )}
+        </div>
+      )}
+
+      {showModal && context.fullMarkdown && (
+        <div
+          className="boot-context-modal-overlay"
+          onClick={() => setShowModal(false)}
+          onTouchStart={(e) => e.stopPropagation()}
+        >
+          <div className="boot-context-modal" onClick={(e) => e.stopPropagation()}>
+            <div className="boot-context-modal-header">
+              <h3>Boot Context (Full Markdown)</h3>
+              <button onClick={() => setShowModal(false)} className="boot-context-modal-close">
+                ✕
+              </button>
+            </div>
+            <pre className="boot-context-modal-content">{context.fullMarkdown}</pre>
+          </div>
         </div>
       )}
     </div>
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -2405,13 +2405,19 @@ textarea:focus {
   background: var(--bg-primary);
 }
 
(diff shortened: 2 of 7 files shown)


frontend/src/components/BootContextPill.tsx
```diff
@@ -106,6 +119,24 @@ export function BootContextPill({ context }: Props) {
           )}
         </div>
       )}
+
+      {showModal && context.fullMarkdown && (
```
dimakis, line 136: 🟡 **bugs**: The modal has no Escape key handler. A `useEffect` that listens for `keydown` when `showModal` is true should call `setShowModal(false)` on Escape. The codebase already uses this pattern (e.g., TelosSection.tsx). Without it, keyboard-only users cannot dismiss the modal. `[fixable]`

dimakis, line 136: 🔵 **style**: The modal overlay div lacks `role="dialog"` and `aria-modal="true"` attributes, and focus is not trapped inside the modal. Consider adding basic ARIA attributes for screen reader support. `[fixable]`

server/chat.ts
```diff
@@ -851,6 +851,7 @@ async function _startChatInner(
 
       const included = extractSections(rawIncluded);
       const trimmed = extractSections(rawTrimmed);
+      const fullMarkdown = typeof obj.result === 'string' ? obj.result : undefined;
```
dimakis: 🔵 **unsafe_assumptions**: `obj.result` (the full compiled markdown from ContexGin) could be very large — potentially tens of thousands of characters for repos with extensive context sources. This is sent over the WebSocket and stored in the client's state. Consider whether a size cap or truncation would be appropriate, or at minimum documenting the expected size range. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid feature addition — sticky positioning and modal are correctly implemented. Main gaps are missing protocol-parser tests for the new `fullMarkdown` field and no Escape-key dismissal for the modal._

- 🟡 **missing_tests**: The existing boot_context test cases in protocol-parser.test.ts do not cover the new `fullMarkdown` field. There should be at least one test verifying that `fullMarkdown` passes through to SET_BOOT_CONTEXT, and one verifying it is omitted when absent or non-string. `[fixable]`


### `frontend/src/components/BootContextPill.tsx`
_Solid feature addition — sticky positioning and modal are correctly implemented. Main gaps are missing protocol-parser tests for the new `fullMarkdown` field and no Escape-key dismissal for the modal._

- 🟡 **bugs** (L123): The modal has no Escape key handler. A `useEffect` that listens for `keydown` when `showModal` is true should call `setShowModal(false)` on Escape. The codebase already uses this pattern (e.g., TelosSection.tsx). Without it, keyboard-only users cannot dismiss the modal. `[fixable]`
- 🔵 **style** (L123): The modal overlay div lacks `role="dialog"` and `aria-modal="true"` attributes, and focus is not trapped inside the modal. Consider adding basic ARIA attributes for screen reader support. `[fixable]`


### `server/chat.ts`
_Solid feature addition — sticky positioning and modal are correctly implemented. Main gaps are missing protocol-parser tests for the new `fullMarkdown` field and no Escape-key dismissal for the modal._

- 🔵 **unsafe_assumptions** (L854): `obj.result` (the full compiled markdown from ContexGin) could be very large — potentially tens of thousands of characters for repos with extensive context sources. This is sent over the WebSocket and stored in the client's state. Consider whether a size cap or truncation would be appropriate, or at minimum documenting the expected size range. `[fixable]`

dimakis:
## Centaur Review

### `frontend/src/components/BootContextPill.tsx`
_Solid feature work with one HTML validity bug (nested buttons) that should be fixed before merge, plus missing test coverage for the new fullMarkdown protocol field._

- 🔴 **bugs** (L65): Nested `<button>` inside `<button>`. The view-full button (line 65) is a child of the pill-header button (line 59). This is invalid HTML per spec — browsers handle it inconsistently (some extract the inner button from the DOM, some fire both handlers). The `e.stopPropagation()` mitigates double-firing in compliant browsers, but the structure should be fixed. Replace the outer `<button>` with a `<div role="button" tabIndex={0}>` or move the view-full button outside the header button as a sibling. `[fixable]`
- 🟡 **unsafe_assumptions** (L123): The modal overlay (position: fixed) is rendered inside `.boot-context-pill` (position: sticky). While this works today because no ancestor element has `transform`, `will-change`, or `filter`, any future CSS change adding those properties to the scroll container or its ancestors will silently break the modal's viewport-relative positioning. Consider rendering the modal as a sibling outside the sticky container, or using a React portal. `[fixable]`
- 🔵 **style** (L123): The modal lacks an Escape key handler. Users expect to dismiss modals with Escape. Add a `useEffect` with a `keydown` listener when `showModal` is true, calling `setShowModal(false)` on Escape. `[fixable]`
- 🔵 **style** (L129): The modal is missing accessibility attributes: `role="dialog"`, `aria-modal="true"`, and an `aria-label` on `.boot-context-modal`. Minor gap, but easy to add. `[fixable]`


_Solid feature work with one HTML validity bug (nested buttons) that should be fixed before merge, plus missing test coverage for the new fullMarkdown protocol field._

- 🟡 **missing_tests** (L709): The boot_context test suite (6 tests) has no coverage for the new `fullMarkdown` field. Needs at least: (1) a positive test showing `fullMarkdown: 'some content'` is passed through to the SET_BOOT_CONTEXT action, and (2) the existing 'defaults missing fields' test (line 835) updated to verify `fullMarkdown` is `undefined` when absent. Existing tests still pass (toEqual treats undefined as missing), but the new feature path is untested. `[fixable]`

server/chat.ts
```diff
@@ -851,6 +851,7 @@ async function _startChatInner(
```
dimakis, line 851: 🔴 **bugs**: `obj.result` does not exist on the contexgin `compile()` return value. The compiled object has `{ sources, included, trimmed, bootTokens }` — there is no `result` field. `fullMarkdown` will always be `undefined`, making the modal button never render. The server needs to extract the compiled markdown from wherever contexgin actually stores it (possibly `obj.compiled` or via a separate method). `[fixable]`

server/chat.ts
```diff
@@ -861,6 +862,7 @@ async function _startChatInner(
         sources,
         included,
         trimmed,
```
dimakis, line 864: 🟡 **regressions**: The fallback/error `boot_context` messages (lines 792-801 and 864-873) do not include `fullMarkdown`. While not a functional regression since the field is optional, it's inconsistent — the protocol parser handles the missing field gracefully, but any future code expecting symmetry between success and fallback payloads would need to account for this. `[fixable]`

frontend/src/components/BootContextPill.tsx
```diff
@@ -53,59 +54,108 @@ export function BootContextPill({ context }: Props) {
...
+                  {KIND_LABELS[src.kind] ?? src.kind}
                 </span>
-              </button>
-              {showTrimmed &&
-                context.trimmed.map((section, idx) => (
-                  <SectionRow key={idx} section={section} dimmed />
+                <span className="boot-context-pill-source-path">{src.path}</span>
+              </div>
+            ))}
+
```
dimakis, line 100: 🟡 **unsafe_assumptions**: The modal renders `context.fullMarkdown` in a raw `<pre>` tag. If the markdown content is very large (contexgin compiles from multiple sources), this could cause performance issues on mobile — the entire string is rendered at once with no virtualization. Consider whether the content could realistically be large enough to matter (e.g., 50k+ chars). `[fixable]`

frontend/src/components/BootContextPill.tsx
```diff
@@ -53,59 +54,108 @@ export function BootContextPill({ context }: Props) {
...
-          {context.sources.map((src, idx) => (
-            <div key={idx} className="boot-context-pill-source-row">
-              <span className={`boot-context-pill-kind boot-context-pill-kind--${src.kind}`}>
-                {KIND_LABELS[src.kind] ?? src.kind}
-              </span>
-              <span className="boot-context-pill-source-path">{src.path}</span>
-            </div>
-          ))}
+  // Escape key handler for modal
+  useEffect(() => {
```
dimakis, line 58: 🔵 **style**: The `useEffect` for Escape key handling is a common pattern that could use `useCallback` for the handler, but more importantly: the effect runs and cleans up on every `showModal` toggle. This is fine functionally but adding the listener only when the modal opens (which it does via the `if (!showModal) return` guard) is good practice — no issue here, just confirming the pattern is correct.

frontend/src/styles/global.css
```diff
@@ -2399,6 +2399,16 @@ textarea:focus {
   border-radius: 6px;
   overflow: hidden;
   margin-bottom: 0.3rem;
```
dimakis, line 2401: 🟡 **bugs**: The `position: sticky; top: 0` on `.boot-context-pill` is inside a flex container (`.chat-messages` has `display: flex; flex-direction: column`). Sticky positioning inside a flex column works, but the pill will only stick within its own flex item bounds — once the scroll passes the pill's original position plus its height, it scrolls away. Since the pill is the first child and has no siblings wrapping it, sticky will work as long as the scroll container itself has `overflow-y: auto` (which `.chat-messages` does). However, `background: var(--bg-primary)` may not fully cover the underlying content if the scroll container has padding (`padding: var(--space-4) var(--space-3)`), leaving a visible gap at the edges. `[fixable]`

frontend/src/styles/global.css
```diff
@@ -2597,6 +2607,96 @@ textarea:focus {
   opacity: 0.7;
 }
```
dimakis, line 2612: 🔵 **style**: `border-left: 1px solid rgba(255, 255, 255, 0.1)` on `.boot-context-pill-view-full` uses a hardcoded white-based color instead of a CSS variable. Other borders in this file use `rgba(255, 255, 255, 0.1)` too, so this is consistent with existing style — but it will look wrong on a light theme if one is ever added. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_The `obj.result` field used to extract `fullMarkdown` on the server does not exist on the contexgin compile output, so the modal button will never render — this is the critical blocker. The frontend implementation (sticky + modal) is solid otherwise, but needs server-side fix and component tests._

- 🔴 **bugs** (L851): `obj.result` does not exist on the contexgin `compile()` return value. The compiled object has `{ sources, included, trimmed, bootTokens }` — there is no `result` field. `fullMarkdown` will always be `undefined`, making the modal button never render. The server needs to extract the compiled markdown from wherever contexgin actually stores it (possibly `obj.compiled` or via a separate method). `[fixable]`
- 🟡 **regressions** (L864): The fallback/error `boot_context` messages (lines 792-801 and 864-873) do not include `fullMarkdown`. While not a functional regression since the field is optional, it's inconsistent — the protocol parser handles the missing field gracefully, but any future code expecting symmetry between success and fallback payloads would need to account for this. `[fixable]`


### `frontend/src/components/BootContextPill.tsx`
_The `obj.result` field used to extract `fullMarkdown` on the server does not exist on the contexgin compile output, so the modal button will never render — this is the critical blocker. The frontend implementation (sticky + modal) is solid otherwise, but needs server-side fix and component tests._

- 🟡 **missing_tests**: No component-level tests exist for BootContextPill. This PR adds significant UI behavior (sticky positioning, modal with Escape key handling, overlay click-to-dismiss, aria attributes) with zero component test coverage. At minimum, tests should verify: modal opens on button click, modal closes on Escape, modal closes on overlay click, modal button is hidden when `fullMarkdown` is absent. `[fixable]`
- 🟡 **unsafe_assumptions** (L100): The modal renders `context.fullMarkdown` in a raw `<pre>` tag. If the markdown content is very large (contexgin compiles from multiple sources), this could cause performance issues on mobile — the entire string is rendered at once with no virtualization. Consider whether the content could realistically be large enough to matter (e.g., 50k+ chars). `[fixable]`
- 🔵 **style** (L58): The `useEffect` for Escape key handling is a common pattern that could use `useCallback` for the handler, but more importantly: the effect runs and cleans up on every `showModal` toggle. This is fine functionally but adding the listener only when the modal opens (which it does via the `if (!showModal) return` guard) is good practice — no issue here, just confirming the pattern is correct.


_The `obj.result` field used to extract `fullMarkdown` on the server does not exist on the contexgin compile output, so the modal button will never render — this is the critical blocker. The frontend implementation (sticky + modal) is solid otherwise, but needs server-side fix and component tests._

- 🟡 **bugs** (L2401): The `position: sticky; top: 0` on `.boot-context-pill` is inside a flex container (`.chat-messages` has `display: flex; flex-direction: column`). Sticky positioning inside a flex column works, but the pill will only stick within its own flex item bounds — once the scroll passes the pill's original position plus its height, it scrolls away. Since the pill is the first child and has no siblings wrapping it, sticky will work as long as the scroll container itself has `overflow-y: auto` (which `.chat-messages` does). However, `background: var(--bg-primary)` may not fully cover the underlying content if the scroll container has padding (`padding: var(--space-4) var(--space-3)`), leaving a visible gap at the edges. `[fixable]`
- 🔵 **style** (L2609): `border-left: 1px solid rgba(255, 255, 255, 0.1)` on `.boot-context-pill-view-full` uses a hardcoded white-based color instead of a CSS variable. Other borders in this file use `rgba(255, 255, 255, 0.1)` too, so this is consistent with existing style — but it will look wrong on a light theme if one is ever added. `[fixable]`

dimakis:
## Centaur Review

Found **4** issue(s).


_Clean implementation of sticky pill + modal view with good defensive parsing, proper accessibility attributes, escape-key dismissal, and solid test coverage. Only minor style nits — no bugs or regressions._

- 🔵 **style** (L2398): Duplicate `width` declaration — `width: 100%` on line 2398 is overridden by `width: calc(100% + 2 * var(--space-3))` on line 2408. The first one is now dead code and should be removed. `[fixable]`
- 🔵 **style** (L2414): `gap: 0` is the default value for flexbox and is redundant. Remove or replace with a meaningful value. `[fixable]`


_Clean implementation of sticky pill + modal view with good defensive parsing, proper accessibility attributes, escape-key dismissal, and solid test coverage. Only minor style nits — no bugs or regressions._

- 🔵 **missing_tests** (L748): The first 'rich source metadata' test uses strict `toEqual` but omits `fullMarkdown` from the expected object, while the 'defaults missing fields' test at line 842 explicitly asserts `fullMarkdown: undefined`. Both pass (Vitest treats undefined-valued properties as absent in `toEqual`), but the inconsistency could confuse a future reader. Consider adding `fullMarkdown: undefined` to the first test's expected output for consistency. `[fixable]`


### `frontend/src/components/BootContextPill.tsx`
_Clean implementation of sticky pill + modal view with good defensive parsing, proper accessibility attributes, escape-key dismissal, and solid test coverage. Only minor style nits — no bugs or regressions._

- 🔵 **unsafe_assumptions** (L136): The modal is rendered inside the `.chat-messages` scroll container (via the Fragment). While `position: fixed` on the overlay generally escapes scroll containers, on older iOS Safari with `-webkit-overflow-scrolling: touch` it can misbehave. If iOS rendering issues surface, consider using a React portal to render the modal at `document.body`. `[fixable]`

server/chat.ts
dimakis, line 851: 🟡 **unsafe_assumptions**: `obj.bootPayload` is read from the ContexGin compile result, but no other code in the repo references this property name. Other fields on `obj` are `sources`, `included`, `trimmed`, `bootTokens`. If ContexGin uses a different name (e.g. `payload`, `markdown`, `compiledMarkdown`), `fullMarkdown` will silently always be `undefined`. The `typeof` guard makes this safe but possibly dead code — verify the property name against ContexGin's actual output. `[fixable]`

frontend/src/styles/global.css
```diff
@@ -2399,6 +2399,19 @@ textarea:focus {
...
   overflow: hidden;
   margin-bottom: 0.3rem;
+  position: sticky;
+  top: 0;
+  z-index: 10;
+  background: var(--bg-primary);
+  margin-inline: calc(-1 * var(--space-3));
+  padding-inline: var(--space-3);
+  width: calc(100% + 2 * var(--space-3));
+}
```
dimakis, line 2409: 🔵 **style**: `gap: 0` on `.boot-context-pill-header-wrapper` is the CSS default and can be removed. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Clean implementation: sticky positioning is correct (parent has overflow-y: auto), modal has proper accessibility attrs and Escape handling, protocol changes are defensively typed with good test coverage. Main concern is verifying that ContexGin actually outputs a `bootPayload` property — otherwise the fullMarkdown feature is silently inert._

- 🟡 **unsafe_assumptions** (L851): `obj.bootPayload` is read from the ContexGin compile result, but no other code in the repo references this property name. Other fields on `obj` are `sources`, `included`, `trimmed`, `bootTokens`. If ContexGin uses a different name (e.g. `payload`, `markdown`, `compiledMarkdown`), `fullMarkdown` will silently always be `undefined`. The `typeof` guard makes this safe but possibly dead code — verify the property name against ContexGin's actual output. `[fixable]`


_Clean implementation: sticky positioning is correct (parent has overflow-y: auto), modal has proper accessibility attrs and Escape handling, protocol changes are defensively typed with good test coverage. Main concern is verifying that ContexGin actually outputs a `bootPayload` property — otherwise the fullMarkdown feature is silently inert._

- 🔵 **missing_tests**: Protocol-level tests for `fullMarkdown` parsing are solid (present, missing, non-string). However, there are no component-level tests for the new modal behavior (open/close, Escape key, overlay click dismiss). This is consistent with the existing codebase which has no BootContextPill render tests, but the modal adds meaningful interactive logic that could regress.


_Clean implementation: sticky positioning is correct (parent has overflow-y: auto), modal has proper accessibility attrs and Escape handling, protocol changes are defensively typed with good test coverage. Main concern is verifying that ContexGin actually outputs a `bootPayload` property — otherwise the fullMarkdown feature is silently inert._

- 🔵 **style** (L2409): `gap: 0` on `.boot-context-pill-header-wrapper` is the CSS default and can be removed. `[fixable]`

PR #342, opened 2026-05-18
fix(ws): preserve model selection across interrupt/resume

dimakis:
## Problem

The model dropdown selection was not being respected after an interrupt/resume cycle. When a session was interrupted (stop generation or send while running) and needed to resume via `startChat()`, the model parameter was completely dropped, causing the SDK to fall back to its default model (`claude-sonnet-4-5`) instead of using the user's selected model (e.g., `claude-opus-4-6`).

This created a confusing UX where:
1. User selects "Opus 4.6" in the dropdown
2. Session starts correctly with Opus 4.6
3. User interrupts (stops mid-generation or sends while running)
4. Session resumes **silently switching to Sonnet 4.5**
5. Model self-reports as Sonnet despite UI showing Opus

Additionally, the active session path (`sendToChat()`) completely ignored model changes, so changing the dropdown had no effect until the session was fully stopped and restarted.

## Root Cause

**Bug 1: Interrupt resume drops model**
In `server/ws-handler-v2.ts` line 564, the interrupt_resume `startChat()` call did not pass the `model` parameter:

```typescript
startChat(transport, sessionClientId, msg.prompt, {
  resume: msg.sessionId,
  images: msg.images,
  contextBlocks: msg.contextBlocks,
  clientMsgId: msg.clientMsgId,
  // model: ← MISSING
});
```

**Bug 2: Active session ignores model changes**
The `sendToChat()` path (used when pushing prompts to a running session) has no model parameter at all — it just pushes to the existing SDK query. Model changes in the dropdown are silently ignored for the rest of that session.

### Protocol layer
- Added `model` field to `V2InterruptMessage` schema in `packages/protocol/src/ws-schemas-v2.ts`

### Client layer
- Send `model` in interrupt WS messages (`packages/client/src/store.ts`)
- Pass `modelState` to interrupt calls in `ChatView.tsx` and `DesktopChatView.tsx`

### Server layer
- Added `model` field to `ManagedSession` interface (`packages/harness/src/session-registry.ts`)
- Store `options.model` on session at start (`server/chat.ts`)
- Pass model in interrupt_resume path with fallback chain: `msg.model ?? found.session?.model` (`server/ws-handler-v2.ts`)

The fallback ensures that even if the message doesn't include a model (e.g., from an older client or edge case), we use the model from the previous session's state rather than silently dropping to the SDK default.

## Testing

- Type-checked with `tsc -b`
- All 48 existing tests pass
- Manual testing confirmed model selection is now preserved across interrupt/resume cycles

## Note on Bug 2

The active session model-change issue (Bug 2) is a **separate, known limitation** — the SDK doesn't support changing models mid-conversation. This PR doesn't fix that (would require restarting the SDK query), but it does ensure the model stays consistent when a session is interrupted and resumed, which is the main UX blocker.

Future work could grey out the dropdown during active sessions to make this limitation explicit.

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -178,7 +178,7 @@ export function ChatView() {
 
   function handleInterrupt(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): void {
     voice.stopSpeaking();
-    storeInterruptMessage(text, { images, contextBlocks: ctxBlocks });
+    storeInterruptMessage(text, { images, contextBlocks: ctxBlocks, model: modelState });
     forceScrollToBottom();
   }
 
--- a/frontend/src/pages/DesktopChatView.tsx
+++ b/frontend/src/pages/DesktopChatView.tsx
@@ -142,7 +142,7 @@ export function DesktopChatView() {
 
   function handleInterrupt(text: string, images?: ImageAttachment[], ctxBlocks?: string[]): void {
     voice.stopSpeaking();
-    storeInterruptMessage(text, { images, contextBlocks: ctxBlocks });
+    storeInterruptMessage(text, { images, contextBlocks: ctxBlocks, model: modelState });
     forceScrollToBottom();
   }
 
--- a/packages/client/__tests__/store.test.ts
+++ b/packages/client/__tests__/store.test.ts
@@ -721,6 +721,58 @@ describe('sendMessage — session expired recovery', () => {
   });
 });
 
+describe('interruptMessage', () => {
+  it('includes model from opts when provided', () => {
+    const store = createReadyStore();
+    store.getState().sendMessage('first');
+    lastWs.simulateMessage({ type: 'session_id', sessionId: 'sess-int' });
+
+    store.getState().interruptMessage('urgent', { model: 'claude-opus-4-6' });
+
+    const interrupts = lastWs.parsedSent().filter((m) => m.type === 'interrupt');
+    expect(interrupts).toHaveLength(1);
+    expect(interrupts[0].model).toBe('claude-opus-4-6');
+  });
+
+  it('falls back to config.modelId when opts.model is not provided', () => {
+    const store = createReadyStore();
+    store.getState().setModel('claude-sonnet-4-6');
+    store.getState().sendMessage('first');
+    lastWs.simulateMessage({ type: 'session_id', sessionId: 'sess-int2' });
+
+    store.getState().interruptMessage('urgent');
+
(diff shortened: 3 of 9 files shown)


dimakis:
## Centaur Review

_Correct approach for preserving model across interrupt-resume, but the active-interrupt path (line 555) doesn't propagate the model, creating an inconsistency where model changes only stick when the session is idle/detached. Tests need updating to cover the new model forwarding._

- 🟡 **missing_tests**: Existing interrupt-resume tests (lines 725-755, 1996-2031) use `expect.objectContaining({ resume: 'sess-1' })` and don't verify the `model` field. Need tests for: (1) `msg.model` is forwarded to `startChat` on resume, (2) `found.session.model` is used as fallback when `msg.model` is absent, (3) `msg.model` takes precedence over `found.session.model`. `[fixable]`


_Correct approach for preserving model across interrupt-resume, but the active-interrupt path (line 555) doesn't propagate the model, creating an inconsistency where model changes only stick when the session is idle/detached. Tests need updating to cover the new model forwarding._

- 🟡 **missing_tests**: No tests verify that `interruptMessage()` includes `model` from opts or falls back to `config.modelId`. The `sendMessage` path likely has model tests that should be mirrored for the interrupt path. `[fixable]`


### `server/ws-handler-v2.ts`
_Correct approach for preserving model across interrupt-resume, but the active-interrupt path (line 555) doesn't propagate the model, creating an inconsistency where model changes only stick when the session is idle/detached. Tests need updating to cover the new model forwarding._

- 🟡 **bugs** (L555): Active-session interrupts (line 555) call `interruptChat(activeClientId, msg.prompt, msg.images, msg.contextBlocks, msg.clientMsgId)` which does not accept or forward the `model` parameter. If a user changes the model and interrupts an *active* session, the model change is silently dropped — the session continues with the original model. The `session.model` field also isn't updated in this path, so a subsequent resume would still use the old model. This is a partial fix: model is only preserved on the resume path, not on active interrupts. `[fixable]`


### `server/chat.ts`
_Correct approach for preserving model across interrupt-resume, but the active-interrupt path (line 555) doesn't propagate the model, creating an inconsistency where model changes only stick when the session is idle/detached. Tests need updating to cover the new model forwarding._

- 🔵 **style** (L679): `session.model = options.model` unconditionally overwrites with `undefined` when `options.model` isn't provided (e.g., non-interrupt `startChat` calls or calls that don't pass model). This is currently harmless because the only consumer (`found.session?.model`) tolerates `undefined`, but `session.model = options.model ?? session.model` would be more defensive for future callers that may rely on the stored value persisting. `[fixable]`

PR #343, opened 2026-05-19
feat(ui): syntax highlighting + unified pill pattern + pop-out

dimakis:
### Commit 1: Syntax highlighting infrastructure
- **CodeBlock component** — reusable syntax-highlighted code viewer using highlight.js core with 17 registered languages and `github-dark-dimmed` theme
- **Tool pill upgrades** — Read results, Write inputs, Edit diffs, and Bash commands all render with syntax highlighting
- **Chat code blocks** — wired `rehype-highlight` into TextBubble for fenced code blocks
- **Protocol layer** — extended `RawToolInput` with `'read'` type and `language` field; new `languageFromPath()` util (50+ extensions)
- **Worktree fix** — `@mitzo/protocol` path aliases for correct resolution

### Commit 2: Unified pill pattern + pop-out
- **SubagentCard → pill** — now renders as `tool-pill--agent` with green left border, tool count badge, k-formatted tokens, text blocks rendered as markdown
- **ThinkingBlock → pill** — same dot/chevron/expand pattern, italic label, dimmed done dot
- **Pop-out button** — `↗` in CodeBlock headers and tool pill details, navigates to `/files?path=` file viewer for Read/Write/Edit tools
- **CSS cleanup** — removed old `subagent-*` and `thinking-block-*` classes, all blocks now use unified `tool-pill` pattern

## Architecture

Adding a new language: one entry in `language.ts` + one `registerLanguage` in `CodeBlock.tsx`.

## Test plan

- [x] Protocol tests pass (196/196)
- [x] Server tests pass (23/23)
- [x] Frontend type-checks cleanly
- [x] Vite production build succeeds
- [ ] Visual: expand Read/Write/Edit/Bash pills → syntax-highlighted code
- [ ] Visual: fenced code blocks in chat have syntax colors
- [ ] Visual: Agent pills show green left border + tool count badge
- [ ] Visual: Thinking blocks use pill pattern with italic label
- [ ] Visual: pop-out `↗` navigates to file viewer

### DIFF
--- a/frontend/src/components/CodeBlock.tsx
+++ b/frontend/src/components/CodeBlock.tsx
@@ -166,5 +166,6 @@ function escapeHtml(text: string): string {
     .replace(/&/g, '&amp;')
     .replace(/</g, '&lt;')
     .replace(/>/g, '&gt;')
-    .replace(/"/g, '&quot;');
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
 }
--- a/frontend/src/components/CodeBlock.tsx
+++ b/frontend/src/components/CodeBlock.tsx
@@ -154,10 +154,7 @@ export function CodeBlock({
           )}
         </div>
       )}
-      <pre
-        className="code-block-highlight-pre hljs"
-        style={{ maxHeight: `${maxHeight}px` }}
-      >
+      <pre className="code-block-highlight-pre hljs" style={{ maxHeight: `${maxHeight}px` }}>
         <code dangerouslySetInnerHTML={{ __html: highlighted }} />
       </pre>
     </div>
--- a/frontend/src/components/CodeBlock.tsx
+++ b/frontend/src/components/CodeBlock.tsx
@@ -1,45 +1,77 @@
 import { useMemo } from 'react';
 import hljs from 'highlight.js/lib/core';
+import { CopyButton } from './CopyButton';
 
 // Register languages on demand — add new ones here.
 // Each import adds ~2-10 KB gzipped; tree-shaking drops unused ones.
-import javascript from 'highlight.js/lib/languages/javascript';
-import typescript from 'highlight.js/lib/languages/typescript';
-import python from 'highlight.js/lib/languages/python';
+// IMPORTANT: Keep in sync with EXT_MAP in packages/protocol/src/language.ts
 import bash from 'highlight.js/lib/languages/bash';
-import json from 'highlight.js/lib/languages/json';
-import yaml from 'highlight.js/lib/languages/yaml';
-import xml from 'highlight.js/lib/languages/xml';
+import c from 'highlight.js/lib/languages/c';
+import cmake from 'highlight.js/lib/languages/cmake';
+import cpp from 'highlight.js/lib/languages/cpp';
+import csharp from 'highlight.js/lib/languages/csharp';
 import css from 'highlight.js/lib/languages/css';
-import sql from 'highlight.js/lib/languages/sql';
-import rust from 'highlight.js/lib/languages/rust';
-import go from 'highlight.js/lib/languages/go';
-import java from 'highlight.js/lib/languages/java';
-import swift from 'highlight.js/lib/languages/swift';
 import diff from 'highlight.js/lib/languages/diff';
-import markdown from 'highlight.js/lib/languages/markdown';
+import dockerfile from 'highlight.js/lib/languages/dockerfile';
+import go from 'highlight.js/lib/languages/go';
+import graphql from 'highlight.js/lib/languages/graphql';
 import ini from 'highlight.js/lib/languages/ini';
+import java from 'highlight.js/lib/languages/java';
+import javascript from 'highlight.js/lib/languages/javascript';
+import json from 'highlight.js/lib/languages/json';
+import kotlin from 'highlight.js/lib/languages/kotlin';
+import latex from 'highlight.js/lib/languages/latex';
+import less from 'highlight.js/lib/languages/less';
+import lua from 'highlight.js/lib/languages/lua';
+import makefile from 'highlight.js/lib/languages/makefile';
+import markdown from 'highlight.js/lib/languages/markdown';
+import perl from 'highlight.js/lib/languages/perl';
+import php from 'highlight.js/lib/languages/php';
+import protobuf from 'highlight.js/lib/languages/protobuf';
+import python from 'highlight.js/lib/languages/python';
+import r from 'highlight.js/lib/languages/r';
+import ruby from 'highlight.js/lib/languages/ruby';
+import rust from 'highlight.js/lib/languages/rust';
 import scss from 'highlight.js/lib/languages/scss';
+import sql from 'highlight.js/lib/languages/sql';
+import swift from 'highlight.js/lib/languages/swift';
+import typescript from 'highlight.js/lib/languages/typescript';
+import xml from 'highlight.js/lib/languages/xml';
+import yaml from 'highlight.js/lib/languages/yaml';
 
-hljs.registerLanguage('javascript', javascript);
-hljs.registerLanguage('typescript', typescript);
-hljs.registerLanguage('python', python);
 hljs.registerLanguage('bash', bash);
-hljs.registerLanguage('json', json);
-hljs.registerLanguage('yaml', yaml);
-hljs.registerLanguage('xml', xml);
+hljs.registerLanguage('c', c);
+hljs.registerLanguage('cmake', cmake);
+hljs.registerLanguage('cpp', cpp);
+hljs.registerLanguage('csharp', csharp);
 hljs.registerLanguage('css', css);
-hljs.registerLanguage('sql', sql);
-hljs.registerLanguage('rust', rust);
-hljs.registerLanguage('go', go);
-hljs.registerLanguage('java', java);
-hljs.registerLanguage('swift', swift);
 hljs.registerLanguage('diff', diff);
-hljs.registerLanguage('markdown', markdown);
+hljs.registerLanguage('dockerfile', dockerfile);
+hljs.registerLanguage('go', go);
+hljs.registerLanguage('graphql', graphql);
 hljs.registerLanguage('ini', ini);
+hljs.registerLanguage('java', java);
+hljs.registerLanguage('javascript', javascript);
+hljs.registerLanguage('json', json);
+hljs.registerLanguage('kotlin', kotlin);
+hljs.registerLanguage('latex', latex);
+hljs.registerLanguage('less', less);
+hljs.registerLanguage('lua', lua);
+hljs.registerLanguage('makefile', makefile);
+hljs.registerLanguage('markdown', markdown);
+hljs.registerLanguage('perl', perl);
+hljs.registerLanguage('php', php);
+hljs.registerLanguage('protobuf', protobuf);
+hljs.registerLanguage('python', python);
+hljs.registerLanguage('r', r);
+hljs.registerLanguage('ruby', ruby);
+hljs.registerLanguage('rust', rust);
 hljs.registerLanguage('scss', scss);
-
-import { CopyButton } from './CopyButton';
+hljs.registerLanguage('sql', sql);
+hljs.registerLanguage('swift', swift);
+hljs.registerLanguage('typescript', typescript);
+hljs.registerLanguage('xml', xml);
+hljs.registerLanguage('yaml', yaml);
 
 interface CodeBlockProps {
   /** The source code to display. */
--- a/frontend/src/components/CodeBlock.tsx
+++ b/frontend/src/components/CodeBlock.tsx
@@ -1,0 +1,114 @@
+import { useMemo } from 'react';
+import hljs from 'highlight.js/lib/core';
+
+// Register languages on demand — add new ones here.
+// Each import adds ~2-10 KB gzipped; tree-shaking drops unused ones.
+import javascript from 'highlight.js/lib/languages/javascript';
+import typescript from 'highlight.js/lib/languages/typescript';
+import python from 'highlight.js/lib/languages/python';
+import bash from 'highlight.js/lib/languages/bash';
+import json from 'highlight.js/lib/languages/json';
+import yaml from 'highlight.js/lib/languages/yaml';
+import xml from 'highlight.js/lib/languages/xml';
+import css from 'highlight.js/lib/languages/css';
+import sql from 'highlight.js/lib/languages/sql';
+import rust from 'highlight.js/lib/languages/rust';
+import go from 'highlight.js/lib/languages/go';
+import java from 'highlight.js/lib/languages/java';
+import swift from 'highlight.js/lib/languages/swift';
+import diff from 'highlight.js/lib/languages/diff';
+import markdown from 'highlight.js/lib/languages/markdown';
+import ini from 'highlight.js/lib/languages/ini';
+import scss from 'highlight.js/lib/languages/scss';
+
+hljs.registerLanguage('javascript', javascript);
+hljs.registerLanguage('typescript', typescript);
+hljs.registerLanguage('python', python);
+hljs.registerLanguage('bash', bash);
+hljs.registerLanguage('json', json);
+hljs.registerLanguage('yaml', yaml);
+hljs.registerLanguage('xml', xml);
+hljs.registerLanguage('css', css);
+hljs.registerLanguage('sql', sql);
+hljs.registerLanguage('rust', rust);
+hljs.registerLanguage('go', go);
+hljs.registerLanguage('java', java);
+hljs.registerLanguage('swift', swift);
+hljs.registerLanguage('diff', diff);
+hljs.registerLanguage('markdown', markdown);
+hljs.registerLanguage('ini', ini);
+hljs.registerLanguage('scss', scss);
+
+import { CopyButton } from './CopyButton';
+
+interface CodeBlockProps {
+  /** The source code to display. */
+  code: string;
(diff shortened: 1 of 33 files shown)


packages/protocol/src/language.ts
```diff
@@ -0,0 +1,109 @@
+// Maps file extensions to highlight.js language identifiers.
+// To add a new language, add an entry here — no other changes needed.
```
dimakis: 🟡 **bugs**: EXT_MAP comment says "To add a new language, add an entry here — no other changes needed" but CodeBlock.tsx has a separate language registration list. Languages like c, cpp, kotlin, csharp, ruby, lua, perl, php, r, latex, hcl, dockerfile, graphql, protobuf, makefile, cmake, and less are in EXT_MAP but NOT registered in CodeBlock.tsx. Files with those extensions will get a `language` prop set but hljs won't recognise it, silently falling back to plain text. The comment is misleading and the two lists will diverge further over time. `[fixable]`

packages/protocol/src/language.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  // Infrastructure
+  tf: 'hcl',
+  hcl: 'hcl',
+  Dockerfile: 'dockerfile',
+  dockerfile: 'dockerfile',
+
+  // SQL
+  sql: 'sql',
+
```
dimakis, line 77: 🔵 **style**: Inconsistent basename handling: `Dockerfile` has both case variants (`Dockerfile` and `dockerfile`) as keys, but `makefile` only has the lowercase variant. Both work due to the lowerBasename fallback, but the inconsistency makes the map harder to reason about. Consider using only lowercase keys and relying on the lowerBasename check uniformly. `[fixable]`

frontend/src/components/CodeBlock.tsx
```diff
@@ -0,0 +1,114 @@
...
+hljs.registerLanguage('rust', rust);
+hljs.registerLanguage('go', go);
+hljs.registerLanguage('java', java);
+hljs.registerLanguage('swift', swift);
+hljs.registerLanguage('diff', diff);
+hljs.registerLanguage('markdown', markdown);
+hljs.registerLanguage('ini', ini);
+hljs.registerLanguage('scss', scss);
+
+import { CopyButton } from './CopyButton';
```
dimakis: 🔵 **style**: The `CopyButton` import is placed after the hljs.registerLanguage() calls (line 42) instead of being grouped with the other imports at the top of the file. Move it up with the other component imports for conventional ordering. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/package.json`
_Well-structured syntax highlighting feature with good test coverage on the protocol and model-forwarding changes, but highlight.js should be an explicit dependency and the language registration gap between EXT_MAP (~35 languages) and CodeBlock.tsx (17 registered) will silently fail for many file types._

- 🟡 **unsafe_assumptions**: highlight.js is imported directly in CodeBlock.tsx and main.tsx (CSS theme) but is not listed as a direct dependency — it's only a transitive dependency of rehype-highlight. If rehype-highlight changes its internals or npm deduplication moves the package, these imports will break. Add `highlight.js` as an explicit dependency. `[fixable]`


### `packages/protocol/src/language.ts`
_Well-structured syntax highlighting feature with good test coverage on the protocol and model-forwarding changes, but highlight.js should be an explicit dependency and the language registration gap between EXT_MAP (~35 languages) and CodeBlock.tsx (17 registered) will silently fail for many file types._

- 🟡 **bugs** (L2): EXT_MAP comment says "To add a new language, add an entry here — no other changes needed" but CodeBlock.tsx has a separate language registration list. Languages like c, cpp, kotlin, csharp, ruby, lua, perl, php, r, latex, hcl, dockerfile, graphql, protobuf, makefile, cmake, and less are in EXT_MAP but NOT registered in CodeBlock.tsx. Files with those extensions will get a `language` prop set but hljs won't recognise it, silently falling back to plain text. The comment is misleading and the two lists will diverge further over time. `[fixable]`
- 🔵 **style** (L80): Inconsistent basename handling: `Dockerfile` has both case variants (`Dockerfile` and `dockerfile`) as keys, but `makefile` only has the lowercase variant. Both work due to the lowerBasename fallback, but the inconsistency makes the map harder to reason about. Consider using only lowercase keys and relying on the lowerBasename check uniformly. `[fixable]`


### `frontend/src/components/CodeBlock.tsx`
_Well-structured syntax highlighting feature with good test coverage on the protocol and model-forwarding changes, but highlight.js should be an explicit dependency and the language registration gap between EXT_MAP (~35 languages) and CodeBlock.tsx (17 registered) will silently fail for many file types._

- 🟡 **missing_tests**: No tests for the CodeBlock component. The project requires TDD (CLAUDE.md). At minimum, test that: (1) known languages produce highlighted output, (2) unknown/unregistered languages fall back to escaped plain text, (3) the escapeHtml fallback handles all relevant characters, (4) empty code returns empty string. `[fixable]`
- 🔵 **style** (L42): The `CopyButton` import is placed after the hljs.registerLanguage() calls (line 42) instead of being grouped with the other imports at the top of the file. Move it up with the other component imports for conventional ordering. `[fixable]`


### `frontend/src/components/ToolPill.tsx`
_Well-structured syntax highlighting feature with good test coverage on the protocol and model-forwarding changes, but highlight.js should be an explicit dependency and the language registration gap between EXT_MAP (~35 languages) and CodeBlock.tsx (17 registered) will silently fail for many file types._

- 🔵 **missing_tests**: The new ToolResult component and the updated RawInputDetail (especially the new 'read' type early-return) have no unit tests. The ToolResult rendering logic — which conditionally applies syntax highlighting for read results vs plain text for others — would benefit from test coverage. `[fixable]`

dimakis:
## Centaur Review

Found **8** issue(s) (2 warning).


### `server/index.ts`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🟡 **regressions** (L781): The v1 WebSocket interrupt handler does not forward the `model` parameter to `interruptChat()`. The v1 path is still active (used when clients skip the `hello` handshake). While v1 clients likely don't send `model`, the omission means model context is silently lost if a v1 interrupt resumes a session that had a model set — the 6th arg defaults to `undefined` so `session.model` won't be updated. Not a breaking change (parameter is optional), but inconsistent with the v2 path. `[fixable]`


### `frontend/src/components/ToolPill.tsx`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🟡 **bugs** (L85): `ToolResult` renders nothing when `block.toolResult` is `undefined`, but it also renders nothing for falsy string values like `""` (empty string). The condition `if (raw?.type === 'read' && block.toolResult)` skips empty-string results, falling through to the generic `<pre>` branch which also renders an empty string. Minor: the old code always rendered the "Result" section when `done` was true, giving a visual cue that the tool finished even with empty output. Now an empty-result tool shows no result section at all. `[fixable]`


### `frontend/src/components/__tests__/ToolPill.test.tsx`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🔵 **missing_tests**: No test coverage for the new `ToolResult` component or the `CodeBlock` integration within `ToolPill`. Specifically: no test for the 'read' type rendering path (read tool with result showing syntax-highlighted CodeBlock), no test for the 'diff' type rendering with old/new CodeBlocks, and no test for the 'command' type rendering. The existing tests only verify pop-out button presence. `[fixable]`


### `frontend/src/components/CodeBlock.tsx`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🔵 **missing_tests**: CodeBlock is a new component with non-trivial logic (language detection, HTML escaping fallback, variant styling, conditional header/float-actions rendering) but has no dedicated unit tests. At minimum, tests should cover: rendering with/without a language, the escapeHtml fallback path, variant classes, and the pop-out button callback. `[fixable]`
- 🔵 **style** (L138): The `escapeHtml` fallback does not escape single quotes (`'` → `&#39;`). While not exploitable in this specific context (output is placed inside a `<code>` element, not an attribute), it's a defense-in-depth gap. Standard HTML escaping includes single quotes. `[fixable]`


### `packages/protocol/__tests__/language.test.ts`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🔵 **missing_tests**: No test for `Makefile` detection. The EXT_MAP has `Dockerfile` with an explicit test, but `Makefile` relies on the `lowerBasename` fallback (`makefile` key) with no test proving it works. Also no tests for other basename-matched entries like `cmake`. `[fixable]`


### `packages/protocol/src/language.ts`
_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🔵 **style** (L75): Inconsistent casing strategy: `Dockerfile` has both `Dockerfile` and `dockerfile` entries for explicit matching, but `Makefile` only has `makefile`. Both work because of the `lowerBasename` fallback, but the asymmetry is confusing. Either rely on the fallback for both or add explicit entries for both. `[fixable]`


_Solid feature PR: syntax highlighting, unified pill pattern, and model forwarding on interrupt are well-implemented with good test coverage. Main concern is the v1 interrupt handler not forwarding the new `model` parameter, and missing unit tests for the new CodeBlock component._

- 🔵 **regressions**: The old `.subagent-card` CSS block is entirely deleted but the comment says "kept for compat". There is no actual CSS kept — just a comment. If any external consumer (e.g. custom CSS overrides, browser extensions) targeted `.subagent-card` classes, they will break silently. The comment is misleading since no compat CSS is retained. `[fixable]`

frontend/src/components/SubagentCard.tsx
```diff
@@ -18,42 +20,47 @@ export function SubagentCard({ subagent }: SubagentCardProps) {
...
-        <span className="subagent-summary">{summary}</span>
-        {usage && <span className="subagent-tokens">{formatTokens(usage)}</span>}
-        <span className="subagent-chevron">{expanded ? '▾' : '▸'}</span>
+        <span className="tool-pill-name">Agent</span>
+        <span className="tool-pill-input">{summary}</span>
+        {toolCount > 0 && <span className="tool-pill-badge">{toolCount}</span>}
+        {usage && <span className="tool-pill-tokens">{formatTokens(usage)}</span>}
+        {!done && <span className="tool-pill-status">Running...</span>}
+        <span className="tool-pill-chevron">{expanded ? '▾' : '▸'}</span>
       </button>
```
dimakis, line 52: 🟡 **regressions**: Subagent text blocks now render through `TextBubble` (ReactMarkdown + rehype-highlight + file-path linkification) instead of a plain `<div>{block.content}</div>`. This is a significant behavior change — markdown formatting, link handling, and navigate hooks are now active in subagent text output. If subagent text ever contains unintended markdown (e.g. `**bold**`, `[link](url)`, backtick fences), it will render differently than before. Worth confirming this is the desired behavior since the PR description doesn't call it out.

frontend/src/components/MessageBubble.tsx
```diff
@@ -106,6 +107,7 @@ export function TextBubble({ content, streaming = false, timestamp, readAloud }:
       <div className="msg-bubble-markdown" ref={contentRef}>
         <ReactMarkdown
           remarkPlugins={[remarkGfm]}
+          rehypePlugins={[rehypeHighlight]}
```
dimakis, line 110: 🔵 **style**: Adding `rehype-highlight` here loads highlight.js languages via lowlight (which includes all common languages by default), while `CodeBlock.tsx` separately does selective manual imports from `highlight.js/lib/core`. Both end up in the bundle. Consider either: (a) using a shared configured lowlight/hljs instance, or (b) passing a custom lowlight to rehype-highlight to share the same selective language set as CodeBlock. `[fixable]`

frontend/src/components/ToolPill.tsx
```diff
@@ -1,46 +1,133 @@
...
-        <pre className="tool-pill-pre tool-pill-code">{raw.command}</pre>
+        <CodeBlock code={raw.command || ''} language="bash" label="Command" maxHeight={200} />
       </div>
     );
   }
   return null;
 }
 
+/** Render tool result — syntax-highlighted for read/write tools, plain for others. */
+function ToolResult({
```
dimakis, line 80: 🔵 **style**: `useNavigate()` and `useLocation()` are called on every ToolPill render, even for tools like Grep or Glob that will never use pop-out. This adds a Router subscription to every pill. Minor perf concern in sessions with many tool calls. Consider moving the navigate logic into the CodeBlock's onPopOut callback or extracting a wrapper that only mounts when expanded. `[fixable]`

packages/protocol/src/language.ts
```diff
@@ -0,0 +1,106 @@
...
+  const lowerBasename = basename.toLowerCase();
+
+  // Check full basename first (Dockerfile, Makefile, etc.)
+  if (EXT_MAP[basename]) return EXT_MAP[basename];
+  if (EXT_MAP[lowerBasename]) return EXT_MAP[lowerBasename];
+
+  // Then check extension
+  const dotIdx = basename.lastIndexOf('.');
+  if (dotIdx < 0) return undefined;
+  const ext = basename.slice(dotIdx + 1);
```
dimakis, line 104: 🔵 **unsafe_assumptions**: Extension lookup is case-sensitive (`EXT_MAP[ext]`), which is correct for case-sensitive filesystems but may miss on case-insensitive ones. For example, `.PY` or `.Ts` won't match. The basename check has a lowercase fallback (line 99) but the extension path doesn't. Consider `EXT_MAP[ext] ?? EXT_MAP[ext.toLowerCase()]` for robustness, especially since file paths come from agent SDK tool calls which may preserve original casing. `[fixable]`

dimakis (commented):
## Centaur Review

### `frontend/src/components/ThinkingBlock.tsx`
_Solid feature PR with good test coverage for new components. Main concerns: ThinkingBlock refactor has zero test coverage, and SubagentCard's switch to TextBubble (markdown rendering) is a subtle behavioral change worth confirming as intentional._

- 🟡 **missing_tests**: ThinkingBlock was significantly refactored — all CSS class names changed from `thinking-block-*` to `tool-pill-*`, the DOM structure changed (button/span hierarchy), and redacted thinking layout changed. There is no ThinkingBlock test file anywhere in the repo. The new class names (e.g. `tool-pill-header`, `tool-pill-dot--pending`, `tool-pill--thinking`) could silently regress without test coverage. `[fixable]`


_Solid feature PR with good test coverage for new components. Main concerns: ThinkingBlock refactor has zero test coverage, and SubagentCard's switch to TextBubble (markdown rendering) is a subtle behavioral change worth confirming as intentional._

- 🟡 **regressions** (L52): Subagent text blocks now render through `TextBubble` (ReactMarkdown + rehype-highlight + file-path linkification) instead of a plain `<div>{block.content}</div>`. This is a significant behavior change — markdown formatting, link handling, and navigate hooks are now active in subagent text output. If subagent text ever contains unintended markdown (e.g. `**bold**`, `[link](url)`, backtick fences), it will render differently than before. Worth confirming this is the desired behavior since the PR description doesn't call it out.


_Solid feature PR with good test coverage for new components. Main concerns: ThinkingBlock refactor has zero test coverage, and SubagentCard's switch to TextBubble (markdown rendering) is a subtle behavioral change worth confirming as intentional._

- 🔵 **style** (L110): Adding `rehype-highlight` here loads highlight.js languages via lowlight (which includes all common languages by default), while `CodeBlock.tsx` separately does selective manual imports from `highlight.js/lib/core`. Both end up in the bundle. Consider either: (a) using a shared configured lowlight/hljs instance, or (b) passing a custom lowlight to rehype-highlight to share the same selective language set as CodeBlock. `[fixable]`


### `frontend/src/components/ToolPill.tsx`
_Solid feature PR with good test coverage for new components. Main concerns: ThinkingBlock refactor has zero test coverage, and SubagentCard's switch to TextBubble (markdown rendering) is a subtle behavioral change worth confirming as intentional._

- 🔵 **style** (L80): `useNavigate()` and `useLocation()` are called on every ToolPill render, even for tools like Grep or Glob that will never use pop-out. This adds a Router subscription to every pill. Minor perf concern in sessions with many tool calls. Consider moving the navigate logic into the CodeBlock's onPopOut callback or extracting a wrapper that only mounts when expanded. `[fixable]`


### `packages/protocol/src/language.ts`
_Solid feature PR with good test coverage for new components. Main concerns: ThinkingBlock refactor has zero test coverage, and SubagentCard's switch to TextBubble (markdown rendering) is a subtle behavioral change worth confirming as intentional._

- 🔵 **unsafe_assumptions** (L104): Extension lookup is case-sensitive (`EXT_MAP[ext]`), which is correct for case-sensitive filesystems but may miss on case-insensitive ones. For example, `.PY` or `.Ts` won't match. The basename check has a lowercase fallback (line 99) but the extension path doesn't. Consider `EXT_MAP[ext] ?? EXT_MAP[ext.toLowerCase()]` for robustness, especially since file paths come from agent SDK tool calls which may preserve original casing. `[fixable]`

PR #345, opened 2026-05-20
feat(ui): share/forward files from messages

dimakis:
## Summary
- Adds a share button (↦) next to file path links in assistant messages
- On mobile: opens native iOS/Android share sheet via Web Share API — user picks the app (email, Slack, Messages, AirDrop, etc.)
- On desktop: falls back to browser file download
- New `GET /api/files/download` server endpoint with `isAllowedPath()` security validation

## Changes
- **server/app.ts**: `GET /api/files/download` — serves files with `Content-Disposition: attachment`
- **frontend/src/lib/share-file.ts**: `shareFile()` utility — fetches file from server, tries native share, falls back to download
- **frontend/src/components/ShareButton.tsx**: Button component with idle/busy/done/error states
- **frontend/src/components/MessageBubble.tsx**: Wraps file path links in a `<span>` group with adjacent share button
- **frontend/src/styles/global.css**: Share button styles — hidden by default, visible on hover (always visible on touch devices)

## Test plan
- [x] Server: 4 new route tests (download, 403, 404, directory rejection) — all pass
- [x] Frontend: 3 new share-file tests (download fallback, native share, error handling) — all pass
- [x] TypeScript: no new type errors
- [x] ESLint: no new lint errors
- [ ] Manual: test share sheet on iOS device
- [ ] Manual: test download fallback on desktop Chrome

### DIFF
--- a/frontend/src/components/ShareButton.tsx
+++ b/frontend/src/components/ShareButton.tsx
@@ -41,7 +41,8 @@ export function ShareButton({ filePath, className }: ShareButtonProps) {
           ? 'Failed'
           : 'Share file';
 
-  const icon = state === 'busy' ? '...' : state === 'done' ? '\u2713' : state === 'error' ? '!' : '\u21A6';
+  const icon =
+    state === 'busy' ? '...' : state === 'done' ? '\u2713' : state === 'error' ? '!' : '\u21A6';
 
   return (
     <button
--- a/frontend/src/components/ShareButton.tsx
+++ b/frontend/src/components/ShareButton.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useRef } from 'react';
 import { shareFile } from '../lib/share-file';
 
 interface ShareButtonProps {
@@ -8,13 +8,15 @@ interface ShareButtonProps {
 
 export function ShareButton({ filePath, className }: ShareButtonProps) {
   const [state, setState] = useState<'idle' | 'busy' | 'done' | 'error'>('idle');
+  const busyRef = useRef(false);
 
   const handleShare = useCallback(
     async (e: React.MouseEvent) => {
       e.preventDefault();
       e.stopPropagation();
-      if (state === 'busy') return;
+      if (busyRef.current) return;
 
+      busyRef.current = true;
       setState('busy');
       try {
         await shareFile(filePath);
@@ -23,9 +25,11 @@ export function ShareButton({ filePath, className }: ShareButtonProps) {
       } catch {
         setState('error');
         setTimeout(() => setState('idle'), 2000);
+      } finally {
+        busyRef.current = false;
       }
     },
-    [filePath, state],
+    [filePath],
   );
 
   const label =
--- a/frontend/src/components/ShareButton.tsx
+++ b/frontend/src/components/ShareButton.tsx
@@ -1,0 +1,53 @@
+import { useState, useCallback } from 'react';
+import { shareFile } from '../lib/share-file';
+
+interface ShareButtonProps {
+  filePath: string;
+  className?: string;
+}
+
+export function ShareButton({ filePath, className }: ShareButtonProps) {
+  const [state, setState] = useState<'idle' | 'busy' | 'done' | 'error'>('idle');
+
+  const handleShare = useCallback(
+    async (e: React.MouseEvent) => {
+      e.preventDefault();
+      e.stopPropagation();
+      if (state === 'busy') return;
+
+      setState('busy');
+      try {
+        await shareFile(filePath);
+        setState('done');
+        setTimeout(() => setState('idle'), 1500);
+      } catch {
+        setState('error');
+        setTimeout(() => setState('idle'), 2000);
+      }
+    },
+    [filePath, state],
+  );
+
+  const label =
+    state === 'busy'
+      ? 'Sharing...'
+      : state === 'done'
+        ? 'Shared'
+        : state === 'error'
+          ? 'Failed'
+          : 'Share file';
+
+  const icon = state === 'busy' ? '...' : state === 'done' ? '\u2713' : state === 'error' ? '!' : '\u21A6';
+
+  return (
+    <button
+      className={`share-btn ${className ?? ''} ${state !== 'idle' ? `share-btn--${state}` : ''}`.trim()}
+      aria-label={label}
+      title={label}
+      onClick={handleShare}
+      disabled={state === 'busy'}
+    >
+      {icon}
+    </button>
+  );
+}
--- a/frontend/src/lib/share-file.ts
+++ b/frontend/src/lib/share-file.ts
@@ -71,7 +71,12 @@ export async function shareFile(filePath: string): Promise<boolean> {
   // Try native share (mobile)
   if (canNativeShare(filename, typedBlob)) {
     const file = new File([typedBlob], filename, { type: mime });
-    await navigator.share({ files: [file] });
+    try {
+      await navigator.share({ files: [file] });
+    } catch (err) {
+      if (err instanceof DOMException && err.name === 'AbortError') return true;
+      throw err;
+    }
     return true;
   }
 
--- a/frontend/src/lib/share-file.ts
+++ b/frontend/src/lib/share-file.ts
@@ -1,0 +1,88 @@
+import { apiFetch } from './api-fetch';
+
+/** MIME types for common file extensions. Falls back to application/octet-stream. */
+function mimeFromExt(filename: string): string {
+  const ext = filename.split('.').pop()?.toLowerCase() ?? '';
+  const map: Record<string, string> = {
+    md: 'text/markdown',
+    txt: 'text/plain',
+    json: 'application/json',
+    yaml: 'application/x-yaml',
+    yml: 'application/x-yaml',
+    csv: 'text/csv',
+    html: 'text/html',
+    css: 'text/css',
+    js: 'text/javascript',
+    ts: 'text/typescript',
+    tsx: 'text/typescript',
+    jsx: 'text/javascript',
+    py: 'text/x-python',
+    sh: 'text/x-shellscript',
+    pdf: 'application/pdf',
+    png: 'image/png',
+    jpg: 'image/jpeg',
+    jpeg: 'image/jpeg',
+    gif: 'image/gif',
+    webp: 'image/webp',
+    svg: 'image/svg+xml',
+  };
+  return map[ext] ?? 'application/octet-stream';
+}
+
+/** Extract just the filename from an absolute or relative path. */
+function filenameFromPath(filePath: string): string {
+  return filePath.split('/').pop() ?? 'file';
+}
+
+/** Download file bytes from the server. */
+async function fetchFileBlob(filePath: string): Promise<{ blob: Blob; filename: string }> {
+  const res = await apiFetch(`/api/files/download?path=${encodeURIComponent(filePath)}`);
+  if (!res.ok) {
+    const body = await res.json().catch(() => ({ error: 'Download failed' }));
+    throw new Error(body.error ?? `Download failed (${res.status})`);
+  }
+  const blob = await res.blob();
+  const filename = filenameFromPath(filePath);
+  return { blob, filename };
+}
+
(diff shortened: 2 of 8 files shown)


frontend/src/styles/global.css
```diff
@@ -1885,6 +1891,47 @@ textarea:focus {
...
 
+.share-btn {
+  background: none;
+  border: none;
+  cursor: pointer;
+  color: var(--text-dim);
+  font-size: 0.75em;
+  padding: 1px 3px;
+  border-radius: 3px;
+  transition: color 0.15s, background 0.15s;
```
dimakis: 🟡 **bugs**: Duplicate `transition` property on `.share-btn`: line 1902 (`transition: color 0.15s, background 0.15s;`) is immediately overwritten by line 1906 (`transition: opacity 0.15s, color 0.15s, background 0.15s;`). The first declaration is dead code — remove it. `[fixable]`

frontend/src/lib/share-file.ts
```diff
@@ -0,0 +1,88 @@
...
+  const { blob, filename } = await fetchFileBlob(filePath);
+
+  // Re-type the blob with a proper MIME if the server sent application/octet-stream
+  const mime = blob.type === 'application/octet-stream' ? mimeFromExt(filename) : blob.type;
+  const typedBlob = mime !== blob.type ? new Blob([blob], { type: mime }) : blob;
+
+  // Try native share (mobile)
+  if (canNativeShare(filename, typedBlob)) {
+    const file = new File([typedBlob], filename, { type: mime });
+    await navigator.share({ files: [file] });
```
dimakis: 🟡 **bugs**: When the user cancels the native share sheet, `navigator.share()` rejects with an `AbortError`. This propagates to `ShareButton`'s catch block, which shows the 'Failed' error state — misleading UX since the user intentionally cancelled. Filter out `AbortError` before throwing, e.g. `catch (err) { if (err instanceof DOMException && err.name === 'AbortError') return true; throw err; }`. `[fixable]`

server/app.ts
```diff
@@ -1327,6 +1327,35 @@ app.get('/api/files/read', (req, res) => {
...
+    res.status(404).json({ error: 'File not found' });
+    return;
+  }
+  try {
+    const stat = statSync(filePath);
+    if (stat.isDirectory()) {
+      res.status(400).json({ error: 'Cannot download a directory' });
+      return;
+    }
+    const filename = basename(filePath);
```
dimakis, line 1346: 🟡 **unsafe_assumptions**: The `Content-Disposition` header interpolates `filename` without escaping quotes. A file named `report".txt` would break the header syntax. Use RFC 5987 encoding (`filename*=UTF-8''${encodeURIComponent(filename)}`) or at minimum escape double quotes in the filename. `[fixable]`

frontend/src/components/ShareButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      try {
+        await shareFile(filePath);
+        setState('done');
+        setTimeout(() => setState('idle'), 1500);
+      } catch {
+        setState('error');
+        setTimeout(() => setState('idle'), 2000);
+      }
+    },
+    [filePath, state],
```
dimakis: 🔵 **style**: `useCallback` lists `state` as a dependency, causing the callback to be recreated on every state change. The `state === 'busy'` guard works correctly, but the dependency is only needed for that guard — a `useRef` for the busy flag would avoid unnecessary re-renders of any children receiving this callback. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid feature with good test coverage on the server side. Two functional issues to address: duplicate CSS `transition` (dead code) and `navigator.share` cancellation treated as an error on mobile._

- 🟡 **bugs** (L1902): Duplicate `transition` property on `.share-btn`: line 1902 (`transition: color 0.15s, background 0.15s;`) is immediately overwritten by line 1906 (`transition: opacity 0.15s, color 0.15s, background 0.15s;`). The first declaration is dead code — remove it. `[fixable]`


### `frontend/src/lib/share-file.ts`
_Solid feature with good test coverage on the server side. Two functional issues to address: duplicate CSS `transition` (dead code) and `navigator.share` cancellation treated as an error on mobile._

- 🟡 **bugs** (L74): When the user cancels the native share sheet, `navigator.share()` rejects with an `AbortError`. This propagates to `ShareButton`'s catch block, which shows the 'Failed' error state — misleading UX since the user intentionally cancelled. Filter out `AbortError` before throwing, e.g. `catch (err) { if (err instanceof DOMException && err.name === 'AbortError') return true; throw err; }`. `[fixable]`


### `server/app.ts`
_Solid feature with good test coverage on the server side. Two functional issues to address: duplicate CSS `transition` (dead code) and `navigator.share` cancellation treated as an error on mobile._

- 🟡 **unsafe_assumptions** (L1346): The `Content-Disposition` header interpolates `filename` without escaping quotes. A file named `report".txt` would break the header syntax. Use RFC 5987 encoding (`filename*=UTF-8''${encodeURIComponent(filename)}`) or at minimum escape double quotes in the filename. `[fixable]`


### `frontend/src/lib/__tests__/share-file.test.ts`
_Solid feature with good test coverage on the server side. Two functional issues to address: duplicate CSS `transition` (dead code) and `navigator.share` cancellation treated as an error on mobile._

- 🔵 **missing_tests**: No test for the user-cancellation path (`navigator.share` rejecting with `AbortError`). This is the most common non-happy-path scenario on mobile and currently triggers the wrong UI state. `[fixable]`


### `frontend/src/components/ShareButton.tsx`
_Solid feature with good test coverage on the server side. Two functional issues to address: duplicate CSS `transition` (dead code) and `navigator.share` cancellation treated as an error on mobile._

- 🔵 **missing_tests**: No component tests for ShareButton. The state machine (idle → busy → done/error → idle via setTimeout) and the event handler guards (stopPropagation, disabled during busy) are logic that would benefit from basic render tests. `[fixable]`
- 🔵 **style** (L28): `useCallback` lists `state` as a dependency, causing the callback to be recreated on every state change. The `state === 'busy'` guard works correctly, but the dependency is only needed for that guard — a `useRef` for the busy flag would avoid unnecessary re-renders of any children receiving this callback. `[fixable]`

server/app.ts
```diff
@@ -1327,6 +1327,36 @@ app.get('/api/files/read', (req, res) => {
...
+    const safeFilename = filename.replace(/"/g, '\\"');
```
dimakis, line 1347: 🟡 **unsafe_assumptions**: Content-Disposition `safeFilename` only escapes double-quotes but not backslashes. In RFC 2616 quoted-strings, both `"` and `\` must be escaped. A file named `back\"slash.txt` would produce a malformed header. The `filename*` fallback (encodeURIComponent) is correct, so modern browsers are fine, but older clients may misbehave. Fix: `.replace(/[\\"/]/g, '\\$&')`. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Solid feature with good test coverage and proper path validation. Two minor correctness issues: Content-Disposition backslash escaping and empty-string fallback in filenameFromPath._

- 🟡 **unsafe_assumptions** (L1347): Content-Disposition `safeFilename` only escapes double-quotes but not backslashes. In RFC 2616 quoted-strings, both `"` and `\` must be escaped. A file named `back\"slash.txt` would produce a malformed header. The `filename*` fallback (encodeURIComponent) is correct, so modern browsers are fine, but older clients may misbehave. Fix: `.replace(/[\\"/]/g, '\\$&')`. `[fixable]`
- 🔵 **bugs** (L1349): TOCTOU gap: `statSync` at line 1341 reads the file size for `Content-Length`, but `sendFile` reads the file asynchronously. If the file is modified between those calls, the client receives a mismatched Content-Length. Additionally, `sendFile` errors (e.g. file deleted between stat and send) are not caught by the synchronous try/catch — they go to Express's default error handler. Low probability but worth noting.


### `frontend/src/lib/share-file.ts`
_Solid feature with good test coverage and proper path validation. Two minor correctness issues: Content-Disposition backslash escaping and empty-string fallback in filenameFromPath._

- 🟡 **bugs** (L34): `filenameFromPath` returns empty string for trailing-slash paths (e.g. `/workspace/dir/`) because `''.split('/').pop()` is `''` and the `??` fallback only triggers on `null`/`undefined`, not empty string. This would produce a File/download with an empty name. Use `|| 'file'` instead of `?? 'file'`, or strip trailing slashes first. `[fixable]`
- 🔵 **style** (L6): The MIME map is recreated on every call to `mimeFromExt`. Since it's a constant, hoist it to module scope to avoid repeated allocation. `[fixable]`


### `frontend/src/lib/__tests__/share-file.test.ts`
_Solid feature with good test coverage and proper path validation. Two minor correctness issues: Content-Disposition backslash escaping and empty-string fallback in filenameFromPath._

- 🔵 **missing_tests**: No unit tests for `mimeFromExt` or `filenameFromPath` helper functions. Edge cases like extensionless files, dotfiles (`.gitignore`), and trailing-slash paths are untested. Consider extracting and testing these utilities. `[fixable]`


_Solid feature with good test coverage and proper path validation. Two minor correctness issues: Content-Disposition backslash escaping and empty-string fallback in filenameFromPath._

- 🔵 **missing_tests** (L663): Download route tests don't cover filenames with special characters (spaces, quotes, unicode) to verify Content-Disposition header correctness. Adding a test with e.g. `'file "name".txt'` would exercise the escaping logic. `[fixable]`

server/app.ts
```diff
@@ -1327,6 +1327,36 @@ app.get('/api/files/read', (req, res) => {
...
+    res.setHeader('Content-Disposition', `attachment; filename="${safeFilename}"; filename*=UTF-8''${encodeURIComponent(filename)}`);
+    res.setHeader('Content-Length', stat.size);
```
dimakis, line 1349: 🔵 **bugs**: TOCTOU gap: `statSync` at line 1341 reads the file size for `Content-Length`, but `sendFile` reads the file asynchronously. If the file is modified between those calls, the client receives a mismatched Content-Length. Additionally, `sendFile` errors (e.g. file deleted between stat and send) are not caught by the synchronous try/catch — they go to Express's default error handler. Low probability but worth noting.

frontend/src/lib/share-file.ts
```diff
@@ -0,0 +1,93 @@
...
+    gif: 'image/gif',
+    webp: 'image/webp',
+    svg: 'image/svg+xml',
+  };
+  return map[ext] ?? 'application/octet-stream';
+}
+
+/** Extract just the filename from an absolute or relative path. */
+function filenameFromPath(filePath: string): string {
+  return filePath.split('/').pop() ?? 'file';
```
dimakis, line 34: 🟡 **bugs**: `filenameFromPath` returns empty string for trailing-slash paths (e.g. `/workspace/dir/`) because `''.split('/').pop()` is `''` and the `??` fallback only triggers on `null`/`undefined`, not empty string. This would produce a File/download with an empty name. Use `|| 'file'` instead of `?? 'file'`, or strip trailing slashes first. `[fixable]`

frontend/src/lib/share-file.ts
```diff
@@ -0,0 +1,93 @@
+import { apiFetch } from './api-fetch';
+
+/** MIME types for common file extensions. Falls back to application/octet-stream. */
+function mimeFromExt(filename: string): string {
+  const ext = filename.split('.').pop()?.toLowerCase() ?? '';
+  const map: Record<string, string> = {
```
dimakis, line 6: 🔵 **style**: The MIME map is recreated on every call to `mimeFromExt`. Since it's a constant, hoist it to module scope to avoid repeated allocation. `[fixable]`

server/__tests__/routes.test.ts
```diff
@@ -659,6 +659,42 @@ describe('file routes', () => {
       .send({ path: '/tmp/outside/file.txt', content: 'nope' });
     expect(res.status).toBe(403);
   });
+
+  it('GET /api/files/download — downloads file with Content-Disposition', async () => {
```
dimakis, line 663: 🔵 **missing_tests**: Download route tests don't cover filenames with special characters (spaces, quotes, unicode) to verify Content-Disposition header correctness. Adding a test with e.g. `'file "name".txt'` would exercise the escaping logic. `[fixable]`

PR #350, opened 2026-05-22
fix(ui): sync model preference into store on startup

dimakis:
## Summary

- The Zustand store initialized `config.modelId` as `null` while the UI dropdown read the preference from localStorage
- `sendMessage()` reads from the store, so the `model` field was never included in the WS payload — the SDK silently fell back to Sonnet
- One-line fix: call `setModel(getPreferredModel())` after store creation to sync localStorage into the store

Selected model now persists across all sessions until the dropdown is changed.

## Test plan

- [ ] Select Opus in dropdown, send a message — verify commit Co-Authored-By says Opus
- [ ] Reload the app without touching the dropdown — verify the model still sends correctly
- [ ] Change to Sonnet, send a message — verify it switches
- [ ] Change back to Opus — verify it persists on next session


dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #351, opened 2026-05-22
fix(server): resolve SDK session ID for headless resume

dimakis:
## Summary

- Headless sessions (created via `POST /api/sessions` with `initialPrompt`) fail silently — Claude never responds
- Root cause: query loop passes the worktree session ID (e.g. `2026-05-22-xxxx`) to the SDK's `--resume`, but the SDK expects a UUID
- Add `getSessionSdkId()` to look up the stored SDK UUID from the session index
- Use it in `chat.ts` when constructing the resume parameter, falling back to the original value

## Context

PR Shepherd agent in mgmt creates per-PR Mitzo sessions via `POST /api/sessions`. Sessions were created successfully but Claude never executed because the resume call failed with:
```
Error: --resume requires a valid session ID or session title.
Provided value "2026-05-22-41327e2d562c" is not a UUID.
```

## Test plan

- [ ] Create a headless session via `POST /api/sessions` with `initialPrompt`
- [ ] Verify Claude responds in the session
- [ ] Verify session appears on mobile with assistant reply
- [ ] Verify normal (non-headless) session resume still works

### DIFF
--- a/frontend/src/__tests__/client-store.test.ts
+++ b/frontend/src/__tests__/client-store.test.ts
@@ -1,0 +1,19 @@
+import { describe, it, expect, beforeAll } from 'vitest';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+// Structural test: client-store uses model-preference sync at module scope,
+// which accesses localStorage. The `typeof window` guard prevents crashes
+// when the module is imported in a non-browser context (e.g. Vitest server tests).
+describe('client-store window guard', () => {
+  let source: string;
+
+  beforeAll(() => {
+    source = readFileSync(join(import.meta.dirname, '..', 'client-store.ts'), 'utf-8');
+  });
+
+  it('guards model preference sync with typeof window check', () => {
+    expect(source).toContain("typeof window !== 'undefined'");
+    expect(source).toContain('getPreferredModel()');
+  });
+});
--- a/frontend/src/client-store.ts
+++ b/frontend/src/client-store.ts
@@ -28,7 +28,9 @@ export const clientStore = createMitzoStore({
 });
 
 // Sync localStorage model preference into the store so sendMessage() includes it
-clientStore.getState().setModel(getPreferredModel());
+if (typeof window !== 'undefined') {
+  clientStore.getState().setModel(getPreferredModel());
+}
 
 // Wire Capacitor app lifecycle → force WS reconnect on resume, send suspend on background
 registerCapacitorLifecycle(
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -345,13 +345,20 @@ describe('resume resolves SDK session ID', () => {
     expect(chatSource).toContain('getSessionSdkId(BASE_REPO, options.resume)');
   });
 
-  it('guards BASE_REPO before calling getSessionSdkId', () => {
-    expect(chatSource).toContain('BASE_REPO && getSessionSdkId(BASE_REPO');
+  it('guards BASE_REPO with ternary to avoid empty-string falsy bug', () => {
+    // Must use ternary (BASE_REPO ? ...) not && (which returns '' for empty string)
+    expect(chatSource).toContain('BASE_REPO ? getSessionSdkId(BASE_REPO');
   });
 
   it('falls back to raw options.resume when lookup returns undefined', () => {
-    // The ?? operator provides the fallback
-    expect(chatSource).toMatch(/getSessionSdkId\(.*\)\)\s*\?\?\s*options\.resume/);
+    expect(chatSource).toContain('?? options.resume');
+  });
+
+  it('resolvedResume is computed before the query() call', () => {
+    const resolveIdx = chatSource.indexOf('let resolvedResume');
+    const queryIdx = chatSource.indexOf('const q = query(');
+    expect(resolveIdx).toBeGreaterThan(-1);
+    expect(queryIdx).toBeGreaterThan(resolveIdx);
   });
 
   it('warns when REPO_PATH is unset during resume', () => {
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -332,6 +332,8 @@ describe('startChat stores user message for resumed sessions', () => {
   });
 });
 
+// Structural tests: resume resolution requires the Agent SDK query() pipeline,
+// so we assert against the source rather than invoking startChat() directly.
 describe('resume resolves SDK session ID', () => {
   let chatSource: string;
 
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -332,6 +332,35 @@ describe('startChat stores user message for resumed sessions', () => {
   });
 });
 
+describe('resume resolves SDK session ID', () => {
+  let chatSource: string;
+
+  beforeAll(async () => {
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+  });
+
+  it('calls getSessionSdkId before passing resume to query()', () => {
+    expect(chatSource).toContain('getSessionSdkId(BASE_REPO, options.resume)');
+  });
+
+  it('guards BASE_REPO before calling getSessionSdkId', () => {
+    expect(chatSource).toContain('BASE_REPO && getSessionSdkId(BASE_REPO');
+  });
+
+  it('falls back to raw options.resume when lookup returns undefined', () => {
+    // The ?? operator provides the fallback
+    expect(chatSource).toMatch(/getSessionSdkId\(.*\)\)\s*\?\?\s*options\.resume/);
+  });
+
+  it('warns when REPO_PATH is unset during resume', () => {
+    expect(chatSource).toContain(
+      'REPO_PATH unset — resume will use raw worktree ID, SDK may reject it',
+    );
+  });
+});
+
 describe('isIsolationEnabled', () => {
   let originalEnv: string | undefined;
 
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -63,7 +63,12 @@ export function setSessionChangeCallback(cb: SessionChangeCallback): void {
 import { EventStore } from './event-store.js';
 import { capturePromptComparison } from './prompt-compare.js';
 import { shouldAutoRename, extractRecentPrompts, generateSessionName } from './auto-rename.js';
-import { registerSession, updateSessionTitle, finalizeCloseout, getSessionSdkId } from './session-index.js';
+import {
+  registerSession,
+  updateSessionTitle,
+  finalizeCloseout,
+  getSessionSdkId,
+} from './session-index.js';
 import { createLogger } from './logger.js';
 import { withSpan, withSpanAsync } from './tracing.js';
 
@@ -904,7 +909,11 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: getSessionSdkId(BASE_REPO, options.resume) ?? options.resume } : {}),
+        ...(options.resume
+          ? {
+              resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
+            }
+          : {}),
         ...(Object.keys(allMcpServers).length > 0 ? { mcpServers: allMcpServers } : {}),
         ...(hooks ? { hooks } : {}),
         canUseTool: buildPermissionHandler(clientId, registry, {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -910,9 +910,14 @@ async function _startChatInner(
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
         ...(options.resume
-          ? {
-              resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
-            }
+          ? (() => {
+              if (!BASE_REPO) {
+                log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
+              }
+              return {
+                resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
+              };
+            })()
           : {}),
         ...(Object.keys(allMcpServers).length > 0 ? { mcpServers: allMcpServers } : {}),
         ...(hooks ? { hooks } : {}),
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -63,7 +63,7 @@ export function setSessionChangeCallback(cb: SessionChangeCallback): void {
 import { EventStore } from './event-store.js';
 import { capturePromptComparison } from './prompt-compare.js';
 import { shouldAutoRename, extractRecentPrompts, generateSessionName } from './auto-rename.js';
-import { registerSession, updateSessionTitle, finalizeCloseout } from './session-index.js';
+import { registerSession, updateSessionTitle, finalizeCloseout, getSessionSdkId } from './session-index.js';
 import { createLogger } from './logger.js';
 import { withSpan, withSpanAsync } from './tracing.js';
 
@@ -904,7 +904,7 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: options.resume } : {}),
+        ...(options.resume ? { resume: getSessionSdkId(BASE_REPO, options.resume) ?? options.resume } : {}),
         ...(Object.keys(allMcpServers).length > 0 ? { mcpServers: allMcpServers } : {}),
         ...(hooks ? { hooks } : {}),
         canUseTool: buildPermissionHandler(clientId, registry, {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -891,6 +891,16 @@ async function _startChatInner(
   })();
   capturePromptComparison(wtId, cwd, systemPromptAppend, repoWorktrees).catch(() => {});
 
+  // Resolve SDK session UUID for resume — worktree IDs are not valid SDK session IDs
+  let resolvedResume: string | undefined;
+  if (options.resume) {
+    if (!BASE_REPO) {
+      log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
+    }
+    resolvedResume =
+      (BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume;
+  }
+
   try {
     const q = query({
       prompt: inputQueue as AsyncIterable<SDKUserMessage>,
@@ -909,16 +919,7 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume
-          ? (() => {
-              if (!BASE_REPO) {
-                log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
-              }
-              return {
-                resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
-              };
-            })()
-          : {}),
+        ...(resolvedResume ? { resume: resolvedResume } : {}),
         ...(Object.keys(allMcpServers).length > 0 ? { mcpServers: allMcpServers } : {}),
         ...(hooks ? { hooks } : {}),
         canUseTool: buildPermissionHandler(clientId, registry, {
--- a/server/session-index.ts
+++ b/server/session-index.ts
@@ -156,10 +156,7 @@ export function updateSessionSdkId(repoPath: string, wtId: string, sdkSessionId:
   upsertEntry(repoPath, { id: wtId, sdk_session_id: sdkSessionId });
 }
 
-/**
- * Get the SDK session ID for a given worktree ID.
- * Returns undefined if the entry doesn't exist or sdk_session_id is not set.
- */
+/** Look up the SDK session UUID stored for a worktree ID. */
 export function getSessionSdkId(repoPath: string, wtId: string): string | undefined {
   const entries = readIndex(repoPath);
   const entry = entries.find((e) => e.id === wtId);
--- a/server/session-index.ts
+++ b/server/session-index.ts
@@ -155,3 +155,13 @@ export function updateSessionSdkId(repoPath: string, wtId: string, sdkSessionId:
 
   upsertEntry(repoPath, { id: wtId, sdk_session_id: sdkSessionId });
 }
+
+/**
+ * Get the SDK session ID for a given worktree ID.
+ * Returns undefined if the entry doesn't exist or sdk_session_id is not set.
+ */
+export function getSessionSdkId(repoPath: string, wtId: string): string | undefined {
+  const entries = readIndex(repoPath);
+  const entry = entries.find((e) => e.id === wtId);
+  return entry?.sdk_session_id;
+}
--- a/server/__tests__/session-index.test.ts
+++ b/server/__tests__/session-index.test.ts
@@ -8,6 +8,7 @@ import {
   registerSession,
   updateSessionTitle,
   updateSessionSdkId,
+  getSessionSdkId,
   finalizeCloseout,
 } from '../session-index.js';
 
@@ -303,6 +304,25 @@ describe('session-index', () => {
     it('is a no-op if session does not exist', () => {
       updateSessionSdkId(repoPath, 'nonexistent', 'sdk-abc');
       expect(readIndex(repoPath)).toEqual([]);
+    });
+  });
+
+  describe('getSessionSdkId', () => {
+    it('returns the stored SDK session ID', () => {
+      upsertEntry(repoPath, { id: 'sess-1', status: 'active' });
+      updateSessionSdkId(repoPath, 'sess-1', 'f2fd42cf-5c9f-4524-be08-9b019c1cc3a2');
+
+      expect(getSessionSdkId(repoPath, 'sess-1')).toBe('f2fd42cf-5c9f-4524-be08-9b019c1cc3a2');
+    });
+
+    it('returns undefined when entry does not exist', () => {
+      expect(getSessionSdkId(repoPath, 'nonexistent')).toBeUndefined();
+    });
+
+    it('returns undefined when sdk_session_id is not set', () => {
+      upsertEntry(repoPath, { id: 'sess-1', status: 'active' });
+
+      expect(getSessionSdkId(repoPath, 'sess-1')).toBeUndefined();
     });
   });
 });


server/session-index.ts
```diff
@@ -155,3 +155,13 @@ export function updateSessionSdkId(repoPath: string, wtId: string, sdkSessionId:
 
   upsertEntry(repoPath, { id: wtId, sdk_session_id: sdkSessionId });
 }
+
+/**
+ * Get the SDK session ID for a given worktree ID.
+ * Returns undefined if the entry doesn't exist or sdk_session_id is not set.
+ */
+export function getSessionSdkId(repoPath: string, wtId: string): string | undefined {
```
dimakis, line 160: 🟡 **missing_tests**: No test file exists for session-index.ts at all. The new `getSessionSdkId` function (and the existing `updateSessionSdkId`) lack unit tests. Given the project's TDD requirement, this should have tests covering: (1) returns undefined when entry doesn't exist, (2) returns undefined when sdk_session_id is not set, (3) returns the stored SDK session ID when present. `[fixable]`

server/session-index.ts
```diff
@@ -155,3 +155,13 @@ export function updateSessionSdkId(repoPath: string, wtId: string, sdkSessionId:
 
   upsertEntry(repoPath, { id: wtId, sdk_session_id: sdkSessionId });
 }
+
+/**
```
dimakis: 🔵 **style**: The JSDoc comment on `getSessionSdkId` is a multi-line block that just restates what the type signature already says (returns `string | undefined`). Per project conventions, prefer no comment or a single-line comment only when the WHY is non-obvious. `[fixable]`

server/chat.ts
```diff
@@ -904,7 +904,7 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: options.resume } : {}),
+        ...(options.resume ? { resume: getSessionSdkId(BASE_REPO, options.resume) ?? options.resume } : {}),
```
dimakis: 🔵 **unsafe_assumptions**: `getSessionSdkId(BASE_REPO, ...)` is called without guarding that `BASE_REPO` is non-empty. `BASE_REPO` is `process.env.REPO_PATH || ''`. Other resume-related code in this file guards with `if (options.resume && BASE_REPO)` (e.g., line 637). When `BASE_REPO` is empty, `readIndex('')` constructs a path relative to CWD (`./.claude/sessions/index.yaml`), which will return `[]` and the `??` fallback safely fires — so this isn't a bug, but it performs a needless filesystem read. A guard like `(BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume` would be more consistent with the rest of the file. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/session-index.ts`
_Correct fix — resolves SDK session ID for headless resume with a safe fallback. Main gap is missing unit tests for session-index.ts (project requires TDD). Minor style and consistency nits._

- 🟡 **missing_tests** (L163): No test file exists for session-index.ts at all. The new `getSessionSdkId` function (and the existing `updateSessionSdkId`) lack unit tests. Given the project's TDD requirement, this should have tests covering: (1) returns undefined when entry doesn't exist, (2) returns undefined when sdk_session_id is not set, (3) returns the stored SDK session ID when present. `[fixable]`
- 🔵 **style** (L159): The JSDoc comment on `getSessionSdkId` is a multi-line block that just restates what the type signature already says (returns `string | undefined`). Per project conventions, prefer no comment or a single-line comment only when the WHY is non-obvious. `[fixable]`


### `server/chat.ts`
_Correct fix — resolves SDK session ID for headless resume with a safe fallback. Main gap is missing unit tests for session-index.ts (project requires TDD). Minor style and consistency nits._

- 🔵 **unsafe_assumptions** (L907): `getSessionSdkId(BASE_REPO, ...)` is called without guarding that `BASE_REPO` is non-empty. `BASE_REPO` is `process.env.REPO_PATH || ''`. Other resume-related code in this file guards with `if (options.resume && BASE_REPO)` (e.g., line 637). When `BASE_REPO` is empty, `readIndex('')` constructs a path relative to CWD (`./.claude/sessions/index.yaml`), which will return `[]` and the `??` fallback safely fires — so this isn't a bug, but it performs a needless filesystem read. A guard like `(BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume` would be more consistent with the rest of the file. `[fixable]`

server/chat.ts
```diff
@@ -904,7 +909,11 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: options.resume } : {}),
+        ...(options.resume
```
dimakis: 🟡 **missing_tests**: The SDK session ID resolution logic in `_startChatInner` (line 914) is not tested at integration level. The ws-handler-v2 tests mock `startChat` entirely, so the `getSessionSdkId` lookup path is never exercised end-to-end. A unit test for `_startChatInner` (or at minimum a test that verifies the `resume` value passed to `query()` is the resolved SDK UUID rather than the raw worktree ID) would catch regressions in this mapping. `[fixable]`

server/chat.ts
```diff
@@ -904,7 +909,11 @@ async function _startChatInner(
...
+          ? {
+              resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
```
dimakis: 🔵 **unsafe_assumptions**: `BASE_REPO` is `process.env.REPO_PATH || ''` (line 101). The empty-string fallback is falsy, so the `&&` guard works correctly to skip the lookup when `REPO_PATH` is unset. However, this means the feature silently degrades — resume will pass the worktree ID directly to the SDK (which may fail) instead of surfacing a clear error. This is acceptable for now since `REPO_PATH` is documented as required, but worth noting.

frontend/src/client-store.ts
```diff
@@ -26,6 +27,9 @@ export const clientStore = createMitzoStore({
   },
 });
 
+// Sync localStorage model preference into the store so sendMessage() includes it
```
dimakis, line 30: 🔵 **style**: The comment `// Sync localStorage model preference into the store so sendMessage() includes it` explains the 'why' well, which is good. However, this is a module-level side effect that runs on import. If `client-store.ts` is ever imported in a test or SSR context, it will silently read from localStorage. Current usage is fine (single-page app entry point), but if this file's import surface grows, consider guarding with a `typeof window !== 'undefined'` check.

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Clean, well-scoped fix. The SDK session ID resolution logic is correct with proper fallback, the new `getSessionSdkId` function is well-tested, and the model preference sync addresses a real bug. Main gap is the lack of integration-level tests for the resume resolution path in chat.ts._

- 🟡 **missing_tests** (L912): The SDK session ID resolution logic in `_startChatInner` (line 914) is not tested at integration level. The ws-handler-v2 tests mock `startChat` entirely, so the `getSessionSdkId` lookup path is never exercised end-to-end. A unit test for `_startChatInner` (or at minimum a test that verifies the `resume` value passed to `query()` is the resolved SDK UUID rather than the raw worktree ID) would catch regressions in this mapping. `[fixable]`
- 🔵 **unsafe_assumptions** (L914): `BASE_REPO` is `process.env.REPO_PATH || ''` (line 101). The empty-string fallback is falsy, so the `&&` guard works correctly to skip the lookup when `REPO_PATH` is unset. However, this means the feature silently degrades — resume will pass the worktree ID directly to the SDK (which may fail) instead of surfacing a clear error. This is acceptable for now since `REPO_PATH` is documented as required, but worth noting.


_Clean, well-scoped fix. The SDK session ID resolution logic is correct with proper fallback, the new `getSessionSdkId` function is well-tested, and the model preference sync addresses a real bug. Main gap is the lack of integration-level tests for the resume resolution path in chat.ts._

- 🔵 **style** (L30): The comment `// Sync localStorage model preference into the store so sendMessage() includes it` explains the 'why' well, which is good. However, this is a module-level side effect that runs on import. If `client-store.ts` is ever imported in a test or SSR context, it will silently read from localStorage. Current usage is fine (single-page app entry point), but if this file's import surface grows, consider guarding with a `typeof window !== 'undefined'` check.

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_The core `getSessionSdkId` helper and tests are clean, but the fallback expression in chat.ts has a subtle `??` vs `||` bug when `REPO_PATH` is unset (empty string is not coalesced by `??`)._

- 🔴 **bugs** (L914): `(BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume` — when `BASE_REPO` is `''` (REPO_PATH unset), the `&&` short-circuits to `''`, and `??` does NOT coalesce empty strings (only null/undefined), so `resume` becomes `''` instead of falling back to `options.resume`. Use `||` instead of `??`, or restructure: `(BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume`. `[fixable]`
- 🟡 **missing_tests** (L912): The SDK session ID resolution logic in `_startChatInner` has no unit test. The ws-handler-v2 tests mock `startChat` entirely, so the `getSessionSdkId` lookup is never exercised. A test (even with mocked `getSessionSdkId`) verifying the correct value reaches `query()` options would catch the `??` vs `||` bug above. `[fixable]`


_The core `getSessionSdkId` helper and tests are clean, but the fallback expression in chat.ts has a subtle `??` vs `||` bug when `REPO_PATH` is unset (empty string is not coalesced by `??`)._

- 🔵 **style** (L31): The model preference sync is tangentially related to the headless resume fix. Consider splitting it into its own PR (it already has its own commit `c4cfe5e`) to keep the PR focused on the session ID resolution.

server/chat.ts
dimakis: 🔴 **bugs**: `(BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume` — when `BASE_REPO` is `''` (REPO_PATH unset), the `&&` short-circuits to `''`, and `??` does NOT coalesce empty strings (only null/undefined), so `resume` becomes `''` instead of falling back to `options.resume`. Use `||` instead of `??`, or restructure: `(BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume`. `[fixable]`

server/chat.ts
dimakis: 🟡 **missing_tests**: The SDK session ID resolution logic in `_startChatInner` has no unit test. The ws-handler-v2 tests mock `startChat` entirely, so the `getSessionSdkId` lookup is never exercised. A test (even with mocked `getSessionSdkId`) verifying the correct value reaches `query()` options would catch the `??` vs `||` bug above. `[fixable]`

frontend/src/client-store.ts
```diff
@@ -26,6 +27,9 @@ export const clientStore = createMitzoStore({
...
+clientStore.getState().setModel(getPreferredModel());
```
dimakis: 🔵 **style**: The model preference sync is tangentially related to the headless resume fix. Consider splitting it into its own PR (it already has its own commit `c4cfe5e`) to keep the PR focused on the session ID resolution.

server/chat.ts
```diff
@@ -904,7 +909,16 @@ async function _startChatInner(
...
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: options.resume } : {}),
+        ...(options.resume
+          ? (() => {
+              if (!BASE_REPO) {
+                log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
+              }
+              return {
+                resume: (BASE_REPO && getSessionSdkId(BASE_REPO, options.resume)) ?? options.resume,
```
dimakis: 🔴 **bugs**: When `BASE_REPO` is `''` (REPO_PATH unset), `'' && getSessionSdkId(...)` evaluates to `''`, and `'' ?? options.resume` returns `''` (since `??` only checks null/undefined, not falsy values). This passes an empty string as the `resume` value instead of falling back to `options.resume`. Use a ternary instead: `(BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume`. `[fixable]`

server/chat.ts
```diff
@@ -904,7 +909,16 @@ async function _startChatInner(
         allowedTools: [...modeAllowed, ...mcpAllowed, ...extraTools],
         thinking: resolveThinking(options.model),
         ...(options.model ? { model: parseModelSpec(options.model).model } : {}),
-        ...(options.resume ? { resume: options.resume } : {}),
+        ...(options.resume
```
dimakis: 🔵 **style**: The IIFE `(() => { ... })()` inside a spread is harder to read than a simple ternary or a local variable computed before the `query()` call. Consider extracting `const resolvedResume = options.resume ? (BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume : undefined` above the call and spreading `...(resolvedResume ? { resume: resolvedResume } : {})`. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -332,6 +332,35 @@ describe('startChat stores user message for resumed sessions', () => {
   });
 });
 
+describe('resume resolves SDK session ID', () => {
```
dimakis, line 337: 🟡 **missing_tests**: The 'resume resolves SDK session ID' tests are source-text assertions (string-matching `chat.ts` source code). While this pattern is established in the file for other hard-to-mock SDK flows, these tests won't catch the `??` vs `||` bug on line 918 — they verify substring presence but not runtime semantics. A unit test for `getSessionSdkId` itself (which exists in session-index.test.ts) is solid, but the integration between the lookup and the fallback is only structurally verified, not behaviorally.

frontend/src/client-store.ts
```diff
@@ -26,6 +27,11 @@ export const clientStore = createMitzoStore({
   },
 });
 
+// Sync localStorage model preference into the store so sendMessage() includes it
+if (typeof window !== 'undefined') {
```
dimakis, line 31: 🔵 **missing_tests**: The new `setModel(getPreferredModel())` bootstrap has no test coverage. A unit test verifying that `clientStore.getState().config.modelId` matches the localStorage value after module initialization would lock in the behavior. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Critical bug: when REPO_PATH is unset, `&&` short-circuits to `''` which `??` doesn't catch, so resume gets an empty string instead of the original worktree ID. The rest of the PR is sound — `getSessionSdkId` is well-tested and the frontend model-sync fix is correct._

- 🔴 **bugs** (L918): When `BASE_REPO` is `''` (REPO_PATH unset), `'' && getSessionSdkId(...)` evaluates to `''`, and `'' ?? options.resume` returns `''` (since `??` only checks null/undefined, not falsy values). This passes an empty string as the `resume` value instead of falling back to `options.resume`. Use a ternary instead: `(BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume`. `[fixable]`
- 🔵 **style** (L912): The IIFE `(() => { ... })()` inside a spread is harder to read than a simple ternary or a local variable computed before the `query()` call. Consider extracting `const resolvedResume = options.resume ? (BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume : undefined` above the call and spreading `...(resolvedResume ? { resume: resolvedResume } : {})`. `[fixable]`


_Critical bug: when REPO_PATH is unset, `&&` short-circuits to `''` which `??` doesn't catch, so resume gets an empty string instead of the original worktree ID. The rest of the PR is sound — `getSessionSdkId` is well-tested and the frontend model-sync fix is correct._

- 🟡 **missing_tests** (L335): The 'resume resolves SDK session ID' tests are source-text assertions (string-matching `chat.ts` source code). While this pattern is established in the file for other hard-to-mock SDK flows, these tests won't catch the `??` vs `||` bug on line 918 — they verify substring presence but not runtime semantics. A unit test for `getSessionSdkId` itself (which exists in session-index.test.ts) is solid, but the integration between the lookup and the fallback is only structurally verified, not behaviorally.


_Critical bug: when REPO_PATH is unset, `&&` short-circuits to `''` which `??` doesn't catch, so resume gets an empty string instead of the original worktree ID. The rest of the PR is sound — `getSessionSdkId` is well-tested and the frontend model-sync fix is correct._

- 🔵 **missing_tests** (L31): The new `setModel(getPreferredModel())` bootstrap has no test coverage. A unit test verifying that `clientStore.getState().config.modelId` matches the localStorage value after module initialization would lock in the behavior. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -332,6 +332,42 @@ describe('startChat stores user message for resumed sessions', () => {
...
```
dimakis, line 337: 🔵 **style**: The earlier source-reading test block (line 285-288) has a rationale comment explaining why behavioral testing is impractical and source-code assertions are used instead. This new `describe` block follows the same pattern but omits the rationale, making it less obvious to future readers why these tests read raw source instead of exercising the function. `[fixable]`

frontend/src/client-store.ts
```diff
@@ -28,7 +28,9 @@ export const clientStore = createMitzoStore({
 });
 
 // Sync localStorage model preference into the store so sendMessage() includes it
-clientStore.getState().setModel(getPreferredModel());
+if (typeof window !== 'undefined') {
```
dimakis, line 31: 🔵 **missing_tests**: The `typeof window` guard is a reasonable defensive fix (likely for Vitest imports in a non-browser environment), but there is no corresponding test verifying that importing `client-store.ts` in a windowless context no longer throws. If this guard was added to fix a specific failure, a regression test would lock it in. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-scoped fix. The core logic (getSessionSdkId + resolvedResume fallback with ternary guard) is correct, tests follow established source-reading patterns, and backward compatibility is preserved via nullish coalescing fallback. Two minor suggestions, no blocking issues._

- 🔵 **style** (L335): The earlier source-reading test block (line 285-288) has a rationale comment explaining why behavioral testing is impractical and source-code assertions are used instead. This new `describe` block follows the same pattern but omits the rationale, making it less obvious to future readers why these tests read raw source instead of exercising the function. `[fixable]`


_Clean, well-scoped fix. The core logic (getSessionSdkId + resolvedResume fallback with ternary guard) is correct, tests follow established source-reading patterns, and backward compatibility is preserved via nullish coalescing fallback. Two minor suggestions, no blocking issues._

- 🔵 **missing_tests** (L31): The `typeof window` guard is a reasonable defensive fix (likely for Vitest imports in a non-browser environment), but there is no corresponding test verifying that importing `client-store.ts` in a windowless context no longer throws. If this guard was added to fix a specific failure, a regression test would lock it in. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -332,6 +332,44 @@ describe('startChat stores user message for resumed sessions', () => {
...
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+  });
+
+  it('calls getSessionSdkId before passing resume to query()', () => {
+    expect(chatSource).toContain('getSessionSdkId(BASE_REPO, options.resume)');
+  });
+
+  it('guards BASE_REPO with ternary to avoid empty-string falsy bug', () => {
```
dimakis, line 350: 🔵 **style**: The comment '// Must use ternary (BASE_REPO ? ...) not && (which returns '' for empty string)' is helpful rationale, but the assertion `expect(chatSource).toContain('BASE_REPO ? getSessionSdkId(BASE_REPO')` is fragile — any whitespace or formatting change (e.g. Prettier wrapping the line) will break this test without any behavioral change. This is an inherent limitation of all the structural tests here, but this particular assertion is especially tight since it spans across a ternary expression.

server/__tests__/chat.test.ts
```diff
@@ -332,6 +332,44 @@ describe('startChat stores user message for resumed sessions', () => {
   });
 });
 
+// Structural tests: resume resolution requires the Agent SDK query() pipeline,
+// so we assert against the source rather than invoking startChat() directly.
+describe('resume resolves SDK session ID', () => {
```
dimakis, line 337: 🔵 **missing_tests**: The structural tests verify that resume resolution code exists in the source text, but there is no behavioral test confirming that when `options.resume` is a worktree ID with a stored `sdk_session_id`, the correct UUID actually reaches the SDK `query()` call. The structural tests are a reasonable pragmatic choice given the difficulty of mocking the Agent SDK (and the project already uses this pattern), but a comment acknowledging this gap or a future TODO would help.

server/chat.ts
```diff
@@ -886,6 +891,16 @@ async function _startChatInner(
...
   capturePromptComparison(wtId, cwd, systemPromptAppend, repoWorktrees).catch(() => {});
 
+  // Resolve SDK session UUID for resume — worktree IDs are not valid SDK session IDs
+  let resolvedResume: string | undefined;
+  if (options.resume) {
+    if (!BASE_REPO) {
+      log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
+    }
+    resolvedResume =
+      (BASE_REPO ? getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume;
```
dimakis, line 901: 🟡 **unsafe_assumptions**: `getSessionSdkId` performs synchronous YAML file I/O (`readFileSync` inside `readIndex`) on every resume call. This is fine for the current call volume, but if resume frequency increases, this will block the event loop. The existing `updateSessionSdkId`/`readIndex` callers already do synchronous I/O, so this is consistent with the codebase — flagging it as a known trade-off, not a blocker.

frontend/src/__tests__/client-store.test.ts
```diff
@@ -0,0 +1,19 @@
...
+import { join } from 'path';
+
+// Structural test: client-store uses model-preference sync at module scope,
+// which accesses localStorage. The `typeof window` guard prevents crashes
+// when the module is imported in a non-browser context (e.g. Vitest server tests).
+describe('client-store window guard', () => {
+  let source: string;
+
+  beforeAll(() => {
+    source = readFileSync(join(import.meta.dirname, '..', 'client-store.ts'), 'utf-8');
```
dimakis, line 12: 🔵 **style**: `import.meta.dirname` is a Node 21+ feature (and supported by Vitest). The server tests already use it, so this is consistent. Just noting it requires Node >= 21 at test-time.

dimakis (commented):
## Centaur Review

_Clean, well-scoped fix: adds a simple lookup function, correctly handles the empty-string falsy edge case, and has reasonable test coverage via structural assertions consistent with the project's existing patterns. No bugs or regressions found._

- 🔵 **style** (L350): The comment '// Must use ternary (BASE_REPO ? ...) not && (which returns '' for empty string)' is helpful rationale, but the assertion `expect(chatSource).toContain('BASE_REPO ? getSessionSdkId(BASE_REPO')` is fragile — any whitespace or formatting change (e.g. Prettier wrapping the line) will break this test without any behavioral change. This is an inherent limitation of all the structural tests here, but this particular assertion is especially tight since it spans across a ternary expression.
- 🔵 **missing_tests** (L337): The structural tests verify that resume resolution code exists in the source text, but there is no behavioral test confirming that when `options.resume` is a worktree ID with a stored `sdk_session_id`, the correct UUID actually reaches the SDK `query()` call. The structural tests are a reasonable pragmatic choice given the difficulty of mocking the Agent SDK (and the project already uses this pattern), but a comment acknowledging this gap or a future TODO would help.


### `server/chat.ts`
_Clean, well-scoped fix: adds a simple lookup function, correctly handles the empty-string falsy edge case, and has reasonable test coverage via structural assertions consistent with the project's existing patterns. No bugs or regressions found._

- 🟡 **unsafe_assumptions** (L901): `getSessionSdkId` performs synchronous YAML file I/O (`readFileSync` inside `readIndex`) on every resume call. This is fine for the current call volume, but if resume frequency increases, this will block the event loop. The existing `updateSessionSdkId`/`readIndex` callers already do synchronous I/O, so this is consistent with the codebase — flagging it as a known trade-off, not a blocker.


### `frontend/src/__tests__/client-store.test.ts`
_Clean, well-scoped fix: adds a simple lookup function, correctly handles the empty-string falsy edge case, and has reasonable test coverage via structural assertions consistent with the project's existing patterns. No bugs or regressions found._

- 🔵 **style** (L12): `import.meta.dirname` is a Node 21+ feature (and supported by Vitest). The server tests already use it, so this is consistent. Just noting it requires Node >= 21 at test-time.

server/__tests__/chat.test.ts
```diff
@@ -332,6 +332,44 @@ describe('startChat stores user message for resumed sessions', () => {
...
+  it('calls getSessionSdkId before passing resume to query()', () => {
+    expect(chatSource).toContain('getSessionSdkId(BASE_REPO, options.resume)');
+  });
+
+  it('guards BASE_REPO with ternary to avoid empty-string falsy bug', () => {
+    // Must use ternary (BASE_REPO ? ...) not && (which returns '' for empty string)
+    expect(chatSource).toContain('BASE_REPO ? getSessionSdkId(BASE_REPO');
+  });
+
+  it('falls back to raw options.resume when lookup returns undefined', () => {
```
dimakis, line 355: 🔵 **style**: The assertion `expect(chatSource).toContain('?? options.resume')` is fragile — it matches any occurrence of `?? options.resume` anywhere in chat.ts, not specifically in the resume resolution block. If another use of `?? options.resume` is added elsewhere, this test would still pass even if the actual resolution fallback were removed. Consider matching a more specific string like `getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume` to anchor it to the correct code path. `[fixable]`

server/__tests__/chat.test.ts
dimakis, line 337: 🔵 **missing_tests**: There is no structural test asserting that the resolved UUID (not the worktree ID) is logged or traceable on fallback. When `getSessionSdkId` returns `undefined` and the raw worktree ID is used, the only signal is the absence of the UUID — no log line distinguishes 'resolved successfully' from 'fell back to worktree ID'. A debug/info log on successful resolution would aid troubleshooting and could be structurally asserted. `[fixable]`

server/chat.ts
```diff
@@ -886,6 +891,16 @@ async function _startChatInner(
   })();
   capturePromptComparison(wtId, cwd, systemPromptAppend, repoWorktrees).catch(() => {});
 
+  // Resolve SDK session UUID for resume — worktree IDs are not valid SDK session IDs
+  let resolvedResume: string | undefined;
+  if (options.resume) {
+    if (!BASE_REPO) {
+      log.warn('REPO_PATH unset — resume will use raw worktree ID, SDK may reject it');
+    }
+    resolvedResume =
```
dimakis, line 900: 🟡 **bugs**: The `getSessionSdkId` call is synchronous I/O (reads YAML from disk via `readFileSync` inside `readIndex`). This runs on the main event loop inside an `async` function that could have used an async read. For a single resume this is unlikely to cause issues, but it's worth noting — if the session index file grows large or disk is slow, this blocks the event loop. Not critical for current usage patterns but worth considering if resume frequency increases. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean, well-structured fix with appropriate fallback semantics. The core logic is correct: worktree IDs are resolved to SDK session UUIDs before being passed to `query()`, with safe fallback to the raw ID. The structural test pattern matches existing conventions. Minor suggestions around test specificity and observability on the fallback path._

- 🔵 **style** (L355): The assertion `expect(chatSource).toContain('?? options.resume')` is fragile — it matches any occurrence of `?? options.resume` anywhere in chat.ts, not specifically in the resume resolution block. If another use of `?? options.resume` is added elsewhere, this test would still pass even if the actual resolution fallback were removed. Consider matching a more specific string like `getSessionSdkId(BASE_REPO, options.resume) : undefined) ?? options.resume` to anchor it to the correct code path. `[fixable]`
- 🔵 **missing_tests** (L337): There is no structural test asserting that the resolved UUID (not the worktree ID) is logged or traceable on fallback. When `getSessionSdkId` returns `undefined` and the raw worktree ID is used, the only signal is the absence of the UUID — no log line distinguishes 'resolved successfully' from 'fell back to worktree ID'. A debug/info log on successful resolution would aid troubleshooting and could be structurally asserted. `[fixable]`


### `server/chat.ts`
_Clean, well-structured fix with appropriate fallback semantics. The core logic is correct: worktree IDs are resolved to SDK session UUIDs before being passed to `query()`, with safe fallback to the raw ID. The structural test pattern matches existing conventions. Minor suggestions around test specificity and observability on the fallback path._

- 🟡 **bugs** (L900): The `getSessionSdkId` call is synchronous I/O (reads YAML from disk via `readFileSync` inside `readIndex`). This runs on the main event loop inside an `async` function that could have used an async read. For a single resume this is unlikely to cause issues, but it's worth noting — if the session index file grows large or disk is slow, this blocks the event loop. Not critical for current usage patterns but worth considering if resume frequency increases. `[fixable]`

PR #353, opened 2026-05-22
fix(client): restore running state on reconnect + keyboard interrupt

dimakis:
## Summary

- **Running state lost on reconnect:** `protocol-parser.ts` only dispatched `SET_RUNNING` for `running === false`, silently dropping the `true` case. After iOS WS reconnect, the UI showed idle send button instead of interrupt/queue/stop — so messages queued via `sendToChat()` with no `stopTask()` cascade, leaving subagents stuck indefinitely.
- **iOS keyboard bypass:** The keyboard Send/Enter key always called `handleSend()`, bypassing the on-screen interrupt/queue/stop buttons. Now `handleKeyDown` routes through `handleInterrupt()` when the session is running.

## Test plan

- [x] `protocol-parser.test.ts` updated — `reconnected` with `running: true` now asserts `SET_RUNNING: true` + `setWsRunning` called
- [ ] Manual: background iOS app while subagent is running → reopen → verify interrupt/stop buttons appear
- [ ] Manual: type message + tap iOS keyboard Send while agent is running → verify it interrupts (not queues)

### DIFF
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -139,8 +139,9 @@ export function parseServerMessage(
         )
       ) {
         const active = sessions.find((s) => s.sessionId === state.currentSessionId);
-        if (active && active.running === false) {
-          result.messagesActions.push({ type: 'SET_RUNNING', running: false });
+        if (active) {
+          result.messagesActions.push({ type: 'SET_RUNNING', running: active.running });
+          if (active.running) callbacks.setWsRunning?.(poolKey, true);
         }
       }
       callbacks.onReconnected?.();
--- a/frontend/src/components/ChatInput.tsx
+++ b/frontend/src/components/ChatInput.tsx
@@ -151,7 +151,11 @@ export function ChatInput({
   function handleKeyDown(e: KeyboardEvent) {
     if (e.key === 'Enter' && !e.shiftKey) {
       e.preventDefault();
-      handleSend();
+      if (running && onInterrupt) {
+        handleInterrupt();
+      } else {
+        handleSend();
+      }
     }
   }
 
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -627,15 +627,17 @@ describe('reconnected', () => {
     expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: false });
   });
 
-  it('does not dispatch SET_RUNNING when active session is still running', () => {
+  it('dispatches SET_RUNNING true when active session is still running', () => {
+    const setWsRunning = vi.fn();
     const state = makeState({ currentSessionId: 'sid-1' });
     const r = parseServerMessage(
       { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 0, running: true }] },
       state,
-      makeCallbacks(),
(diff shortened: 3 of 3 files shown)


dimakis:
## Centaur Review

Found **2** issue(s).


_Clean, correct fix: restores running state on reconnect and routes Enter to interrupt when running. Both changes are well-tested at the protocol level; minor gaps in negative assertions and ChatInput test coverage._

- 🔵 **missing_tests** (L619): The test for `running: false` on reconnect does not assert that `setWsRunning` is NOT called. The implementation asymmetrically calls `setWsRunning` only for `true` but not `false`. A negative assertion (`expect(setWsRunning).not.toHaveBeenCalled()`) in the existing `running: false` test would document this intentional asymmetry. `[fixable]`


### `frontend/src/components/ChatInput.tsx`
_Clean, correct fix: restores running state on reconnect and routes Enter to interrupt when running. Both changes are well-tested at the protocol level; minor gaps in negative assertions and ChatInput test coverage._

- 🔵 **missing_tests** (L154): No unit/integration tests exist for ChatInput. The new Enter-key routing (interrupt vs send based on `running` + `onInterrupt`) is untested. Key edge case: `running=true` but `onInterrupt` undefined should still fall through to `handleSend`. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix. Restores running state on reconnect and routes Enter key to interrupt — both changes are correct with good test coverage including edge cases._

- 🔵 **missing_tests**: No test for the case where currentSessionId is set but the session is not found in the reconnected sessions list (active === undefined). The .find() returns undefined and the if(active) guard skips SET_RUNNING — this is correct behavior but worth a test to document the contract. `[fixable]`


_Clean, well-tested fix. Restores running state on reconnect and routes Enter key to interrupt — both changes are correct with good test coverage including edge cases._

- 🔵 **style** (L144): setWsRunning is only called when active.running is true, not when false — asymmetric with other message handlers (e.g. reattach_failed calls setWsRunning(false)). Since setWsRunning is deprecated and SET_RUNNING handles both directions, this is functionally fine, but could be simplified to `callbacks.setWsRunning?.(poolKey, active.running)` for symmetry. Low priority given the deprecation. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -139,8 +139,9 @@ export function parseServerMessage(
         )
       ) {
         const active = sessions.find((s) => s.sessionId === state.currentSessionId);
-        if (active && active.running === false) {
-          result.messagesActions.push({ type: 'SET_RUNNING', running: false });
+        if (active) {
+          result.messagesActions.push({ type: 'SET_RUNNING', running: active.running });
+          if (active.running) callbacks.setWsRunning?.(poolKey, true);
```
dimakis, line 144: 🔵 **style**: setWsRunning is only called when active.running is true, not when false — asymmetric with other message handlers (e.g. reattach_failed calls setWsRunning(false)). Since setWsRunning is deprecated and SET_RUNNING handles both directions, this is functionally fine, but could be simplified to `callbacks.setWsRunning?.(poolKey, active.running)` for symmetry. Low priority given the deprecation. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #355, opened 2026-05-22
fix(server): skip --resume on first headless query

dimakis:
## Summary

- Headless sessions (`POST /api/sessions` with `initialPrompt`) passed the Mitzo worktree ID as `--resume` to the Claude Code SDK, which rejects non-UUID values — causing 100% failure on all headless dispatch (PR Shepherd, etc.)
- Fix: omit `resume` on the first headless query since no SDK session exists yet. The UUID is captured via `updateSessionSdkId()` after the first assistant message and used for subsequent queries.
- Design doc: `local_features/headless-session-resume/design.md` (in mgmt)

## Test plan

- [x] Structural test: headless `startChat()` call omits `resume` option
- [x] Structural test: interactive resume path still resolves SDK session ID via `getSessionSdkId()`
- [ ] Manual: `POST /api/sessions` with `initialPrompt` → first query succeeds
- [ ] Manual: PR Shepherd dispatch creates sessions that run the skill
- [ ] Manual: user reply from phone resumes the SDK session correctly

### DIFF
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -459,6 +459,43 @@ describe('discoverSessionWorktrees integration', () => {
   });
 });
 
+describe('headless session does not pass resume on first query', () => {
+  let appSource: string;
+  let chatSource: string;
+
+  beforeAll(async () => {
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    appSource = readFileSync(join(import.meta.dirname, '..', 'app.ts'), 'utf-8');
+    chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+  });
+
+  it('headless startChat call omits resume option', () => {
+    // Find the headless startChat block (identified by NullTransport + headless clientId)
+    const headlessMarker = 'const clientId = `headless:${wtId}`';
+    const headlessIdx = appSource.indexOf(headlessMarker);
+    expect(headlessIdx).toBeGreaterThan(-1);
+
+    // Extract the startChat call after the headless marker
+    const startChatIdx = appSource.indexOf('await startChat(', headlessIdx);
+    expect(startChatIdx).toBeGreaterThan(-1);
+
+    // Get the options object passed to startChat (up to the closing paren + semicolon)
+    const callEnd = appSource.indexOf(');', startChatIdx);
+    const callRegion = appSource.slice(startChatIdx, callEnd);
+
+    // Must NOT contain resume as an option key (resume: ...)
+    expect(callRegion).not.toMatch(/resume\s*[,:]/);
+    expect(callRegion).not.toMatch(/resume\s*\?/);
+  });
+
+  it('interactive resume path still resolves SDK session ID', () => {
+    // The resume resolution logic in startChat must still exist for interactive sessions
+    expect(chatSource).toContain('getSessionSdkId(BASE_REPO, options.resume)');
+    expect(chatSource).toContain('resolvedResume');
+  });
+});
+
 describe('validateResumable', () => {
   it('returns valid for a CWD that passes git check', async () => {
     const { validateResumable } = await import('../chat.js');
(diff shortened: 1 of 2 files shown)


server/__tests__/chat.test.ts
```diff
@@ -459,6 +459,43 @@ describe('discoverSessionWorktrees integration', () => {
...
+  let chatSource: string;
+
+  beforeAll(async () => {
+    const { readFileSync } = await import('fs');
+    const { join } = await import('path');
+    appSource = readFileSync(join(import.meta.dirname, '..', 'app.ts'), 'utf-8');
+    chatSource = readFileSync(join(import.meta.dirname, '..', 'chat.ts'), 'utf-8');
+  });
+
+  it('headless startChat call omits resume option', () => {
```
dimakis, line 471: 🟡 **style**: Source-code-scanning tests (reading app.ts/chat.ts as strings and asserting on substrings) are fragile — any refactor that changes whitespace, variable names, or call structure will break them silently or produce false positives. The `indexOf(');', startChatIdx)` heuristic fails if the options object contains a nested `);` sequence (e.g. in a comment or string). Consider testing the actual behavior (e.g. mock `startChat` and assert the options it receives) rather than parsing source text.

server/__tests__/chat.test.ts
```diff
@@ -459,6 +459,43 @@ describe('discoverSessionWorktrees integration', () => {
...
+    // Get the options object passed to startChat (up to the closing paren + semicolon)
+    const callEnd = appSource.indexOf(');', startChatIdx);
+    const callRegion = appSource.slice(startChatIdx, callEnd);
+
+    // Must NOT contain resume as an option key (resume: ...)
+    expect(callRegion).not.toMatch(/resume\s*[,:]/);
+    expect(callRegion).not.toMatch(/resume\s*\?/);
+  });
+
+  it('interactive resume path still resolves SDK session ID', () => {
```
dimakis: 🔵 **missing_tests**: The second test ('interactive resume path still resolves SDK session ID') duplicates the existing test at line ~346 which already asserts `chatSource.toContain('getSessionSdkId(BASE_REPO, options.resume)')`. This is redundant coverage — both read chat.ts source and check the same substring. `[fixable]`

dimakis (commented):
## Centaur Review

_Correct one-line fix (removing `resume: wtId` for first headless query). The logic is sound and well-commented, but the tests rely on fragile source-code string scanning rather than behavioral assertions._

- 🟡 **style** (L473): Source-code-scanning tests (reading app.ts/chat.ts as strings and asserting on substrings) are fragile — any refactor that changes whitespace, variable names, or call structure will break them silently or produce false positives. The `indexOf(');', startChatIdx)` heuristic fails if the options object contains a nested `);` sequence (e.g. in a comment or string). Consider testing the actual behavior (e.g. mock `startChat` and assert the options it receives) rather than parsing source text.
- 🔵 **missing_tests** (L492): The second test ('interactive resume path still resolves SDK session ID') duplicates the existing test at line ~346 which already asserts `chatSource.toContain('getSessionSdkId(BASE_REPO, options.resume)')`. This is redundant coverage — both read chat.ts source and check the same substring. `[fixable]`

server/app.ts
```diff
@@ -480,7 +480,10 @@ async function handleSessionCreate(
       const clientId = `headless:${wtId}`;
       const transport = new NullTransport();
       await startChat(transport, clientId, initialPrompt, {
-        resume: wtId,
+        // No resume — this is the first query, no SDK session exists yet.
```
dimakis, line 483: 🔵 **style**: The 4-line comment explaining why `resume` is omitted is helpful for the initial commit but may become noise long-term. A single line like `// First query — no SDK session to resume yet` would suffice given that the commit message and the `updateSessionSdkId` call site already document the lifecycle. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `server/app.ts`
_Correct fix — the headless path was passing a worktree ID as a resume token before any SDK session existed, which would cause the SDK to reject the query. The test follows an established source-scanning pattern in this repo. LGTM with a minor comment verbosity nit._

- 🔵 **style** (L483): The 4-line comment explaining why `resume` is omitted is helpful for the initial commit but may become noise long-term. A single line like `// First query — no SDK session to resume yet` would suffice given that the commit message and the `updateSessionSdkId` call site already document the lifecycle. `[fixable]`

PR #358, opened 2026-05-22
refactor(server): fetch boot context from ContexGin HTTP API

dimakis:
## Summary

- Replace inline `import('contexgin')` library call with HTTP fetch to running ContexGin server (`GET /api/agents/:name/context`)
- The old approach used a hardcoded 8k token budget fallback and read the agent recipe from the worktree path (often stale/missing). The server already compiles correctly with the recipe's 12k budget.
- Extract `fetchBootContext()` as a testable function, reducing ~140 lines of library import + recipe parsing + response mapping to a single HTTP call

## Root cause

`chat.ts` imported `contexgin` as a Node library and called the legacy `compile()` function with `DEFAULT_TOKEN_BUDGET = 8000`. It attempted to read `tokenBudget` from `.agents/mitzo-conversational.yaml` in `cwd` (the worktree), which often didn't have the file or had a stale version. Meanwhile, the running ContexGin server at `CONTEXGIN_URL` was already compiling correctly with 12k. Mitzo was bypassing its own infrastructure.

## Test plan

- [x] 7 new tests: success path, unreachable server, non-200 response, missing boot field, empty sources, non-string source filtering, env var URL
- [x] Existing `contexgin-smoke`, `context-baseline`, and `prompt-compare` tests still pass (21/21)
- [x] Full test suite: no new failures (2 pre-existing on main)
- [ ] Manual: start a Mitzo session, verify boot context shows >8k budget in source explorer

### DIFF
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -5,6 +5,25 @@ import type { BootContextMessage } from '../chat.js';
 const mockFetch = vi.fn();
 vi.stubGlobal('fetch', mockFetch);
 
+// Mock child_process for local fallback tests
+const mockExecFile = vi.fn();
+vi.mock('child_process', () => ({
+  execFileSync: vi.fn(),
+  execFile: mockExecFile,
+}));
+
+// Mock util.promisify to return our mock
+vi.mock('util', async (importOriginal) => {
+  const actual = (await importOriginal()) as Record<string, unknown>;
+  return {
+    ...actual,
+    promisify: (fn: unknown) => {
+      if (fn === mockExecFile) return mockExecFile;
+      return (actual.promisify as (fn: unknown) => unknown)(fn);
+    },
+  };
+});
+
 // Mock logger to suppress output
 vi.mock('../logger.js', () => ({
   createLogger: () => ({
@@ -93,38 +112,66 @@ describe('fetchBootContext', () => {
     });
   });
 
-  it('returns local-fallback when ContexGin is unreachable', async () => {
+  it('falls back to local Python script when ContexGin is unreachable', async () => {
     mockFetch.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: '# Local boot context\nFallback content.' }),
+    });
 
-    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
-    expect(result.sourceCount).toBe(0);
-    expect(result.tokenCount).toBe(0);
+    expect(result.sourceCount).toBe(5);
+    expect(result.fullMarkdown).toBe('# Local boot context\nFallback content.');
+    expect(result.tokenCount).toBeGreaterThan(0);
+    expect(result.sources).toHaveLength(5);
+    expect(result.sources[0]).toEqual({
+      path: 'memory/Profile/Working Style.md',
+      kind: 'profile',
+    });
   });
 
-  it('returns local-fallback on non-200 response', async () => {
+  it('falls back to local Python script on non-200 response', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: false,
       status: 404,
       text: async () => '{"error":"Agent not found"}',
     });
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: 'fallback content' }),
+    });
 
-    const result = await fetchBootContext('nonexistent-agent', CONTEXGIN_URL);
+    const result = await fetchBootContext('nonexistent-agent', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
-    expect(result.sourceCount).toBe(0);
+    expect(result.fullMarkdown).toBe('fallback content');
   });
 
-  it('returns local-fallback when response lacks boot field', async () => {
+  it('falls back to local Python script when response lacks boot field', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: true,
       json: async () => ({ agent: 'test', identity: {} }),
     });
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: 'fallback' }),
+    });
 
-    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
+    expect(result.fullMarkdown).toBe('fallback');
+  });
+
+  it('returns zero-value fallback when both ContexGin and Python script fail', async () => {
+    mockFetch.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+    mockExecFile.mockRejectedValueOnce(new Error('script not found'));
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
+
+    expect(result.source).toBe('local-fallback');
+    expect(result.sourceCount).toBe(0);
+    expect(result.tokenCount).toBe(0);
+    expect(result.fullMarkdown).toBeUndefined();
   });
 
   it('handles empty sources array', async () => {
@@ -169,6 +216,7 @@ describe('fetchBootContext', () => {
     process.env.CONTEXGIN_URL = 'http://test-host:9999';
 
     mockFetch.mockRejectedValueOnce(new Error('timeout'));
+    mockExecFile.mockRejectedValueOnce(new Error('no script'));
 
     await fetchBootContext('mitzo-conversational');
 
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -1,0 +1,186 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { BootContextMessage } from '../chat.js';
+
+// Mock fetch globally
+const mockFetch = vi.fn();
+vi.stubGlobal('fetch', mockFetch);
+
+// Mock logger to suppress output
+vi.mock('../logger.js', () => ({
+  createLogger: () => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  }),
+}));
+
+// Mock the event store to avoid SQLite initialization
+class FakeEventStore {
+  recordEvent = vi.fn();
+  getEvents = vi.fn().mockReturnValue([]);
+  getSession = vi.fn().mockReturnValue(null);
+  upsertSession = vi.fn();
+}
+vi.mock('../event-store.js', () => ({
+  EventStore: FakeEventStore,
+}));
+
+// Mock repo-config
+vi.mock('../repo-config.js', () => ({
+  loadRepoConfig: vi.fn(() => ({
+    contextBlocks: {},
+    quickActions: [],
+    venvPaths: [],
+    resolvedVenvPaths: [],
+    allowedPaths: [],
+    roots: [],
+    toolTierOverrides: {},
+    inboxPath: '',
+    resolvedInboxPath: '',
+    repos: {},
+    isolation: true,
+  })),
+}));
+
+const { fetchBootContext } = await import('../chat.js');
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe('fetchBootContext', () => {
+  const CONTEXGIN_URL = 'http://localhost:4195';
+
+  it('returns contexgin boot context on successful response', async () => {
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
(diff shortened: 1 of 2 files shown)


server/chat.ts
```diff
@@ -748,147 +831,8 @@ async function _startChatInner(
...
-        tokenCount: 0,
-        tokenBudget: tokenBudget,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
-    }
-  })();
+  // Fire-and-forget: fetch boot context from running ContexGin server
+  fetchBootContext('mitzo-conversational').then((msg) => send(transport, msg));
```
dimakis: 🟡 **bugs**: Missing `.catch()` on fire-and-forget promise chain. If `transport.send()` throws after the async gap (e.g., transport closed between fetch and send), this produces an unhandled promise rejection. The adjacent `capturePromptComparison` call correctly uses `.catch(() => {})`. Fix: `fetchBootContext('mitzo-conversational').then((msg) => send(transport, msg)).catch(() => {});` `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,89 @@ export const eventStore = initEventStore();
...
+      .map((s) => ({ path: s, kind: 'reference' }));
+
+    const fullMarkdown = typeof boot.content === 'string' ? (boot.content as string) : undefined;
+
+    return {
+      type: 'boot_context',
+      source: 'contexgin',
+      sourceCount: sources.length,
+      tokenCount: bootTokens,
+      tokenBudget: bootTokens, // server compiles to its own budget — report actual
```
dimakis, line 212: 🟡 **regressions**: `included` and `trimmed` are now always empty arrays. The old code populated these from `compiled.included`/`compiled.trimmed` with section-level detail (source, heading, token count, content). The frontend's `BootContextPill` renders expandable 'Included (N)' and 'N sections trimmed' sections from these fields — those UI sections will never appear now. If intentional (ContexGin manages compilation internally), these fields on `BootContextMessage` are effectively dead code and should be removed from the interface to avoid confusion.

dimakis, line 212: 🔵 **regressions**: `tokenBudget` is now always equal to `tokenCount` (both set to `bootTokens`). The old code distinguished configured budget (from agent recipe YAML, default 8000) from actual usage. The frontend displays these as 'X.Xk / X.Xk' (count / budget) which is now always identical, making the budget half of the label redundant. In the fallback case, budget changed from 8000 to 0, so the pill shows '0 / 0' instead of '0 / 8.0k'.

server/__tests__/boot-context.test.ts
```diff
@@ -0,0 +1,186 @@
...
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
+    expect(result.sources).toEqual([
+      { path: 'valid.md', kind: 'reference' },
+      { path: 'also-valid.md', kind: 'reference' },
+    ]);
+    expect(result.sourceCount).toBe(2);
+  });
+
+  it('uses default URL from env when not provided', async () => {
```
dimakis: 🔵 **unsafe_assumptions**: Env var cleanup in the last test is not wrapped in `try/finally`. If an assertion fails before line 180, `process.env.CONTEXGIN_URL` leaks its test value into subsequent tests. Use `try/finally` or Vitest's `vi.stubEnv()` for safer cleanup. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Clean refactor from library import to HTTP API. Main concern is a missing `.catch()` on the fire-and-forget promise (unhandled rejection risk), and a UI regression where the `included`/`trimmed` section detail and budget-vs-actual distinction are silently lost from BootContextPill._

- 🟡 **bugs** (L835): Missing `.catch()` on fire-and-forget promise chain. If `transport.send()` throws after the async gap (e.g., transport closed between fetch and send), this produces an unhandled promise rejection. The adjacent `capturePromptComparison` call correctly uses `.catch(() => {})`. Fix: `fetchBootContext('mitzo-conversational').then((msg) => send(transport, msg)).catch(() => {});` `[fixable]`
- 🟡 **regressions** (L160): `included` and `trimmed` are now always empty arrays. The old code populated these from `compiled.included`/`compiled.trimmed` with section-level detail (source, heading, token count, content). The frontend's `BootContextPill` renders expandable 'Included (N)' and 'N sections trimmed' sections from these fields — those UI sections will never appear now. If intentional (ContexGin manages compilation internally), these fields on `BootContextMessage` are effectively dead code and should be removed from the interface to avoid confusion.
- 🔵 **regressions** (L160): `tokenBudget` is now always equal to `tokenCount` (both set to `bootTokens`). The old code distinguished configured budget (from agent recipe YAML, default 8000) from actual usage. The frontend displays these as 'X.Xk / X.Xk' (count / budget) which is now always identical, making the budget half of the label redundant. In the fallback case, budget changed from 8000 to 0, so the pill shows '0 / 0' instead of '0 / 8.0k'.


_Clean refactor from library import to HTTP API. Main concern is a missing `.catch()` on the fire-and-forget promise (unhandled rejection risk), and a UI regression where the `included`/`trimmed` section detail and budget-vs-actual distinction are silently lost from BootContextPill._

- 🔵 **missing_tests**: Missing test for malformed JSON response — when `res.json()` throws (e.g., body is not valid JSON despite `ok: true`), the outer catch block should return `FALLBACK_BOOT_CONTEXT`. This edge case is handled correctly by the implementation but not verified by tests. `[fixable]`
- 🔵 **unsafe_assumptions** (L167): Env var cleanup in the last test is not wrapped in `try/finally`. If an assertion fails before line 180, `process.env.CONTEXGIN_URL` leaks its test value into subsequent tests. Use `try/finally` or Vitest's `vi.stubEnv()` for safer cleanup. `[fixable]`

server/chat.ts
```diff
@@ -748,147 +876,8 @@ async function _startChatInner(
...
-        tokenCount: 0,
-        tokenBudget: tokenBudget,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
-    }
-  })();
+  // Fire-and-forget: fetch boot context from running ContexGin server
+  fetchBootContext('mitzo-conversational').then((msg) => send(transport, msg));
```
dimakis: 🟡 **bugs**: Unhandled rejection risk: `fetchBootContext` never rejects, but `send(transport, msg)` inside `.then()` could throw (e.g., if `transport.send()` throws on a closed/errored socket). Add `.catch()` to match the pattern used on the adjacent line (`capturePromptComparison(...).catch(() => {})`). `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,134 @@ export const eventStore = initEventStore();
...
+      .map((s) => ({ path: s, kind: 'reference' }));
+
+    const fullMarkdown = typeof boot.content === 'string' ? (boot.content as string) : undefined;
+
+    return {
+      type: 'boot_context',
+      source: 'contexgin',
+      sourceCount: sources.length,
+      tokenCount: bootTokens,
+      tokenBudget: bootTokens, // server compiles to its own budget — report actual
```
dimakis, line 212: 🟡 **regressions**: The old code reported a separate `tokenBudget` read from the agent recipe YAML (`.agents/mitzo-conversational.yaml`), which could differ from actual token count. The new code sets `tokenBudget: bootTokens` — always equal to `tokenCount`. If the frontend uses `tokenBudget` to show budget vs. actual usage, it will now always show 100% utilization. Verify the frontend `BootContextPill` doesn't rely on budget != count. `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,134 @@ export const eventStore = initEventStore();
...
+    const bootTokens = typeof boot.tokens === 'number' ? boot.tokens : 0;
+    const rawSources = Array.isArray(boot.sources) ? boot.sources : [];
+
+    const sources: Array<{ path: string; kind: string }> = rawSources
+      .filter((s): s is string => typeof s === 'string')
+      .map((s) => ({ path: s, kind: 'reference' }));
+
+    const fullMarkdown = typeof boot.content === 'string' ? (boot.content as string) : undefined;
+
+    return {
```
dimakis, line 207: 🔵 **regressions**: The old code populated `included` and `trimmed` arrays with section-level metadata (source, heading, tokens, content) from the contexgin compile result. The new HTTP API path always sends empty arrays for both. If ContexGin's HTTP response includes this data, it's being silently dropped. If it doesn't, the `BootContextMessage` interface could be simplified to make `included`/`trimmed` optional. `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,134 @@ export const eventStore = initEventStore();
...
+      timeout: 5000,
+    });
+    const parsed = JSON.parse(stdout) as { additionalContext?: string };
+    const content = parsed.additionalContext ?? '';
+    // Rough token estimate: ~4 chars per token
+    const tokens = Math.ceil(content.length / 4);
+    return {
+      type: 'boot_context',
+      source: 'local-fallback',
+      sourceCount: 5, // hardcoded source count from build_boot_context.py SOURCE_PATHS
```
dimakis: 🔵 **unsafe_assumptions**: Hardcoded `sourceCount: 5` and a hardcoded list of 5 source paths in `localBootContextFallback`. If the Python script's `SOURCE_PATHS` changes, this will silently drift. Consider parsing the actual sources from the script output if it provides them, or adding a comment noting this coupling. `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,134 @@ export const eventStore = initEventStore();
...
+ * Falls back to build_boot_context.py if the server is unreachable.
+ * Never throws — returns a local-fallback message on any error.
+ */
+export async function fetchBootContext(
+  agentName: string,
+  contexginUrl: string = process.env.CONTEXGIN_URL || 'http://localhost:8321',
+  repoRoot: string = BASE_REPO,
+): Promise<BootContextMessage> {
+  try {
+    const url = `${contexginUrl}/api/agents/${agentName}/context`;
```
dimakis: 🔵 **unsafe_assumptions**: `agentName` is interpolated directly into the URL path without encoding. Currently hardcoded to `'mitzo-conversational'` at the call site, so safe in practice. But if the function is reused with user-supplied names containing slashes or special characters, the URL would be malformed. Consider `encodeURIComponent(agentName)` for defense in depth. `[fixable]`

server/chat.ts
```diff
@@ -87,6 +87,134 @@ export const eventStore = initEventStore();
...
+    }
+
+    const bootTokens = typeof boot.tokens === 'number' ? boot.tokens : 0;
+    const rawSources = Array.isArray(boot.sources) ? boot.sources : [];
+
+    const sources: Array<{ path: string; kind: string }> = rawSources
+      .filter((s): s is string => typeof s === 'string')
+      .map((s) => ({ path: s, kind: 'reference' }));
+
+    const fullMarkdown = typeof boot.content === 'string' ? (boot.content as string) : undefined;
```
dimakis: 🔵 **style**: Redundant cast: `(boot.content as string)` is already inside a `typeof boot.content === 'string'` guard, so the `as string` is unnecessary. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Clean refactoring from npm-imported contexgin to HTTP API with solid fallback chain and good test coverage. Main concerns: missing `.catch()` on fire-and-forget promise, and `tokenBudget` now always equals `tokenCount` which may be a behavioral regression for the frontend budget display._

- 🟡 **bugs** (L880): Unhandled rejection risk: `fetchBootContext` never rejects, but `send(transport, msg)` inside `.then()` could throw (e.g., if `transport.send()` throws on a closed/errored socket). Add `.catch()` to match the pattern used on the adjacent line (`capturePromptComparison(...).catch(() => {})`). `[fixable]`
- 🟡 **regressions** (L205): The old code reported a separate `tokenBudget` read from the agent recipe YAML (`.agents/mitzo-conversational.yaml`), which could differ from actual token count. The new code sets `tokenBudget: bootTokens` — always equal to `tokenCount`. If the frontend uses `tokenBudget` to show budget vs. actual usage, it will now always show 100% utilization. Verify the frontend `BootContextPill` doesn't rely on budget != count. `[fixable]`
- 🔵 **regressions** (L200): The old code populated `included` and `trimmed` arrays with section-level metadata (source, heading, tokens, content) from the contexgin compile result. The new HTTP API path always sends empty arrays for both. If ContexGin's HTTP response includes this data, it's being silently dropped. If it doesn't, the `BootContextMessage` interface could be simplified to make `included`/`trimmed` optional. `[fixable]`
- 🔵 **unsafe_assumptions** (L135): Hardcoded `sourceCount: 5` and a hardcoded list of 5 source paths in `localBootContextFallback`. If the Python script's `SOURCE_PATHS` changes, this will silently drift. Consider parsing the actual sources from the script output if it provides them, or adding a comment noting this coupling. `[fixable]`
- 🔵 **unsafe_assumptions** (L167): `agentName` is interpolated directly into the URL path without encoding. Currently hardcoded to `'mitzo-conversational'` at the call site, so safe in practice. But if the function is reused with user-supplied names containing slashes or special characters, the URL would be malformed. Consider `encodeURIComponent(agentName)` for defense in depth. `[fixable]`
- 🔵 **style** (L198): Redundant cast: `(boot.content as string)` is already inside a `typeof boot.content === 'string'` guard, so the `as string` is unnecessary. `[fixable]`


_Clean refactoring from npm-imported contexgin to HTTP API with solid fallback chain and good test coverage. Main concerns: missing `.catch()` on fire-and-forget promise, and `tokenBudget` now always equals `tokenCount` which may be a behavioral regression for the frontend budget display._

- 🔵 **missing_tests**: No test for when `fetch` succeeds but `res.json()` throws (e.g., invalid JSON body). The implementation's outer catch handles it, but an explicit test would document this contract. `[fixable]`

PR #360, opened 2026-05-30
feat(session): persist boot context across reconnect and switch

dimakis:
## Summary
- Boot context pills disappeared on iOS reconnect and session switch — the `boot_context` WS message was fire-and-forget, never persisted
- Adds three-layer persistence: ManagedSession in-memory cache (hot), SQLite `agent_name` + `boot_context` columns (cold), re-send on reconnect/switch
- No frontend changes needed — protocol-parser already handles `boot_context` messages

## Test plan
- [ ] All existing tests pass (438 across 5 suites — verified)
- [ ] Start session → verify pills appear
- [ ] Background iOS → foreground → pills persist
- [ ] Switch to running session → pills appear
- [ ] Switch to ended (cold) session → pills appear from DB
- [ ] Server restart → resume → pills recompile

### DIFF
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -48,6 +48,8 @@ interface SessionRow {
   last_speaker_at: number | null;
   state: string | null;
   last_state_change: number | null;
+  agent_name: string | null;
+  boot_context: string | null;
   created_at: number;
   updated_at: number;
 }
@@ -122,6 +124,7 @@ export class EventStore {
     this.migrateCloseTracking(db);
     this.migrateAttentionTracking(db);
     this.migrateSessionState(db);
+    this.migrateBootContext(db);
 
     this.log.info('EventStore initialized', { dbPath });
 
@@ -281,6 +284,19 @@ export class EventStore {
     }
   }
 
+  private migrateBootContext(db: Database.Database): void {
+    const columns = db.prepare("PRAGMA table_info('sessions')").all() as Array<{ name: string }>;
+    const columnNames = new Set(columns.map((c) => c.name));
+    if (!columnNames.has('agent_name')) {
+      db.exec('ALTER TABLE sessions ADD COLUMN agent_name TEXT');
+      this.log.info('migrated sessions table: added agent_name');
+    }
+    if (!columnNames.has('boot_context')) {
+      db.exec('ALTER TABLE sessions ADD COLUMN boot_context TEXT');
+      this.log.info('migrated sessions table: added boot_context');
+    }
+  }
+
   close(): void {
     if (this.db) {
       this.db.close();
@@ -351,6 +367,14 @@ export class EventStore {
         fields.push('closed_by = ?');
         values.push(meta.closedBy);
       }
+      if (meta.agentName !== undefined) {
+        fields.push('agent_name = ?');
+        values.push(meta.agentName);
+      }
+      if (meta.bootContext !== undefined) {
+        fields.push('boot_context = ?');
+        values.push(meta.bootContext);
+      }
       if (meta.updatedAt !== undefined) {
         fields.push('updated_at = ?');
         values.push(meta.updatedAt);
@@ -374,6 +398,8 @@ export class EventStore {
         'goal_id',
         'telos_task_id',
         'closed_by',
+        'agent_name',
+        'boot_context',
       ];
       const vals: unknown[] = [
         meta.sessionId,
@@ -387,6 +413,8 @@ export class EventStore {
         meta.goalId ?? null,
         meta.telosTaskId ?? null,
         meta.closedBy ?? null,
+        meta.agentName ?? null,
+        meta.bootContext ?? null,
       ];
       if (meta.updatedAt !== undefined) {
         cols.push('updated_at');
@@ -653,6 +681,8 @@ function rowToSession(row: SessionRow): SessionMeta {
     lastSpeakerAt: row.last_speaker_at ?? null,
     state: (row.state as SessionMeta['state']) ?? null,
     lastStateChange: row.last_state_change ?? null,
+    agentName: row.agent_name ?? null,
+    bootContext: row.boot_context ?? null,
     createdAt: row.created_at,
     updatedAt: row.updated_at,
   };
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -675,6 +675,9 @@ async function _startChatInner(
   const mcpAllowed = buildMcpAllowedTools(clientId);
   const extraTools = options.extraTools ? options.extraTools.split(',').map((t) => t.trim()) : [];
 
+  // Resolve agent name early — needed for registration, resume upsert, and boot context.
+  const agentName = options.agentName ?? DEFAULT_AGENT_NAME;
+
   // Streaming-input queue — kept open for the session lifetime.
   const inputQueue = new AsyncQueue<SDKUserMessage>();
   inputQueue.push(makeUserMessage(fullPrompt, 'now'));
@@ -687,6 +690,7 @@ async function _startChatInner(
     wtId,
     sessionAllowList: new Set<string>(),
     worktreePath,
+    agentName,
     // Set sessionId early so pre-assistant events are persisted (iOS reconnect).
     ...(options.resume ? { sessionId: options.resume } : {}),
     ...(options.telosTaskId ? { telosTaskId: options.telosTaskId } : {}),
@@ -726,6 +730,7 @@ async function _startChatInner(
       ...(worktreePath ? { wtId } : {}),
       ...(options.telosTaskId ? { telosTaskId: options.telosTaskId } : {}),
       ...(existingMeta ? { updatedAt: existingMeta.updatedAt } : {}),
+      agentName,
     });
   }
 
@@ -741,7 +746,6 @@ async function _startChatInner(
   // Build session env with worktree paths for the agent (all repos including primary)
   const sessionEnv = sdkEnv();
   sessionEnv.MITZO_SESSION_ID = wtId;
-  const agentName = options.agentName ?? DEFAULT_AGENT_NAME;
   sessionEnv.MITZO_AGENT_NAME = agentName;
   for (const [name, { path }] of repoWorktrees) {
     sessionEnv[`MITZO_REPO_${name.toUpperCase()}`] = path;
@@ -777,16 +781,18 @@ async function _startChatInner(
     } catch (importErr: unknown) {
       const msg = importErr instanceof Error ? importErr.message : String(importErr);
       log.info('contexgin not available, using fallback', { error: msg });
-      send(transport, {
-        type: 'boot_context',
-        source: 'local-fallback',
+      const fallback = {
+        source: 'local-fallback' as const,
         sourceCount: 0,
         tokenCount: 0,
         tokenBudget: DEFAULT_TOKEN_BUDGET,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
+        sources: [] as Array<{ path: string; kind: string }>,
+        included: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+        trimmed: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+      };
+      send(transport, { type: 'boot_context', ...fallback });
+      const s = registry.get(clientId);
+      if (s) s.bootContext = fallback;
       return;
     }
 
@@ -819,16 +825,18 @@ async function _startChatInner(
       // Validate the compiled object shape
       if (!compiled || typeof compiled !== 'object') {
         log.warn('contexgin compile() returned unexpected shape', { compiled });
-        send(transport, {
-          type: 'boot_context',
-          source: 'local-fallback',
+        const fallback = {
+          source: 'local-fallback' as const,
           sourceCount: 0,
           tokenCount: 0,
           tokenBudget: tokenBudget,
-          sources: [],
-          included: [],
-          trimmed: [],
-        });
+          sources: [] as Array<{ path: string; kind: string }>,
+          included: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+          trimmed: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+        };
+        send(transport, { type: 'boot_context', ...fallback });
+        const s = registry.get(clientId);
+        if (s) s.bootContext = fallback;
         return;
       }
 
@@ -879,9 +887,8 @@ async function _startChatInner(
       const trimmed = extractSections(rawTrimmed);
       const fullMarkdown = typeof obj.bootPayload === 'string' ? obj.bootPayload : undefined;
 
-      send(transport, {
-        type: 'boot_context',
-        source: 'contexgin',
+      const bootPayload = {
+        source: 'contexgin' as const,
         sourceCount: sources.length,
         tokenCount: bootTokens,
         tokenBudget: tokenBudget,
@@ -889,20 +896,27 @@ async function _startChatInner(
         included,
         trimmed,
         fullMarkdown,
-      });
+      };
+      send(transport, { type: 'boot_context', ...bootPayload });
+
+      // Cache in ManagedSession for replay on reconnect/switch
+      const s = registry.get(clientId);
+      if (s) s.bootContext = bootPayload;
     } catch (err: unknown) {
       const msg = err instanceof Error ? err.message : String(err);
       log.warn('boot context compilation failed', { error: msg });
-      send(transport, {
-        type: 'boot_context',
-        source: 'local-fallback',
+      const fallbackPayload = {
+        source: 'local-fallback' as const,
         sourceCount: 0,
         tokenCount: 0,
         tokenBudget: tokenBudget,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
+        sources: [] as Array<{ path: string; kind: string }>,
+        included: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+        trimmed: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+      };
+      send(transport, { type: 'boot_context', ...fallbackPayload });
+      const s = registry.get(clientId);
+      if (s) s.bootContext = fallbackPayload;
     }
   })();
   capturePromptComparison(wtId, cwd, systemPromptAppend, repoWorktrees).catch(() => {});
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -831,8 +831,18 @@ async function _startChatInner(
           tokenCount: 0,
           tokenBudget: tokenBudget,
           sources: [] as Array<{ path: string; kind: string }>,
-          included: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
-          trimmed: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
+          included: [] as Array<{
+            source: string;
+            heading: string;
+            tokens: number;
+            content: string;
+          }>,
+          trimmed: [] as Array<{
+            source: string;
+            heading: string;
+            tokens: number;
+            content: string;
+          }>,
         };
         send(transport, { type: 'boot_context', ...fallback });
         const s = registry.get(clientId);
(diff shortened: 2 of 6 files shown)


server/ws-handler-v2.ts
```diff
@@ -280,6 +280,14 @@ export function handleReconnect(
           running,
         });
 
+        // Re-send cached boot_context so pills reappear after reconnect
```
dimakis, line 283: 🟡 **bugs**: Reconnect boot_context replay lacks `sessionId`. The reconnect handler loops over all tracked sessions and sends boot_context for every running session. These messages have no `sessionId`, so the client's event filter treats them as global and applies each to the current messages state (SET_BOOT_CONTEXT overwrites). If a client tracks multiple running sessions, a non-active session's boot_context will overwrite the active session's. Fix: add `sessionId: entry.sessionId` to the replayed boot_context message, matching how other per-session events are tagged. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -371,6 +379,27 @@ export async function handleSwitchSession(
         },
       });
 
+      // Re-send boot_context so pills appear on session switch.
+      // Hot path: running session in SessionRegistry (in-memory cache).
+      // Cold path: ended session — read serialized JSON from EventStore.
+      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
+      if (found?.session?.bootContext) {
```
dimakis, line 386: 🟡 **bugs**: Same sessionId omission on switch_session boot_context replay. Both the hot path (in-memory) and cold path (JSON.parse) send boot_context without a `sessionId`. This works today because switchSession resets messages state before the reply arrives, but it's inconsistent with the multiplexed protocol where all session-scoped events should carry `sessionId` for client demuxing. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -371,6 +379,27 @@ export async function handleSwitchSession(
...
+        ctx.connRegistry.get(connectionId)?.transport.send({
+          type: 'boot_context',
+          ...found.session.bootContext,
+        });
+      } else if (sessionMeta.bootContext) {
+        try {
+          const parsed = JSON.parse(sessionMeta.bootContext);
+          ctx.connRegistry.get(connectionId)?.transport.send({
+            type: 'boot_context',
+            ...parsed,
```
dimakis, line 396: 🟡 **unsafe_assumptions**: Cold-path JSON.parse has no shape validation. `sessionMeta.bootContext` is parsed and spread directly into the WS message. If the stored JSON was written by an older code version with a different shape, or is corrupted, the client receives an unvalidated payload. The protocol-parser on the client side is defensive (checks each field type), so this isn't exploitable, but a basic typeof check on `parsed` (must be a non-null object) would prevent spreading a string or array into the message. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -371,6 +379,27 @@ export async function handleSwitchSession(
...
+          });
+        } catch {
+          // Invalid JSON — skip
+        }
+      }
```
dimakis, line 401: 🔵 **style**: Silent catch on JSON.parse — project conventions use `log.warn()` for recoverable error paths. Adding a warn log with sessionId and the error message would help diagnose boot_context replay failures. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -281,6 +284,19 @@ export class EventStore {
     }
   }
 
+  private migrateBootContext(db: Database.Database): void {
```
dimakis, line 287: 🔵 **style**: Migration method `migrateBootContext` also adds the `agent_name` column. Consider renaming to `migrateBootContextAndAgentName` or `migrateSessionAgentFields` to accurately reflect its scope. `[fixable]`

dimakis (commented):
## Centaur Review

Found **7** issue(s) (5 warning).


### `server/ws-handler-v2.ts`
_Solid implementation of boot_context persistence with correct serialization paths. Main concern: replayed boot_context messages lack `sessionId`, which breaks the multiplexed protocol contract and can cause cross-session bleed during multi-session reconnect. Missing test coverage for the new replay behaviors._

- 🟡 **bugs** (L283): Reconnect boot_context replay lacks `sessionId`. The reconnect handler loops over all tracked sessions and sends boot_context for every running session. These messages have no `sessionId`, so the client's event filter treats them as global and applies each to the current messages state (SET_BOOT_CONTEXT overwrites). If a client tracks multiple running sessions, a non-active session's boot_context will overwrite the active session's. Fix: add `sessionId: entry.sessionId` to the replayed boot_context message, matching how other per-session events are tagged. `[fixable]`
- 🟡 **bugs** (L386): Same sessionId omission on switch_session boot_context replay. Both the hot path (in-memory) and cold path (JSON.parse) send boot_context without a `sessionId`. This works today because switchSession resets messages state before the reply arrives, but it's inconsistent with the multiplexed protocol where all session-scoped events should carry `sessionId` for client demuxing. `[fixable]`
- 🟡 **unsafe_assumptions** (L396): Cold-path JSON.parse has no shape validation. `sessionMeta.bootContext` is parsed and spread directly into the WS message. If the stored JSON was written by an older code version with a different shape, or is corrupted, the client receives an unvalidated payload. The protocol-parser on the client side is defensive (checks each field type), so this isn't exploitable, but a basic typeof check on `parsed` (must be a non-null object) would prevent spreading a string or array into the message. `[fixable]`
- 🔵 **style** (L401): Silent catch on JSON.parse — project conventions use `log.warn()` for recoverable error paths. Adding a warn log with sessionId and the error message would help diagnose boot_context replay failures. `[fixable]`


_Solid implementation of boot_context persistence with correct serialization paths. Main concern: replayed boot_context messages lack `sessionId`, which breaks the multiplexed protocol contract and can cause cross-session bleed during multi-session reconnect. Missing test coverage for the new replay behaviors._

- 🟡 **missing_tests**: No test coverage for boot_context replay in handleReconnect or handleSwitchSession. Both hot path (in-memory ManagedSession.bootContext) and cold path (EventStore JSON parse) are untested. Given the project's TDD mandate, these behaviors should have tests: (1) reconnect sends cached boot_context for running sessions, (2) switch_session sends boot_context from registry (hot) or EventStore (cold), (3) switch_session handles invalid JSON gracefully. `[fixable]`


_Solid implementation of boot_context persistence with correct serialization paths. Main concern: replayed boot_context messages lack `sessionId`, which breaks the multiplexed protocol contract and can cause cross-session bleed during multi-session reconnect. Missing test coverage for the new replay behaviors._

- 🟡 **missing_tests**: No test for agentName and bootContext persistence round-trip in EventStore. The existing upsertSession tests cover goalId, wtId, branch, cwd, etc. — the new fields should follow suit with insert, update, and retrieval assertions. `[fixable]`


_Solid implementation of boot_context persistence with correct serialization paths. Main concern: replayed boot_context messages lack `sessionId`, which breaks the multiplexed protocol contract and can cause cross-session bleed during multi-session reconnect. Missing test coverage for the new replay behaviors._

- 🔵 **style** (L287): Migration method `migrateBootContext` also adds the `agent_name` column. Consider renaming to `migrateBootContextAndAgentName` or `migrateSessionAgentFields` to accurately reflect its scope. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -280,6 +280,14 @@ export function handleReconnect(
           running,
         });
 
+        // Re-send cached boot_context so pills reappear after reconnect
+        if (found && running && found.session?.bootContext) {
```
dimakis, line 284: 🟡 **bugs**: handleReconnect only replays boot_context for running sessions (hot path). Non-running sessions have no cold path — unlike handleSwitchSession which falls back to EventStore's serialized bootContext. Since boot_context is sent via raw `send()` (not `sendOrBuffer`), it is NOT stored in the EventStore event stream, so reconnecting to an ended session will never show boot context pills. Add an `else` branch that reads `ctx.eventStore.getSession(entry.sessionId)?.bootContext` and parses it, mirroring the cold path in handleSwitchSession (lines 391–401). `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -371,6 +379,27 @@ export async function handleSwitchSession(
...
+      // Cold path: ended session — read serialized JSON from EventStore.
+      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
+      if (found?.session?.bootContext) {
+        ctx.connRegistry.get(connectionId)?.transport.send({
+          type: 'boot_context',
+          ...found.session.bootContext,
+        });
+      } else if (sessionMeta.bootContext) {
+        try {
+          const parsed = JSON.parse(sessionMeta.bootContext);
```
dimakis, line 393: 🟡 **unsafe_assumptions**: The cold path in handleSwitchSession does `JSON.parse(sessionMeta.bootContext)` and spreads the result into a WS message with no shape validation. If the stored JSON has a `type` key (which bootContext payloads don't, but could via corruption or future changes), it would overwrite the `type: 'boot_context'` set on line 394. Consider parsing into a variable and deleting `parsed.type` before spreading, or validating the parsed shape. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -371,6 +379,27 @@ export async function handleSwitchSession(
...
+          ctx.connRegistry.get(connectionId)?.transport.send({
+            type: 'boot_context',
+            ...parsed,
+          });
+        } catch {
```
dimakis, line 398: 🔵 **style**: The `catch` block silently swallows JSON parse errors with no logging. Add `log.warn('failed to parse stored bootContext', { sessionId: msg.sessionId })` for debuggability — corrupted bootContext would otherwise be invisible. `[fixable]`

server/chat.ts
```diff
@@ -726,6 +730,7 @@ async function _startChatInner(
       ...(worktreePath ? { wtId } : {}),
       ...(options.telosTaskId ? { telosTaskId: options.telosTaskId } : {}),
       ...(existingMeta ? { updatedAt: existingMeta.updatedAt } : {}),
+      agentName,
```
dimakis, line 733: 🟡 **bugs**: The resume upsert at line 724–734 persists `agentName` but not `bootContext`. On resume, if the prior session had a cached bootContext in the EventStore, this upsert overwrites session metadata without preserving bootContext. The field won't be lost (SQL UPDATE only sets provided fields), but this is an asymmetry worth noting — the resume path reconstructs the boot context from scratch (the IIFE runs again), so the stale stored value will eventually be overwritten. No data loss, but if the boot context IIFE fails on resume, the old value persists, which is actually correct. No fix needed — noting for clarity.

server/chat.ts
```diff
@@ -777,16 +781,18 @@ async function _startChatInner(
     } catch (importErr: unknown) {
       const msg = importErr instanceof Error ? importErr.message : String(importErr);
       log.info('contexgin not available, using fallback', { error: msg });
-      send(transport, {
-        type: 'boot_context',
-        source: 'local-fallback',
+      const fallback = {
```
dimakis, line 784: 🔵 **style**: The fallback boot_context object is duplicated three times (lines 784–792, 830–845, 921–928) with identical shape but verbose inline type annotations (`[] as Array<{ source: string; heading: string; tokens: number; content: string }>`). Extract a helper like `makeEmptyBootContext(tokenBudget: number)` to reduce duplication and the risk of shape drift between the three copies. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Solid feature with correct serialization boundaries, but handleReconnect is missing the cold path for ended sessions (boot_context pills won't appear after reconnecting to a finished session), and there are no tests for the new reconnect/switch replay logic._

- 🟡 **bugs** (L284): handleReconnect only replays boot_context for running sessions (hot path). Non-running sessions have no cold path — unlike handleSwitchSession which falls back to EventStore's serialized bootContext. Since boot_context is sent via raw `send()` (not `sendOrBuffer`), it is NOT stored in the EventStore event stream, so reconnecting to an ended session will never show boot context pills. Add an `else` branch that reads `ctx.eventStore.getSession(entry.sessionId)?.bootContext` and parses it, mirroring the cold path in handleSwitchSession (lines 391–401). `[fixable]`
- 🟡 **unsafe_assumptions** (L393): The cold path in handleSwitchSession does `JSON.parse(sessionMeta.bootContext)` and spreads the result into a WS message with no shape validation. If the stored JSON has a `type` key (which bootContext payloads don't, but could via corruption or future changes), it would overwrite the `type: 'boot_context'` set on line 394. Consider parsing into a variable and deleting `parsed.type` before spreading, or validating the parsed shape. `[fixable]`
- 🔵 **style** (L398): The `catch` block silently swallows JSON parse errors with no logging. Add `log.warn('failed to parse stored bootContext', { sessionId: msg.sessionId })` for debuggability — corrupted bootContext would otherwise be invisible. `[fixable]`
- 🟡 **missing_tests**: No tests cover boot_context replay on reconnect or session switch. Key scenarios to test: (1) handleReconnect re-sends bootContext from in-memory ManagedSession for running sessions, (2) handleSwitchSession re-sends from in-memory cache (hot path), (3) handleSwitchSession re-sends from EventStore JSON (cold path), (4) cold path gracefully handles invalid JSON. The existing test files for ws-handler-v2 and event-store have no bootContext coverage. `[fixable]`


### `server/chat.ts`
_Solid feature with correct serialization boundaries, but handleReconnect is missing the cold path for ended sessions (boot_context pills won't appear after reconnecting to a finished session), and there are no tests for the new reconnect/switch replay logic._

- 🟡 **bugs** (L733): The resume upsert at line 724–734 persists `agentName` but not `bootContext`. On resume, if the prior session had a cached bootContext in the EventStore, this upsert overwrites session metadata without preserving bootContext. The field won't be lost (SQL UPDATE only sets provided fields), but this is an asymmetry worth noting — the resume path reconstructs the boot context from scratch (the IIFE runs again), so the stale stored value will eventually be overwritten. No data loss, but if the boot context IIFE fails on resume, the old value persists, which is actually correct. No fix needed — noting for clarity.
- 🔵 **style** (L784): The fallback boot_context object is duplicated three times (lines 784–792, 830–845, 921–928) with identical shape but verbose inline type annotations (`[] as Array<{ source: string; heading: string; tokens: number; content: string }>`). Extract a helper like `makeEmptyBootContext(tokenBudget: number)` to reduce duplication and the risk of shape drift between the three copies. `[fixable]`


_Solid feature with correct serialization boundaries, but handleReconnect is missing the cold path for ended sessions (boot_context pills won't appear after reconnecting to a finished session), and there are no tests for the new reconnect/switch replay logic._

- 🔵 **missing_tests**: The migrateBootContext migration and the bootContext/agentName fields in upsertSession and rowToSession have no test coverage. A round-trip test (upsert with bootContext JSON string → getSession → verify deserialized fields) would prevent regressions in the persistence layer. `[fixable]`

dimakis (commented):
Please review this PR

server/ws-handler-v2.ts
dimakis, line 284: 🟡 **bugs**: handleReconnect only replays boot_context from in-memory SessionRegistry (hot path) when `running` is true, but does NOT fall back to EventStore for ended sessions. In contrast, handleSwitchSession (line 391) correctly implements a cold path fallback via `sessionMeta.bootContext`. A user reconnecting after a session has ended will not see boot_context pills, even though the data is persisted in the EventStore. The fix would be to add an `else` branch that reads from `ctx.eventStore.getSession(entry.sessionId)?.bootContext` and parses it, mirroring the switch_session logic. `[fixable]`

server/ws-handler-v2.ts
dimakis, line 393: 🔵 **unsafe_assumptions**: JSON.parse of `sessionMeta.bootContext` could be exploited if the database value is tampered with or corrupted, resulting in arbitrary object shapes being spread into a WebSocket message. Consider validating the parsed shape (e.g., checking for expected keys like `source`, `sourceCount`) before sending to the client, or at minimum using a Zod schema. Low risk since the DB is local/trusted, but worth noting for defense-in-depth. `[fixable]`

server/chat.ts
dimakis, line 784: 🔵 **style**: The fallback boot_context object is constructed identically in three places (import error at ~784, invalid compile result at ~828, and catch block at ~918), each with verbose inline type annotations like `[] as Array<{ source: string; heading: string; tokens: number; content: string }>`. Extract a shared `makeFallbackBootContext(tokenBudget: number)` helper to eliminate the repetition and reduce the diff noise. `[fixable]`

server/chat.ts
```diff
@@ -819,16 +825,28 @@ async function _startChatInner(
...
+          trimmed: [] as Array<{
+            source: string;
+            heading: string;
+            tokens: number;
+            content: string;
+          }>,
+        };
+        send(transport, { type: 'boot_context', ...fallback });
+        const s = registry.get(clientId);
+        if (s) s.bootContext = fallback;
```
dimakis, line 849: 🔵 **style**: The pattern `const s = registry.get(clientId); if (s) s.bootContext = fallbackPayload;` is repeated three times in the boot context compilation block (lines ~795, ~849, ~929). This could be a single helper or moved after the try/catch to always set whichever payload was sent. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Solid feature with correct serialization flow, but handleReconnect is missing the cold-path fallback for ended sessions (asymmetric with handleSwitchSession), and no test coverage was added for the new persistence or replay logic._

- 🟡 **bugs** (L284): handleReconnect only replays boot_context from in-memory SessionRegistry (hot path) when `running` is true, but does NOT fall back to EventStore for ended sessions. In contrast, handleSwitchSession (line 391) correctly implements a cold path fallback via `sessionMeta.bootContext`. A user reconnecting after a session has ended will not see boot_context pills, even though the data is persisted in the EventStore. The fix would be to add an `else` branch that reads from `ctx.eventStore.getSession(entry.sessionId)?.bootContext` and parses it, mirroring the switch_session logic. `[fixable]`
- 🔵 **unsafe_assumptions** (L393): JSON.parse of `sessionMeta.bootContext` could be exploited if the database value is tampered with or corrupted, resulting in arbitrary object shapes being spread into a WebSocket message. Consider validating the parsed shape (e.g., checking for expected keys like `source`, `sourceCount`) before sending to the client, or at minimum using a Zod schema. Low risk since the DB is local/trusted, but worth noting for defense-in-depth. `[fixable]`


_Solid feature with correct serialization flow, but handleReconnect is missing the cold-path fallback for ended sessions (asymmetric with handleSwitchSession), and no test coverage was added for the new persistence or replay logic._

- 🟡 **missing_tests**: No tests cover boot_context replay on reconnect or session switch. The existing ws-handler-v2 test suite has no assertions for bootContext. Key scenarios to cover: (1) reconnect with running session replays boot_context, (2) switch to running session uses hot path, (3) switch to ended session uses cold path (JSON.parse from EventStore), (4) malformed JSON in EventStore is gracefully skipped. `[fixable]`


_Solid feature with correct serialization flow, but handleReconnect is missing the cold-path fallback for ended sessions (asymmetric with handleSwitchSession), and no test coverage was added for the new persistence or replay logic._

- 🟡 **missing_tests**: No tests cover the new `agent_name` and `boot_context` columns — neither the migration (`migrateBootContext`) nor the upsert/read round-trip for these fields. The existing event-store test suite has no bootContext assertions. `[fixable]`


### `server/chat.ts`
_Solid feature with correct serialization flow, but handleReconnect is missing the cold-path fallback for ended sessions (asymmetric with handleSwitchSession), and no test coverage was added for the new persistence or replay logic._

- 🔵 **style** (L784): The fallback boot_context object is constructed identically in three places (import error at ~784, invalid compile result at ~828, and catch block at ~918), each with verbose inline type annotations like `[] as Array<{ source: string; heading: string; tokens: number; content: string }>`. Extract a shared `makeFallbackBootContext(tokenBudget: number)` helper to eliminate the repetition and reduce the diff noise. `[fixable]`
- 🔵 **style** (L849): The pattern `const s = registry.get(clientId); if (s) s.bootContext = fallbackPayload;` is repeated three times in the boot context compilation block (lines ~795, ~849, ~929). This could be a single helper or moved after the try/catch to always set whichever payload was sent. `[fixable]`

PR #361, opened 2026-05-30
refactor(server): fetch boot context from ContexGin HTTP API

dimakis:
## Summary

- Replace file-based boot context loading with HTTP fetch from running ContexGin server
- Local Python fallback when ContexGin is unreachable
- Centaur review findings addressed

Replaces #358 (CI was stuck on stale branch).

## Test plan

- [ ] Start session with ContexGin running — verify boot context loads via HTTP
- [ ] Start session without ContexGin — verify local fallback works
- [ ] Verify boot context pill renders correctly

### DIFF
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -83,6 +83,7 @@ describe('fetchBootContext', () => {
         boot: {
           content: '# Boot payload\nContext here.',
           tokens: 11297,
+          tokenBudget: 11297,
           sources: ['CONSTITUTION.md', 'memory/Profile/Principles.md'],
         },
       }),
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -5,6 +5,25 @@ import type { BootContextMessage } from '../chat.js';
 const mockFetch = vi.fn();
 vi.stubGlobal('fetch', mockFetch);
 
+// Mock child_process for local fallback tests
+const mockExecFile = vi.fn();
+vi.mock('child_process', () => ({
+  execFileSync: vi.fn(),
+  execFile: mockExecFile,
+}));
+
+// Mock util.promisify to return our mock
+vi.mock('util', async (importOriginal) => {
+  const actual = (await importOriginal()) as Record<string, unknown>;
+  return {
+    ...actual,
+    promisify: (fn: unknown) => {
+      if (fn === mockExecFile) return mockExecFile;
+      return (actual.promisify as (fn: unknown) => unknown)(fn);
+    },
+  };
+});
+
 // Mock logger to suppress output
 vi.mock('../logger.js', () => ({
   createLogger: () => ({
@@ -93,38 +112,66 @@ describe('fetchBootContext', () => {
     });
   });
 
-  it('returns local-fallback when ContexGin is unreachable', async () => {
+  it('falls back to local Python script when ContexGin is unreachable', async () => {
     mockFetch.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: '# Local boot context\nFallback content.' }),
+    });
 
-    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
-    expect(result.sourceCount).toBe(0);
-    expect(result.tokenCount).toBe(0);
+    expect(result.sourceCount).toBe(5);
+    expect(result.fullMarkdown).toBe('# Local boot context\nFallback content.');
+    expect(result.tokenCount).toBeGreaterThan(0);
+    expect(result.sources).toHaveLength(5);
+    expect(result.sources[0]).toEqual({
+      path: 'memory/Profile/Working Style.md',
+      kind: 'profile',
+    });
   });
 
-  it('returns local-fallback on non-200 response', async () => {
+  it('falls back to local Python script on non-200 response', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: false,
       status: 404,
       text: async () => '{"error":"Agent not found"}',
     });
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: 'fallback content' }),
+    });
 
-    const result = await fetchBootContext('nonexistent-agent', CONTEXGIN_URL);
+    const result = await fetchBootContext('nonexistent-agent', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
-    expect(result.sourceCount).toBe(0);
+    expect(result.fullMarkdown).toBe('fallback content');
   });
 
-  it('returns local-fallback when response lacks boot field', async () => {
+  it('falls back to local Python script when response lacks boot field', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: true,
       json: async () => ({ agent: 'test', identity: {} }),
     });
+    mockExecFile.mockResolvedValueOnce({
+      stdout: JSON.stringify({ additionalContext: 'fallback' }),
+    });
 
-    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
 
     expect(result.source).toBe('local-fallback');
+    expect(result.fullMarkdown).toBe('fallback');
+  });
+
+  it('returns zero-value fallback when both ContexGin and Python script fail', async () => {
+    mockFetch.mockRejectedValueOnce(new Error('ECONNREFUSED'));
+    mockExecFile.mockRejectedValueOnce(new Error('script not found'));
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
+
+    expect(result.source).toBe('local-fallback');
+    expect(result.sourceCount).toBe(0);
+    expect(result.tokenCount).toBe(0);
+    expect(result.fullMarkdown).toBeUndefined();
   });
 
   it('handles empty sources array', async () => {
@@ -169,6 +216,7 @@ describe('fetchBootContext', () => {
     process.env.CONTEXGIN_URL = 'http://test-host:9999';
 
     mockFetch.mockRejectedValueOnce(new Error('timeout'));
+    mockExecFile.mockRejectedValueOnce(new Error('no script'));
 
     await fetchBootContext('mitzo-conversational');
 
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -1,0 +1,186 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { BootContextMessage } from '../chat.js';
+
+// Mock fetch globally
+const mockFetch = vi.fn();
+vi.stubGlobal('fetch', mockFetch);
+
+// Mock logger to suppress output
+vi.mock('../logger.js', () => ({
+  createLogger: () => ({
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  }),
+}));
+
+// Mock the event store to avoid SQLite initialization
+class FakeEventStore {
+  recordEvent = vi.fn();
+  getEvents = vi.fn().mockReturnValue([]);
+  getSession = vi.fn().mockReturnValue(null);
+  upsertSession = vi.fn();
+}
+vi.mock('../event-store.js', () => ({
+  EventStore: FakeEventStore,
+}));
+
+// Mock repo-config
+vi.mock('../repo-config.js', () => ({
+  loadRepoConfig: vi.fn(() => ({
+    contextBlocks: {},
+    quickActions: [],
+    venvPaths: [],
+    resolvedVenvPaths: [],
+    allowedPaths: [],
+    roots: [],
+    toolTierOverrides: {},
+    inboxPath: '',
+    resolvedInboxPath: '',
+    repos: {},
+    isolation: true,
+  })),
+}));
+
+const { fetchBootContext } = await import('../chat.js');
+
+beforeEach(() => {
+  vi.clearAllMocks();
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe('fetchBootContext', () => {
+  const CONTEXGIN_URL = 'http://localhost:4195';
+
+  it('returns contexgin boot context on successful response', async () => {
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
+        agent: 'mitzo-conversational',
+        boot: {
+          content: '# Boot payload\nContext here.',
+          tokens: 11297,
+          sources: ['CONSTITUTION.md', 'memory/Profile/Principles.md'],
+        },
+      }),
+    });
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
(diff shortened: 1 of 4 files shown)


dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid simplification from dynamic-import library to HTTP API with graceful fallback chain. Main concerns: `tokenBudget` semantics changed (now always equals `tokenCount` instead of reflecting configured budget), and `included`/`trimmed` section metadata is permanently empty in the new path._

- 🟡 **regressions** (L897): The call site `fetchBootContext(agentName)` omits `contexginUrl` and `repoRoot`, so both use defaults. The default `repoRoot` is `BASE_REPO` (which could be `''` if `REPO_PATH` is unset). If ContexGin is down and `BASE_REPO` is empty, `localBootContextFallback` will try to exec `scripts/build_boot_context.py` relative to cwd — but this is caught and returns the zero-value fallback, so it degrades gracefully. Not a bug, but worth noting the old code passed `cwd` (the resolved working directory) rather than `BASE_REPO` for the fallback script. If ContexGin is offline and the session uses a worktree `cwd`, the fallback script runs against `BASE_REPO` (the parent repo) instead of the worktree — this is likely correct (the script reads canonical source files), but the behavior changed from the old code which used `cwd`.
- 🟡 **regressions** (L212): The old code set `tokenBudget` from the agent recipe's `context.boot.tokenBudget` field (defaulting to 8000). The new code sets `tokenBudget` equal to `tokenCount` (the actual token count from the server response). This means `tokenBudget` no longer reflects the configured budget — it just mirrors `tokenCount`. Any frontend logic that compares `tokenCount` vs `tokenBudget` (e.g., to show utilization percentage) will always see 100%. The client-side protocol parser at `packages/client/src/protocol-parser.ts:260` does forward `tokenBudget` to the store, so if anything renders budget vs. used, it's now a no-op comparison. `[fixable]`
- 🟡 **unsafe_assumptions** (L141): The `localBootContextFallback` hardcodes `sourceCount: 5` and a fixed list of 5 source paths, with a comment saying it's coupled to `SOURCE_PATHS` in `scripts/build_boot_context.py`. If that script changes its source list, this will silently drift out of sync. Consider reading the script's output for source metadata instead of duplicating the list, or at minimum making the hardcoded count derive from the array length (`sources.length` instead of the literal `5`). `[fixable]`
- 🔵 **style** (L898): The double cast `msg as unknown as Record<string, unknown>` is unnecessarily defensive. `BootContextMessage` is a plain interface with only primitive/array fields — a single `msg as Record<string, unknown>` suffices since the types are structurally compatible. Alternatively, widen the `send()` helper to accept `BootContextMessage` directly. `[fixable]`
- 🔵 **regressions** (L214): The old code populated `included` and `trimmed` arrays with section-level metadata (source file, heading path, token estimates, content) from the contexgin compile result. The new HTTP API path always returns empty arrays for both. If the ContexGin HTTP API does (or will) return this data, it's being silently dropped. If the API intentionally omits it, this is fine — but the `BootContextMessage` interface still declares these fields, and the client parser still parses them, so there's dead plumbing.


_Solid simplification from dynamic-import library to HTTP API with graceful fallback chain. Main concerns: `tokenBudget` semantics changed (now always equals `tokenCount` instead of reflecting configured budget), and `included`/`trimmed` section metadata is permanently empty in the new path._

- 🔵 **missing_tests**: No test covers the `AbortSignal.timeout(5000)` behavior — e.g., verifying that a slow ContexGin response triggers the fallback within the timeout window. This is difficult to test in unit tests but worth noting as an untested edge case.

server/chat.ts
```diff
@@ -764,147 +893,10 @@ async function _startChatInner(
...
-        tokenCount: 0,
-        tokenBudget: tokenBudget,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
-    }
-  })();
+  // Fire-and-forget: fetch boot context from running ContexGin server
+  fetchBootContext(agentName)
```
dimakis, line 914: 🟡 **regressions**: The call site `fetchBootContext(agentName)` omits `contexginUrl` and `repoRoot`, so both use defaults. The default `repoRoot` is `BASE_REPO` (which could be `''` if `REPO_PATH` is unset). If ContexGin is down and `BASE_REPO` is empty, `localBootContextFallback` will try to exec `scripts/build_boot_context.py` relative to cwd — but this is caught and returns the zero-value fallback, so it degrades gracefully. Not a bug, but worth noting the old code passed `cwd` (the resolved working directory) rather than `BASE_REPO` for the fallback script. If ContexGin is offline and the session uses a worktree `cwd`, the fallback script runs against `BASE_REPO` (the parent repo) instead of the worktree — this is likely correct (the script reads canonical source files), but the behavior changed from the old code which used `cwd`.

server/chat.ts
```diff
@@ -93,6 +93,135 @@ export const eventStore = initEventStore();
...
+      .map((s) => ({ path: s, kind: 'reference' }));
+
+    const fullMarkdown = typeof boot.content === 'string' ? boot.content : undefined;
+
+    return {
+      type: 'boot_context',
+      source: 'contexgin',
+      sourceCount: sources.length,
+      tokenCount: bootTokens,
+      tokenBudget: bootTokens, // server compiles to its own budget — report actual
```
dimakis: 🟡 **regressions**: The old code set `tokenBudget` from the agent recipe's `context.boot.tokenBudget` field (defaulting to 8000). The new code sets `tokenBudget` equal to `tokenCount` (the actual token count from the server response). This means `tokenBudget` no longer reflects the configured budget — it just mirrors `tokenCount`. Any frontend logic that compares `tokenCount` vs `tokenBudget` (e.g., to show utilization percentage) will always see 100%. The client-side protocol parser at `packages/client/src/protocol-parser.ts:260` does forward `tokenBudget` to the store, so if anything renders budget vs. used, it's now a no-op comparison. `[fixable]`

server/chat.ts
```diff
@@ -93,6 +93,135 @@ export const eventStore = initEventStore();
...
+      timeout: 5000,
+    });
+    const parsed = JSON.parse(stdout) as { additionalContext?: string };
+    const content = parsed.additionalContext ?? '';
+    // Rough token estimate: ~4 chars per token
+    const tokens = Math.ceil(content.length / 4);
+    return {
+      type: 'boot_context',
+      source: 'local-fallback',
+      // Coupled to SOURCE_PATHS in scripts/build_boot_context.py — update if that list changes
```
dimakis: 🟡 **unsafe_assumptions**: The `localBootContextFallback` hardcodes `sourceCount: 5` and a fixed list of 5 source paths, with a comment saying it's coupled to `SOURCE_PATHS` in `scripts/build_boot_context.py`. If that script changes its source list, this will silently drift out of sync. Consider reading the script's output for source metadata instead of duplicating the list, or at minimum making the hardcoded count derive from the array length (`sources.length` instead of the literal `5`). `[fixable]`

server/chat.ts
```diff
@@ -764,147 +893,10 @@ async function _startChatInner(
...
-        tokenBudget: tokenBudget,
-        sources: [],
-        included: [],
-        trimmed: [],
-      });
-    }
-  })();
+  // Fire-and-forget: fetch boot context from running ContexGin server
+  fetchBootContext(agentName)
+    .then((msg) => send(transport, msg as unknown as Record<string, unknown>))
```
dimakis: 🔵 **style**: The double cast `msg as unknown as Record<string, unknown>` is unnecessarily defensive. `BootContextMessage` is a plain interface with only primitive/array fields — a single `msg as Record<string, unknown>` suffices since the types are structurally compatible. Alternatively, widen the `send()` helper to accept `BootContextMessage` directly. `[fixable]`

server/chat.ts
```diff
@@ -93,6 +93,135 @@ export const eventStore = initEventStore();
...
+    const fullMarkdown = typeof boot.content === 'string' ? boot.content : undefined;
+
+    return {
+      type: 'boot_context',
+      source: 'contexgin',
+      sourceCount: sources.length,
+      tokenCount: bootTokens,
+      tokenBudget: bootTokens, // server compiles to its own budget — report actual
+      sources,
+      included: [],
```
dimakis: 🔵 **regressions**: The old code populated `included` and `trimmed` arrays with section-level metadata (source file, heading path, token estimates, content) from the contexgin compile result. The new HTTP API path always returns empty arrays for both. If the ContexGin HTTP API does (or will) return this data, it's being silently dropped. If the API intentionally omits it, this is fine — but the `BootContextMessage` interface still declares these fields, and the client parser still parses them, so there's dead plumbing.

server/__tests__/boot-context.test.ts
```diff
@@ -0,0 +1,253 @@
...
+      `${CONTEXGIN_URL}/api/agents/mitzo-conversational/context`,
+      { signal: expect.any(AbortSignal) },
+    );
+
+    expect(result).toEqual<BootContextMessage>({
+      type: 'boot_context',
+      source: 'contexgin',
+      sourceCount: 2,
+      tokenCount: 11297,
+      tokenBudget: 11297,
```
dimakis, line 105: 🔴 **bugs**: Test expects `tokenBudget: 11297` but the mock response at line 84 has no `tokenBudget` field in `boot`. The implementation (chat.ts:199) defaults to `8000` when `boot.tokenBudget` is absent: `typeof boot.tokenBudget === 'number' ? boot.tokenBudget : 8000`. This test will fail. Either add `tokenBudget: 11297` to the mock's `boot` object, or change the expectation to `8000`. `[fixable]`

server/chat.ts
```diff
@@ -238,8 +380,8 @@ export function parseModelSpec(spec?: string): { model: string; effort: string |
 // --- Pure helper functions ---
 
 /** Send data via transport (isOpen guard is inside the transport). */
-function send(transport: SessionTransport, data: Record<string, unknown>) {
-  if (transport.isOpen()) transport.send(data);
+function send(transport: SessionTransport, data: Record<string, unknown> | BootContextMessage) {
+  if (transport.isOpen()) transport.send(data as Record<string, unknown>);
```
dimakis, line 384: 🔵 **style**: The `as Record<string, unknown>` cast works around `SessionTransport.send()` only accepting `Record<string, unknown>`. It's safe since `BootContextMessage` is a plain object, but the union type on `send()` (line 383) is misleading — it suggests type safety that the cast immediately discards. Consider either widening `SessionTransport.send()` to accept `unknown` or keeping `send()` typed as just `Record<string, unknown>` and casting at the call site instead. `[fixable]`

server/chat.ts
```diff
@@ -93,6 +93,148 @@ export const eventStore = initEventStore();
...
+    const execFileAsync = promisify(execFile);
+    const { stdout } = await execFileAsync('python3', [scriptPath, '--json'], {
+      cwd: repoRoot,
+      timeout: 5000,
+    });
+    const parsed = JSON.parse(stdout) as { additionalContext?: string };
+    const content = parsed.additionalContext ?? '';
+    // Rough token estimate: ~4 chars per token
+    const tokens = Math.ceil(content.length / 4);
+    const sources = [
```
dimakis, line 138: 🔵 **unsafe_assumptions**: The hardcoded source list in `localBootContextFallback` (lines 138-144) assumes the mgmt workspace structure (`memory/Profile/Working Style.md`, `CONSTITUTION.md`, `SERVICES.md`). These paths won't exist in other repos that might use Mitzo. Since this is explicitly a fallback for when ContexGin is down and scoped to the `REPO_PATH` workspace, it's acceptable — but a comment noting this is mgmt-specific would help future readers.

dimakis (commented):
## Centaur Review

Found **5** issue(s) (1 critical) (1 warning).


_The tokenBudget test/implementation mismatch (mock omits `tokenBudget` but test expects `11297`, implementation defaults to `8000`) will cause a test failure. The core refactor from npm-imported contexgin to HTTP API is sound with good fallback chains, but needs test coverage for `tokenBudget` parsing and `included`/`trimmed` section parsing._

- 🔴 **bugs** (L104): Test expects `tokenBudget: 11297` but the mock response at line 84 has no `tokenBudget` field in `boot`. The implementation (chat.ts:199) defaults to `8000` when `boot.tokenBudget` is absent: `typeof boot.tokenBudget === 'number' ? boot.tokenBudget : 8000`. This test will fail. Either add `tokenBudget: 11297` to the mock's `boot` object, or change the expectation to `8000`. `[fixable]`
- 🟡 **missing_tests**: No test case verifies that `tokenBudget` is correctly parsed when the API response *does* include `boot.tokenBudget`. The existing happy-path test omits it, so the `typeof boot.tokenBudget === 'number'` branch is never exercised. Add a test where the mock includes an explicit `tokenBudget` value distinct from `tokens` to confirm both fields are parsed independently. `[fixable]`
- 🔵 **missing_tests**: No test covers the `included`/`trimmed` section parsing (`parseSections` at chat.ts:210-218). The happy-path test receives no `included`/`trimmed` in the mock, so the array-filtering and field-coercion logic in `parseSections` is untested. A test with mock `included`/`trimmed` arrays containing objects with partial fields would cover the `String(s.source ?? '')` coercion paths. `[fixable]`


### `server/chat.ts`
_The tokenBudget test/implementation mismatch (mock omits `tokenBudget` but test expects `11297`, implementation defaults to `8000`) will cause a test failure. The core refactor from npm-imported contexgin to HTTP API is sound with good fallback chains, but needs test coverage for `tokenBudget` parsing and `included`/`trimmed` section parsing._

- 🔵 **style** (L384): The `as Record<string, unknown>` cast works around `SessionTransport.send()` only accepting `Record<string, unknown>`. It's safe since `BootContextMessage` is a plain object, but the union type on `send()` (line 383) is misleading — it suggests type safety that the cast immediately discards. Consider either widening `SessionTransport.send()` to accept `unknown` or keeping `send()` typed as just `Record<string, unknown>` and casting at the call site instead. `[fixable]`
- 🔵 **unsafe_assumptions** (L138): The hardcoded source list in `localBootContextFallback` (lines 138-144) assumes the mgmt workspace structure (`memory/Profile/Working Style.md`, `CONSTITUTION.md`, `SERVICES.md`). These paths won't exist in other repos that might use Mitzo. Since this is explicitly a fallback for when ContexGin is down and scoped to the `REPO_PATH` workspace, it's acceptable — but a comment noting this is mgmt-specific would help future readers.

server/chat.ts
dimakis: Fixed — added `DEFAULT_TOKEN_BUDGET = 8000` constant. The ContexGin path now reads `boot.tokenBudget` from the response (falling back to 8000), and the local fallback uses the constant. Budget no longer mirrors tokenCount.

server/chat.ts
dimakis: Fixed — extracted the sources array to a variable, now uses `sources.length` instead of the hardcoded literal `5`.

server/chat.ts
dimakis: Addressed — widened `send()` to accept `Record<string, unknown> | BootContextMessage`, removing the double cast at the call site. The internal `as Record<string, unknown>` remains because `SessionTransport.send()` requires it.

server/chat.ts
dimakis: Acknowledged — the fallback running against `BASE_REPO` instead of worktree `cwd` is indeed the correct behavior since the script reads canonical source files. No change needed.

server/chat.ts
dimakis: Fixed — `included` and `trimmed` are now parsed from the ContexGin HTTP response when present, instead of always returning empty arrays.

server/__tests__/boot-context.test.ts
dimakis: Fixed — added `tokenBudget: 11297` to the mock ContexGin response so the test validates end-to-end budget propagation.

server/chat.ts
dimakis: Fair point — the union type on `send()` is a compromise. `SessionTransport.send()` is typed as `Record<string, unknown>` (harness package), so we can't avoid the internal cast without widening the harness interface. Keeping the union at `send()` at least makes the call sites cleaner.

server/chat.ts
dimakis: Acknowledged — these paths are mgmt-specific and only used when ContexGin is offline. This is an acceptable coupling since Mitzo's primary deployment is the mgmt workspace.

PR #362, opened 2026-05-30
fix: state-based session routing to eliminate double-message bug

dimakis:
## Summary

- **Root cause**: `handleSwitchSession` didn't call `watch()`, so events from the query loop never reached the reconnected client. Combined with a `staleInMemory` check that could incorrectly kill a running query loop when EventStore and SessionRegistry diverged, the first message after navigating back to a session was silently swallowed.
- Replace the `staleInMemory` boolean check with Phase 3 state-based routing using `getSessionState()` as the single source of truth in both `handleSendV2` and `handleInterruptV2`
- Add `watch()` call to `handleSwitchSession` + return `running` state to client
- Use `stopChat()` (abort signal) instead of `registry.remove()` for zombie sessions

## Test plan

- [ ] 125/125 ws-handler-v2 tests pass (verified)
- [ ] Navigate away from active session and back — first message should get a response immediately
- [ ] Repeat with session that was idle >48h (ENDED state) — should resume cleanly
- [ ] Verify interrupt works after session switch without double-send

### DIFF
--- a/server/ws-handler-v2.ts
+++ b/server/ws-handler-v2.ts
@@ -361,7 +361,20 @@ export async function handleSwitchSession(
       if (prev && prev !== msg.sessionId) {
         ctx.connRegistry.unwatch(connectionId, prev);
       }
+      // Watch the session immediately so events from a running query loop
+      // reach this client before the first send. Without this, the client
+      // sits in a blind spot between switch_session and the first send —
+      // any events emitted by the query loop during that window are lost.
+      ctx.connRegistry.watch(connectionId, msg.sessionId);
       ctx.connRegistry.setActive(connectionId, msg.sessionId);
+
+      // Determine running state so the client can restore its UI correctly.
+      // Without this, the client shows an idle input after switching to a
+      // session whose query loop is still generating — leading to the
+      // "double message" bug where the first send queues behind a stale
+      // running=false state.
+      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
+      const running = found ? ctx.sessionRegistry.isActive(found.clientId) : false;
 
       ctx.connRegistry.get(connectionId)?.transport.send({
         type: 'session_switched',
@@ -370,6 +383,7 @@ export async function handleSwitchSession(
         cwd: sessionMeta.cwd,
         branch: sessionMeta.branch,
         wtId: sessionMeta.wtId,
+        running,
         tokens: {
           input: sessionMeta.inputTokens,
           output: sessionMeta.outputTokens,
@@ -382,7 +396,6 @@ export async function handleSwitchSession(
       // Re-send boot_context so pills appear on session switch.
       // Hot path: running session in SessionRegistry (in-memory cache).
       // Cold path: ended session — read serialized JSON from EventStore.
-      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
       if (found?.session?.bootContext) {
         ctx.connRegistry.get(connectionId)?.transport.send({
           type: 'boot_context',
@@ -400,7 +413,7 @@ export async function handleSwitchSession(
         }
       }
 
-      log.info('switch_session', { connectionId, sessionId: msg.sessionId });
+      log.info('switch_session', { connectionId, sessionId: msg.sessionId, running });
     },
   );
 }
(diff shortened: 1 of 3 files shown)


dimakis:
## Centaur Review

### `server/ws-handler-v2.ts`
_Solid fix for the double-message bug via state-based routing. The main concern is that handleReconnect still uses the old isActive boolean pattern, leaving a consistency gap. Missing test coverage for the new `running` field in handleSwitchSession._

- 🟡 **regressions** (L215): handleReconnect still uses the old `storeMeta.isActive` boolean + `sessionRegistry.remove()` pattern (lines 215-223) while handleSendV2/handleInterruptV2 have been migrated to the state-based `storeState` + `stopChat()` approach. This inconsistency means the same stale-state divergence that caused the double-message bug in send/interrupt could still manifest during reconnect. Consider migrating handleReconnect to the same state-based routing for consistency. `[fixable]`
- 🔵 **style** (L482): Case 1 comment lists states "ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED" but the condition `storeState !== 'ENDED'` also matches CLOSING. The comment should include CLOSING for accuracy, since CLOSING sessions are routable (they accept messages during graceful closeout). `[fixable]`


_Solid fix for the double-message bug via state-based routing. The main concern is that handleReconnect still uses the old isActive boolean pattern, leaving a consistency gap. Missing test coverage for the new `running` field in handleSwitchSession._

- 🟡 **missing_tests** (L362): handleSwitchSession tests don't assert the new `running` field in the `session_switched` response. The existing test uses `expect.objectContaining` which silently ignores unasserted fields. Should add test cases for: (1) running=true when session is active in registry, (2) running=false when session is not in registry. `[fixable]`


### `server/chat.ts`
_Solid fix for the double-message bug via state-based routing. The main concern is that handleReconnect still uses the old isActive boolean pattern, leaving a consistency gap. Missing test coverage for the new `running` field in handleSwitchSession._

- 🔵 **style** (L898): Double cast `msg as unknown as Record<string, unknown>` is unnecessarily permissive. `BootContextMessage` is an object type with string keys, so a single `as Record<string, unknown>` would suffice and preserve more type safety. The `as unknown` intermediate effectively turns off the type checker. `[fixable]`
- 🔵 **unsafe_assumptions** (L142): localBootContextFallback hardcodes a list of 5 source paths (lines 142-148) with the comment "coupled to SOURCE_PATHS in scripts/build_boot_context.py". If the Python script's source list changes, these will silently diverge. Consider extracting to a shared constant or reading the script output's actual source list. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -484,62 +497,83 @@ export function handleSendV2(
...
+          //
+          // Use the durable state column as the single source of truth
+          // for routing decisions. This replaces the old staleInMemory
+          // check that used the isActive boolean — that check could
+          // incorrectly kill a running query loop when the EventStore
+          // and SessionRegistry diverged, causing the "double message"
+          // bug on reattach.
+
+          // Case 1: Registry has the session AND state says it should
+          // be running (ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED).
```
dimakis: 🟡 **bugs**: Comment lists states routed to Case 1 as "ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED" but the condition `storeState !== 'ENDED' && storeState !== null` also admits CLOSING. A session in CLOSING state is winding down (CLOSING can only transition to ENDED per VALID_TRANSITIONS). Routing a new message via sendToChat to a closing session may race with the closeout — the agent could receive a new prompt while it's trying to commit/summarize. Either add CLOSING to the exclusion list alongside ENDED (treating it as a zombie) or document in the comment that CLOSING is intentionally treated as active. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -631,61 +665,72 @@ export function handleInterruptV2(
...
       }
 
-      if (isActive(found.clientId)) {
-        const storeMeta = ctx.eventStore.getSession(msg.sessionId);
-        const staleInMemory = storeMeta && !storeMeta.isActive;
+      // --- State-based routing (Phase 3) ---
+      //
+      // Same logic as handleSendV2: use the durable state column as
+      // the single source of truth for routing decisions.
+
```
dimakis: 🟡 **bugs**: Same CLOSING-state gap in handleInterruptV2: the comment says "ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED" but the condition also admits CLOSING. Should be consistent with whatever decision is made for handleSendV2. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -484,62 +497,83 @@ export function handleSendV2(
             span.setAttribute('session.state_mismatch', mismatch.details ?? 'unknown');
```
dimakis, line 498: 🔵 **style**: The block comments explaining the state-based routing (lines 497-511, 563-565, 576, and similarly in handleInterruptV2) are quite verbose for this codebase's conventions. CLAUDE.md says "default to writing no comments" and "only add one when the WHY is non-obvious." The initial Phase 3 header comment is useful context, but the per-case comments restate what the code already expresses through the condition structure. Consider trimming to one-line comments per case. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -361,15 +361,29 @@ export async function handleSwitchSession(
...
+      ctx.connRegistry.watch(connectionId, msg.sessionId);
       ctx.connRegistry.setActive(connectionId, msg.sessionId);
 
+      // Determine running state so the client can restore its UI correctly.
+      // Without this, the client shows an idle input after switching to a
+      // session whose query loop is still generating — leading to the
+      // "double message" bug where the first send queues behind a stale
+      // running=false state.
+      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
+      const running = found ? ctx.sessionRegistry.isActive(found.clientId) : false;
```
dimakis: 🔵 **unsafe_assumptions**: handleSwitchSession determines `running` via `ctx.sessionRegistry.isActive(found.clientId)` — the in-memory registry. But the state-based routing in handleSendV2/handleInterruptV2 explicitly avoids trusting the in-memory registry alone (the whole point of this PR). A zombie session would show `running: true` here even though the EventStore state is ENDED. Consider cross-referencing `eventStore.getSessionState()` for consistency, similar to how handleReconnect already does zombie detection (test at line 896). `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Solid fix for the double-message bug with clean state-machine routing, but the CLOSING state falls through to the active path unintentionally, and the new handleSwitchSession changes (running field, early watch) lack dedicated test coverage._

- 🟡 **bugs** (L510): Comment lists states routed to Case 1 as "ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED" but the condition `storeState !== 'ENDED' && storeState !== null` also admits CLOSING. A session in CLOSING state is winding down (CLOSING can only transition to ENDED per VALID_TRANSITIONS). Routing a new message via sendToChat to a closing session may race with the closeout — the agent could receive a new prompt while it's trying to commit/summarize. Either add CLOSING to the exclusion list alongside ENDED (treating it as a zombie) or document in the comment that CLOSING is intentionally treated as active. `[fixable]`
- 🟡 **bugs** (L672): Same CLOSING-state gap in handleInterruptV2: the comment says "ACTIVE, DETACHED, SUSPENDED, STARTING, CREATED" but the condition also admits CLOSING. Should be consistent with whatever decision is made for handleSendV2. `[fixable]`
- 🔵 **style** (L497): The block comments explaining the state-based routing (lines 497-511, 563-565, 576, and similarly in handleInterruptV2) are quite verbose for this codebase's conventions. CLAUDE.md says "default to writing no comments" and "only add one when the WHY is non-obvious." The initial Phase 3 header comment is useful context, but the per-case comments restate what the code already expresses through the condition structure. Consider trimming to one-line comments per case. `[fixable]`
- 🔵 **unsafe_assumptions** (L377): handleSwitchSession determines `running` via `ctx.sessionRegistry.isActive(found.clientId)` — the in-memory registry. But the state-based routing in handleSendV2/handleInterruptV2 explicitly avoids trusting the in-memory registry alone (the whole point of this PR). A zombie session would show `running: true` here even though the EventStore state is ENDED. Consider cross-referencing `eventStore.getSessionState()` for consistency, similar to how handleReconnect already does zombie detection (test at line 896). `[fixable]`


_Solid fix for the double-message bug with clean state-machine routing, but the CLOSING state falls through to the active path unintentionally, and the new handleSwitchSession changes (running field, early watch) lack dedicated test coverage._

- 🟡 **missing_tests**: No test verifies that the `session_switched` response includes the `running` field. The existing handleSwitchSession tests (lines 362-426) assert on mode/cwd/branch/wtId/tokens but not `running`. A test with an active session (isActive → true) should assert `running: true`, and a test with an ended session should assert `running: false`. `[fixable]`
- 🔵 **missing_tests**: No test covers the early `watch()` call added in handleSwitchSession (line 368). The intent — preventing a blind spot between switch and first send — is important enough to verify that `connRegistry.watch` is called before the response is sent. A spy-ordering test or at minimum an assertion that watch was called with the correct args would prevent a regression. `[fixable]`
- 🔵 **missing_tests**: No test covers the CLOSING state for either handleSendV2 or handleInterruptV2. Given the ambiguity noted above, a test that sets `getSessionState` to 'CLOSING' would document the intended behavior. `[fixable]`

dimakis (commented):
## Centaur Review (R2)

Clean fix following the Phase 3 design doc. State-based routing logic is sound, CLOSING exclusion is correct, zombie abort via stopChat() prevents leaked query loops.

**No blocking issues.**

Minor notes for follow-up:
- `handleReconnect` still uses old `getSession().isActive` pattern (different code path, not a regression)
- No protocol-parser test for `SET_RUNNING` dispatch (low risk)

130/130 tests passing. **Approved.**

PR #354, opened 2026-05-22
feat(ui): make What's Next feed scrollable beyond 5 items

dimakis:
## Summary

- Remove the hard `MAX_ITEMS=5` cap from `useAttentionFeed` hook — all attention items are now returned
- Component shows first 5 items by default with a "Show all N" / "Show less" toggle button
- Header badge now shows the true tier-1 count (was previously capped at 5)

## Test plan

- [x] Updated `useAttentionFeed` test: verifies all 10 items returned (was asserting 5-cap)
- [ ] Verify on mobile: "What's Next" shows 5 items, tap "Show all" to expand, "Show less" to collapse
- [ ] Verify badge count shows true total when >5 sessions awaiting reply

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #366, opened 2026-06-02
fix(frontend): reset AttentionFeed showAll state bugs

dimakis:
## Summary
- **showAll sticks when items shrink:** Added `useEffect` to auto-reset `showAll` when `items.length` drops to/below `DEFAULT_VISIBLE_COUNT` (5)
- **showAll persists across collapse/expand:** `toggleOpen` now resets `showAll` to `false` on collapse, so reopening always shows the truncated view
- **Component tests:** Added 13 tests covering show-more/less toggle, collapse reset, item-shrink reset, card rendering, navigation, summary/badge, and loading state

Follow-up to PR #354.

## Test plan
- [x] All 13 new component tests pass (`vitest run AttentionFeed.test.tsx`)
- [x] Existing hook tests still pass (`vitest run useAttentionFeed.test.ts`)
- [ ] Manual: expand "Show all", then trigger items to shrink — verify auto-collapse
- [ ] Manual: expand "Show all", collapse section, reopen — verify truncated view

### DIFF
--- a/frontend/src/components/__tests__/AttentionFeed.test.tsx
+++ b/frontend/src/components/__tests__/AttentionFeed.test.tsx
@@ -1,0 +1,202 @@
+// @vitest-environment jsdom
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { render, screen, fireEvent, cleanup, act } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
+import type { AttentionItem } from '../../hooks/useAttentionFeed';
+
+// ─── Mocks ────────────────────────────────────────────────────────────────
+
+let mockFeedReturn = {
+  items: [] as AttentionItem[],
+  tier1Count: 0,
+  loading: false,
+};
+
+vi.mock('../../hooks/useAttentionFeed', () => ({
+  useAttentionFeed: () => mockFeedReturn,
+}));
+
+vi.mock('../../lib/haptics', () => ({
+  selectionChanged: vi.fn(),
+}));
+
+const mockNavigate = vi.fn();
+vi.mock('react-router-dom', async () => {
+  const actual = await vi.importActual('react-router-dom');
+  return { ...actual, useNavigate: () => mockNavigate };
+});
+
+import { AttentionFeed } from '../AttentionFeed';
+
+// ─── Helpers ──────────────────────────────────────────────────────────────
+
+function makeItem(id: string, overrides: Partial<AttentionItem> = {}): AttentionItem {
+  return {
+    id,
+    source: 'telos',
+    tier: 1,
+    subPriority: 3,
+    title: `Item ${id}`,
+    meta: '1d',
+    accentColor: '#fbbf24',
+    icon: '\u2605',
+    navigateTo: `/todos/${id}`,
+    updatedAt: Date.now(),
+    ...overrides,
+  };
+}
+
+function makeItems(count: number): AttentionItem[] {
+  return Array.from({ length: count }, (_, i) => makeItem(`item-${i}`));
+}
+
+function setFeed(items: AttentionItem[], tier1Count = 0, loading = false) {
+  mockFeedReturn = { items, tier1Count, loading };
+}
+
+function renderFeed() {
+  return render(
+    <MemoryRouter>
+      <AttentionFeed />
+    </MemoryRouter>,
+  );
+}
+
+afterEach(() => {
+  cleanup();
+  mockNavigate.mockClear();
+  setFeed([]);
+});
+
+// ─── Tests ────────────────────────────────────────────────────────────────
+
+describe('AttentionFeed', () => {
+  it('renders empty state when no items', () => {
+    setFeed([]);
+    renderFeed();
+    expect(screen.getByText('Nothing needs your attention right now.')).toBeTruthy();
+  });
+
+  it('renders items as cards', () => {
+    setFeed([makeItem('a'), makeItem('b')]);
+    renderFeed();
+    expect(screen.getByText('Item a')).toBeTruthy();
+    expect(screen.getByText('Item b')).toBeTruthy();
+  });
+
+  it('shows only DEFAULT_VISIBLE_COUNT items initially', () => {
+    setFeed(makeItems(8));
+    renderFeed();
+    const cards = screen.getAllByText(/^Item item-/);
+    expect(cards).toHaveLength(5);
+  });
+
+  describe('show more / show less toggle', () => {
+    it('shows "Show all N" button when items exceed threshold', () => {
+      setFeed(makeItems(8));
+      renderFeed();
+      expect(screen.getByText('Show all 8')).toBeTruthy();
+    });
+
+    it('does not show toggle when items are at or below threshold', () => {
+      setFeed(makeItems(5));
+      renderFeed();
+      expect(screen.queryByText(/Show all/)).toBeNull();
+    });
+
+    it('expands to show all items on click', () => {
+      setFeed(makeItems(8));
+      renderFeed();
+      fireEvent.click(screen.getByText('Show all 8'));
+      const cards = screen.getAllByText(/^Item item-/);
+      expect(cards).toHaveLength(8);
+      expect(screen.getByText('Show less')).toBeTruthy();
+    });
+
+    it('collapses back on second click', () => {
+      setFeed(makeItems(8));
+      renderFeed();
+      fireEvent.click(screen.getByText('Show all 8'));
+      fireEvent.click(screen.getByText('Show less'));
+      const cards = screen.getAllByText(/^Item item-/);
+      expect(cards).toHaveLength(5);
+    });
+  });
+
+  describe('showAll resets on collapse', () => {
+    it('resets expanded state when section is collapsed and reopened', () => {
+      setFeed(makeItems(8));
+      renderFeed();
+
+      // Expand items
+      fireEvent.click(screen.getByText('Show all 8'));
+      expect(screen.getAllByText(/^Item item-/)).toHaveLength(8);
+
+      // Collapse section via header
+      fireEvent.click(screen.getByText("What's Next"));
+      // Cards should be gone
+      expect(screen.queryByText('Item item-0')).toBeNull();
+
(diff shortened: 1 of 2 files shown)


server/chat.ts
```diff
@@ -238,8 +380,8 @@ export function parseModelSpec(spec?: string): { model: string; effort: string |
 // --- Pure helper functions ---
 
 /** Send data via transport (isOpen guard is inside the transport). */
-function send(transport: SessionTransport, data: Record<string, unknown>) {
-  if (transport.isOpen()) transport.send(data);
+function send(transport: SessionTransport, data: Record<string, unknown> | BootContextMessage) {
```
dimakis, line 383: 🔵 **style**: The union type `Record<string, unknown> | BootContextMessage` in `send()` is immediately erased by `data as Record<string, unknown>`. `BootContextMessage` is already structurally assignable to `Record<string, unknown>`, so the union and cast are both unnecessary — the original signature works as-is. `[fixable]`

server/chat.ts
```diff
@@ -768,167 +910,15 @@ async function _startChatInner(
...
-        sources: [] as Array<{ path: string; kind: string }>,
-        included: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
-        trimmed: [] as Array<{ source: string; heading: string; tokens: number; content: string }>,
-      };
-      send(transport, { type: 'boot_context', ...fallbackPayload });
-      const s = registry.get(clientId);
-      if (s) s.bootContext = fallbackPayload;
-    }
-  })();
+      if (s) s.bootContext = msg as unknown as Record<string, unknown>;
```
dimakis, line 921: 🔵 **style**: Double cast `msg as unknown as Record<string, unknown>` to store `bootContext`. Since `BootContextMessage` is assignable to `Record<string, unknown>`, a single cast (`msg as Record<string, unknown>`) suffices. The `as unknown` intermediate makes it look like there's a real type incompatibility when there isn't one. `[fixable]`

dimakis, line 921: 🟡 **regressions**: The cached `bootContext` now includes `type: 'boot_context'` (from `BootContextMessage`), whereas previously it was stored without `type`. Replay in `ws-handler-v2.ts:285-288` does `{ type: 'boot_context', ...found.session.bootContext }` — the spread overwrites the literal `type` with the same value so it works, but the redundancy is accidental. If `BootContextMessage.type` ever diverges from the literal, replay would silently send the wrong type. `[fixable]`

server/chat.ts
```diff
@@ -93,6 +93,148 @@ export const eventStore = initEventStore();
...
+  trimmed: Array<{ source: string; heading: string; tokens: number; content: string }>;
+  fullMarkdown?: string;
+}
+
+const FALLBACK_BOOT_CONTEXT: BootContextMessage = {
+  type: 'boot_context',
+  source: 'local-fallback',
+  sourceCount: 0,
+  tokenCount: 0,
+  tokenBudget: 0,
```
dimakis, line 114: 🟡 **regressions**: `FALLBACK_BOOT_CONTEXT.tokenBudget` is `0`, while the old code always fell back to `8000` (via `DEFAULT_TOKEN_BUDGET`). This changes UI behavior when both ContexGin and the Python script fail — the boot context pill will show 0 budget instead of 8000. Minor since it only affects the double-failure case, but worth confirming this is intentional. `[fixable]`

server/chat.ts
```diff
@@ -93,6 +93,148 @@ export const eventStore = initEventStore();
...
+ * Local fallback: run build_boot_context.py --json to get deterministic boot context.
+ * This is the "old way" — reads canonical source files directly, no server dependency.
+ */
+async function localBootContextFallback(repoRoot: string): Promise<BootContextMessage> {
+  const scriptPath = join(repoRoot, 'scripts', 'build_boot_context.py');
+  try {
+    const { execFile } = await import('child_process');
+    const { promisify } = await import('util');
+    const execFileAsync = promisify(execFile);
+    const { stdout } = await execFileAsync('python3', [scriptPath, '--json'], {
```
dimakis, line 130: 🔵 **unsafe_assumptions**: `localBootContextFallback` uses dynamic `import('child_process')` and `import('util')` for modules that are already statically imported at the top of the file (`execFileSync` from `child_process`). The dynamic import adds unnecessary complexity — since `child_process` is a Node built-in that's already loaded, a static import would be simpler and more predictable. `[fixable]`

server/__tests__/boot-context.test.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL, '/fake/repo');
+
+    expect(result.source).toBe('local-fallback');
+    expect(result.sourceCount).toBe(0);
+    expect(result.tokenCount).toBe(0);
+    expect(result.fullMarkdown).toBeUndefined();
+  });
+
+  it('handles empty sources array', async () => {
```
dimakis, line 178: 🔵 **missing_tests**: The `tokenBudget` default of `8000` (line 199 of chat.ts: `typeof boot.tokenBudget === 'number' ? boot.tokenBudget : 8000`) is not tested. The 'handles empty sources array' test omits `tokenBudget` from the response but doesn't assert the resulting `tokenBudget` value. Adding `expect(result.tokenBudget).toBe(8000)` would cover this default. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Clean refactor from library-based to HTTP-based boot context with good test coverage and a well-tested AttentionFeed fix. Main concerns are minor: cosmetic type casts, a tokenBudget default change from 8000→0 in the zero-value fallback, and an untested default path._

- 🔵 **style** (L383): The union type `Record<string, unknown> | BootContextMessage` in `send()` is immediately erased by `data as Record<string, unknown>`. `BootContextMessage` is already structurally assignable to `Record<string, unknown>`, so the union and cast are both unnecessary — the original signature works as-is. `[fixable]`
- 🔵 **style** (L919): Double cast `msg as unknown as Record<string, unknown>` to store `bootContext`. Since `BootContextMessage` is assignable to `Record<string, unknown>`, a single cast (`msg as Record<string, unknown>`) suffices. The `as unknown` intermediate makes it look like there's a real type incompatibility when there isn't one. `[fixable]`
- 🟡 **regressions** (L919): The cached `bootContext` now includes `type: 'boot_context'` (from `BootContextMessage`), whereas previously it was stored without `type`. Replay in `ws-handler-v2.ts:285-288` does `{ type: 'boot_context', ...found.session.bootContext }` — the spread overwrites the literal `type` with the same value so it works, but the redundancy is accidental. If `BootContextMessage.type` ever diverges from the literal, replay would silently send the wrong type. `[fixable]`
- 🟡 **regressions** (L114): `FALLBACK_BOOT_CONTEXT.tokenBudget` is `0`, while the old code always fell back to `8000` (via `DEFAULT_TOKEN_BUDGET`). This changes UI behavior when both ContexGin and the Python script fail — the boot context pill will show 0 budget instead of 8000. Minor since it only affects the double-failure case, but worth confirming this is intentional. `[fixable]`
- 🔵 **unsafe_assumptions** (L130): `localBootContextFallback` uses dynamic `import('child_process')` and `import('util')` for modules that are already statically imported at the top of the file (`execFileSync` from `child_process`). The dynamic import adds unnecessary complexity — since `child_process` is a Node built-in that's already loaded, a static import would be simpler and more predictable. `[fixable]`


_Clean refactor from library-based to HTTP-based boot context with good test coverage and a well-tested AttentionFeed fix. Main concerns are minor: cosmetic type casts, a tokenBudget default change from 8000→0 in the zero-value fallback, and an untested default path._

- 🔵 **missing_tests** (L178): The `tokenBudget` default of `8000` (line 199 of chat.ts: `typeof boot.tokenBudget === 'number' ? boot.tokenBudget : 8000`) is not tested. The 'handles empty sources array' test omits `tokenBudget` from the response but doesn't assert the resulting `tokenBudget` value. Adding `expect(result.tokenBudget).toBe(8000)` would cover this default. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `frontend/src/components/__tests__/AttentionFeed.test.tsx`
_Clean, well-scoped fix with solid test coverage. Two minor boundary cases in tests worth considering but nothing blocking._

- 🔵 **missing_tests** (L148): The shrink test goes from 8 → 3 but doesn't test the boundary: items shrinking to exactly DEFAULT_VISIBLE_COUNT (5). Since the condition is `<=`, a test with 8 → 5 would verify the boundary is correct and guard against an off-by-one if someone changes it to `<`. `[fixable]`
- 🔵 **missing_tests** (L148): No test for the case where items shrink but remain above threshold (e.g. 8 → 6 while showAll is true). In that scenario showAll stays true and the user sees all 6 items with a 'Show less' button — worth a test to document that this is intentional behavior, not an oversight. `[fixable]`

frontend/src/components/__tests__/AttentionFeed.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+      // Re-open — should show truncated (5), not all 8
+      fireEvent.click(screen.getByText("What's Next"));
+      expect(screen.getAllByText(/^Item item-/)).toHaveLength(5);
+      expect(screen.getByText('Show all 8')).toBeTruthy();
+    });
+  });
+
+  describe('showAll resets when items shrink', () => {
+    it('auto-resets showAll when items drop below threshold', () => {
```
dimakis, line 148: 🔵 **missing_tests**: The shrink test goes from 8 → 3 but doesn't test the boundary: items shrinking to exactly DEFAULT_VISIBLE_COUNT (5). Since the condition is `<=`, a test with 8 → 5 would verify the boundary is correct and guard against an off-by-one if someone changes it to `<`. `[fixable]`

dimakis, line 148: 🔵 **missing_tests**: No test for the case where items shrink but remain above threshold (e.g. 8 → 6 while showAll is true). In that scenario showAll stays true and the user sees all 6 items with a 'Show less' button — worth a test to document that this is intentional behavior, not an oversight. `[fixable]`

PR #368, opened 2026-06-05
fix(client): always force WS reconnect on iOS foreground return

dimakis:
## Summary

- iOS kills WebSocket connections during background without updating `readyState`
- `checkAndReconnect()` trusted `readyState` — if the socket appeared OPEN, it did nothing
- First message sent on the dead socket was silently lost (regression)
- Fix: always tear down the old socket and force a fresh connection on foreground return
- Sends during the reconnect window queue in `pendingSends` and flush after the welcome handshake

## Root cause

`connection.ts:328-336` — the `readyState !== OPEN` guard:

```typescript
// BEFORE (broken)
checkAndReconnect(): void {
    if (!this.ws || this.ws.readyState !== WS_READY_STATE.OPEN) {
        // only reconnects if socket is visibly dead
    }
    // if readyState appears OPEN → does NOTHING
}
```

iOS kills the TCP connection during background but the JS `readyState` property stays `OPEN` until a send/recv actually fails. The first user message fires into the void.

## Test plan

- [ ] Background Mitzo on iOS, wait 30+ seconds, return to foreground
- [ ] Send a message — verify it reaches the agent on the first try
- [ ] Verify rapid foreground/background cycling doesn't cause double connections (reconnectTimer guard)
- [ ] Verify desktop browser tab switching still works (visibilitychange path)

> Re-opened from #367 which was merged without Centaur reviews.

### DIFF
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -324,14 +324,21 @@ export class MitzoConnection {
     }
   }
 
-  /** Force a reconnect check — call from native lifecycle hooks (e.g. Capacitor appStateChange). */
+  /**
+   * Force a reconnect — call from native lifecycle hooks (e.g. Capacitor
+   * appStateChange) and browser visibility events.
+   *
+   * iOS kills WebSocket connections during background without updating
+   * readyState, so we NEVER trust the old socket after a foreground
+   * transition. Always tear down and create a fresh connection. Any
+   * sends that arrive during the reconnect window queue in pendingSends
+   * and flush after the welcome handshake completes.
+   */
   checkAndReconnect(): void {
-    if (!this.ws || this.ws.readyState !== WS_READY_STATE.OPEN) {
-      if (this.reconnectTimer) return;
-      this.defuseOldWs();
-      this._connected = false;
-      this.listener?.({ type: '_close' });
-      this.doConnect();
-    }
+    if (this.reconnectTimer) return;
+    this.defuseOldWs();
+    this._connected = false;
+    this.listener?.({ type: '_close' });
+    this.doConnect();
   }
 }
--- a/packages/client/src/connection.ts
+++ b/packages/client/src/connection.ts
@@ -170,6 +170,14 @@ export class MitzoConnection {
       this.ws.onclose = null;
       this.ws.onmessage = null;
       this.ws.onerror = null;
+      // Explicitly close the socket to avoid orphaning a live connection
+      // on the server side (e.g. desktop tab switch where the socket is
+      // still healthy). No-op if the socket is already dead.
+      try {
+        this.ws.close();
+      } catch {
+        // Socket may already be in a broken state — ignore.
+      }
       this.ws = null;
     }
   }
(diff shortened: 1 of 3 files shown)


dimakis (commented):
## Centaur Review

_Correct fix for the iOS stale-socket problem, but orphans live sockets without closing them and forces unnecessary reconnect churn on desktop — both worth addressing._

- 🟡 **regressions** (L337): On desktop browsers, every tab switch (visibilitychange → visible) now tears down a healthy WebSocket and creates a new one. The old code preserved live sockets. This adds reconnect latency, a brief disconnected flash, and extra server load on every foreground return — even when the socket is fine. The reconnect + seq replay machinery makes this safe from a correctness standpoint, but it's wasteful on non-iOS platforms. Consider guarding with a Capacitor/iOS platform check, or at minimum accepting this as a documented trade-off. `[fixable]`
- 🟡 **bugs** (L339): defuseOldWs() nulls out handlers and drops the reference but never calls ws.close(). On desktop where the socket IS still alive, this orphans an open WebSocket — the server keeps the old connection object until TCP timeout. Previously this was rare (only triggered when readyState was already broken), but now it happens on every foreground transition. The orphaned socket should be explicitly closed before being abandoned to avoid server-side connection leaks. `[fixable]`
- 🟡 **missing_tests**: MitzoConnection has zero test coverage — no test or spec files exist under packages/client/. This PR changes reconnect behavior on all platforms (not just iOS). Key scenarios that should be tested: (1) checkAndReconnect always calls doConnect even when ws.readyState is OPEN, (2) checkAndReconnect is a no-op when reconnectTimer is set, (3) sends during the reconnect window queue in pendingSends and flush after welcome, (4) defuseOldWs prevents stale onclose from clobbering the new socket. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -324,14 +324,21 @@ export class MitzoConnection {
...
+   * Force a reconnect — call from native lifecycle hooks (e.g. Capacitor
+   * appStateChange) and browser visibility events.
+   *
+   * iOS kills WebSocket connections during background without updating
+   * readyState, so we NEVER trust the old socket after a foreground
+   * transition. Always tear down and create a fresh connection. Any
+   * sends that arrive during the reconnect window queue in pendingSends
+   * and flush after the welcome handshake completes.
+   */
   checkAndReconnect(): void {
```
dimakis: 🟡 **regressions**: On desktop browsers, every tab switch (visibilitychange → visible) now tears down a healthy WebSocket and creates a new one. The old code preserved live sockets. This adds reconnect latency, a brief disconnected flash, and extra server load on every foreground return — even when the socket is fine. The reconnect + seq replay machinery makes this safe from a correctness standpoint, but it's wasteful on non-iOS platforms. Consider guarding with a Capacitor/iOS platform check, or at minimum accepting this as a documented trade-off. `[fixable]`

packages/client/src/connection.ts
```diff
@@ -324,14 +324,21 @@ export class MitzoConnection {
...
-    if (!this.ws || this.ws.readyState !== WS_READY_STATE.OPEN) {
-      if (this.reconnectTimer) return;
-      this.defuseOldWs();
-      this._connected = false;
-      this.listener?.({ type: '_close' });
-      this.doConnect();
-    }
+    if (this.reconnectTimer) return;
+    this.defuseOldWs();
```
dimakis, line 350: 🟡 **bugs**: defuseOldWs() nulls out handlers and drops the reference but never calls ws.close(). On desktop where the socket IS still alive, this orphans an open WebSocket — the server keeps the old connection object until TCP timeout. Previously this was rare (only triggered when readyState was already broken), but now it happens on every foreground transition. The orphaned socket should be explicitly closed before being abandoned to avoid server-side connection leaks. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #369, opened 2026-06-05
fix: persist boot context in system prompt and sticky banner

dimakis:
## Summary

- **Server:** Boot context was fire-and-forget — sent as a UI event but never included in the system prompt. On SDK context compaction the model lost it entirely. Now `fetchBootContext()` is awaited and `fullMarkdown` is appended to `systemPromptAppend` so it survives compaction.
- **Frontend:** SessionBanner sat outside the scroll container as a flex sibling, causing it to be pushed offscreen on mobile Safari as messages accumulated. Moved inside `ChatArea`'s scroll container with `position: sticky; top: 0` so it pins to the top regardless of scroll position.

## Test plan

- [ ] Start a new session on mobile — verify boot context banner appears
- [ ] Send 3-4 messages — verify banner stays pinned at top of scroll area
- [ ] Scroll through long conversation — banner stays visible
- [ ] Verify boot context still displays on desktop
- [ ] Verify model retains boot context behavior after extended conversation (compaction)

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -2,7 +2,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import { useParams, useSearchParams, useNavigate } from 'react-router-dom';
 import { ChatArea } from '../components/ChatArea';
 import { ChatInput } from '../components/ChatInput';
-import { SessionBanner } from '../components/SessionBanner';
+
 import { VoiceSettings } from '../components/VoiceSettings';
 import { MitzoLogo } from '../components/MitzoLogo';
 import { useMessages, useConnection, useTokens, useMitzoStore } from '@mitzo/client/hooks';
@@ -272,8 +272,6 @@ export function ChatView() {
         )}
       </header>
 
-      <SessionBanner bootContext={bootContext} sessionContext={sessionContext} />
-
       <ChatArea
         messages={messages.messages}
         current={messages.current}
@@ -283,6 +281,8 @@ export function ChatView() {
         scrollRef={scrollRef}
         progressByToolId={progressByToolId}
         voice={voice}
+        bootContext={bootContext}
+        sessionContext={sessionContext}
       />
 
       <ChatInput
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -2,7 +2,6 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import { useParams, useSearchParams, useNavigate } from 'react-router-dom';
 import { ChatArea } from '../components/ChatArea';
 import { ChatInput } from '../components/ChatInput';
-
 import { VoiceSettings } from '../components/VoiceSettings';
 import { MitzoLogo } from '../components/MitzoLogo';
 import { useMessages, useConnection, useTokens, useMitzoStore } from '@mitzo/client/hooks';
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -255,6 +255,46 @@ describe('fetchBootContext', () => {
     expect(result.fullMarkdown).toBe('json fallback');
   });
 
+  it('returns fullMarkdown suitable for system prompt append', async () => {
+    const markdown = '# Identity\nYou are a helpful assistant.\n\n# Preferences\nBe concise.';
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
+        boot: {
+          content: markdown,
+          tokens: 200,
+          tokenBudget: 8000,
+          sources: ['profile.md'],
+        },
+      }),
+    });
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
+    expect(result.fullMarkdown).toBe(markdown);
+    // Verify the append pattern matches what chat.ts does:
+    // `\n\n# Boot Context\n${fullMarkdown}`
+    const append = result.fullMarkdown ? `\n\n# Boot Context\n${result.fullMarkdown}` : '';
+    expect(append).toContain('# Boot Context');
+    expect(append).toContain('# Identity');
+    expect(append).toContain('Be concise.');
+  });
+
+  it('returns empty string for append when fullMarkdown is missing', async () => {
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
+        boot: { tokens: 0, sources: [] },
+      }),
+    });
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
+    expect(result.fullMarkdown).toBeUndefined();
+    const append = result.fullMarkdown ? `\n\n# Boot Context\n${result.fullMarkdown}` : '';
+    expect(append).toBe('');
+  });
+
   it('uses default URL from env when not provided', async () => {
     const origUrl = process.env.CONTEXGIN_URL;
     process.env.CONTEXGIN_URL = 'http://test-host:9999';
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -271,16 +271,13 @@ describe('fetchBootContext', () => {
 
     const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
 
+    // fullMarkdown must be the exact content string — chat.ts appends it to the system prompt
     expect(result.fullMarkdown).toBe(markdown);
-    // Verify the append pattern matches what chat.ts does:
-    // `\n\n# Boot Context\n${fullMarkdown}`
-    const append = result.fullMarkdown ? `\n\n# Boot Context\n${result.fullMarkdown}` : '';
-    expect(append).toContain('# Boot Context');
-    expect(append).toContain('# Identity');
-    expect(append).toContain('Be concise.');
+    expect(typeof result.fullMarkdown).toBe('string');
+    expect(result.fullMarkdown!.length).toBeGreaterThan(0);
   });
 
-  it('returns empty string for append when fullMarkdown is missing', async () => {
+  it('returns undefined fullMarkdown when boot.content is missing', async () => {
     mockFetch.mockResolvedValueOnce({
       ok: true,
       json: async () => ({
@@ -290,9 +287,8 @@ describe('fetchBootContext', () => {
 
     const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
 
+    // chat.ts checks `bootContextMsg.fullMarkdown` — undefined means no append
     expect(result.fullMarkdown).toBeUndefined();
-    const append = result.fullMarkdown ? `\n\n# Boot Context\n${result.fullMarkdown}` : '';
-    expect(append).toBe('');
   });
 
   it('uses default URL from env when not provided', async () => {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -919,17 +919,25 @@ async function _startChatInner(
 
   // Fetch boot context BEFORE building system prompt so it's part of the
   // system prompt append and survives SDK context compaction.
-  const bootContextMsg = await fetchBootContext(agentName);
+  // fetchBootContext never throws and has a 5s AbortSignal timeout internally.
+  // Race with a 2s deadline so session startup isn't blocked when ContexGin is slow.
+  const bootContextMsg = await Promise.race([
+    fetchBootContext(agentName),
+    new Promise<Awaited<ReturnType<typeof fetchBootContext>>>((resolve) =>
+      setTimeout(() => resolve({ ...FALLBACK_BOOT_CONTEXT }), 2000),
+    ),
+  ]);
   const bootContextAppend = bootContextMsg.fullMarkdown
     ? `\n\n# Boot Context\n${bootContextMsg.fullMarkdown}`
     : '';
 
-  // Send boot context to UI for display (separate from system prompt injection).
-  // stateSessionId is the worktree-level session ID, already computed above.
+  // Send boot context to UI immediately (sessionId may be undefined for new sessions — OK,
+  // it's a display-only hint; the client doesn't key on it for boot context).
   send(transport, { ...bootContextMsg, ...(stateSessionId ? { sessionId: stateSessionId } : {}) });
   // Cache in ManagedSession for replay on reconnect/switch
   session.bootContext = bootContextMsg as unknown as Record<string, unknown>;
-  // Persist to EventStore for cold-path recovery
+  // For resumed sessions, persist immediately (sessionId is known).
+  // For new sessions, persist in onSessionResolved once SDK assigns the ID.
   if (stateSessionId) {
     eventStore.upsertSession({
       sessionId: stateSessionId,
@@ -1020,7 +1028,14 @@ async function _startChatInner(
       options.resume ? undefined : fullPrompt,
       {
         connRegistry: _connRegistry ?? undefined,
-        onSessionResolved: options.onSessionResolved,
+        onSessionResolved: (sessionId: string) => {
+          // Persist boot context to EventStore now that we have a real session ID
+          eventStore.upsertSession({
+            sessionId,
+            bootContext: JSON.stringify(bootContextMsg),
+          });
+          options.onSessionResolved?.(sessionId);
+        },
         onInitialPrompt: (sessionId: string) => {
           tryAutoRename(sessionId, clientId).catch(() => {
             /* errors logged internally */
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -924,15 +924,17 @@ async function _startChatInner(
     ? `\n\n# Boot Context\n${bootContextMsg.fullMarkdown}`
     : '';
 
-  // Send boot context to UI for display (separate from system prompt injection)
-  const bootSid = options.resume ?? registry.get(clientId)?.sessionId;
-  send(transport, { ...bootContextMsg, ...(bootSid ? { sessionId: bootSid } : {}) });
+  // Send boot context to UI for display (separate from system prompt injection).
+  // stateSessionId is the worktree-level session ID, already computed above.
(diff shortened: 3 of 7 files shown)


server/chat.ts
```diff
@@ -900,6 +900,24 @@ async function _startChatInner(
   // Load project hooks from .claude/settings.json (e.g. SessionStart boot context)
   const hooks = loadProjectHooks(cwd);
 
+  // Fetch boot context BEFORE building system prompt so it's part of the
+  // system prompt append and survives SDK context compaction.
+  const bootContextMsg = await fetchBootContext(agentName);
```
dimakis: 🟡 **unsafe_assumptions**: Changing fetchBootContext from fire-and-forget to `await` adds up to 10 seconds of latency before the SDK `query()` call (5s ContexGin timeout + 5s Python fallback timeout in the worst case). The old fire-and-forget approach let the session start immediately while boot context loaded in parallel. This tradeoff is intentional (system prompt needs the content), but worth noting: if ContexGin is slow/down and the Python fallback is also slow, session start stalls noticeably.

server/chat.ts
```diff
@@ -900,6 +900,24 @@ async function _startChatInner(
...
+  const bootContextAppend = bootContextMsg.fullMarkdown
+    ? `\n\n# Boot Context\n${bootContextMsg.fullMarkdown}`
+    : '';
+
+  // Send boot context to UI for display (separate from system prompt injection)
+  const bootSid = options.resume ?? registry.get(clientId)?.sessionId;
+  send(transport, { ...bootContextMsg, ...(bootSid ? { sessionId: bootSid } : {}) });
```
dimakis: 🔵 **regressions**: The boot context WS message is now sent before `query()` starts, so there's no SDK session ID yet for new (non-resume) sessions. `bootSid` falls through to `registry.get(clientId)?.sessionId`, which is the Mitzo worktree ID (set during session registration), not the SDK session UUID. This is consistent with pre-existing behavior (the old `.then()` also used the same fallback), so not a regression — but the comment 'Include sessionId for consistency' from the old code was removed, losing that documentation.

server/chat.ts
```diff
@@ -900,6 +900,24 @@ async function _startChatInner(
   // Load project hooks from .claude/settings.json (e.g. SessionStart boot context)
   const hooks = loadProjectHooks(cwd);
 
+  // Fetch boot context BEFORE building system prompt so it's part of the
+  // system prompt append and survives SDK context compaction.
+  const bootContextMsg = await fetchBootContext(agentName);
+  const bootContextAppend = bootContextMsg.fullMarkdown
```
dimakis, line 932: 🟡 **missing_tests**: The new behavior — boot context markdown being appended to the system prompt — has no test coverage. The existing `boot-context.test.ts` tests `fetchBootContext` in isolation, but no test verifies that `systemPromptAppend` includes the boot context content or that it's correctly gated on `fullMarkdown` being truthy. An integration-level test (or a unit test for the prompt assembly logic) would catch regressions if the append format changes. `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -2,7 +2,7 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 import { useParams, useSearchParams, useNavigate } from 'react-router-dom';
 import { ChatArea } from '../components/ChatArea';
 import { ChatInput } from '../components/ChatInput';
-import { SessionBanner } from '../components/SessionBanner';
+
```
dimakis: 🔵 **style**: The removal of the `SessionBanner` import leaves a blank line between the `ChatInput` import and the `VoiceSettings` import. Minor nit — the blank line is harmless but may be caught by a formatter rule expecting consistent spacing between import groups. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid refactor that correctly moves boot context from fire-and-forget to awaited (safe — fetchBootContext never throws) and promotes SessionBanner into ChatArea for sticky scroll behavior. Main concern is the added latency on session start (up to 10s worst case) and missing test coverage for the system prompt append logic._

- 🟡 **unsafe_assumptions** (L905): Changing fetchBootContext from fire-and-forget to `await` adds up to 10 seconds of latency before the SDK `query()` call (5s ContexGin timeout + 5s Python fallback timeout in the worst case). The old fire-and-forget approach let the session start immediately while boot context loaded in parallel. This tradeoff is intentional (system prompt needs the content), but worth noting: if ContexGin is slow/down and the Python fallback is also slow, session start stalls noticeably.
- 🔵 **regressions** (L912): The boot context WS message is now sent before `query()` starts, so there's no SDK session ID yet for new (non-resume) sessions. `bootSid` falls through to `registry.get(clientId)?.sessionId`, which is the Mitzo worktree ID (set during session registration), not the SDK session UUID. This is consistent with pre-existing behavior (the old `.then()` also used the same fallback), so not a regression — but the comment 'Include sessionId for consistency' from the old code was removed, losing that documentation.
- 🟡 **missing_tests** (L906): The new behavior — boot context markdown being appended to the system prompt — has no test coverage. The existing `boot-context.test.ts` tests `fetchBootContext` in isolation, but no test verifies that `systemPromptAppend` includes the boot context content or that it's correctly gated on `fullMarkdown` being truthy. An integration-level test (or a unit test for the prompt assembly logic) would catch regressions if the append format changes. `[fixable]`


_Solid refactor that correctly moves boot context from fire-and-forget to awaited (safe — fetchBootContext never throws) and promotes SessionBanner into ChatArea for sticky scroll behavior. Main concern is the added latency on session start (up to 10s worst case) and missing test coverage for the system prompt append logic._

- 🔵 **style** (L5): The removal of the `SessionBanner` import leaves a blank line between the `ChatInput` import and the `VoiceSettings` import. Minor nit — the blank line is harmless but may be caught by a formatter rule expecting consistent spacing between import groups. `[fixable]`

server/chat.ts
```diff
@@ -900,6 +900,26 @@ async function _startChatInner(
   // Load project hooks from .claude/settings.json (e.g. SessionStart boot context)
   const hooks = loadProjectHooks(cwd);
 
+  // Fetch boot context BEFORE building system prompt so it's part of the
+  // system prompt append and survives SDK context compaction.
+  const bootContextMsg = await fetchBootContext(agentName);
```
dimakis: 🟡 **regressions**: Changing fetchBootContext from fire-and-forget to awaited adds up to 5 seconds (the AbortSignal.timeout) to session startup latency when ContexGin is slow or unreachable, plus additional time for the local fallback subprocess. The old code let the session start immediately while boot context loaded in the background. Consider whether this latency tradeoff is acceptable — the user will see a blank chat for longer on ContexGin timeouts. `[fixable]`

server/chat.ts
```diff
@@ -900,6 +900,26 @@ async function _startChatInner(
...
+  const bootContextAppend = bootContextMsg.fullMarkdown
+    ? `\n\n# Boot Context\n${bootContextMsg.fullMarkdown}`
+    : '';
+
+  // Send boot context to UI for display (separate from system prompt injection).
+  // stateSessionId is the worktree-level session ID, already computed above.
+  send(transport, { ...bootContextMsg, ...(stateSessionId ? { sessionId: stateSessionId } : {}) });
```
dimakis, line 938: 🟡 **regressions**: The old fire-and-forget code looked up `session.sessionId` via `registry.get(clientId)` at resolution time (after the SDK query started and assigned the real sessionId). The new code uses `stateSessionId` which is `options.resume ?? session.sessionId` computed before the SDK query starts. For non-resume sessions, `session.sessionId` is undefined at this point (it gets set later by the query loop), so the boot_context message sent to the UI and the EventStore persistence will both lack a sessionId. The old code had the same race for the initial send but at least had a chance of resolving it if the fetch was slow enough. Verify this doesn't break client-side session demuxing. `[fixable]`

server/__tests__/boot-context.test.ts
```diff
@@ -229,6 +229,46 @@ describe('fetchBootContext', () => {
     expect(result.fullMarkdown).toBe('json fallback');
   });
 
+  it('returns fullMarkdown suitable for system prompt append', async () => {
```
dimakis, line 258: 🔵 **missing_tests**: The two new tests reconstruct the append pattern inline and assert against it, but they only test `fetchBootContext` return values — not that `chat.ts` actually appends the result to `systemPromptAppend`. The append logic in chat.ts (lines 906-908, 932) has no direct test coverage. An integration test that verifies the system prompt passed to `query()` contains the boot context would be more valuable. `[fixable]`

server/__tests__/boot-context.test.ts
```diff
@@ -229,6 +229,46 @@ describe('fetchBootContext', () => {
...
+          tokenBudget: 8000,
+          sources: ['profile.md'],
+        },
+      }),
+    });
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
+    expect(result.fullMarkdown).toBe(markdown);
+    // Verify the append pattern matches what chat.ts does:
```
dimakis: 🔵 **style**: The test at lines 249-254 is tautological — it constructs a template string and then asserts it contains substrings that are obviously present. It doesn't test any production code path; it tests the test's own string interpolation. The meaningful assertion is on line 248 (`result.fullMarkdown === markdown`), which the existing test structure already covers. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_The core change (boot context in system prompt for compaction survival) is sound, but awaiting the fetch adds up to 5s startup latency on ContexGin failures, and the sessionId may be undefined for non-resume sessions at the point of send/persist._

- 🟡 **regressions** (L905): Changing fetchBootContext from fire-and-forget to awaited adds up to 5 seconds (the AbortSignal.timeout) to session startup latency when ContexGin is slow or unreachable, plus additional time for the local fallback subprocess. The old code let the session start immediately while boot context loaded in the background. Consider whether this latency tradeoff is acceptable — the user will see a blank chat for longer on ContexGin timeouts. `[fixable]`
- 🟡 **regressions** (L912): The old fire-and-forget code looked up `session.sessionId` via `registry.get(clientId)` at resolution time (after the SDK query started and assigned the real sessionId). The new code uses `stateSessionId` which is `options.resume ?? session.sessionId` computed before the SDK query starts. For non-resume sessions, `session.sessionId` is undefined at this point (it gets set later by the query loop), so the boot_context message sent to the UI and the EventStore persistence will both lack a sessionId. The old code had the same race for the initial send but at least had a chance of resolving it if the fetch was slow enough. Verify this doesn't break client-side session demuxing. `[fixable]`


_The core change (boot context in system prompt for compaction survival) is sound, but awaiting the fetch adds up to 5s startup latency on ContexGin failures, and the sessionId may be undefined for non-resume sessions at the point of send/persist._

- 🔵 **missing_tests** (L232): The two new tests reconstruct the append pattern inline and assert against it, but they only test `fetchBootContext` return values — not that `chat.ts` actually appends the result to `systemPromptAppend`. The append logic in chat.ts (lines 906-908, 932) has no direct test coverage. An integration test that verifies the system prompt passed to `query()` contains the boot context would be more valuable. `[fixable]`
- 🔵 **style** (L249): The test at lines 249-254 is tautological — it constructs a template string and then asserts it contains substrings that are obviously present. It doesn't test any production code path; it tests the test's own string interpolation. The meaningful assertion is on line 248 (`result.fullMarkdown === markdown`), which the existing test structure already covers. `[fixable]`

server/chat.ts
```diff
@@ -900,6 +900,34 @@ async function _startChatInner(
...
+    : '';
+
+  // Send boot context to UI immediately (sessionId may be undefined for new sessions — OK,
+  // it's a display-only hint; the client doesn't key on it for boot context).
+  send(transport, { ...bootContextMsg, ...(stateSessionId ? { sessionId: stateSessionId } : {}) });
+  // Cache in ManagedSession for replay on reconnect/switch
+  session.bootContext = bootContextMsg as unknown as Record<string, unknown>;
+  // For resumed sessions, persist immediately (sessionId is known).
+  // For new sessions, persist in onSessionResolved once SDK assigns the ID.
+  if (stateSessionId) {
```
dimakis, line 943: 🟡 **bugs**: Double-persist for resumed sessions: when `options.resume` is set, `stateSessionId` is truthy so `eventStore.upsertSession` fires here (line 924-928). Then `onSessionResolved` also fires in the query loop (line 1023-1028) and calls `upsertSession` again with the same data. Harmless (idempotent write) but wasteful — consider guarding the `onSessionResolved` upsert with `if (!options.resume)` or removing the early persist and relying solely on the callback. `[fixable]`

server/chat.ts
```diff
@@ -900,6 +900,34 @@ async function _startChatInner(
   // Load project hooks from .claude/settings.json (e.g. SessionStart boot context)
   const hooks = loadProjectHooks(cwd);
 
+  // Fetch boot context BEFORE building system prompt so it's part of the
+  // system prompt append and survives SDK context compaction.
+  // fetchBootContext never throws and has a 5s AbortSignal timeout internally.
+  // Race with a 2s deadline so session startup isn't blocked when ContexGin is slow.
+  const bootContextMsg = await Promise.race([
+    fetchBootContext(agentName),
+    new Promise<Awaited<ReturnType<typeof fetchBootContext>>>((resolve) =>
```
dimakis: 🔵 **unsafe_assumptions**: The `setTimeout` in the `Promise.race` is never cleared. When `fetchBootContext` resolves first (the common/fast path), the 2s timer lingers until it fires and resolves a dead promise. Consider storing the timer ID and clearing it after the race, or using `AbortSignal.timeout(2000)` to cancel the fetch itself instead of racing with a fallback. `[fixable]`

server/chat.ts
```diff
@@ -900,6 +900,34 @@ async function _startChatInner(
   // Load project hooks from .claude/settings.json (e.g. SessionStart boot context)
   const hooks = loadProjectHooks(cwd);
 
+  // Fetch boot context BEFORE building system prompt so it's part of the
+  // system prompt append and survives SDK context compaction.
+  // fetchBootContext never throws and has a 5s AbortSignal timeout internally.
+  // Race with a 2s deadline so session startup isn't blocked when ContexGin is slow.
+  const bootContextMsg = await Promise.race([
```
dimakis, line 925: 🟡 **regressions**: Session startup is now blocking on boot context fetch (up to 2s). The old fire-and-forget approach had zero startup latency impact. The trade-off (boot context in system prompt, survives compaction) is sound and well-commented, but the 2s worst-case delay on every session start is worth measuring in practice — especially on mobile where perceived latency matters. If ContexGin is routinely slow, users will feel this.

server/chat.ts
```diff
@@ -900,6 +900,34 @@ async function _startChatInner(
...
+  ]);
+  const bootContextAppend = bootContextMsg.fullMarkdown
+    ? `\n\n# Boot Context\n${bootContextMsg.fullMarkdown}`
+    : '';
+
+  // Send boot context to UI immediately (sessionId may be undefined for new sessions — OK,
+  // it's a display-only hint; the client doesn't key on it for boot context).
+  send(transport, { ...bootContextMsg, ...(stateSessionId ? { sessionId: stateSessionId } : {}) });
+  // Cache in ManagedSession for replay on reconnect/switch
+  session.bootContext = bootContextMsg as unknown as Record<string, unknown>;
```
dimakis, line 940: 🔵 **style**: `session.bootContext = bootContextMsg as unknown as Record<string, unknown>` — the double-cast through `unknown` suggests the `bootContext` field on `ManagedSession` should be typed as `BootContextMessage` instead of `Record<string, unknown>`. This would eliminate the cast and make the type relationship explicit. Consider updating the `ManagedSession` type. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid architectural improvement — moving boot context from fire-and-forget to system-prompt-embedded ensures it survives SDK compaction. Main concerns: redundant double-persist for resumed sessions, uncleaned timer in Promise.race, and missing test coverage for the system prompt append logic._

- 🟡 **bugs** (L924): Double-persist for resumed sessions: when `options.resume` is set, `stateSessionId` is truthy so `eventStore.upsertSession` fires here (line 924-928). Then `onSessionResolved` also fires in the query loop (line 1023-1028) and calls `upsertSession` again with the same data. Harmless (idempotent write) but wasteful — consider guarding the `onSessionResolved` upsert with `if (!options.resume)` or removing the early persist and relying solely on the callback. `[fixable]`
- 🔵 **unsafe_assumptions** (L909): The `setTimeout` in the `Promise.race` is never cleared. When `fetchBootContext` resolves first (the common/fast path), the 2s timer lingers until it fires and resolves a dead promise. Consider storing the timer ID and clearing it after the race, or using `AbortSignal.timeout(2000)` to cancel the fetch itself instead of racing with a fallback. `[fixable]`
- 🟡 **regressions** (L907): Session startup is now blocking on boot context fetch (up to 2s). The old fire-and-forget approach had zero startup latency impact. The trade-off (boot context in system prompt, survives compaction) is sound and well-commented, but the 2s worst-case delay on every session start is worth measuring in practice — especially on mobile where perceived latency matters. If ContexGin is routinely slow, users will feel this.
- 🔵 **style** (L921): `session.bootContext = bootContextMsg as unknown as Record<string, unknown>` — the double-cast through `unknown` suggests the `bootContext` field on `ManagedSession` should be typed as `BootContextMessage` instead of `Record<string, unknown>`. This would eliminate the cast and make the type relationship explicit. Consider updating the `ManagedSession` type. `[fixable]`


_Solid architectural improvement — moving boot context from fire-and-forget to system-prompt-embedded ensures it survives SDK compaction. Main concerns: redundant double-persist for resumed sessions, uncleaned timer in Promise.race, and missing test coverage for the system prompt append logic._

- 🟡 **missing_tests**: The two new test cases verify `fullMarkdown` field presence/absence, which is good. But the core behavior change — appending `fullMarkdown` to the system prompt via `bootContextAppend` — has no test coverage. A test asserting that `systemPromptAppend` includes the boot context markdown (and omits it when `fullMarkdown` is undefined) would catch regressions in the append logic. This would need to be in a chat.ts-level test or integration test. `[fixable]`
- 🔵 **missing_tests**: No test for the `Promise.race` timeout behavior — e.g., that a slow `fetchBootContext` (>2s) results in the fallback being used and the session still starting correctly. `[fixable]`

PR #370, opened 2026-06-08
fix(chat): use actual token count as boot budget fallback

dimakis:
## Summary
- When ContexGin didn't return `tokenBudget` in its boot response, Mitzo fell back to a hardcoded `8000` — making the SessionBanner always show "8k"
- Changed fallback from `8000` to `bootTokens` (actual compiled token count), so the UI reflects reality even before ContexGin adds the field
- Paired with dimakis/contexgin#(pending) which adds the `tokenBudget` field to the ContexGin response

## Test plan
- [x] `tsc --noEmit` passes
- [x] No new test failures (37 pre-existing, all unrelated)
- [ ] Verify SessionBanner shows ~11k (or actual count) instead of 8k

### DIFF
--- a/server/__tests__/boot-context.test.ts
+++ b/server/__tests__/boot-context.test.ts
@@ -191,6 +191,32 @@ describe('fetchBootContext', () => {
     expect(result.tokenCount).toBe(100);
   });
 
+  it('falls back to bootTokens when tokenBudget is absent', async () => {
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
+        boot: { content: 'payload', tokens: 11297, sources: ['a.md'] },
+      }),
+    });
+
+    const result = await fetchBootContext('mitzo-conversational', CONTEXGIN_URL);
+
+    expect(result.tokenBudget).toBe(11297);
+  });
+
+  it('uses tokenBudget from ContexGin when provided', async () => {
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      json: async () => ({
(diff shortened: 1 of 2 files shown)


dimakis:
## Centaur Review

_Correct fix — replaces an arbitrary 8000 magic number with the actual token count. One existing test already exercises the fallback path but doesn't assert the budget value; adding that assertion would lock in the new behavior._

- 🟡 **missing_tests** (L192): The 'handles empty sources array' test sends a response without `tokenBudget` (line 182), which exercises the fallback path changed by this PR. But it never asserts `result.tokenBudget`. Add `expect(result.tokenBudget).toBe(100)` to verify the new fallback-to-bootTokens behavior. Without this assertion, the test passes regardless of whether the fallback is 8000, 100, or 0. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


_Correct fix — aligns the ContexGin path's missing-tokenBudget fallback with the local-fallback path (which already uses actual tokens as budget). Just needs an explicit test assertion to lock in the new behavior._

- 🔵 **missing_tests**: No test explicitly covers the missing-tokenBudget fallback path. The 'empty sources' test (line 178) and 'filters non-string sources' test (line 194) both omit tokenBudget from the mock response and exercise this code path, but neither asserts on result.tokenBudget. Adding an assertion (e.g., expect(result.tokenBudget).toBe(100) in the empty-sources test) would lock in the new behavior and prevent silent regression. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #371, opened 2026-06-09
fix(client): clear WS pending sends on session transition

dimakis:
## Summary

- Messages queued in `connection.pendingSends` during WS downtime (iOS background) were flushed to the wrong session after reconnect
- Add `clearPendingSends()` to `MitzoConnection` and call it from `newSession()` and `switchSession()`
- Three new tests covering the drain, empty-queue safety, and end-to-end store-level leak prevention

## Test plan

- [x] `npx vitest run packages/client/` — 309 tests pass
- [ ] Manual: send message on iOS → background app → start new session → verify old message doesn't appear

### DIFF
--- a/packages/client/__tests__/connection.test.ts
+++ b/packages/client/__tests__/connection.test.ts
@@ -520,6 +520,44 @@ describe('MitzoConnection', () => {
     });
   });
 
+  describe('clearPendingSends', () => {
+    it('drains queued messages so they are not flushed on reconnect', () => {
+      vi.useFakeTimers();
+      const conn = createConnection();
+      conn.onMessage(() => {});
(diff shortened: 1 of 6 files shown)


dimakis (commented):
## Centaur Review

Found **1** issue(s).


_Clean, well-scoped fix for a real iOS cross-session message leak. Both the WS queue drain and the tokenBudget fallback change are correct and well-tested; only a minor coverage gap for switchSession._

- 🔵 **missing_tests**: switchSession also calls clearPendingSends (store.ts line 254) but only newSession has a store-level integration test for this behavior. A parallel test under the 'switchSession' describe block would close the coverage gap and catch regressions if the call is accidentally removed from one path but not the other. `[fixable]`

PR #372, opened 2026-06-09
fix(client): clear WS pending sends on session transition

dimakis:
## Summary

- Messages queued in `connection.pendingSends` during WS downtime (iOS background) were flushed to the wrong session after reconnect
- Add `clearPendingSends()` to `MitzoConnection` and call it from `newSession()` and `switchSession()`
- Three new tests covering the drain, empty-queue safety, and end-to-end store-level leak prevention

Reopened from #371 (merged prematurely without reviews — reverted).

## Test plan

- [x] `npx vitest run packages/client/` — 309 tests pass
- [ ] Manual: send message on iOS → background app → start new session → verify old message doesn't appear

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -250,6 +250,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
+      // Drain stale pending sends from the previous session's WS queue.
+      connection.clearPendingSends();
 
       set((s) => ({
         sessions: { ...s.sessions, active: id },
@@ -281,6 +283,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       parserState.currentSessionId = undefined;
       parserState.pendingSend = [];
       clearPendingSendTimer();
+      // Drain the WS pending-send queue to prevent stale messages from the
+      // old session leaking into the new one on reconnect (iOS background).
+      connection.clearPendingSends();
       connection.send({ type: 'switch_session', sessionId: null });
       set({
         sessions: { ...get().sessions, active: null },
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -250,7 +250,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
-      // Drain stale pending sends from the previous session's WS queue.
+      parserState.pendingSend = [];
+      clearPendingSendTimer();
       connection.clearPendingSends();
 
       set((s) => ({
@@ -283,8 +284,6 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       parserState.currentSessionId = undefined;
       parserState.pendingSend = [];
       clearPendingSendTimer();
-      // Drain the WS pending-send queue to prevent stale messages from the
-      // old session leaking into the new one on reconnect (iOS background).
       connection.clearPendingSends();
       connection.send({ type: 'switch_session', sessionId: null });
       set({
--- a/packages/client/__tests__/connection.test.ts
+++ b/packages/client/__tests__/connection.test.ts
@@ -520,6 +520,44 @@ describe('MitzoConnection', () => {
     });
   });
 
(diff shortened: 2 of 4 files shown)


packages/client/src/store.ts
```diff
@@ -281,6 +283,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       parserState.currentSessionId = undefined;
       parserState.pendingSend = [];
       clearPendingSendTimer();
+      // Drain the WS pending-send queue to prevent stale messages from the
```
dimakis: 🔵 **style**: The two-line comment duplicates the JSDoc on clearPendingSends() itself ('to avoid cross-session message leaks'). Given the project convention of writing no comments unless the WHY is non-obvious, and the method name + JSDoc already convey intent, this comment could be removed. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-scoped fix for a real iOS background race condition. Minor gap: switchSession path lacks a dedicated test for the new clearPendingSends call._

- 🔵 **missing_tests**: switchSession also calls connection.clearPendingSends() (store.ts:254) but only newSession has a test for it. A parallel test under the switchSession describe block would verify the same cross-session leak protection applies when switching between existing sessions while the WS is down. `[fixable]`


_Clean, well-scoped fix for a real iOS background race condition. Minor gap: switchSession path lacks a dedicated test for the new clearPendingSends call._

- 🔵 **style** (L286): The two-line comment duplicates the JSDoc on clearPendingSends() itself ('to avoid cross-session message leaks'). Given the project convention of writing no comments unless the WHY is non-obvious, and the method name + JSDoc already convey intent, this comment could be removed. `[fixable]`

packages/client/src/store.ts
```diff
@@ -250,6 +250,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
+      // Drain stale pending sends from the previous session's WS queue.
+      connection.clearPendingSends();
```
dimakis, line 255: 🔵 **regressions**: switchSession clears connection.pendingSends but does not clear parserState.pendingSend or cancel pendingSendTimer (unlike newSession at lines 284-285). If a user sends a message while the agent is running and then switches sessions before session_end arrives, the drainOne timer will still fire and call connection.send() with the stale payload + SET_RUNNING on the new session's UI. This is a pre-existing gap (not introduced by this PR) but the fix is incomplete without it — consider adding parserState.pendingSend = [] and clearPendingSendTimer() in switchSession. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix for a real iOS reconnect bug. Two gaps: switchSession's clearPendingSends call lacks a test, and it doesn't clear the store-level pendingSend queue (parserState.pendingSend) the way newSession does._

- 🔵 **missing_tests**: switchSession also calls clearPendingSends() (store.ts:254) but has no test verifying that stale WS-queued messages from the old session are drained on switch. The newSession path is tested; the switchSession path should be as well for parity. `[fixable]`


_Clean, well-tested fix for a real iOS reconnect bug. Two gaps: switchSession's clearPendingSends call lacks a test, and it doesn't clear the store-level pendingSend queue (parserState.pendingSend) the way newSession does._

- 🔵 **regressions** (L254): switchSession clears connection.pendingSends but does not clear parserState.pendingSend or cancel pendingSendTimer (unlike newSession at lines 284-285). If a user sends a message while the agent is running and then switches sessions before session_end arrives, the drainOne timer will still fire and call connection.send() with the stale payload + SET_RUNNING on the new session's UI. This is a pre-existing gap (not introduced by this PR) but the fix is incomplete without it — consider adding parserState.pendingSend = [] and clearPendingSendTimer() in switchSession. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


_Clean, well-targeted fix for a real iOS cross-session message leak. Implementation is correct; one call site (`switchSession`) lacks a store-level test._

- 🔵 **missing_tests**: The PR adds `clearPendingSends()` in both `switchSession` and `newSession`, but only `newSession` has a store-level test verifying the WS queue is drained. A parallel test under the `switchSession` describe block (e.g. WS dies → user sends → `switchSession('new-id')` → reconnect → stale message not flushed) would cover the other call site. `[fixable]`

PR #373, opened 2026-06-09
fix(chat): resolve SSH agent socket fresh per session on macOS

dimakis:
## Summary

- Mitzo's server process inherits `SSH_AUTH_SOCK` at launch, but macOS invalidates the socket path after sleep/wake
- SDK sessions spawned after wake get a stale socket → git push/fetch fails with "Permission denied (publickey)"
- Fix: resolve the live socket via `launchctl getenv SSH_AUTH_SOCK` in `sdkEnv()` at session spawn time
- No-op on non-macOS, graceful fallback if launchctl fails

## Test plan

- [x] Server tests pass (pre-existing failures unchanged)
- [ ] Deploy, sleep/wake Mac, verify git push works in a new Mitzo session

### DIFF
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -529,26 +529,67 @@ describe('validateResumable', () => {
 });
 
 describe('resolveSshAuthSock', () => {
-  it('returns a string on macOS when the agent is running', async () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('returns trimmed socket path on macOS', async () => {
+    vi.resetModules();
+    vi.doMock('os', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return { ...actual, platform: () => 'darwin' };
+    });
+    vi.doMock('child_process', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return {
+        ...actual,
+        execFileSync: vi.fn().mockReturnValue('  /private/tmp/com.apple.launchd.xxx/Listeners  \n'),
+      };
+    });
     const { resolveSshAuthSock } = await import('../chat.js');
-    const result = resolveSshAuthSock();
-    if (process.platform === 'darwin') {
-      // On macOS CI/dev, the agent should be running
-      expect(typeof result === 'string' || result === null).toBe(true);
-      if (result) {
-        expect(result).toContain('/');
-      }
-    } else {
-      expect(result).toBeNull();
-    }
+    expect(resolveSshAuthSock()).toBe('/private/tmp/com.apple.launchd.xxx/Listeners');
+  });
+
+  it('returns null when launchctl returns empty string', async () => {
+    vi.resetModules();
+    vi.doMock('os', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return { ...actual, platform: () => 'darwin' };
+    });
+    vi.doMock('child_process', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return { ...actual, execFileSync: vi.fn().mockReturnValue('  \n') };
+    });
+    const { resolveSshAuthSock } = await import('../chat.js');
+    expect(resolveSshAuthSock()).toBeNull();
+  });
+
+  it('returns null when execFileSync throws', async () => {
+    vi.resetModules();
+    vi.doMock('os', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return { ...actual, platform: () => 'darwin' };
+    });
+    vi.doMock('child_process', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return {
+        ...actual,
+        execFileSync: vi.fn().mockImplementation(() => {
+          throw new Error('launchctl not found');
+        }),
+      };
+    });
+    const { resolveSshAuthSock } = await import('../chat.js');
+    expect(resolveSshAuthSock()).toBeNull();
   });
 
   it('returns null on non-macOS platforms', async () => {
-    // On macOS this test validates the shape; on Linux CI it validates the null path
+    vi.resetModules();
+    vi.doMock('os', async (importOriginal) => {
+      const actual = (await importOriginal()) as Record<string, unknown>;
+      return { ...actual, platform: () => 'linux' };
+    });
     const { resolveSshAuthSock } = await import('../chat.js');
-    const result = resolveSshAuthSock();
-    if (process.platform !== 'darwin') {
-      expect(result).toBeNull();
-    }
+    expect(resolveSshAuthSock()).toBeNull();
   });
 });
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -527,3 +527,28 @@ describe('validateResumable', () => {
     expect(result).toEqual({ valid: false });
   });
 });
+
+describe('resolveSshAuthSock', () => {
+  it('returns a string on macOS when the agent is running', async () => {
+    const { resolveSshAuthSock } = await import('../chat.js');
+    const result = resolveSshAuthSock();
+    if (process.platform === 'darwin') {
+      // On macOS CI/dev, the agent should be running
+      expect(typeof result === 'string' || result === null).toBe(true);
+      if (result) {
+        expect(result).toContain('/');
+      }
+    } else {
+      expect(result).toBeNull();
+    }
+  });
+
+  it('returns null on non-macOS platforms', async () => {
+    // On macOS this test validates the shape; on Linux CI it validates the null path
+    const { resolveSshAuthSock } = await import('../chat.js');
+    const result = resolveSshAuthSock();
+    if (process.platform !== 'darwin') {
+      expect(result).toBeNull();
+    }
+  });
+});
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -386,6 +386,25 @@ function send(transport: SessionTransport, data: Record<string, unknown> | BootC
 
 const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
 
+/**
+ * Resolve the live SSH agent socket on macOS via launchctl.
+ * The server process may have started with a socket that became stale after
+ * sleep/wake — launchctl always returns the current one.
+ * Returns null on non-macOS or if the agent isn't running.
+ */
+export function resolveSshAuthSock(): string | null {
+  if (platform() !== 'darwin') return null;
+  try {
+    const sock = execFileSync('launchctl', ['getenv', 'SSH_AUTH_SOCK'], {
+      encoding: 'utf8',
+      timeout: 500,
+    }).trim();
+    return sock || null;
+  } catch {
+    return null;
+  }
+}
+
 function sdkEnv(): Record<string, string> {
   const env = { ...process.env } as Record<string, string>;
   env.CLAUDE_CODE_USE_VERTEX = process.env.CLAUDE_CODE_USE_VERTEX || '1';
@@ -406,20 +425,8 @@ function sdkEnv(): Record<string, string> {
   const venvPaths = getRepoConfig().resolvedVenvPaths;
   env.PATH = [...venvPaths, existingPath].join(':');
 
-  // Resolve the current SSH agent socket on macOS. The server process may have
-  // started with a socket that became stale after sleep/wake — launchctl always
-  // returns the live one.
-  if (platform() === 'darwin') {
-    try {
-      const sock = execFileSync('launchctl', ['getenv', 'SSH_AUTH_SOCK'], {
-        encoding: 'utf8',
-        timeout: 2000,
-      }).trim();
-      if (sock) env.SSH_AUTH_SOCK = sock;
-    } catch {
-      // launchctl unavailable or no agent — leave inherited value
-    }
-  }
+  const sock = resolveSshAuthSock();
+  if (sock) env.SSH_AUTH_SOCK = sock;
 
   delete env.AUTH_PASSPHRASE;
   delete env.AUTH_SECRET;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -12,7 +12,7 @@ import { readFileSync, writeFileSync, mkdirSync, readdirSync, existsSync } from 
 import { join, resolve, dirname } from 'path';
 import { fileURLToPath } from 'url';
 import { randomUUID } from 'crypto';
-import { homedir } from 'os';
+import { homedir, platform } from 'os';
 import {
   createWorktree,
   createWorktreeAsync,
@@ -405,6 +405,21 @@ function sdkEnv(): Record<string, string> {
   const existingPath = env.PATH || '/usr/bin:/bin:/usr/local/bin';
   const venvPaths = getRepoConfig().resolvedVenvPaths;
   env.PATH = [...venvPaths, existingPath].join(':');
+
+  // Resolve the current SSH agent socket on macOS. The server process may have
+  // started with a socket that became stale after sleep/wake — launchctl always
+  // returns the live one.
+  if (platform() === 'darwin') {
+    try {
+      const sock = execFileSync('launchctl', ['getenv', 'SSH_AUTH_SOCK'], {
+        encoding: 'utf8',
+        timeout: 2000,
+      }).trim();
+      if (sock) env.SSH_AUTH_SOCK = sock;
+    } catch {
+      // launchctl unavailable or no agent — leave inherited value
+    }
+  }
 
   delete env.AUTH_PASSPHRASE;
   delete env.AUTH_SECRET;
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -386,12 +386,7 @@ function send(transport: SessionTransport, data: Record<string, unknown> | BootC
 
 const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
 
-/**
- * Resolve the live SSH agent socket on macOS via launchctl.
- * The server process may have started with a socket that became stale after
- * sleep/wake — launchctl always returns the current one.
- * Returns null on non-macOS or if the agent isn't running.
- */
+// launchctl resolves the live socket — the inherited one goes stale after sleep/wake.
 export function resolveSshAuthSock(): string | null {
   if (platform() !== 'darwin') return null;
   try {


server/chat.ts
```diff
@@ -406,6 +406,21 @@ function sdkEnv(): Record<string, string> {
   const venvPaths = getRepoConfig().resolvedVenvPaths;
   env.PATH = [...venvPaths, existingPath].join(':');
 
+  // Resolve the current SSH agent socket on macOS. The server process may have
+  // started with a socket that became stale after sleep/wake — launchctl always
+  // returns the live one.
+  if (platform() === 'darwin') {
```
dimakis: 🔵 **missing_tests**: The SSH agent socket resolution via `execFileSync('launchctl', ...)` in `sdkEnv()` has no test coverage. While the function is difficult to unit test (platform-specific, synchronous exec), this is a new side effect on every session start that could mask failures silently. The `sdkEnv` function is not exported or tested at all — acceptable for now given its nature, but worth noting.

server/chat.ts
```diff
@@ -406,6 +406,21 @@ function sdkEnv(): Record<string, string> {
...
+    try {
+      const sock = execFileSync('launchctl', ['getenv', 'SSH_AUTH_SOCK'], {
```
dimakis: 🔵 **unsafe_assumptions**: `execFileSync` is called synchronously on every `sdkEnv()` invocation (i.e., every session start). If `launchctl` hangs or is slow, the 2-second timeout blocks the Node.js event loop. Consider caching the result with a short TTL (e.g., 30s) or using `execFile` (async). In practice this is unlikely to be a problem since `launchctl getenv` is near-instant, but it's worth being aware of. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Clean, well-motivated fix for two real issues (cross-session message leaks and stale SSH sockets). The client-side changes are well-tested for `newSession` but missing a symmetric test for `switchSession`. The server-side `execFileSync` is pragmatic but runs synchronously on every session start._

- 🔵 **missing_tests**: The store test only covers `newSession` clearing pending sends. There is no equivalent test for `switchSession`, which also received the same three cleanup lines (`parserState.pendingSend = []`, `clearPendingSendTimer()`, `connection.clearPendingSends()`). Since `switchSession` had zero pending-send cleanup before this PR, it's the more important path to test. `[fixable]`


### `server/chat.ts`
_Clean, well-motivated fix for two real issues (cross-session message leaks and stale SSH sockets). The client-side changes are well-tested for `newSession` but missing a symmetric test for `switchSession`. The server-side `execFileSync` is pragmatic but runs synchronously on every session start._

- 🔵 **missing_tests** (L412): The SSH agent socket resolution via `execFileSync('launchctl', ...)` in `sdkEnv()` has no test coverage. While the function is difficult to unit test (platform-specific, synchronous exec), this is a new side effect on every session start that could mask failures silently. The `sdkEnv` function is not exported or tested at all — acceptable for now given its nature, but worth noting.
- 🔵 **unsafe_assumptions** (L414): `execFileSync` is called synchronously on every `sdkEnv()` invocation (i.e., every session start). If `launchctl` hangs or is slow, the 2-second timeout blocks the Node.js event loop. Consider caching the result with a short TTL (e.g., 30s) or using `execFile` (async). In practice this is unlikely to be a problem since `launchctl getenv` is near-instant, but it's worth being aware of. `[fixable]`

server/chat.ts
dimakis: 🟡 **missing_tests**: The SSH agent socket resolution via `execFileSync('launchctl', ...)` has no test coverage. Since `sdkEnv()` is a private function called once per session start, consider extracting the launchctl lookup into a small testable helper, or at minimum adding an integration note. The 2-second timeout on a synchronous call in the session-start hot path could block the event loop under pathological conditions (e.g. launchd unresponsive during wake). `[fixable]`

server/chat.ts
dimakis: 🔵 **unsafe_assumptions**: `execFileSync` is a blocking call in `sdkEnv()`, which runs on every session start. If `launchctl` hangs (e.g. during a macOS sleep/wake race), it blocks the Node event loop for up to 2 seconds. Consider using `execFile` (async) with a callback/promise, or caching the result with a short TTL so repeated session starts don't each pay the cost. `[fixable]`

packages/client/src/store.ts
```diff
@@ -281,6 +284,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       parserState.currentSessionId = undefined;
       parserState.pendingSend = [];
       clearPendingSendTimer();
+      connection.clearPendingSends();
       connection.send({ type: 'switch_session', sessionId: null });
```
dimakis, line 288: 🔵 **regressions**: In `newSession()`, `connection.clearPendingSends()` is called before `connection.send({ type: 'switch_session', sessionId: null })`. If the WS is currently disconnected, the `switch_session` message itself will be queued into `pendingSends`. This ordering is correct (clear stale, then enqueue fresh). Just confirming this is intentional — the `switchSession` path follows the same pattern at line 255/266.

dimakis (commented):
## Centaur Review

_Clean, well-tested fix for two real bugs (cross-session message leaks and stale SSH sockets). Minor gaps: no test for `switchSession` pending-send clearing, and the synchronous `execFileSync` in `sdkEnv()` could block the event loop under edge conditions._

- 🔵 **missing_tests**: The `switchSession` path also gained `connection.clearPendingSends()` (store.ts:255), but there is no analogous test covering it — only `newSession` is tested. A similar test simulating queued messages during disconnect followed by `switchSession('new-id')` would close the gap. `[fixable]`


### `server/chat.ts`
_Clean, well-tested fix for two real bugs (cross-session message leaks and stale SSH sockets). Minor gaps: no test for `switchSession` pending-send clearing, and the synchronous `execFileSync` in `sdkEnv()` could block the event loop under edge conditions._

- 🟡 **missing_tests** (L412): The SSH agent socket resolution via `execFileSync('launchctl', ...)` has no test coverage. Since `sdkEnv()` is a private function called once per session start, consider extracting the launchctl lookup into a small testable helper, or at minimum adding an integration note. The 2-second timeout on a synchronous call in the session-start hot path could block the event loop under pathological conditions (e.g. launchd unresponsive during wake). `[fixable]`
- 🔵 **unsafe_assumptions** (L414): `execFileSync` is a blocking call in `sdkEnv()`, which runs on every session start. If `launchctl` hangs (e.g. during a macOS sleep/wake race), it blocks the Node event loop for up to 2 seconds. Consider using `execFile` (async) with a callback/promise, or caching the result with a short TTL so repeated session starts don't each pay the cost. `[fixable]`


_Clean, well-tested fix for two real bugs (cross-session message leaks and stale SSH sockets). Minor gaps: no test for `switchSession` pending-send clearing, and the synchronous `execFileSync` in `sdkEnv()` could block the event loop under edge conditions._

- 🔵 **regressions** (L288): In `newSession()`, `connection.clearPendingSends()` is called before `connection.send({ type: 'switch_session', sessionId: null })`. If the WS is currently disconnected, the `switch_session` message itself will be queued into `pendingSends`. This ordering is correct (clear stale, then enqueue fresh). Just confirming this is intentional — the `switchSession` path follows the same pattern at line 255/266.

server/chat.ts
dimakis: 🟡 **unsafe_assumptions**: `execFileSync` blocks the Node.js event loop for up to 2 seconds (the configured timeout). While `sdkEnv()` is only called once per session start, multiple concurrent session starts could stack these blocking calls. Consider using `execFile` (async) with a cached result or `spawnSync` with a shorter timeout. `[fixable]`

server/chat.ts
dimakis: 🔵 **missing_tests**: No tests for the SSH agent socket resolution logic. The `sdkEnv()` function is not exported or tested. Consider extracting the `launchctl` lookup into a testable helper, or at minimum adding a test that verifies `sdkEnv()` returns a valid env object on macOS with and without a working `launchctl`. `[fixable]`

packages/client/src/store.ts
```diff
@@ -281,6 +284,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       parserState.currentSessionId = undefined;
       parserState.pendingSend = [];
       clearPendingSendTimer();
+      connection.clearPendingSends();
```
dimakis, line 287: 🟡 **regressions**: In `newSession()`, `connection.clearPendingSends()` (line 287) runs immediately before `connection.send({ type: 'switch_session', sessionId: null })` (line 288). If the WebSocket is disconnected at this point, the `switch_session` message gets queued in `connection.pendingSends`. A rapid subsequent `newSession()` or `switchSession()` call would then clear that queued `switch_session` message, causing the server to never learn about the session transition. The same pattern exists in `switchSession()` (lines 255 vs 266) but is less risky due to the async gap. Consider sending `switch_session` before clearing, or making control messages bypass the pending queue. `[fixable]`

packages/client/src/store.ts
```diff
@@ -250,6 +250,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
         connection.clearSession(oldId);
       }
       parserState.currentSessionId = id;
```
dimakis, line 252: 🔵 **style**: The three-line clearing sequence (`parserState.pendingSend = []; clearPendingSendTimer(); connection.clearPendingSends();`) is now duplicated identically in both `switchSession` and `newSession`. Consider extracting a `clearAllPendingQueues()` helper to reduce duplication and ensure both paths stay in sync. `[fixable]`

packages/client/__tests__/store.test.ts
```diff
@@ -254,6 +254,45 @@ describe('newSession', () => {
     expect(store.getState().progress.toolIndex).toEqual({});
   });
 
+  it('clears WS pending sends to prevent cross-session message leaks', () => {
```
dimakis, line 257: 🔵 **missing_tests**: The store test only covers `newSession()` clearing pending sends. There is no corresponding test for `switchSession()` clearing pending sends, which is the other call site added by this PR (store.ts line 255). Add a parallel test that verifies `switchSession` also drains stale messages. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid fix for cross-session message leaks with good test coverage on the client side. The server-side `execFileSync` for SSH socket resolution is correct but blocks the event loop; the `switch_session` message after `clearPendingSends` has a subtle drop risk under rapid session transitions while disconnected._

- 🟡 **unsafe_assumptions** (L414): `execFileSync` blocks the Node.js event loop for up to 2 seconds (the configured timeout). While `sdkEnv()` is only called once per session start, multiple concurrent session starts could stack these blocking calls. Consider using `execFile` (async) with a cached result or `spawnSync` with a shorter timeout. `[fixable]`
- 🔵 **missing_tests** (L412): No tests for the SSH agent socket resolution logic. The `sdkEnv()` function is not exported or tested. Consider extracting the `launchctl` lookup into a testable helper, or at minimum adding a test that verifies `sdkEnv()` returns a valid env object on macOS with and without a working `launchctl`. `[fixable]`


_Solid fix for cross-session message leaks with good test coverage on the client side. The server-side `execFileSync` for SSH socket resolution is correct but blocks the event loop; the `switch_session` message after `clearPendingSends` has a subtle drop risk under rapid session transitions while disconnected._

- 🟡 **regressions** (L287): In `newSession()`, `connection.clearPendingSends()` (line 287) runs immediately before `connection.send({ type: 'switch_session', sessionId: null })` (line 288). If the WebSocket is disconnected at this point, the `switch_session` message gets queued in `connection.pendingSends`. A rapid subsequent `newSession()` or `switchSession()` call would then clear that queued `switch_session` message, causing the server to never learn about the session transition. The same pattern exists in `switchSession()` (lines 255 vs 266) but is less risky due to the async gap. Consider sending `switch_session` before clearing, or making control messages bypass the pending queue. `[fixable]`
- 🔵 **style** (L252): The three-line clearing sequence (`parserState.pendingSend = []; clearPendingSendTimer(); connection.clearPendingSends();`) is now duplicated identically in both `switchSession` and `newSession`. Consider extracting a `clearAllPendingQueues()` helper to reduce duplication and ensure both paths stay in sync. `[fixable]`


_Solid fix for cross-session message leaks with good test coverage on the client side. The server-side `execFileSync` for SSH socket resolution is correct but blocks the event loop; the `switch_session` message after `clearPendingSends` has a subtle drop risk under rapid session transitions while disconnected._

- 🔵 **missing_tests** (L257): The store test only covers `newSession()` clearing pending sends. There is no corresponding test for `switchSession()` clearing pending sends, which is the other call site added by this PR (store.ts line 255). Add a parallel test that verifies `switchSession` also drains stale messages. `[fixable]`

server/chat.ts
```diff
@@ -406,6 +406,21 @@ function sdkEnv(): Record<string, string> {
   const venvPaths = getRepoConfig().resolvedVenvPaths;
   env.PATH = [...venvPaths, existingPath].join(':');
 
+  // Resolve the current SSH agent socket on macOS. The server process may have
+  // started with a socket that became stale after sleep/wake — launchctl always
+  // returns the live one.
+  if (platform() === 'darwin') {
+    try {
```
dimakis: 🔵 **missing_tests**: The `sdkEnv()` SSH agent socket resolution has no test coverage. While testing `launchctl` directly is impractical, `sdkEnv()` could be refactored to accept an injectable resolver (or `platform` override) so the branch logic is testable. Not blocking given the graceful fallback, but worth noting. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Clean, well-tested fix for two distinct issues (stale SSH sockets on macOS and cross-session message leaks). Minor gaps: no test for `switchSession` pending-send clearing, and the synchronous `launchctl` call could be made async._

- 🔵 **missing_tests**: The new `switchSession` code path also clears `parserState.pendingSend`, the pending-send timer, and `connection.clearPendingSends()`, but there is no test verifying that `switchSession` drains pending sends. The existing test only covers `newSession`. A parallel test under `describe('switchSession')` would prevent regressions if the two code paths diverge. `[fixable]`


### `server/chat.ts`
_Clean, well-tested fix for two distinct issues (stale SSH sockets on macOS and cross-session message leaks). Minor gaps: no test for `switchSession` pending-send clearing, and the synchronous `launchctl` call could be made async._

- 🔵 **missing_tests** (L413): The `sdkEnv()` SSH agent socket resolution has no test coverage. While testing `launchctl` directly is impractical, `sdkEnv()` could be refactored to accept an injectable resolver (or `platform` override) so the branch logic is testable. Not blocking given the graceful fallback, but worth noting. `[fixable]`
- 🔵 **unsafe_assumptions** (L414): `execFileSync` blocks the event loop for up to 2000ms on every session start. In practice `launchctl getenv` is sub-millisecond, so the risk is low, but if the system is under load or `launchctl` hangs (e.g. during a macOS update), it would stall the server. Consider using `execFile` (async) to avoid blocking, consistent with other async patterns in `_startChatInner`. `[fixable]`

server/chat.ts
dimakis: 🔵 **unsafe_assumptions**: `execFileSync` blocks the event loop for up to 2000ms on every session start. In practice `launchctl getenv` is sub-millisecond, so the risk is low, but if the system is under load or `launchctl` hangs (e.g. during a macOS update), it would stall the server. Consider using `execFile` (async) to avoid blocking, consistent with other async patterns in `_startChatInner`. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-scoped fix. The client-side pending-send clearing is correct and well-tested for `newSession`; the SSH socket refresh is safe with appropriate fallback. Two minor test coverage gaps worth closing._

- 🔵 **missing_tests**: The store test covers `newSession` clearing pending sends but there is no analogous test for `switchSession`, which also gained `parserState.pendingSend = []`, `clearPendingSendTimer()`, and `connection.clearPendingSends()` in this PR. The `switchSession` path has a different shape (it keeps one session active, sends `switch_session` with a real ID, and awaits a REST restore), so a dedicated test would guard against regressions there. `[fixable]`


### `server/chat.ts`
_Clean, well-scoped fix. The client-side pending-send clearing is correct and well-tested for `newSession`; the SSH socket refresh is safe with appropriate fallback. Two minor test coverage gaps worth closing._

- 🔵 **missing_tests** (L412): The `sdkEnv()` SSH_AUTH_SOCK resolution via `launchctl getenv` has no test coverage. A unit test with a stubbed `execFileSync` (or extracted helper) would cover: (a) happy path sets env, (b) empty output preserves inherited value, (c) launchctl failure preserves inherited value, (d) non-darwin skips entirely. `sdkEnv` is currently untested altogether so this is a pre-existing gap, but the new platform-specific branch is worth covering. `[fixable]`

server/chat.ts
dimakis: 🔵 **missing_tests**: The `sdkEnv()` SSH_AUTH_SOCK resolution via `launchctl getenv` has no test coverage. A unit test with a stubbed `execFileSync` (or extracted helper) would cover: (a) happy path sets env, (b) empty output preserves inherited value, (c) launchctl failure preserves inherited value, (d) non-darwin skips entirely. `sdkEnv` is currently untested altogether so this is a pre-existing gap, but the new platform-specific branch is worth covering. `[fixable]`

server/__tests__/chat.test.ts
```diff
@@ -527,3 +527,28 @@ describe('validateResumable', () => {
     expect(result).toEqual({ valid: false });
   });
 });
+
+describe('resolveSshAuthSock', () => {
```
dimakis, line 531: 🟡 **missing_tests**: Both tests are platform-conditional and degenerate to no-ops on the wrong platform: the first test's `else` branch asserts null (same as the second test), and the second test does nothing on macOS. Neither test mocks `execFileSync` or `platform()`, so there's no way to exercise the error/catch path or verify the launchctl command arguments. Consider mocking `child_process.execFileSync` to test: (1) the happy path returns trimmed output, (2) empty string from launchctl returns null, (3) execFileSync throwing returns null, (4) the timeout option is passed correctly. `[fixable]`

server/chat.ts
```diff
@@ -386,6 +386,25 @@ function send(transport: SessionTransport, data: Record<string, unknown> | BootC
 
 const IPV4_PRELOAD = join(dirname(fileURLToPath(import.meta.url)), 'ipv4-preload.cjs');
 
+/**
```
dimakis: 🔵 **style**: The multi-line JSDoc comment is more verbose than the project convention (CLAUDE.md: 'default to writing no comments, one short line max'). The function name and return type already communicate the intent. Consider reducing to a single-line comment or removing entirely. `[fixable]`

server/chat.ts
```diff
@@ -386,6 +386,25 @@ function send(transport: SessionTransport, data: Record<string, unknown> | BootC
...
+ * Resolve the live SSH agent socket on macOS via launchctl.
+ * The server process may have started with a socket that became stale after
+ * sleep/wake — launchctl always returns the current one.
+ * Returns null on non-macOS or if the agent isn't running.
+ */
+export function resolveSshAuthSock(): string | null {
+  if (platform() !== 'darwin') return null;
+  try {
+    const sock = execFileSync('launchctl', ['getenv', 'SSH_AUTH_SOCK'], {
```
dimakis, line 393: 🔵 **unsafe_assumptions**: The 500ms timeout is synchronous (`execFileSync`), meaning it blocks the Node.js event loop on every `sdkEnv()` call. In practice `launchctl getenv` is near-instant, but if `launchctl` hangs (e.g., during macOS update/recovery), the server blocks for 500ms per session start. This is acceptable given the context (session start is already heavy), but worth noting.

dimakis (commented):
## Centaur Review

_Solid, well-scoped fix for stale SSH sockets after macOS sleep/wake. Tests exist but are platform-conditional no-ops — mocking would make them deterministic and cover the error path._

- 🟡 **missing_tests** (L531): Both tests are platform-conditional and degenerate to no-ops on the wrong platform: the first test's `else` branch asserts null (same as the second test), and the second test does nothing on macOS. Neither test mocks `execFileSync` or `platform()`, so there's no way to exercise the error/catch path or verify the launchctl command arguments. Consider mocking `child_process.execFileSync` to test: (1) the happy path returns trimmed output, (2) empty string from launchctl returns null, (3) execFileSync throwing returns null, (4) the timeout option is passed correctly. `[fixable]`


### `server/chat.ts`
_Solid, well-scoped fix for stale SSH sockets after macOS sleep/wake. Tests exist but are platform-conditional no-ops — mocking would make them deterministic and cover the error path._

- 🔵 **style** (L389): The multi-line JSDoc comment is more verbose than the project convention (CLAUDE.md: 'default to writing no comments, one short line max'). The function name and return type already communicate the intent. Consider reducing to a single-line comment or removing entirely. `[fixable]`
- 🔵 **unsafe_assumptions** (L398): The 500ms timeout is synchronous (`execFileSync`), meaning it blocks the Node.js event loop on every `sdkEnv()` call. In practice `launchctl getenv` is near-instant, but if `launchctl` hangs (e.g., during macOS update/recovery), the server blocks for 500ms per session start. This is acceptable given the context (session start is already heavy), but worth noting.

server/__tests__/chat.test.ts
dimakis, line 531: 🟡 **missing_tests**: Both tests are environment-dependent and conditionally skip their real assertions via `if (process.platform)` guards. On macOS, the second test ('returns null on non-macOS platforms') asserts nothing. On Linux, the first test's macOS branch is dead code. Neither test mocks `platform()` or `execFileSync`, so there is no deterministic coverage of the error/fallback path (e.g. launchctl not found, launchctl returning empty string, launchctl timing out). Consider mocking `os.platform` and `child_process.execFileSync` to exercise all branches on every platform. `[fixable]`

server/chat.ts
```diff
@@ -406,6 +425,9 @@ function sdkEnv(): Record<string, string> {
   const venvPaths = getRepoConfig().resolvedVenvPaths;
   env.PATH = [...venvPaths, existingPath].join(':');
 
+  const sock = resolveSshAuthSock();
```
dimakis, line 423: 🔵 **unsafe_assumptions**: `resolveSshAuthSock()` is called synchronously inside `sdkEnv()`, which runs on every `startChat()`. `execFileSync` with a 500ms timeout blocks the event loop. Under normal conditions this is fast, but if launchd is slow (e.g. during wake) this stalls all incoming requests. The current 500ms timeout is reasonable but worth noting — if this becomes a problem, consider caching the result for a short TTL or resolving asynchronously. `[fixable]`

server/chat.ts
dimakis: 🔵 **style**: The project's CLAUDE.md says 'Default to writing no comments. Only add one when the WHY is non-obvious.' The 5-line JSDoc on `resolveSshAuthSock` largely repeats what the function name and code already say. The stale-after-sleep motivation is the only non-obvious part — a single-line comment would suffice. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid, minimal fix for a real macOS sleep/wake problem. Main gap is that the tests are platform-conditional and don't mock, so no single CI environment exercises all branches._

- 🟡 **missing_tests** (L531): Both tests are environment-dependent and conditionally skip their real assertions via `if (process.platform)` guards. On macOS, the second test ('returns null on non-macOS platforms') asserts nothing. On Linux, the first test's macOS branch is dead code. Neither test mocks `platform()` or `execFileSync`, so there is no deterministic coverage of the error/fallback path (e.g. launchctl not found, launchctl returning empty string, launchctl timing out). Consider mocking `os.platform` and `child_process.execFileSync` to exercise all branches on every platform. `[fixable]`


### `server/chat.ts`
_Solid, minimal fix for a real macOS sleep/wake problem. Main gap is that the tests are platform-conditional and don't mock, so no single CI environment exercises all branches._

- 🔵 **unsafe_assumptions** (L428): `resolveSshAuthSock()` is called synchronously inside `sdkEnv()`, which runs on every `startChat()`. `execFileSync` with a 500ms timeout blocks the event loop. Under normal conditions this is fast, but if launchd is slow (e.g. during wake) this stalls all incoming requests. The current 500ms timeout is reasonable but worth noting — if this becomes a problem, consider caching the result for a short TTL or resolving asynchronously. `[fixable]`
- 🔵 **style** (L389): The project's CLAUDE.md says 'Default to writing no comments. Only add one when the WHY is non-obvious.' The 5-line JSDoc on `resolveSshAuthSock` largely repeats what the function name and code already say. The stale-after-sleep motivation is the only non-obvious part — a single-line comment would suffice. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #378, opened 2026-06-09
fix(chat): ensure agentName persists across all session paths

dimakis:
## Summary

- **interrupt_resume** path in `ws-handler-v2.ts` now forwards `agentName` and `telosTaskId` from the existing session
- **REST headless session creation** in `app.ts` now includes `DEFAULT_AGENT_NAME` in both the DB upsert and `startChat` call
- **Boot context persistence** in `chat.ts` now includes `agentName` as a safety net in the fire-and-forget upsert, preventing null agent_name when sessions end before the callback fires

Fixes 6 sessions/day running without boot context due to null `agent_name` in the DB.

## Test plan
- [x] `boot-context.test.ts` — 10/10 passing
- [x] `ws-handler-v2.test.ts` — 135/135 passing
- [ ] Manual: create new session, verify `agent_name` is set in DB
- [ ] Manual: interrupt a session, verify `agent_name` persists
- [ ] Manual: verify boot context pill shows sources on session start

### DIFF
--- a/server/app.ts
+++ b/server/app.ts
@@ -35,7 +35,7 @@ import {
   createSessionWorktrees as createAllWorktrees,
   listWorktrees,
 } from './worktree.js';
-import { GIT_BRANCH_TIMEOUT_MS } from './constants.js';
+import { DEFAULT_AGENT_NAME, GIT_BRANCH_TIMEOUT_MS } from './constants.js';
 import { INTERNAL_TOKEN } from './internal-token.js';
 import { getLocalCommit, isUpdateAvailable } from './git-version.js';
 import { resolvePending } from './permissions.js';
@@ -472,6 +472,7 @@ async function handleSessionCreate(
         initialPrompt,
         isActive: true,
         mode: mode ?? 'agent',
+        agentName: DEFAULT_AGENT_NAME,
       });
 
       // Start the session headlessly — NullTransport discards WS messages
@@ -487,6 +488,7 @@ async function handleSessionCreate(
         mode: mode ?? 'agent',
         model,
         isolation: true,
+        agentName: DEFAULT_AGENT_NAME,
       });
 
       log.info('headless session started', { sessionId: wtId, prompt: initialPrompt });
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -927,19 +927,27 @@ async function _startChatInner(
     buildWorktreeSystemPrompt(repoWorktrees) +
     buildTaskPromptForSession(clientId);
 
-  // Fire-and-forget: fetch boot context from running ContexGin server
+  // Fire-and-forget: fetch boot context from running ContexGin server.
+  // The callback may fire after the session ends (registry cleaned up) or
+  // before the SDK assigns a sessionId (new sessions). We include agentName
+  // in the upsert as a safety net so it's never null in the DB.
   fetchBootContext(agentName)
     .then((msg) => {
-      // Include sessionId for consistency with reconnect/switch replay paths
-      const sid = options.resume ?? registry.get(clientId)?.sessionId;
-      send(transport, { ...msg, ...(sid ? { sessionId: sid } : {}) });
-      // Cache in ManagedSession for replay on reconnect/switch
-      const s = registry.get(clientId);
-      if (s) s.bootContext = msg as unknown as Record<string, unknown>;
-      // Persist to EventStore for cold-path recovery (resumed sessions
-      // may not run the query loop long enough for it to upsert).
+      const session = registry.get(clientId);
+      const sid = options.resume ?? session?.sessionId;
+      // Send to client if transport still connected
+      if (session) {
+        send(transport, { ...msg, ...(sid ? { sessionId: sid } : {}) });
+        session.bootContext = msg as unknown as Record<string, unknown>;
+      }
+      // Persist to EventStore — include agentName as safety net since
+      // this upsert may be the last write for short-lived sessions.
       if (sid) {
-        eventStore.upsertSession({ sessionId: sid, bootContext: JSON.stringify(msg) });
+        eventStore.upsertSession({
+          sessionId: sid,
+          bootContext: JSON.stringify(msg),
+          agentName,
+        });
       }
     })
     .catch((err: unknown) => {
--- a/server/__tests__/event-store.test.ts
+++ b/server/__tests__/event-store.test.ts
@@ -235,6 +235,63 @@ describe('EventStore', () => {
       expect(session).not.toBeNull();
       expect(session!.cwd).toBeNull();
     });
+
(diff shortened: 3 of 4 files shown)


dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `server/app.ts`
_Clean, well-scoped fix that ensures agentName is persisted across all session creation paths (REST headless, boot-context safety net, interrupt-resume). The fallback at chat.ts:838 (`?? DEFAULT_AGENT_NAME`) makes the undefined case safe. No bugs found; main gap is test coverage for the new paths._

- 🔵 **missing_tests** (L475): No tests cover `handleSessionCreate` or verify that `agentName` is persisted in the EventStore upsert and passed to `startChat` for headless sessions. The existing test files (`chat.test.ts`, `boot-context.test.ts`) don't exercise these paths. Given that this is the core fix (ensuring agentName lands in the DB for externally-created sessions), a test confirming the upsert payload would increase confidence. `[fixable]`


### `server/chat.ts`
_Clean, well-scoped fix that ensures agentName is persisted across all session creation paths (REST headless, boot-context safety net, interrupt-resume). The fallback at chat.ts:838 (`?? DEFAULT_AGENT_NAME`) makes the undefined case safe. No bugs found; main gap is test coverage for the new paths._

- 🔵 **regressions** (L939): The old code sent boot context to the transport unconditionally (even if the session had been cleaned up from the registry). The new code guards with `if (session)`, which is strictly better — but it also means that if the registry is cleaned up between the `fetchBootContext` resolve and the send, the client never receives boot context for that session. This is a minor behavioral change; the client can still recover via EventStore on next switch/reattach, so this is fine in practice, just worth noting.

server/app.ts
```diff
@@ -472,6 +472,7 @@ async function handleSessionCreate(
         initialPrompt,
         isActive: true,
         mode: mode ?? 'agent',
+        agentName: DEFAULT_AGENT_NAME,
```
dimakis, line 475: 🔵 **missing_tests**: No tests cover `handleSessionCreate` or verify that `agentName` is persisted in the EventStore upsert and passed to `startChat` for headless sessions. The existing test files (`chat.test.ts`, `boot-context.test.ts`) don't exercise these paths. Given that this is the core fix (ensuring agentName lands in the DB for externally-created sessions), a test confirming the upsert payload would increase confidence. `[fixable]`

server/chat.ts
```diff
@@ -927,19 +927,27 @@ async function _startChatInner(
...
-      send(transport, { ...msg, ...(sid ? { sessionId: sid } : {}) });
-      // Cache in ManagedSession for replay on reconnect/switch
-      const s = registry.get(clientId);
-      if (s) s.bootContext = msg as unknown as Record<string, unknown>;
-      // Persist to EventStore for cold-path recovery (resumed sessions
-      // may not run the query loop long enough for it to upsert).
+      const session = registry.get(clientId);
+      const sid = options.resume ?? session?.sessionId;
+      // Send to client if transport still connected
+      if (session) {
```
dimakis, line 939: 🔵 **regressions**: The old code sent boot context to the transport unconditionally (even if the session had been cleaned up from the registry). The new code guards with `if (session)`, which is strictly better — but it also means that if the registry is cleaned up between the `fetchBootContext` resolve and the send, the client never receives boot context for that session. This is a minor behavioral change; the client can still recover via EventStore on next switch/reattach, so this is fine in practice, just worth noting.

dimakis (commented):
## Centaur Review (2/2)

Found **3** issue(s).

### `server/chat.ts`
_Guards the fire-and-forget boot context callback against registry-cleaned sessions, and includes `agentName` in the EventStore upsert as a belt-and-suspenders write._

- 🟢 **correctness** (L936-951): The `if (session)` guard before `send()` is strictly correct — sending to a transport after registry cleanup could throw or silently fail. The old code sent unconditionally. This is a behavior improvement, not just a cleanup.

- 🟢 **correctness** (L949): Including `agentName` in the boot-context upsert is the right safety net. For short-lived sessions that end before the async `fetchBootContext` resolves, this may be the only write that persists `agentName`. The `agentName` variable is captured from the enclosing scope at session start (line 838: `const agentName = options.agentName ?? DEFAULT_AGENT_NAME`), so it's always defined — no null risk.

### `server/ws-handler-v2.ts`
_Forwards `agentName` and `telosTaskId` from the in-memory session to `startChat` in the interrupt_resume path._

- 🟢 **correctness** (L754-755): The interrupt message schema (`V2InterruptMessage`) doesn't include `agentName` or `telosTaskId`, so reading from `found.session` is the right source. The `send` handler reads from `msg.agentName` because the send schema includes it — this asymmetry is correct and intentional.

- 🔵 **style** (L754): The optional chaining `found.session?.agentName` is defensive but redundant — `findBySessionId` returns `{ clientId: string; session: ManagedSession } | null`, so when `found` is non-null (guaranteed by the early return at L664), `found.session` is always defined. However, this matches the pre-existing pattern at L750 (`found.session?.model`), so consistency wins over pedantry. No change needed.

### `server/app.ts`
_Adds `DEFAULT_AGENT_NAME` to both the EventStore upsert and `startChat` call for REST headless sessions._

- 🟢 **correctness** (L475, L491): Both the DB write and the `startChat` call now include `agentName`. Without this, headless sessions (created via REST API) had null `agent_name` in the DB, and `_startChatInner` would default to `DEFAULT_AGENT_NAME` at runtime but never persist it until the fire-and-forget callback (which races). Now the upsert at L475 writes it eagerly.

- 🔵 **suggestion** (L475, L491): If the REST endpoint ever accepts a user-specified `agentName` (e.g., for different agent profiles), hardcoding `DEFAULT_AGENT_NAME` here would need updating. Currently this is fine — the REST session endpoint doesn't expose agent selection. Just flagging for future awareness.

_Four new tests covering insert, update, combined boot+agent persistence, and the non-clobber guarantee._

- 🟢 **tests** (L239-294): Good coverage of the EventStore layer. The "preserves agentName when updating only bootContext" test (L280) is particularly valuable — it verifies that `upsertSession` with only `bootContext` doesn't null out `agentName`, which is exactly the invariant the chat.ts safety net relies on.

### Cross-cutting

- 🔵 **missing_coverage**: The v1 legacy handler in `index.ts` (lines 719, 745) still doesn't pass `agentName` to `startChat`. This is safe because `_startChatInner` defaults it at line 838, but those sessions will never inherit a session-specific agent name on resume. Since v1 is marked legacy and v2 is the primary path, this is acceptable — just noting for completeness.

**Verdict: Approve.** Clean, well-scoped fix. Each of the three root causes (interrupt_resume, headless REST, fire-and-forget race) is addressed at the right layer. The tests cover the EventStore contract. No regressions. CI passing.

PR #375, opened 2026-06-09
fix(frontend): render HTML tables and images in file viewer

dimakis:
## Summary
- Add `rehype-raw` plugin to `ReactMarkdown` in **FileViewer** and **MarkdownPreviewCard** so raw HTML (e.g. pandas DataFrame `<table>` tags) renders instead of showing as escaped text
- Wrap tables in a scroll container for horizontal overflow on mobile
- Add `.viewer-markdown img` CSS (max-width, border-radius) so images display properly

## Test plan
- [ ] Open a markdown file containing a raw HTML table (e.g. a notebook-generated report) — should render as a styled table, not raw tags
- [ ] Open a markdown file with images — should display inline with rounded corners
- [ ] Verify GFM pipe-syntax tables still render correctly
- [ ] Check MarkdownPreviewCard (inline previews in chat) renders HTML tables too
- [ ] Confirm no XSS risk — content is local files only

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #379, opened 2026-06-10
fix(frontend): extend sanitize schema for img attributes

dimakis:
## Summary
- Extend `rehype-sanitize` default schema to allow `alt`, `width`, `height` on `img` tags
- Default schema only allows `src`, `longDesc`, and aria attributes — missing `alt` means images render without alt text, and `width`/`height` are stripped

Follow-up to #375.

## Test plan
- [ ] Open a markdown file with `alt text` — verify alt text renders on broken/loading images
- [ ] Verify img width/height attributes are preserved in raw HTML images

### DIFF
--- a/frontend/src/lib/markdown-config.tsx
+++ b/frontend/src/lib/markdown-config.tsx
@@ -1,12 +1,20 @@
 import React from 'react';
 import remarkGfm from 'remark-gfm';
 import rehypeRaw from 'rehype-raw';
-import rehypeSanitize from 'rehype-sanitize';
+import rehypeSanitize, { defaultSchema } from 'rehype-sanitize';
 import type { Components } from 'react-markdown';
 import type { PluggableList } from 'unified';
 
+const sanitizeSchema = {
+  ...defaultSchema,
+  attributes: {
+    ...defaultSchema.attributes,
+    img: [...(defaultSchema.attributes?.img ?? []), 'alt', 'width', 'height'],
+  },
+};
+
 export const remarkPlugins: PluggableList = [remarkGfm];
-export const rehypePlugins: PluggableList = [rehypeRaw, rehypeSanitize];
+export const rehypePlugins: PluggableList = [rehypeRaw, [rehypeSanitize, sanitizeSchema]];
 
 export const markdownComponents: Components = {
   table: ({ children, ...props }) => (
--- a/frontend/src/lib/markdown-config.tsx
+++ b/frontend/src/lib/markdown-config.tsx
@@ -9,7 +9,7 @@ const sanitizeSchema = {
   ...defaultSchema,
   attributes: {
     ...defaultSchema.attributes,
-    img: [...(defaultSchema.attributes?.img ?? []), 'alt', 'width', 'height'],
+    img: [...(defaultSchema.attributes?.img ?? []), 'width', 'height'],
   },
 };
 


dimakis:
## Centaur Review

_Good refactoring to centralize markdown config with proper sanitization via rehype-raw + rehype-sanitize. One real bug: double table margins in `.md-preview-card-body` that was fixed for `.viewer-markdown` but missed here. Two minor style nits._

- 🟡 **bugs** (L2968): `.md-preview-card-body table` still has `margin: 0.3em 0` (line 2968 in the full file), but tables are now wrapped in `.table-scroll-wrapper` which adds its own `margin: 0.5em 0`. This causes double vertical margins. The same fix was correctly applied to `.viewer-markdown table` (changed from `margin: 0.5em 0` to `margin: 0`), but was missed here. `[fixable]`


### `frontend/src/lib/markdown-config.tsx`
_Good refactoring to centralize markdown config with proper sanitization via rehype-raw + rehype-sanitize. One real bug: double table margins in `.md-preview-card-body` that was fixed for `.viewer-markdown` but missed here. Two minor style nits._

- 🔵 **style** (L12): The `defaultSchema` from hast-util-sanitize already includes `['alt', 'height', 'src', 'width']` for `img` elements. Adding `'alt', 'width', 'height'` produces duplicates. The spread is harmless but misleading — it implies these attributes aren't already allowed. Either remove the custom `img` entry entirely (since it adds nothing new) or add a comment explaining the intent is to document/pin the allowed list. `[fixable]`
- 🔵 **style** (L1): `import React from 'react'` is unnecessary. The project uses the automatic JSX transform (no other file in `frontend/src/` has this import), so React doesn't need to be in scope for JSX. `[fixable]`

frontend/src/lib/markdown-config.tsx
```diff
@@ -1,12 +1,20 @@
...
-import rehypeSanitize from 'rehype-sanitize';
+import rehypeSanitize, { defaultSchema } from 'rehype-sanitize';
 import type { Components } from 'react-markdown';
 import type { PluggableList } from 'unified';
 
+const sanitizeSchema = {
+  ...defaultSchema,
+  attributes: {
+    ...defaultSchema.attributes,
+    img: [...(defaultSchema.attributes?.img ?? []), 'alt', 'width', 'height'],
```
dimakis: 🔵 **bugs**: `alt` is already included in `defaultSchema.attributes.img` from `hast-util-sanitize` (the GitHub sanitization schema). Spreading the defaults then appending `'alt'` again creates a duplicate entry. Harmless at runtime (the sanitizer checks membership), but misleading — it implies `alt` was previously stripped. Consider removing `'alt'` from the appended list to accurately reflect which attributes this override actually adds. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `frontend/src/lib/markdown-config.tsx`
_Clean, safe change — extends the sanitize schema to preserve img width/height in markdown rendering. One minor redundancy with `alt` already being in the default schema._

- 🔵 **bugs** (L12): `alt` is already included in `defaultSchema.attributes.img` from `hast-util-sanitize` (the GitHub sanitization schema). Spreading the defaults then appending `'alt'` again creates a duplicate entry. Harmless at runtime (the sanitizer checks membership), but misleading — it implies `alt` was previously stripped. Consider removing `'alt'` from the appended list to accurately reflect which attributes this override actually adds. `[fixable]`

PR #380, opened 2026-06-10
fix(chat): deferred boot context persist for new sessions

dimakis:
## Summary

- Boot context was silently lost for most sessions due to a race between the ~100ms ContexGin fetch and the 1-5s SDK session ID assignment. Only **12 of 1117 sessions** had `boot_context` persisted in the EventStore.
- Adds a 5-second deferred retry in the boot context callback — if the session ID becomes available after the initial attempt, persists to EventStore (skips if the query-loop already handled it).
- Makes `agentName` unconditional in the query-loop session-ID upsert — the previous truthiness guard let reconciliation-created rows keep `null` agent_name.

Builds on #378 which fixed the `agentName` gaps in interrupt_resume and headless paths.

## Test plan

- [ ] New `event-store.test.ts` tests pass: deferred persist fills bootContext when query-loop missed it
- [ ] Existing boot-context and event-store tests still pass
- [ ] Manual: start a new Mitzo session, verify `boot_context` column is populated in EventStore within 10s
- [ ] Manual: verify reconnect/switch replays boot context from EventStore

### DIFF
--- a/server/__tests__/event-store.test.ts
+++ b/server/__tests__/event-store.test.ts
@@ -277,6 +277,46 @@ describe('EventStore', () => {
       expect(session!.bootContext).toBe(bootPayload);
     });
 
+    it('deferred persist fills bootContext when query-loop missed it', () => {
+      // Simulates the deferred persist pattern: query-loop creates the session
+      // without bootContext (it wasn't cached yet), then the deferred timer
+      // fires and persists it.
+      store.upsertSession({
+        sessionId: 'sess-deferred',
+        agentName: 'mitzo-conversational',
+        // bootContext not included — simulates query-loop upsert before fetch completed
+      });
+
+      const before = store.getSession('sess-deferred');
+      expect(before!.bootContext).toBeNull();
+
+      // Deferred persist fills in bootContext
+      const bootPayload = JSON.stringify({ type: 'boot_context', source: 'contexgin' });
+      store.upsertSession({
+        sessionId: 'sess-deferred',
+        bootContext: bootPayload,
+        agentName: 'mitzo-conversational',
+      });
+
+      const after = store.getSession('sess-deferred');
+      expect(after!.bootContext).toBe(bootPayload);
+      expect(after!.agentName).toBe('mitzo-conversational');
+    });
+
+    it('deferred persist skips when query-loop already persisted bootContext', () => {
+      // Query-loop persisted boot context successfully — deferred should not overwrite
+      const bootPayload = JSON.stringify({ type: 'boot_context', source: 'contexgin', tokens: 11429 });
+      store.upsertSession({
+        sessionId: 'sess-already',
+        agentName: 'mitzo-conversational',
+        bootContext: bootPayload,
+      });
+
+      const session = store.getSession('sess-already');
+      expect(session!.bootContext).toBe(bootPayload);
+      // Deferred persist would check this and skip
+    });
+
     it('preserves agentName when updating only bootContext', () => {
       store.upsertSession({
         sessionId: 'sess-preserve',
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -931,6 +931,13 @@ async function _startChatInner(
   // The callback may fire after the session ends (registry cleaned up) or
   // before the SDK assigns a sessionId (new sessions). We include agentName
   // in the upsert as a safety net so it's never null in the DB.
+  //
+  // For new sessions the SDK session ID typically arrives 1–5s after query()
+  // starts, while ContexGin responds in ~100ms. The in-memory cache
+  // (session.bootContext) bridges the gap — the query-loop upsert reads it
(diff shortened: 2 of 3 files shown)


dimakis (commented):
## Centaur Review

Found **4** issue(s).


### `server/chat.ts`
_Correct fix for a real race condition. The data-layer change and deferred timer are sound, but the actual fallback path (setTimeout in chat.ts) lacks direct test coverage, and the skip-test is misleading about what it exercises._

- 🔵 **missing_tests** (L963): The deferred setTimeout fallback (the core fix in this PR) has no test coverage. The tests added in event-store.test.ts validate upsert idempotency at the data layer, but the actual race-condition logic — registry lookup after delay, conditional persist via `if (!existing?.bootContext)` — is untested. A unit test with fake timers, a stub registry, and a stub EventStore would exercise this code path directly. `[fixable]`
- 🔵 **style** (L963): The 5000ms delay is a magic number. Consider extracting it to a named constant in `constants.ts` (e.g., `BOOT_CONTEXT_DEFERRED_DELAY_MS`) to make the intent clear and the value easy to find/tune. `[fixable]`
- 🔵 **unsafe_assumptions** (L963): The setTimeout is fire-and-forget — no handle is stored, so it cannot be cancelled if the session is torn down before it fires. The code is safe (optional chaining on `registry.get` handles the missing-session case), but storing the timer handle and clearing it during session cleanup (e.g., in `registry.remove`) would be cleaner and avoid the zombie timer running against a cleaned-up clientId. `[fixable]`


_Correct fix for a real race condition. The data-layer change and deferred timer are sound, but the actual fallback path (setTimeout in chat.ts) lacks direct test coverage, and the skip-test is misleading about what it exercises._

- 🔵 **missing_tests** (L306): The 'deferred persist skips when query-loop already persisted bootContext' test is misleading — it verifies that bootContext survives a round-trip through upsert/getSession, but does not actually test the skip logic (`if (!existing?.bootContext)`). The comment 'Deferred persist would check this and skip' acknowledges this gap. Consider either (a) adding a second upsert without bootContext and asserting the original value is preserved, or (b) renaming the test to clarify it only tests the precondition. `[fixable]`

server/chat.ts
```diff
@@ -948,6 +955,26 @@ async function _startChatInner(
           bootContext: JSON.stringify(msg),
           agentName,
         });
+      } else {
+        // Session ID not yet available (new session, SDK hasn't responded).
+        // The in-memory cache is set above; the query-loop upsert will read
+        // it when the ID arrives. As a belt-and-suspenders fallback, poll
+        // once after a short delay in case the query-loop upsert missed it.
+        setTimeout(() => {
```
dimakis, line 963: 🔵 **missing_tests**: The deferred setTimeout fallback (the core fix in this PR) has no test coverage. The tests added in event-store.test.ts validate upsert idempotency at the data layer, but the actual race-condition logic — registry lookup after delay, conditional persist via `if (!existing?.bootContext)` — is untested. A unit test with fake timers, a stub registry, and a stub EventStore would exercise this code path directly. `[fixable]`

dimakis, line 963: 🔵 **style**: The 5000ms delay is a magic number. Consider extracting it to a named constant in `constants.ts` (e.g., `BOOT_CONTEXT_DEFERRED_DELAY_MS`) to make the intent clear and the value easy to find/tune. `[fixable]`

dimakis, line 963: 🔵 **unsafe_assumptions**: The setTimeout is fire-and-forget — no handle is stored, so it cannot be cancelled if the session is torn down before it fires. The code is safe (optional chaining on `registry.get` handles the missing-session case), but storing the timer handle and clearing it during session cleanup (e.g., in `registry.remove`) would be cleaner and avoid the zombie timer running against a cleaned-up clientId. `[fixable]`

server/__tests__/event-store.test.ts
```diff
@@ -277,6 +277,46 @@ describe('EventStore', () => {
...
+        bootContext: bootPayload,
+        agentName: 'mitzo-conversational',
+      });
+
+      const after = store.getSession('sess-deferred');
+      expect(after!.bootContext).toBe(bootPayload);
+      expect(after!.agentName).toBe('mitzo-conversational');
+    });
+
+    it('deferred persist skips when query-loop already persisted bootContext', () => {
```
dimakis, line 306: 🔵 **missing_tests**: The 'deferred persist skips when query-loop already persisted bootContext' test is misleading — it verifies that bootContext survives a round-trip through upsert/getSession, but does not actually test the skip logic (`if (!existing?.bootContext)`). The comment 'Deferred persist would check this and skip' acknowledges this gap. Consider either (a) adding a second upsert without bootContext and asserting the original value is preserved, or (b) renaming the test to clarify it only tests the precondition. `[fixable]`

dimakis:
Superseded by #369 which takes the comprehensive approach (blocking await + system prompt append). The deferred retry in this PR is no longer needed.

PR #381, opened 2026-06-12
feat(server): SSE + HTTP POST chat transport — Phase 1

dimakis:
## Summary

SSE + HTTP POST transport migration (Phases 1-3 of 4):

### Phase 1 — Server-side SSE endpoints
- **SessionSseRegistry**: Per-connection SSE stream manager with 30s heartbeat, auto-cleanup, seq-based event IDs for replay
- **SseTransport**: SessionTransport adapter (same interface as WsTransport)
- **chat-rest-handler**: 11 HTTP POST endpoints wrapping ws-handler-v2 handlers with Zod validation and X-Connection-ID binding
- **GET /api/chat/events**: SSE stream endpoint with welcome handshake and reconnect-via-query-param

### Phase 2 — Client-side SseConnection
- **SseConnection**: Drop-in replacement for MitzoConnection using EventSource (server→client) + fetch POST (client→server)
- **ChatConnection**: Transport-agnostic interface both MitzoConnection and SseConnection implement
- EventSource auto-reconnects natively — no heartbeat hack needed
- Reconnect URL includes session:seq params for server-side replay

### Phase 3 — Feature flag wiring
- Store accepts optional `sseConfig` — when provided, creates SseConnection instead of MitzoConnection
- localStorage feature flag: `localStorage.setItem('mitzo:transport', 'sse')` + reload
- Default remains WebSocket — enables side-by-side testing

## Motivation

Eliminates the entire class of iOS WebSocket reconnection bugs by replacing WS chat transport with SSE (server→client) + HTTP POST (client→server). SSE connections survive iOS background/foreground transitions that kill WebSocket connections.

## Test plan

- [x] 29 server-side tests passing (SessionSseRegistry, SseTransport, chat-rest-handler)
- [x] 18 client-side tests passing (SseConnection lifecycle, send/receive, queueing, reconnect)
- [x] Type-checks clean
- [ ] Centaur review
- [ ] Manual smoke test: toggle localStorage flag, verify chat works end-to-end over SSE

## Remaining

- Phase 4: Remove WS chat transport (separate PR after SSE transport is validated in production)

### DIFF
--- a/packages/client/src/chat-connection.ts
+++ b/packages/client/src/chat-connection.ts
@@ -1,13 +1,4 @@
-/**
- * ChatConnection — transport-agnostic interface for bidirectional chat.
- *
- * Both MitzoConnection (WebSocket) and SseConnection (SSE + HTTP POST)
- * implement this interface. The store and UI code depend only on this
- * contract, making the transport swappable via feature flag.
- *
- * Provider-agnostic: this interface carries protocol-level events
- * regardless of the inference provider behind them.
- */
+// Transport-agnostic chat connection interface (WS and SSE both implement this).
 
 import type { ConnectionListener } from './connection.js';
 
--- a/packages/client/src/chat-connection.ts
+++ b/packages/client/src/chat-connection.ts
@@ -1,0 +1,45 @@
+/**
+ * ChatConnection — transport-agnostic interface for bidirectional chat.
+ *
+ * Both MitzoConnection (WebSocket) and SseConnection (SSE + HTTP POST)
+ * implement this interface. The store and UI code depend only on this
+ * contract, making the transport swappable via feature flag.
+ *
+ * Provider-agnostic: this interface carries protocol-level events
+ * regardless of the inference provider behind them.
+ */
+
+import type { ConnectionListener } from './connection.js';
+
+export interface ChatConnection {
+  /** Open the transport (start WS connection or SSE EventSource). */
+  connect(): void;
+  /** Close the transport and clean up listeners. */
+  disconnect(): void;
+  /**
+   * Send a protocol message to the server.
+   * Returns true if sent or queued, false if the connection is down
+   * and not recovering.
+   */
+  send(msg: Record<string, unknown>): boolean;
+  /** Register the message listener (only one — last write wins). */
+  onMessage(listener: ConnectionListener): void;
+  /** Whether the transport is connected and ready to send. */
+  isConnected(): boolean;
+  /** Server-assigned connection ID (null before welcome). */
+  getConnectionId(): string | null;
+  /** Track the latest received seq for a session (for reconnect replay). */
+  trackSeq(sessionId: string, seq: number): void;
+  /** Get the last received seq for a session (0 if untracked). */
+  getLastSeq(sessionId: string): number;
+  /** Stop tracking a session (e.g. after close). */
+  clearSession(sessionId: string): void;
+  /** Drain the pending-send queue. */
+  clearPendingSends(): void;
+  /** List all session IDs currently being tracked. */
+  getTrackedSessions(): string[];
+  /** Signal the server that this client is about to be backgrounded. */
+  sendSuspend(): void;
+  /** Check connectivity and reconnect if needed. force=true tears down unconditionally. */
+  checkAndReconnect(force?: boolean): void;
+}
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -198,7 +198,7 @@ export class SseConnection {
     let url = `${this.config.baseUrl}/api/chat/events`;
     if (this._isReconnect && this.seqBySession.size > 0) {
       const sessionsParam = Array.from(this.seqBySession.entries())
-        .map(([sid, seq]) => `${sid}:${seq}`)
+        .map(([sid, seq]) => `${sid}|${seq}`)
         .join(',');
       url += `?sessions=${encodeURIComponent(sessionsParam)}`;
     }
@@ -222,49 +222,22 @@ export class SseConnection {
       this.listener?.({ type: '_open' });
     });
 
-    // All other session events — dispatch to listener
-    const eventTypes = [
-      'message_start',
-      'block_start',
-      'block_delta',
-      'block_stop',
-      'message_stop',
-      'session_end',
-      'session_start',
-      'session_state',
-      'session_list',
-      'permission_request',
-      'task_state',
-      'todo_state',
-      'tool_use',
-      'tool_result',
-      'tool_event',
-      'error',
-      'tokens',
-      'progress',
-      'boot_context_meta',
-      'config',
-      'calendar',
-      'inbox_state',
-    ];
+    // Catch-all for session events. Server sends all non-welcome events as
+    // `event: message`, so es.onmessage handles everything — no allowlist needed.
+    es.onmessage = (e: MessageEvent) => {
+      let msg: Record<string, unknown>;
+      try {
+        msg = JSON.parse(e.data);
+      } catch {
+        return;
+      }
 
-    for (const eventType of eventTypes) {
-      es.addEventListener(eventType, (e: MessageEvent) => {
-        let msg: Record<string, unknown>;
-        try {
-          msg = JSON.parse(e.data);
-        } catch {
-          return;
-        }
+      if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
+        this.seqBySession.set(msg.sessionId as string, msg.seq as number);
+      }
 
-        // Track seq for reconnect replay
-        if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
-          this.seqBySession.set(msg.sessionId as string, msg.seq as number);
-        }
-
-        this.listener?.(msg);
-      });
-    }
+      this.listener?.(msg);
+    };
 
     es.onerror = () => {
       // EventSource auto-reconnects on error. We only need to update
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -14,6 +14,7 @@
  */
 
 import type { ConnectionListener } from './connection.js';
+import type { ChatConnection } from './chat-connection.js';
 
 export interface SseConnectionConfig {
   /** Base URL for API endpoints (e.g. "https://host:3100"). No trailing slash. */
@@ -29,7 +30,7 @@ export interface SseConnectionConfig {
 
 const MAX_PENDING_SENDS = 100;
 
-export class SseConnection {
+export class SseConnection implements ChatConnection {
   private es: EventSource | null = null;
   private _connectionId: string | null = null;
   private _connected = false;
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -1,0 +1,382 @@
+/**
+ * SseConnection — SSE + HTTP POST transport for the v2 protocol.
+ *
+ * Drop-in replacement for MitzoConnection. Same public interface, different
+ * wire transport:
+ *   - Server→client: EventSource (SSE) on GET /api/chat/events
+ *   - Client→server: fetch POST to /api/chat/{send,stop,interrupt,...}
+ *
+ * Eliminates the iOS WebSocket reconnection bug class. EventSource auto-
+ * reconnects natively — no heartbeat hack, no readyState staleness, no
+ * silent kills without onclose firing.
+ *
+ * The server runs both transports in parallel during the migration period.
+ */
+
+import type { ConnectionListener } from './connection.js';
+
+export interface SseConnectionConfig {
+  /** Base URL for API endpoints (e.g. "https://host:3100"). No trailing slash. */
+  baseUrl: string;
+  /** fetch implementation — allows the store to inject apiFetch with auth headers. */
+  fetch: (url: string, init?: RequestInit) => Promise<Response>;
+  /** Factory for EventSource — allows injection for testing. */
+  createEventSource?: (url: string) => EventSource;
+  reconnectDelayMs?: number;
+  /** URL for the sendBeacon suspend fallback. */
+  suspendUrl?: string;
+}
+
+const MAX_PENDING_SENDS = 100;
+
+export class SseConnection {
+  private es: EventSource | null = null;
+  private _connectionId: string | null = null;
+  private _connected = false;
+  private _isReconnect = false;
+  private listener: ConnectionListener | null = null;
+  private seqBySession = new Map<string, number>();
+  private pendingSends: Array<{ endpoint: string; body: Record<string, unknown> }> = [];
+  private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
+  private boundOnVisibility: (() => void) | null = null;
+  private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
+  private boundOnPageHide: (() => void) | null = null;
+  private config: Required<SseConnectionConfig>;
+
+  constructor(config: SseConnectionConfig) {
+    this.config = {
+      createEventSource: (url: string) => new EventSource(url),
+      reconnectDelayMs: 500,
+      suspendUrl: '',
+      ...config,
+    };
+  }
+
+  connect(): void {
+    this.doConnect();
+    this.addBrowserListeners();
+  }
+
+  disconnect(): void {
+    this.removeBrowserListeners();
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer);
+      this.reconnectTimer = null;
+    }
+    if (this.es) {
+      this.es.close();
+      this.es = null;
+    }
+    this._connected = false;
+  }
+
+  /**
+   * Send a message to the server via HTTP POST.
+   *
+   * Maps message types to REST endpoints:
+   *   { type: 'send', ... }      → POST /api/chat/send
+   *   { type: 'stop', ... }      → POST /api/chat/stop
+   *   { type: 'interrupt', ... } → POST /api/chat/interrupt
+   *   etc.
+   *
+   * Returns true if the message was sent or queued, false if not connected
+   * and not reconnecting.
+   */
+  send(msg: Record<string, unknown>): boolean {
+    const endpoint = this.messageTypeToEndpoint(msg.type as string);
+    if (!endpoint) return false;
+
+    if (this._connected && this._connectionId) {
+      this.doPost(endpoint, msg);
+      return true;
+    }
+
+    // Queue if reconnecting
+    if (this.reconnectTimer || this.es) {
+      if (this.pendingSends.length >= MAX_PENDING_SENDS) {
+        this.pendingSends.shift();
+      }
+      this.pendingSends.push({ endpoint, body: msg });
+      return true;
+    }
+
+    return false;
+  }
+
+  onMessage(listener: ConnectionListener): void {
+    this.listener = listener;
+  }
+
+  isConnected(): boolean {
+    return this._connected;
+  }
+
+  getConnectionId(): string | null {
+    return this._connectionId;
+  }
+
+  trackSeq(sessionId: string, seq: number): void {
+    this.seqBySession.set(sessionId, seq);
+  }
+
+  getLastSeq(sessionId: string): number {
+    return this.seqBySession.get(sessionId) ?? 0;
+  }
+
+  clearSession(sessionId: string): void {
+    this.seqBySession.delete(sessionId);
+  }
+
+  clearPendingSends(): void {
+    this.pendingSends = [];
+  }
+
+  getTrackedSessions(): string[] {
+    return Array.from(this.seqBySession.keys());
+  }
+
+  /**
+   * Signal the server that this client is about to be backgrounded.
+   * Uses fetch POST first; falls back to sendBeacon.
+   */
+  sendSuspend(): void {
+    if (this.seqBySession.size === 0) return;
+
+    const sessions = Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+      sessionId,
+      lastSeq,
+    }));
+
+    // Try POST first
+    if (this._connected && this._connectionId) {
+      this.doPost('suspend', { type: 'session_suspend', sessions });
+      return;
+    }
+
+    // sendBeacon fallback
+    if (
+      this.config.suspendUrl &&
+      this._connectionId &&
+      typeof globalThis.navigator?.sendBeacon === 'function'
+    ) {
+      const payload = JSON.stringify({ connectionId: this._connectionId, sessions });
+      globalThis.navigator.sendBeacon(
+        this.config.suspendUrl,
+        new Blob([payload], { type: 'application/json' }),
+      );
+    }
+  }
+
+  /**
+   * Force reconnect — close existing EventSource and reconnect.
+   * Unlike WS, EventSource reconnects automatically, but force=true
+   * tears down and rebuilds for iOS Capacitor lifecycle hooks.
+   */
+  checkAndReconnect(force = false): void {
+    if (!force && this._connected) return;
+    if (this.reconnectTimer) return;
+    if (this.es) {
+      this.es.close();
+      this.es = null;
+    }
+    this._connected = false;
+    this.listener?.({ type: '_close' });
+    this.doConnect();
+  }
+
+  // ─── Internal ──────────────────────────────────────────────────────────────
+
+  private doConnect(): void {
+    if (this.es) return;
+
+    if (this.reconnectTimer) {
+      clearTimeout(this.reconnectTimer);
+      this.reconnectTimer = null;
+    }
+
+    // Build URL with reconnect sessions if this is a reconnect
+    let url = `${this.config.baseUrl}/api/chat/events`;
+    if (this._isReconnect && this.seqBySession.size > 0) {
+      const sessionsParam = Array.from(this.seqBySession.entries())
+        .map(([sid, seq]) => `${sid}:${seq}`)
+        .join(',');
+      url += `?sessions=${encodeURIComponent(sessionsParam)}`;
+    }
+
+    const es = this.config.createEventSource(url);
+    this.es = es;
+
+    // Welcome event — server sends connectionId
+    es.addEventListener('welcome', (e: MessageEvent) => {
+      let msg: Record<string, unknown>;
+      try {
+        msg = JSON.parse(e.data);
+      } catch {
+        return;
+      }
+      this._connectionId = msg.connectionId as string;
+      this._connected = true;
+      this._isReconnect = true;
+
+      this.flushPendingSends();
+      this.listener?.({ type: '_open' });
+    });
+
+    // All other session events — dispatch to listener
+    const eventTypes = [
+      'message_start',
+      'block_start',
+      'block_delta',
+      'block_stop',
+      'message_stop',
+      'session_end',
+      'session_start',
+      'session_state',
+      'session_list',
+      'permission_request',
+      'task_state',
+      'todo_state',
+      'tool_use',
+      'tool_result',
+      'tool_event',
+      'error',
+      'tokens',
+      'progress',
+      'boot_context_meta',
+      'config',
+      'calendar',
+      'inbox_state',
+    ];
+
+    for (const eventType of eventTypes) {
+      es.addEventListener(eventType, (e: MessageEvent) => {
+        let msg: Record<string, unknown>;
+        try {
+          msg = JSON.parse(e.data);
+        } catch {
+          return;
+        }
+
+        // Track seq for reconnect replay
+        if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
+          this.seqBySession.set(msg.sessionId as string, msg.seq as number);
+        }
+
+        this.listener?.(msg);
+      });
+    }
+
+    es.onerror = () => {
+      // EventSource auto-reconnects on error. We only need to update
+      // our state and notify the listener.
+      if (this._connected) {
+        this._connected = false;
+        this.listener?.({ type: '_close' });
+      }
+    };
+
+    // EventSource fires 'open' when the connection is established,
+    // but we wait for the 'welcome' event before marking as connected.
+  }
+
+  private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
+    if (!this._connectionId) return;
+    try {
+      await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Connection-ID': this._connectionId,
+        },
+        body: JSON.stringify(body),
+      });
+    } catch {
+      // POST failures are non-fatal — the server may be temporarily
+      // unreachable. The SSE stream will reconnect and replay missed events.
+    }
+  }
+
+  private flushPendingSends(): void {
+    if (this.pendingSends.length === 0) return;
+    const toFlush = this.pendingSends;
+    this.pendingSends = [];
+    for (const { endpoint, body } of toFlush) {
+      this.doPost(endpoint, body);
+    }
+  }
+
+  /**
+   * Map v2 message types to REST endpoint names.
+   * Returns null for unknown types.
+   */
+  private messageTypeToEndpoint(type: string): string | null {
+    switch (type) {
+      case 'send':
+        return 'send';
+      case 'stop':
+        return 'stop';
+      case 'interrupt':
+        return 'interrupt';
+      case 'permission_response':
+        return 'permission';
+      case 'set_mode':
+        return 'mode';
+      case 'watch':
+        return 'watch';
+      case 'unwatch':
+        return 'unwatch';
+      case 'switch_session':
+        return 'switch';
+      case 'session_suspend':
+        return 'suspend';
+      case 'session_close':
+        return 'close';
+      case 'reconnect':
+        return 'reconnect';
+      default:
+        return null;
+    }
+  }
+
+  // ─── Browser lifecycle ─────────────────────────────────────────────────────
+
+  private addBrowserListeners(): void {
+    if (typeof globalThis.document === 'undefined') return;
+
+    this.boundOnVisibility = () => {
+      if (document.visibilityState === 'visible') {
+        this.checkAndReconnect();
+        this.listener?.({ type: '_foreground' });
+      } else if (document.visibilityState === 'hidden') {
+        this.sendSuspend();
+      }
+    };
+
+    this.boundOnPageShow = (e: PageTransitionEvent) => {
+      if (e.persisted) this.checkAndReconnect();
+    };
+
+    this.boundOnPageHide = () => {
+      this.sendSuspend();
+    };
+
+    document.addEventListener('visibilitychange', this.boundOnVisibility);
+    globalThis.addEventListener('pageshow', this.boundOnPageShow);
+    globalThis.addEventListener('pagehide', this.boundOnPageHide);
+  }
+
+  private removeBrowserListeners(): void {
+    if (this.boundOnVisibility) {
+      document.removeEventListener('visibilitychange', this.boundOnVisibility);
+      this.boundOnVisibility = null;
+    }
+    if (this.boundOnPageShow) {
+      globalThis.removeEventListener('pageshow', this.boundOnPageShow);
+      this.boundOnPageShow = null;
+    }
+    if (this.boundOnPageHide) {
+      globalThis.removeEventListener('pagehide', this.boundOnPageHide);
+      this.boundOnPageHide = null;
+    }
+  }
+}
--- a/server/__tests__/sse-transport.test.ts
+++ b/server/__tests__/sse-transport.test.ts
@@ -1,0 +1,170 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { Response } from 'express';
+import { SessionSseRegistry } from '../session-sse-registry.js';
+import { SseTransport } from '../sse-transport.js';
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+function mockResponse(): Response {
+  const chunks: string[] = [];
+  return {
+    write: vi.fn((data: string) => {
+      chunks.push(data);
+      return true;
+    }),
+    writeHead: vi.fn(),
+    end: vi.fn(),
+    writableEnded: false,
+    // Access captured output for assertions
+    _chunks: chunks,
+  } as unknown as Response & { _chunks: string[] };
+}
+
+// ─── SessionSseRegistry ─────────────────────────────────────────────────────
+
+describe('SessionSseRegistry', () => {
+  let registry: SessionSseRegistry;
+
+  beforeEach(() => {
+    registry = new SessionSseRegistry();
+  });
+
+  it('registers and tracks connections', () => {
+    const res = mockResponse();
+    registry.add('conn-1', res);
+    expect(registry.size).toBe(1);
+    expect(registry.isOpen('conn-1')).toBe(true);
+  });
+
+  it('removes connections', () => {
+    const res = mockResponse();
+    registry.add('conn-1', res);
+    registry.remove('conn-1');
+    expect(registry.size).toBe(0);
+    expect(registry.isOpen('conn-1')).toBe(false);
+  });
+
+  it('closes existing stream when same connectionId re-registers', () => {
+    const res1 = mockResponse();
+    const res2 = mockResponse();
+    registry.add('conn-1', res1);
+    registry.add('conn-1', res2);
+    expect(registry.size).toBe(1);
+    expect(res1.end).toHaveBeenCalled();
+  });
+
+  it('sends SSE event with correct format', () => {
+    const res = mockResponse() as Response & { _chunks: string[] };
+    registry.add('conn-1', res);
+
+    const sent = registry.sendTo('conn-1', { type: 'welcome', protocolVersion: 2 });
+
+    expect(sent).toBe(true);
+    expect(res._chunks).toHaveLength(1);
+    const frame = res._chunks[0];
+    expect(frame).toContain('event: welcome');
+    expect(frame).toContain('data: {"type":"welcome","protocolVersion":2}');
+    expect(frame.endsWith('\n\n')).toBe(true);
+  });
+
+  it('includes SSE id field when seq provided', () => {
+    const res = mockResponse() as Response & { _chunks: string[] };
+    registry.add('conn-1', res);
+
+    registry.sendTo('conn-1', { type: 'block_delta', delta: 'hello' }, 42);
+
+    const frame = res._chunks[0];
+    expect(frame).toContain('id: 42');
+    expect(frame).toContain('event: block_delta');
+  });
+
+  it('omits id field when no seq', () => {
+    const res = mockResponse() as Response & { _chunks: string[] };
+    registry.add('conn-1', res);
+
+    registry.sendTo('conn-1', { type: 'welcome' });
+
+    const frame = res._chunks[0];
+    expect(frame).not.toContain('id:');
+  });
+
+  it('returns false for unknown connection', () => {
+    expect(registry.sendTo('unknown', { type: 'test' })).toBe(false);
+  });
+
+  it('removes connection on write failure', () => {
+    const res = mockResponse();
+    (res.write as ReturnType<typeof vi.fn>).mockImplementation(() => {
+      throw new Error('socket closed');
+    });
+    registry.add('conn-1', res);
+
+    const sent = registry.sendTo('conn-1', { type: 'test' });
+
+    expect(sent).toBe(false);
+    expect(registry.size).toBe(0);
+  });
+
+  it('reports closed when writableEnded', () => {
+    const res = mockResponse();
+    registry.add('conn-1', res);
+    expect(registry.isOpen('conn-1')).toBe(true);
+
+    (res as unknown as { writableEnded: boolean }).writableEnded = true;
+    expect(registry.isOpen('conn-1')).toBe(false);
+  });
+
+  it('destroy closes all streams', () => {
+    const res1 = mockResponse();
+    const res2 = mockResponse();
+    registry.add('conn-1', res1);
+    registry.add('conn-2', res2);
+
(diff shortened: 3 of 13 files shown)


dimakis (commented):
## Centaur Review

### `server/index.ts`
_Sound transport abstraction and good test coverage, but the SSE close handler is missing `removeObserver()` cleanup (session leak risk) and `isSuspended()` check (behavioral parity gap with the WS close handler)._

- 🔴 **bugs** (L430): SSE close handler never calls `registry.removeObserver(transport)`. The WS close handler calls this at line 476 to remove the transport from all session observer sets. Without it, if an SSE transport was the last observer of a detached session, the detach timer never restarts and the session leaks indefinitely. Add `registry.removeObserver(transport)` before `connRegistry.remove(connectionId)`. `[fixable]`
- 🟡 **regressions** (L415): SSE close handler doesn't check `registry.isSuspended(found.clientId)` before calling `detachChat()`. The WS close handler (line 487) skips detach for suspended sessions (iOS backgrounding), delegating lifecycle to the suspend grace timer. Without this check, an SSE disconnect during suspension will prematurely detach the session, breaking the iOS suspend/resume flow. `[fixable]`
- 🟡 **regressions** (L415): SSE close handler checks `isActive(found.clientId)` but the WS close handler (line 493) checks `registry.isAttached(found.clientId)` and also verifies `session.transport === transport` to confirm ownership. The SSE handler's `isActive` check and split-based ownership check are a weaker guard — they don't verify the transport reference matches, so a race between two SSE connections for the same session could detach a session owned by a different transport. `[fixable]`
- 🔵 **style** (L414): Uses `found.clientId.split(':')[0]` to extract the owner connection, but `ws-handler-v2.ts` exports a dedicated `getOwnerConnection()` helper (line 92) that handles edge cases (no colon, multiple colons via `indexOf`). All other call sites use it. Import and use the helper for consistency. `[fixable]`


### `server/chat-rest-handler.ts`
_Sound transport abstraction and good test coverage, but the SSE close handler is missing `removeObserver()` cleanup (session leak risk) and `isSuspended()` check (behavioral parity gap with the WS close handler)._

- 🟡 **unsafe_assumptions** (L148): Endpoints that don't call `getTransport()` (stop, permission, mode, watch, unwatch, switch, suspend, close, reconnect) only validate header presence, not that the connectionId corresponds to a registered connection. Unlike WS where messages only arrive from live sockets, any HTTP client can send an arbitrary X-Connection-ID. The handlers will silently fail when `connRegistry.get(connectionId)` returns undefined. Consider adding a lightweight `connRegistry.has(connectionId)` check after `getConnectionId()`. `[fixable]`
- 🟡 **unsafe_assumptions** (L272): The `/reconnect` endpoint uses manual validation (`!body.sessions || !Array.isArray(body.sessions)`) instead of a Zod schema like all other endpoints. This doesn't validate the shape of individual session entries (sessionId must be string, lastSeq must be number). A malformed entry like `{sessionId: 123}` would pass validation and reach `handleReconnect` with incorrect types. `[fixable]`
- 🔵 **style** (L1): Multi-paragraph docblock at the top of the file (13 lines) and on helper functions violates the project's code style: "default to writing no comments" and "never write multi-paragraph docstrings." The module-level comment describes WHAT (thin wrappers around ws-handler-v2), which the imports and function names already convey. Same applies to session-sse-registry.ts and sse-transport.ts. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,84 @@ const v2Ctx: V2HandlerContext = {
...
+              sessionId,
+              clientId: found.clientId,
+            });
+          }
+        }
+        overviewEmitter.scheduleBroadcast();
+      },
+    );
+
+    connRegistry.remove(connectionId);
```
dimakis, line 410: 🔴 **bugs**: SSE close handler never calls `registry.removeObserver(transport)`. The WS close handler calls this at line 476 to remove the transport from all session observer sets. Without it, if an SSE transport was the last observer of a detached session, the detach timer never restarts and the session leaks indefinitely. Add `registry.removeObserver(transport)` before `connRegistry.remove(connectionId)`. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,84 @@ const v2Ctx: V2HandlerContext = {
...
+    withSpan(
+      'sse.disconnect',
+      { 'sse.connectionId': connectionId },
+      (span) => {
+        // Find all sessions owned by this connection and detach them
+        for (const sessionId of [...(connRegistry.get(connectionId)?.watchedSessions ?? [])]) {
+          const found = registry.findBySessionId(sessionId);
+          if (!found) continue;
+          const ownerConnection = found.clientId.split(':')[0];
+          if (ownerConnection === connectionId && isActive(found.clientId)) {
```
dimakis: 🟡 **regressions**: SSE close handler doesn't check `registry.isSuspended(found.clientId)` before calling `detachChat()`. The WS close handler (line 487) skips detach for suspended sessions (iOS backgrounding), delegating lifecycle to the suspend grace timer. Without this check, an SSE disconnect during suspension will prematurely detach the session, breaking the iOS suspend/resume flow. `[fixable]`

dimakis: 🟡 **regressions**: SSE close handler checks `isActive(found.clientId)` but the WS close handler (line 493) checks `registry.isAttached(found.clientId)` and also verifies `session.transport === transport` to confirm ownership. The SSE handler's `isActive` check and split-based ownership check are a weaker guard — they don't verify the transport reference matches, so a race between two SSE connections for the same session could detach a session owned by a different transport. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,84 @@ const v2Ctx: V2HandlerContext = {
...
+    // Detach owned sessions — same logic as WS close handler
+    withSpan(
+      'sse.disconnect',
+      { 'sse.connectionId': connectionId },
+      (span) => {
+        // Find all sessions owned by this connection and detach them
+        for (const sessionId of [...(connRegistry.get(connectionId)?.watchedSessions ?? [])]) {
+          const found = registry.findBySessionId(sessionId);
+          if (!found) continue;
+          const ownerConnection = found.clientId.split(':')[0];
```
dimakis: 🔵 **style**: Uses `found.clientId.split(':')[0]` to extract the owner connection, but `ws-handler-v2.ts` exports a dedicated `getOwnerConnection()` helper (line 92) that handles edge cases (no colon, multiple colons via `indexOf`). All other call sites use it. Import and use the helper for consistency. `[fixable]`

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,305 @@
...
+    try {
+      handleInterruptV2(connectionId, transport, msg, ctx);
+      res.status(202).json({ ok: true });
+    } catch (err) {
+      log.error('POST /chat/interrupt failed', { connectionId, error: String(err) });
+      res.status(500).json({ ok: false, error: 'Internal server error' });
+    }
+  });
+
+  // POST /api/chat/stop
```
dimakis: 🟡 **unsafe_assumptions**: Endpoints that don't call `getTransport()` (stop, permission, mode, watch, unwatch, switch, suspend, close, reconnect) only validate header presence, not that the connectionId corresponds to a registered connection. Unlike WS where messages only arrive from live sockets, any HTTP client can send an arbitrary X-Connection-ID. The handlers will silently fail when `connRegistry.get(connectionId)` returns undefined. Consider adding a lightweight `connRegistry.has(connectionId)` check after `getConnectionId()`. `[fixable]`

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,305 @@
...
+      res.status(500).json({ ok: false, error: 'Internal server error' });
+    }
+  });
+
+  // POST /api/chat/close
+  router.post('/close', (req, res) => {
+    const connectionId = getConnectionId(req, res);
+    if (!connectionId) return;
+
+    const msg = validateBody(SessionCloseMessage, req.body, res);
```
dimakis, line 238: 🟡 **unsafe_assumptions**: The `/reconnect` endpoint uses manual validation (`!body.sessions || !Array.isArray(body.sessions)`) instead of a Zod schema like all other endpoints. This doesn't validate the shape of individual session entries (sessionId must be string, lastSeq must be number). A malformed entry like `{sessionId: 123}` would pass validation and reach `handleReconnect` with incorrect types. `[fixable]`

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,305 @@
+/**
```
dimakis: 🔵 **style**: Multi-paragraph docblock at the top of the file (13 lines) and on helper functions violates the project's code style: "default to writing no comments" and "never write multi-paragraph docstrings." The module-level comment describes WHAT (thin wrappers around ws-handler-v2), which the imports and function names already convey. Same applies to session-sse-registry.ts and sse-transport.ts. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,382 @@
...
+      this.flushPendingSends();
+      this.listener?.({ type: '_open' });
+    });
+
+    // All other session events — dispatch to listener
+    const eventTypes = [
+      'message_start',
+      'block_start',
+      'block_delta',
+      'block_stop',
```
dimakis: 🔴 **bugs**: Event types `block_stop` and `message_stop` (lines 230-231) do not exist in the v2 protocol. The server emits `block_end` and `message_end` (see query-loop.ts lines 361, 467). Because SSE uses explicit `addEventListener` per event type (unlike WS which receives all messages), these events will be silently dropped — meaning the client will never see block/message completion events over SSE, breaking streaming completely. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,382 @@
...
+      this._connectionId = msg.connectionId as string;
+      this._connected = true;
+      this._isReconnect = true;
+
+      this.flushPendingSends();
+      this.listener?.({ type: '_open' });
+    });
+
+    // All other session events — dispatch to listener
+    const eventTypes = [
```
dimakis: 🔴 **bugs**: The SSE event type allowlist is fundamentally incomplete. Unlike WebSocket (which passes all messages to the listener), SSE silently drops any event type not in this list. Missing critical types include: `boot_context` (listed as `boot_context_meta` — wrong name), `reconnected`, `session_resumed`, `session_switched`, `session_takeover`, `session_cleared`, `session_close_ack`, `watched`, `unwatched`, `mode_changed`, `native_command_result`, `skill_invoked`, `user_message`, `session_id`, `token_update`, and all subagent events (`subagent_start`, `subagent_block_start`, `subagent_block_delta`, `subagent_block_end`, `subagent_end`, `subagent_cancelled`, `subagent_tool_result`). These are all sent by the server via `transport.send()` or `connRegistry.broadcast()` in ws-handler-v2.ts and query-loop.ts. `[fixable]`

dimakis: 🟡 **regressions**: The SSE approach requires maintaining a hardcoded event type allowlist that must stay in sync with every new server event type added in the future. This is a maintenance hazard — any new event added server-side will silently fail on SSE clients unless this list is updated. Consider using a generic `message` event on the SSE stream instead, matching the WS behavior where all events pass through a single channel and are dispatched by `type` field in the JSON payload. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,382 @@
...
+        }
+
+        this.listener?.(msg);
+      });
+    }
+
+    es.onerror = () => {
+      // EventSource auto-reconnects on error. We only need to update
+      // our state and notify the listener.
+      if (this._connected) {
```
dimakis, line 246: 🟡 **unsafe_assumptions**: The `onerror` handler (line 272) relies on EventSource's built-in auto-reconnect, but EventSource auto-reconnect does NOT re-register the named event listeners added in `doConnect()`. When EventSource reconnects internally, it reuses the same object with existing listeners, which is fine. However, if `doConnect()` is called again (via `checkAndReconnect`), a new EventSource is created and listeners are re-registered — but there's no guard against calling `doConnect()` while `es` is non-null in the reconnect path (the guard at line 178 `if (this.es) return` should prevent this, and `checkAndReconnect` sets `this.es = null` first). This is actually correctly handled, disregard the concern about listener duplication — but note that the `_isReconnect` flag is set to `true` on the very first welcome (line 219), meaning the flag name is misleading. It's really `_hasConnectedBefore`.

server/index.ts
```diff
@@ -354,6 +358,84 @@ const v2Ctx: V2HandlerContext = {
   nativeCommands,
 };
 
+// ─── SSE Chat Transport ──────────────────────────────────────────────────────
+// SSE (server→client) + HTTP POST (client→server) transport for chat events.
+// Runs alongside WS during migration. Feature-flagged on the client side.
+
+app.get('/api/chat/events', (req, res) => {
```
dimakis, line 365: 🟡 **missing_tests**: The SSE events endpoint (GET /api/chat/events) — including connection lifecycle, welcome handshake, reconnect query param parsing, and the disconnect/detach logic — has no integration tests. The disconnect handler (lines 402-432) contains significant session-management logic (watched session iteration, suspend check, detach, overview emitter) that mirrors the WS disconnect handler but could diverge. An integration test with supertest would verify the full flow. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,84 @@ const v2Ctx: V2HandlerContext = {
...
+  // Send welcome with connectionId — client uses this in X-Connection-ID header on POSTs
+  chatSseRegistry.sendTo(connectionId, {
+    type: 'welcome',
+    protocolVersion: 2,
+    connectionId,
+  });
+
+  // Handle reconnect if sessions provided as query param
+  const sessionsParam = req.query.sessions as string | undefined;
+  if (sessionsParam) {
```
dimakis, line 388: 🟡 **unsafe_assumptions**: The SSE reconnect query param parser (lines 388-394) uses `entry.split(':')` which will break for session IDs containing colons. While current session IDs likely don't contain colons, this is more fragile than the WS reconnect which sends structured JSON. Consider using a different delimiter or URL-encoding the session IDs. `[fixable]`

server/session-sse-registry.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Close any existing stream for this connection (e.g. stale reconnect)
+    const existing = this.streams.get(connectionId);
+    if (existing) {
+      try {
+        existing.end();
+      } catch {
+        // best effort
+      }
+    }
+
```
dimakis, line 27: 🔵 **style**: The `add()` method logs 'SSE chat stream connected' at line 27, and the SSE endpoint in index.ts also logs 'SSE chat stream connected' at line 400. This produces duplicate log messages for every SSE connection. Remove one of them. `[fixable]`

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,265 @@
+// HTTP POST endpoints for chat operations — thin wrappers around ws-handler-v2.
```
dimakis, line 1: 🔵 **style**: The router has significant boilerplate duplication — each endpoint repeats the same pattern of getConnectionId → requireConnection/getTransport → validateBody → try/catch with identical error handling. A shared middleware or wrapper function would reduce this from ~200 lines to ~80. `[fixable]`

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,265 @@
...
+  schema: { safeParse: (data: unknown) => { success: boolean; data?: T; error?: unknown } },
+  body: unknown,
+  res: Response,
+): T | null {
+  const result = schema.safeParse(body);
+  if (!result.success) {
+    res.status(400).json({ ok: false, error: 'Invalid request body', details: result.error });
+    return null;
+  }
+  return result.data!;
```
dimakis, line 88: 🟡 **bugs**: The `/send` endpoint uses `getTransport()` (which checks both sseRegistry and connRegistry), but `/stop` uses `requireConnection()` (which only checks connRegistry). Both should use the same validation. More importantly, `handleSendV2` receives the `transport` parameter explicitly, but `handleStopV2` doesn't — it looks up the transport via `connRegistry.get(connectionId)`. If the SSE stream dies between the POST arriving and the handler executing, `handleStopV2` will find the connection in connRegistry (registered during SSE setup) but its transport may report `isOpen() = false`. This is a minor race but matches the WS behavior where the same can happen.

dimakis (commented):
## Centaur Review

Found **10** issue(s) (2 critical) (5 warning).


_Critical event type mismatches (`block_stop`/`message_stop` should be `block_end`/`message_end`) and a severely incomplete SSE event allowlist will cause the SSE transport to silently drop most server events, making it non-functional. The server-side plumbing (SessionSseRegistry, SseTransport, chat-rest-handler) is solid._

- 🔴 **bugs** (L230): Event types `block_stop` and `message_stop` (lines 230-231) do not exist in the v2 protocol. The server emits `block_end` and `message_end` (see query-loop.ts lines 361, 467). Because SSE uses explicit `addEventListener` per event type (unlike WS which receives all messages), these events will be silently dropped — meaning the client will never see block/message completion events over SSE, breaking streaming completely. `[fixable]`
- 🔴 **bugs** (L226): The SSE event type allowlist is fundamentally incomplete. Unlike WebSocket (which passes all messages to the listener), SSE silently drops any event type not in this list. Missing critical types include: `boot_context` (listed as `boot_context_meta` — wrong name), `reconnected`, `session_resumed`, `session_switched`, `session_takeover`, `session_cleared`, `session_close_ack`, `watched`, `unwatched`, `mode_changed`, `native_command_result`, `skill_invoked`, `user_message`, `session_id`, `token_update`, and all subagent events (`subagent_start`, `subagent_block_start`, `subagent_block_delta`, `subagent_block_end`, `subagent_end`, `subagent_cancelled`, `subagent_tool_result`). These are all sent by the server via `transport.send()` or `connRegistry.broadcast()` in ws-handler-v2.ts and query-loop.ts. `[fixable]`
- 🟡 **regressions** (L226): The SSE approach requires maintaining a hardcoded event type allowlist that must stay in sync with every new server event type added in the future. This is a maintenance hazard — any new event added server-side will silently fail on SSE clients unless this list is updated. Consider using a generic `message` event on the SSE stream instead, matching the WS behavior where all events pass through a single channel and are dispatched by `type` field in the JSON payload. `[fixable]`
- 🟡 **unsafe_assumptions** (L272): The `onerror` handler (line 272) relies on EventSource's built-in auto-reconnect, but EventSource auto-reconnect does NOT re-register the named event listeners added in `doConnect()`. When EventSource reconnects internally, it reuses the same object with existing listeners, which is fine. However, if `doConnect()` is called again (via `checkAndReconnect`), a new EventSource is created and listeners are re-registered — but there's no guard against calling `doConnect()` while `es` is non-null in the reconnect path (the guard at line 178 `if (this.es) return` should prevent this, and `checkAndReconnect` sets `this.es = null` first). This is actually correctly handled, disregard the concern about listener duplication — but note that the `_isReconnect` flag is set to `true` on the very first welcome (line 219), meaning the flag name is misleading. It's really `_hasConnectedBefore`.


### `server/index.ts`
_Critical event type mismatches (`block_stop`/`message_stop` should be `block_end`/`message_end`) and a severely incomplete SSE event allowlist will cause the SSE transport to silently drop most server events, making it non-functional. The server-side plumbing (SessionSseRegistry, SseTransport, chat-rest-handler) is solid._

- 🟡 **missing_tests** (L365): The SSE events endpoint (GET /api/chat/events) — including connection lifecycle, welcome handshake, reconnect query param parsing, and the disconnect/detach logic — has no integration tests. The disconnect handler (lines 402-432) contains significant session-management logic (watched session iteration, suspend check, detach, overview emitter) that mirrors the WS disconnect handler but could diverge. An integration test with supertest would verify the full flow. `[fixable]`
- 🟡 **unsafe_assumptions** (L388): The SSE reconnect query param parser (lines 388-394) uses `entry.split(':')` which will break for session IDs containing colons. While current session IDs likely don't contain colons, this is more fragile than the WS reconnect which sends structured JSON. Consider using a different delimiter or URL-encoding the session IDs. `[fixable]`


### `server/session-sse-registry.ts`
_Critical event type mismatches (`block_stop`/`message_stop` should be `block_end`/`message_end`) and a severely incomplete SSE event allowlist will cause the SSE transport to silently drop most server events, making it non-functional. The server-side plumbing (SessionSseRegistry, SseTransport, chat-rest-handler) is solid._

- 🔵 **style** (L27): The `add()` method logs 'SSE chat stream connected' at line 27, and the SSE endpoint in index.ts also logs 'SSE chat stream connected' at line 400. This produces duplicate log messages for every SSE connection. Remove one of them. `[fixable]`


### `server/chat-rest-handler.ts`
_Critical event type mismatches (`block_stop`/`message_stop` should be `block_end`/`message_end`) and a severely incomplete SSE event allowlist will cause the SSE transport to silently drop most server events, making it non-functional. The server-side plumbing (SessionSseRegistry, SseTransport, chat-rest-handler) is solid._

- 🔵 **style** (L1): The router has significant boilerplate duplication — each endpoint repeats the same pattern of getConnectionId → requireConnection/getTransport → validateBody → try/catch with identical error handling. A shared middleware or wrapper function would reduce this from ~200 lines to ~80. `[fixable]`
- 🟡 **bugs** (L88): The `/send` endpoint uses `getTransport()` (which checks both sseRegistry and connRegistry), but `/stop` uses `requireConnection()` (which only checks connRegistry). Both should use the same validation. More importantly, `handleSendV2` receives the `transport` parameter explicitly, but `handleStopV2` doesn't — it looks up the transport via `connRegistry.get(connectionId)`. If the SSE stream dies between the POST arriving and the handler executing, `handleStopV2` will find the connection in connRegistry (registered during SSE setup) but its transport may report `isOpen() = false`. This is a minor race but matches the WS behavior where the same can happen.


_Critical event type mismatches (`block_stop`/`message_stop` should be `block_end`/`message_end`) and a severely incomplete SSE event allowlist will cause the SSE transport to silently drop most server events, making it non-functional. The server-side plumbing (SessionSseRegistry, SseTransport, chat-rest-handler) is solid._

- 🔵 **missing_tests**: No test verifies that the event type allowlist in SseConnection matches the actual server event types. A test asserting completeness (or at minimum testing that `block_end` and `message_end` events are dispatched to the listener) would have caught the `block_stop`/`message_stop` naming bug. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,355 @@
...
+  /** Factory for EventSource — allows injection for testing. */
+  createEventSource?: (url: string) => EventSource;
+  reconnectDelayMs?: number;
+  /** URL for the sendBeacon suspend fallback. */
+  suspendUrl?: string;
+}
+
+const MAX_PENDING_SENDS = 100;
+
+export class SseConnection {
```
dimakis: 🔵 **style**: Neither `SseConnection` nor `MitzoConnection` explicitly `implements ChatConnection`. While TypeScript structural typing handles this today, adding `implements ChatConnection` to both classes would catch drift at compile time if the interface gains new methods. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,355 @@
...
+          'X-Connection-ID': this._connectionId,
+        },
+        body: JSON.stringify(body),
+      });
+    } catch {
+      // POST failures are non-fatal — the server may be temporarily
+      // unreachable. The SSE stream will reconnect and replay missed events.
+    }
+  }
+
```
dimakis, line 272: 🔵 **missing_tests**: Browser lifecycle handlers (visibilitychange, pageshow, pagehide) are not tested. The `addBrowserListeners`/`removeBrowserListeners` code could be verified with JSDOM mocks for `document.visibilityState` and `PageTransitionEvent`. The WS `MitzoConnection` similarly lacks these tests, so this is pre-existing — but worth noting for a new transport.

server/index.ts
```diff
@@ -354,6 +358,86 @@ const v2Ctx: V2HandlerContext = {
...
+    try {
+      const sessions = sessionsParam.split(',').map((entry) => {
+        const sep = entry.lastIndexOf('|');
+        const sessionId = sep > 0 ? entry.slice(0, sep) : entry;
+        const seqStr = sep > 0 ? entry.slice(sep + 1) : '0';
+        return { sessionId, lastSeq: parseInt(seqStr, 10) || 0 };
+      });
+      handleReconnect(connectionId, { type: 'reconnect', sessions }, v2Ctx);
```
dimakis, line 396: 🟡 **bugs**: The SSE reconnect query-param parser uses `parseInt(seqStr, 10) || 0`, which means a legitimate `lastSeq` of `NaN` (from a malformed query) silently becomes `0`. This is probably fine since `handleReconnect` replays from seq 0, but it also means a bad `sessions` param like `?sessions=garbage` will silently parse as `[{sessionId:'garbage', lastSeq:0}]` and attempt a watch+replay for a nonexistent session. Unlike the WS path where `ReconnectMessage` Zod validation rejects bad input, this SSE path bypasses schema validation entirely. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,86 @@ const v2Ctx: V2HandlerContext = {
...
+
+app.get('/api/chat/events', (req, res) => {
+  res.writeHead(200, {
+    'Content-Type': 'text/event-stream',
+    'Cache-Control': 'no-cache',
+    Connection: 'keep-alive',
+    'X-Accel-Buffering': 'no',
+  });
+
+  const connectionId = `conn-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
```
dimakis, line 373: 🟡 **unsafe_assumptions**: Connection ID is generated as `conn-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`. The WS path uses a UUID (`randomUUID()`). `Date.now()` + 6 random chars has much lower entropy (~31 bits of randomness) than a UUID (~122 bits). While this isn't a security credential (the auth cookie handles authentication), a collision could cause one SSE client to hijack another's session. Consider using `randomUUID()` for consistency with the WS path. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,86 @@ const v2Ctx: V2HandlerContext = {
...
+  const transport = new SseTransport(connectionId, chatSseRegistry);
+
+  chatSseRegistry.add(connectionId, res);
+  connRegistry.register(connectionId, transport);
```
dimakis, line 377: 🟡 **regressions**: The SSE path calls `connRegistry.register(connectionId, transport)` directly, bypassing `handleHello()`. The WS path goes through `handleHello()` which calls `connRegistry.register()` and then sends the welcome via `conn.transport.send()`. The SSE path sends the welcome via `chatSseRegistry.sendTo()` instead of `transport.send()`. This works, but means any future logic added to `handleHello()` (e.g. logging, metrics, connection limits) won't apply to SSE connections. Consider calling `handleHello()` for SSE too, or extracting shared registration logic. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,86 @@ const v2Ctx: V2HandlerContext = {
   nativeCommands,
 };
 
+// ─── SSE Chat Transport ──────────────────────────────────────────────────────
+// SSE (server→client) + HTTP POST (client→server) transport for chat events.
+// Runs alongside WS during migration. Feature-flagged on the client side.
+
+app.get('/api/chat/events', (req, res) => {
```
dimakis, line 365: 🔵 **missing_tests**: The SSE `/api/chat/events` GET endpoint (connection lifecycle, welcome, reconnect query parsing, disconnect cleanup with session detach) has no integration test. The `chat-rest-handler.test.ts` covers POST endpoints but not the SSE stream setup, welcome handshake, or disconnect-triggered detach logic.

server/chat-rest-handler.ts
```diff
@@ -0,0 +1,265 @@
...
+}
+
```
dimakis, line 90: 🔵 **style**: Every route handler repeats the same try/catch + log.error + 500 pattern. This could be extracted into a small wrapper (e.g. `wrapHandler(name, handler)`) to reduce the 265-line file. Not blocking, but the repetition is notable. `[fixable]`

server/session-sse-registry.ts
```diff
@@ -0,0 +1,133 @@
...
+    return this.streams.size;
+  }
+
+  // ─── Internal ──────────────────────────────────────────────────────────────
+
+  private startHeartbeat(): void {
+    if (this.heartbeatTimer) return;
+
+    log.info('Starting SSE chat heartbeat', { intervalMs: HEARTBEAT_INTERVAL_MS });
+    this.heartbeatTimer = setInterval(() => {
```
dimakis, line 97: 🔵 **bugs**: The heartbeat timer uses `setInterval` but the `startHeartbeat` call only happens in `add()` when `streams.size === 1`. If all connections drop and a new one connects, the heartbeat restarts correctly. However, there's no guard in `destroy()` against double-calls — if `destroy()` is called twice (e.g. during signal handling), the second call iterates an empty map harmlessly, but `stopHeartbeat()` is already null-safe. No actual bug, but the `log.info('Destroying...')` would emit twice.

dimakis (commented):
## Centaur Review

_Well-structured transport abstraction with good test coverage on core components. Main concerns are: SSE reconnect query parsing bypasses Zod validation (unlike WS path), connection ID entropy is lower than WS, and SSE registration bypasses handleHello() which could cause future drift._

- 🔵 **style** (L32): Neither `SseConnection` nor `MitzoConnection` explicitly `implements ChatConnection`. While TypeScript structural typing handles this today, adding `implements ChatConnection` to both classes would catch drift at compile time if the interface gains new methods. `[fixable]`
- 🔵 **missing_tests** (L271): Browser lifecycle handlers (visibilitychange, pageshow, pagehide) are not tested. The `addBrowserListeners`/`removeBrowserListeners` code could be verified with JSDOM mocks for `document.visibilityState` and `PageTransitionEvent`. The WS `MitzoConnection` similarly lacks these tests, so this is pre-existing — but worth noting for a new transport.


### `server/index.ts`
_Well-structured transport abstraction with good test coverage on core components. Main concerns are: SSE reconnect query parsing bypasses Zod validation (unlike WS path), connection ID entropy is lower than WS, and SSE registration bypasses handleHello() which could cause future drift._

- 🟡 **bugs** (L396): The SSE reconnect query-param parser uses `parseInt(seqStr, 10) || 0`, which means a legitimate `lastSeq` of `NaN` (from a malformed query) silently becomes `0`. This is probably fine since `handleReconnect` replays from seq 0, but it also means a bad `sessions` param like `?sessions=garbage` will silently parse as `[{sessionId:'garbage', lastSeq:0}]` and attempt a watch+replay for a nonexistent session. Unlike the WS path where `ReconnectMessage` Zod validation rejects bad input, this SSE path bypasses schema validation entirely. `[fixable]`
- 🟡 **unsafe_assumptions** (L373): Connection ID is generated as `conn-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`. The WS path uses a UUID (`randomUUID()`). `Date.now()` + 6 random chars has much lower entropy (~31 bits of randomness) than a UUID (~122 bits). While this isn't a security credential (the auth cookie handles authentication), a collision could cause one SSE client to hijack another's session. Consider using `randomUUID()` for consistency with the WS path. `[fixable]`
- 🟡 **regressions** (L377): The SSE path calls `connRegistry.register(connectionId, transport)` directly, bypassing `handleHello()`. The WS path goes through `handleHello()` which calls `connRegistry.register()` and then sends the welcome via `conn.transport.send()`. The SSE path sends the welcome via `chatSseRegistry.sendTo()` instead of `transport.send()`. This works, but means any future logic added to `handleHello()` (e.g. logging, metrics, connection limits) won't apply to SSE connections. Consider calling `handleHello()` for SSE too, or extracting shared registration logic. `[fixable]`
- 🔵 **missing_tests** (L365): The SSE `/api/chat/events` GET endpoint (connection lifecycle, welcome, reconnect query parsing, disconnect cleanup with session detach) has no integration test. The `chat-rest-handler.test.ts` covers POST endpoints but not the SSE stream setup, welcome handshake, or disconnect-triggered detach logic.


### `server/chat-rest-handler.ts`
_Well-structured transport abstraction with good test coverage on core components. Main concerns are: SSE reconnect query parsing bypasses Zod validation (unlike WS path), connection ID entropy is lower than WS, and SSE registration bypasses handleHello() which could cause future drift._

- 🔵 **style** (L90): Every route handler repeats the same try/catch + log.error + 500 pattern. This could be extracted into a small wrapper (e.g. `wrapHandler(name, handler)`) to reduce the 265-line file. Not blocking, but the repetition is notable. `[fixable]`


### `server/session-sse-registry.ts`
_Well-structured transport abstraction with good test coverage on core components. Main concerns are: SSE reconnect query parsing bypasses Zod validation (unlike WS path), connection ID entropy is lower than WS, and SSE registration bypasses handleHello() which could cause future drift._

- 🔵 **bugs** (L98): The heartbeat timer uses `setInterval` but the `startHeartbeat` call only happens in `add()` when `streams.size === 1`. If all connections drop and a new one connects, the heartbeat restarts correctly. However, there's no guard in `destroy()` against double-calls — if `destroy()` is called twice (e.g. during signal handling), the second call iterates an empty map harmlessly, but `stopHeartbeat()` is already null-safe. No actual bug, but the `log.info('Destroying...')` would emit twice.

dimakis (commented):
## Centaur Review

Found **9** issue(s) (3 warning).


### `server/chat-rest-handler.ts`
_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🟡 **regressions**: No FIFO message ordering guarantee. The WS handler uses a per-connection promise chain (`dispatchChain`) to ensure async handlers like `handleSwitchSession` complete before the next message is processed. The REST endpoints have no such mechanism — concurrent POSTs from the same client can interleave. For example, a rapid `switch_session` followed by `send` could execute out of order if the switch's async SDK discovery takes time. Consider a per-connection serial queue or document this as a known limitation of the SSE transport. `[fixable]`
- 🔵 **style**: Every route handler has identical try/catch boilerplate (7 lines per endpoint × 11 endpoints = 77 lines). Consider extracting a `wrapHandler` utility that catches errors and logs them, reducing each route to 3-4 lines. `[fixable]`


### `server/index.ts`
_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🟡 **bugs**: SSE `req.on('close')` handler calls `connRegistry.remove(connectionId)` before iterating `watchedSessions`, but the `watchedSessions` snapshot is taken before removal (matching the WS handler pattern). However, in the detach loop, `session.transport === transport` is checked — but for SSE, the session's transport may have been set to a *different* `SseTransport` instance if the same session was reattached by another SSE connection. This is the same pattern as WS so not a regression, but the `SseTransport` instance is created fresh per SSE connection (`new SseTransport(connectionId, chatSseRegistry)`), so identity comparison works correctly. No action needed — noting for reviewer awareness.


### `server/session-sse-registry.ts`
_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🟡 **unsafe_assumptions**: The heartbeat timer is started on first `add()` and only stopped when `streams.size === 0`. If `remove()` is called for a connection that was already removed (e.g. from a write failure in `sendTo` triggering `remove()`, then the `req.on('close')` callback also calling `remove()`), the second `remove()` is a no-op (correct). However, the heartbeat `setInterval` callback iterates `this.streams` and calls `this.remove()` for failures — if `remove()` causes the map to become empty and stops the heartbeat *during* the for-loop iteration, the `clearInterval` is safe because remaining iterations just skip. This is fine but worth a comment.


### `server/__tests__/chat-rest-handler.test.ts`
_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🔵 **missing_tests**: The test mocks all handler functions but doesn't test the `GET /api/chat/events` SSE endpoint (index.ts). The SSE connection setup, welcome event, reconnect query param parsing, and disconnect/detach logic are untested. Consider an integration test using supertest with `Accept: text/event-stream` to verify the SSE stream lifecycle. `[fixable]`


### `packages/client/src/chat-connection.ts`
_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🔵 **style**: The `ChatConnection` interface is well-designed but `MitzoConnection` doesn't explicitly declare `implements ChatConnection`. TypeScript's structural typing makes this work, but an explicit `implements` clause would catch interface drift at compile time (e.g. if a new method is added to `ChatConnection`, `MitzoConnection` would get a compile error). `[fixable]`


_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🔵 **unsafe_assumptions**: The `onerror` handler assumes EventSource will auto-reconnect, which is correct per the spec. However, there's no retry count or backoff — if the server is genuinely down, EventSource will hammer it with reconnect attempts at the browser's default interval (typically 3-5s). The WS transport has `reconnectDelayMs` with explicit reconnect scheduling. For the SSE transport, the native retry is fine for transient errors, but consider adding `retry:` field in the SSE welcome event to control the browser's reconnect interval. `[fixable]`
- 🔵 **bugs**: `doPost()` silently catches all errors. If a `send` POST fails (e.g. 500 from server), the client has no way to know the message wasn't processed. The WS transport would at least get an error event. Consider emitting a `_send_error` event to the listener for critical message types like `send` and `interrupt` so the store can surface the failure to the user. `[fixable]`


_Solid Phase 1 implementation with clean transport abstraction and good test coverage. The main concern is the lack of FIFO message ordering in the REST handler (the WS path serializes async dispatch), which could cause race conditions with rapid client messages during session switches._

- 🔵 **missing_tests**: No test for the `doPost` failure path — what happens when `fetch()` rejects or returns a non-ok response? The current implementation silently swallows errors, which is intentional, but a test should verify the connection remains functional after a POST failure. `[fixable]`

server/index.ts
dimakis, line 365: 🟡 **unsafe_assumptions**: HTTP POST endpoints lack per-request FIFO ordering. The WS handler chains dispatches through a promise chain (`dispatchChain = dispatchChain.then(...)`) to ensure message ordering even with async handlers like `handleSwitchSession`. The REST endpoints handle each request independently — two rapid-fire POSTs (e.g. `/send` then `/switch`) could be processed out of order if the first handler yields. This is a semantic difference from the WS transport that could cause subtle bugs under contention. `[fixable]`

server/index.ts
```diff
@@ -354,6 +358,86 @@ const v2Ctx: V2HandlerContext = {
...
+  connRegistry.register(connectionId, transport);
+
+  // Send welcome with connectionId — client uses this in X-Connection-ID header on POSTs
+  chatSseRegistry.sendTo(connectionId, {
+    type: 'welcome',
+    protocolVersion: 2,
+    connectionId,
+  });
+
+  // Handle reconnect if sessions provided as query param
```
dimakis, line 386: 🔵 **missing_tests**: The `sessions` query parameter parsing on `GET /api/chat/events` (reconnect-on-connect) has no test coverage. Edge cases like malformed entries, missing `|` separator, non-numeric seq values, or empty strings are handled defensively but untested. `[fixable]`

server/__tests__/sse-transport.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    // Access captured output for assertions
+    _chunks: chunks,
+  } as unknown as Response & { _chunks: string[] };
+}
+
+// ─── SessionSseRegistry ─────────────────────────────────────────────────────
+
+describe('SessionSseRegistry', () => {
+  let registry: SessionSseRegistry;
+
```
dimakis, line 27: 🟡 **missing_tests**: SessionSseRegistry tests create a new registry in `beforeEach` but never call `destroy()` in `afterEach`. The `add()` call starts a `setInterval` heartbeat timer that leaks across test runs. Add `afterEach(() => { registry.destroy(); })` to prevent timer leaks and potential flaky test behavior. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid Phase 1 implementation with good test coverage and correct handler reuse. Main concerns: HTTP POST endpoints lack the FIFO dispatch ordering that WS provides, and the `checkAndReconnect` → `_close` → reconnect sequence may race with listener state cleanup._

- 🟡 **unsafe_assumptions** (L365): HTTP POST endpoints lack per-request FIFO ordering. The WS handler chains dispatches through a promise chain (`dispatchChain = dispatchChain.then(...)`) to ensure message ordering even with async handlers like `handleSwitchSession`. The REST endpoints handle each request independently — two rapid-fire POSTs (e.g. `/send` then `/switch`) could be processed out of order if the first handler yields. This is a semantic difference from the WS transport that could cause subtle bugs under contention. `[fixable]`
- 🔵 **missing_tests** (L386): The `sessions` query parameter parsing on `GET /api/chat/events` (reconnect-on-connect) has no test coverage. Edge cases like malformed entries, missing `|` separator, non-numeric seq values, or empty strings are handled defensively but untested. `[fixable]`


### `server/__tests__/sse-transport.test.ts`
_Solid Phase 1 implementation with good test coverage and correct handler reuse. Main concerns: HTTP POST endpoints lack the FIFO dispatch ordering that WS provides, and the `checkAndReconnect` → `_close` → reconnect sequence may race with listener state cleanup._

- 🟡 **missing_tests** (L27): SessionSseRegistry tests create a new registry in `beforeEach` but never call `destroy()` in `afterEach`. The `add()` call starts a `setInterval` heartbeat timer that leaks across test runs. Add `afterEach(() => { registry.destroy(); })` to prevent timer leaks and potential flaky test behavior. `[fixable]`
- 🔵 **missing_tests**: No test coverage for the SSE heartbeat mechanism in SessionSseRegistry — the 30s `:heartbeat\n\n` comment write and cleanup of failed connections during heartbeat are untested. Consider adding a test with fake timers that verifies heartbeat writes and dead connection cleanup. `[fixable]`


### `server/session-sse-registry.ts`
_Solid Phase 1 implementation with good test coverage and correct handler reuse. Main concerns: HTTP POST endpoints lack the FIFO dispatch ordering that WS provides, and the `checkAndReconnect` → `_close` → reconnect sequence may race with listener state cleanup._

- 🔵 **style** (L42): Duplicate log message: `remove()` logs `'SSE chat stream disconnected'` and the `req.on('close')` handler in index.ts also logs the same message. The caller in index.ts should handle logging, or `remove()` should be silent, to avoid duplicate log lines for every disconnect. `[fixable]`


_Solid Phase 1 implementation with good test coverage and correct handler reuse. Main concerns: HTTP POST endpoints lack the FIFO dispatch ordering that WS provides, and the `checkAndReconnect` → `_close` → reconnect sequence may race with listener state cleanup._

- 🟡 **bugs** (L206): In `checkAndReconnect(force=true)`, the method emits `_close` then calls `doConnect()` synchronously. But `doConnect()` will fire `_open` via the EventSource `welcome` handler asynchronously. If the listener reacts to `_close` by cleaning up state (e.g. clearing session tracking), the subsequent reconnect replay via the `sessions` query param will have lost the seq data it needs. The WS version (`MitzoConnection.checkAndReconnect`) preserves `seqBySession` across reconnect — verify the `_close` listener in the store doesn't clear it. `[fixable]`
- 🔵 **style**: SseConnection does not declare `implements ChatConnection`. While TypeScript verifies structural compatibility at the `const connection: ChatConnection` assignment in store.ts, adding `implements ChatConnection` to the class declaration would make the contract explicit and produce better error messages if the interface drifts. `[fixable]`


### `packages/client/src/chat-connection.ts`
_Solid Phase 1 implementation with good test coverage and correct handler reuse. Main concerns: HTTP POST endpoints lack the FIFO dispatch ordering that WS provides, and the `checkAndReconnect` → `_close` → reconnect sequence may race with listener state cleanup._

- 🔵 **style** (L1): The 10-line doc comment at the top of this interface file is unusually verbose for this codebase's convention of minimal comments. The interface name and method signatures are self-documenting; the provider-agnostic note could be a single line. `[fixable]`

server/session-sse-registry.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Remove and clean up an SSE stream.
+   */
+  remove(connectionId: string): void {
+    if (!this.streams.has(connectionId)) return;
+    this.streams.delete(connectionId);
+    log.info('SSE chat stream disconnected', { connectionId, total: this.streams.size });
+
```
dimakis, line 41: 🔵 **style**: Duplicate log message: `remove()` logs `'SSE chat stream disconnected'` and the `req.on('close')` handler in index.ts also logs the same message. The caller in index.ts should handle logging, or `remove()` should be silent, to avoid duplicate log lines for every disconnect. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -0,0 +1,355 @@
...
+    // Build URL with reconnect sessions if this is a reconnect
+    let url = `${this.config.baseUrl}/api/chat/events`;
+    if (this._isReconnect && this.seqBySession.size > 0) {
+      const sessionsParam = Array.from(this.seqBySession.entries())
+        .map(([sid, seq]) => `${sid}|${seq}`)
+        .join(',');
+      url += `?sessions=${encodeURIComponent(sessionsParam)}`;
+    }
+
+    const es = this.config.createEventSource(url);
```
dimakis, line 207: 🟡 **bugs**: In `checkAndReconnect(force=true)`, the method emits `_close` then calls `doConnect()` synchronously. But `doConnect()` will fire `_open` via the EventSource `welcome` handler asynchronously. If the listener reacts to `_close` by cleaning up state (e.g. clearing session tracking), the subsequent reconnect replay via the `sessions` query param will have lost the seq data it needs. The WS version (`MitzoConnection.checkAndReconnect`) preserves `seqBySession` across reconnect — verify the `_close` listener in the store doesn't clear it. `[fixable]`

packages/client/src/chat-connection.ts
```diff
@@ -0,0 +1,45 @@
+/**
```
dimakis: 🔵 **style**: The 10-line doc comment at the top of this interface file is unusually verbose for this codebase's convention of minimal comments. The interface name and method signatures are self-documenting; the provider-agnostic note could be a single line. `[fixable]`

PR #383, opened 2026-06-12
fix(client): SSE reconnect sends POST to restore session watch

dimakis:
## Summary

- EventSource auto-reconnect reuses the original URL (without `?sessions=`), so the server never ran `handleReconnect` — no watch, no reattach, no event replay. Users had to send multiple messages before the session responded.
- Moved reconnect from URL query param to explicit POST in the welcome handler. Fires on every reconnect (auto or explicit), ensuring the server always runs the full flow: watch + reattach + event replay.

## Test plan

- [x] All 329 client tests pass (including 3 new tests for the reconnect behavior)
- [ ] Deploy to dev server and verify: background phone, return, send one message — should get response immediately
- [ ] Verify EventSource auto-reconnect (network drop) triggers reconnect POST
- [ ] Verify explicit `checkAndReconnect(true)` (Capacitor resume) still works

### DIFF
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -195,14 +195,11 @@ export class SseConnection implements ChatConnection {
       this.reconnectTimer = null;
     }
 
-    // Build URL with reconnect sessions if this is a reconnect
-    let url = `${this.config.baseUrl}/api/chat/events`;
-    if (this._isReconnect && this.seqBySession.size > 0) {
-      const sessionsParam = Array.from(this.seqBySession.entries())
-        .map(([sid, seq]) => `${sid}|${seq}`)
-        .join(',');
-      url += `?sessions=${encodeURIComponent(sessionsParam)}`;
-    }
+    // Always use the base URL — reconnect sessions are sent via POST in the
+    // welcome handler. This avoids the bug where EventSource auto-reconnect
+    // reuses the original URL (missing ?sessions=), and eliminates double
+    // handleReconnect when doConnect() AND welcome both trigger it.
+    const url = `${this.config.baseUrl}/api/chat/events`;
 
     const es = this.config.createEventSource(url);
     this.es = es;
@@ -217,6 +214,21 @@ export class SseConnection implements ChatConnection {
       }
       this._connectionId = msg.connectionId as string;
       this._connected = true;
+
+      // Always send reconnect on welcome if we have tracked sessions.
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      if (this._isReconnect && this.seqBySession.size > 0) {
+        this.doPost('reconnect', {
+          type: 'reconnect',
+          sessions: Array.from(this.seqBySession.entries()).map(
+            ([sessionId, lastSeq]) => ({ sessionId, lastSeq }),
+          ),
+        });
+      }
       this._isReconnect = true;
 
       this.flushPendingSends();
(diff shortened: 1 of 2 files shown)


packages/client/src/sse-connection.ts
```diff
@@ -217,6 +214,21 @@ export class SseConnection implements ChatConnection {
...
       this._connectionId = msg.connectionId as string;
       this._connected = true;
+
+      // Always send reconnect on welcome if we have tracked sessions.
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      if (this._isReconnect && this.seqBySession.size > 0) {
```
dimakis, line 224: 🟡 **unsafe_assumptions**: The reconnect POST is fire-and-forget (`doPost` swallows errors silently). If the POST fails (e.g. server hasn't fully registered the connection yet by the time the POST arrives, since welcome is sent before the connection is fully set up), the client has no way to retry and sessions will appear unwatched with no event replay. The old query-param approach was atomic with the connection setup. Consider adding a retry (e.g. one retry after a short delay) or verifying success via the `reconnected` summary message from the server. `[fixable]`

dimakis, line 224: 🔵 **bugs**: Minor ordering consideration: the reconnect POST fires before `flushPendingSends()`. If a user queued a `watch` message for a session that's also in `seqBySession`, the server would process `handleReconnect` (which calls `connRegistry.watch()`) and then the duplicate `watch` POST. This is harmless (watching is idempotent) but worth noting.

dimakis (commented):
## Centaur Review

_Solid fix for the EventSource auto-reconnect URL-reuse bug. The approach is sound — moving session reattach from URL params to a POST on welcome ensures both explicit and native auto-reconnects trigger the full server-side flow. Main concern is the fire-and-forget POST with no retry on failure, and a missing test for the native auto-reconnect scenario (error → auto welcome → POST)._

- 🔵 **missing_tests**: The core bug this PR fixes is native EventSource auto-reconnect (ES fires error, reconnects on its own, sends a new welcome). There's no test simulating that scenario: trigger error → ES auto-fires welcome again → verify reconnect POST is sent. The existing tests only cover explicit `checkAndReconnect(true)`, not the native auto-reconnect path where the same EventSource instance fires a second welcome without `doConnect()` being called. The mock would need to simulate the ES auto-reconnecting (emitting welcome on the same instance after error). `[fixable]`


_Solid fix for the EventSource auto-reconnect URL-reuse bug. The approach is sound — moving session reattach from URL params to a POST on welcome ensures both explicit and native auto-reconnects trigger the full server-side flow. Main concern is the fire-and-forget POST with no retry on failure, and a missing test for the native auto-reconnect scenario (error → auto welcome → POST)._

- 🔵 **regressions**: The server-side SSE handler in `server/index.ts:391-405` still parses `?sessions=` query params from the URL and calls `handleReconnect()`. Now that the client never sends this param, that server code is dead. Not a regression per se (it's harmless), but it's stale code that should be cleaned up in a follow-up to avoid confusion about which reconnect path is active. `[fixable]`
- 🟡 **unsafe_assumptions** (L224): The reconnect POST is fire-and-forget (`doPost` swallows errors silently). If the POST fails (e.g. server hasn't fully registered the connection yet by the time the POST arrives, since welcome is sent before the connection is fully set up), the client has no way to retry and sessions will appear unwatched with no event replay. The old query-param approach was atomic with the connection setup. Consider adding a retry (e.g. one retry after a short delay) or verifying success via the `reconnected` summary message from the server. `[fixable]`
- 🔵 **bugs** (L224): Minor ordering consideration: the reconnect POST fires before `flushPendingSends()`. If a user queued a `watch` message for a session that's also in `seqBySession`, the server would process `handleReconnect` (which calls `connRegistry.watch()`) and then the duplicate `watch` POST. This is harmless (watching is idempotent) but worth noting.

dimakis:
Already merged to main via ef854d9. Closing stale PR.

PR #384, opened 2026-06-14
fix(session-overview): done sessions never transition to idle

dimakis:
## Summary

- **Idle transition timer**: `touch()` now schedules a broadcast at `DONE_TIMEOUT_MS` (5 min) so the done→idle transition fires proactively, even when no other events occur
- **Persistent session timeout**: `persistentToActivity()` now applies the same elapsed-time check instead of hardcoding `state: 'done'` — old attention sessions transition to idle and disappear from the active list

## Root cause

Sessions were stuck as "done" in the Active Sessions list indefinitely because:
1. State derivation was purely on-demand — no timer existed to trigger re-evaluation after 5 minutes of inactivity
2. Sessions from `getAttentionSessions()` (assistant spoke last) were always returned as `done` regardless of age

## Test plan

- [x] Existing tests pass (38/38)
- [x] New test: idle transition timer broadcasts after DONE_TIMEOUT_MS
- [x] New test: subsequent touch resets the idle timer
- [x] New test: persistent sessions transition to idle after timeout
- [x] Updated test: recent persistent sessions still show as done

### DIFF
--- a/server/session-overview.ts
+++ b/server/session-overview.ts
@@ -2,8 +2,9 @@
  * Session Overview — computes SessionActivity[] from server-side registries
  * and broadcasts via SSE with coalescing.
  *
- * State is derived lazily at broadcast time — no stored state, no timers for
- * timeout transitions. This avoids divergence across multiple clients.
+ * State is derived lazily at broadcast time. An idle-transition timer ensures
+ * that "done" sessions are re-evaluated and broadcast as "idle" after
+ * DONE_TIMEOUT_MS, even when no other events occur.
  */
 
 import type { SessionRegistry, ActiveSessionInfo } from '@mitzo/harness';
@@ -82,6 +83,8 @@ export class SessionOverviewEmitter {
   private uncommittedRefreshTimer: ReturnType<typeof setInterval> | null = null;
   /** Guard against overlapping refresh runs. */
   private refreshInFlight = false;
+  /** Timer that fires when the next "done" session should transition to "idle". */
+  private idleTransitionTimer: ReturnType<typeof setTimeout> | null = null;
 
   constructor(deps: SessionOverviewDeps) {
     this.deps = deps;
@@ -91,9 +94,25 @@ export class SessionOverviewEmitter {
   /**
    * Record an event time for a session (call on turn start/end, attach, etc.).
    * Used for Done → Idle timeout derivation.
+   * Also schedules a future broadcast so that done → idle fires on time.
    */
   touch(clientId: string): void {
     this.lastEventTimes.set(clientId, Date.now());
+    this.scheduleIdleTransition();
+  }
+
+  /**
+   * Schedule a broadcast at DONE_TIMEOUT_MS so that sessions entering "done"
+   * transition to "idle" even when no other events occur. Resets any existing
+   * timer — only the most recent touch matters since compute() evaluates all
+   * sessions at broadcast time.
+   */
+  private scheduleIdleTransition(): void {
+    if (this.idleTransitionTimer) clearTimeout(this.idleTransitionTimer);
+    this.idleTransitionTimer = setTimeout(() => {
+      this.idleTransitionTimer = null;
+      this.scheduleBroadcast();
+    }, DONE_TIMEOUT_MS);
   }
 
   /**
@@ -282,27 +301,33 @@ export class SessionOverviewEmitter {
   /**
    * Convert a persistent SessionMeta (from EventStore) to a SessionActivity.
    * These are sessions not currently live but awaiting user reply.
+   * Applies the same DONE_TIMEOUT_MS check as live sessions — persistent
+   * sessions older than the timeout are returned as "idle" so they disappear
+   * from the active list.
    */
   private persistentToActivity(meta: SessionMeta, now: number): SessionActivity {
     const title = meta.summary || meta.sessionId.slice(-8);
     const repo = meta.cwd ? extractRepoName(meta.cwd) : undefined;
     const lastEventAt = meta.lastSpeakerAt ?? meta.updatedAt;
-    const idleMinutes = Math.max(0, Math.round((now - lastEventAt) / 60_000));
+    const elapsed = now - lastEventAt;
+    const idleMinutes = Math.max(0, Math.round(elapsed / 60_000));
 
     // Check for uncommitted work (from background cache)
     const uncommittedWork = meta.cwd ? this.checkUncommittedCached(meta.cwd) : false;
+
+    const state: SessionActivityState = elapsed < DONE_TIMEOUT_MS ? 'done' : 'idle';
 
     return {
       sessionId: meta.sessionId,
       clientId: meta.sessionId, // No live clientId — use sessionId
       title,
       repo,
-      state: 'done',
+      state,
       flags: [],
       lastEventAt,
       taskId: meta.goalId ?? undefined,
       uncommittedWork,
-      awaitingReply: true, // By definition — sourced from getAttentionSessions()
+      awaitingReply: state === 'done', // Only meaningful while still active
       idleMinutes,
     };
   }
@@ -420,6 +445,10 @@ export class SessionOverviewEmitter {
     if (this.coalesceTimer) {
       clearTimeout(this.coalesceTimer);
       this.coalesceTimer = null;
+    }
+    if (this.idleTransitionTimer) {
+      clearTimeout(this.idleTransitionTimer);
+      this.idleTransitionTimer = null;
     }
     if (this.uncommittedRefreshTimer) {
       clearInterval(this.uncommittedRefreshTimer);
--- a/server/session-overview.ts
+++ b/server/session-overview.ts
@@ -102,17 +102,30 @@ export class SessionOverviewEmitter {
   }
 
   /**
-   * Schedule a broadcast at DONE_TIMEOUT_MS so that sessions entering "done"
-   * transition to "idle" even when no other events occur. Resets any existing
-   * timer — only the most recent touch matters since compute() evaluates all
-   * sessions at broadcast time.
+   * Schedule a broadcast when the earliest tracked session should transition
+   * from "done" to "idle". Picks the minimum remaining time across all
+   * tracked sessions so that staggered touches don't delay earlier transitions.
    */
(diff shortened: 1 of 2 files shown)


dimakis:
## Centaur Review

_Solid fix: persistent sessions correctly transition done→idle, and SSE reconnect moves from fragile URL params to a reliable POST. One minor race (unawaited POST before flush) and a timer-precision tradeoff worth noting, but nothing blocking._

- 🟡 **bugs** (L224): The reconnect POST (`doPost`) is fire-and-forget — not awaited before `flushPendingSends()` on line 234. If the client has queued sends, they may arrive at the server before the reconnect POST completes, meaning the server hasn't yet set up watches or reattached sessions. In practice this is likely benign (sends route independently of watch setup, and the server replays missed events), but under high latency the ordering gap widens. Consider awaiting the POST or moving `flushPendingSends` into the POST's `.then()` callback. `[fixable]`


_Solid fix: persistent sessions correctly transition done→idle, and SSE reconnect moves from fragile URL params to a reliable POST. One minor race (unawaited POST before flush) and a timer-precision tradeoff worth noting, but nothing blocking._

- 🔵 **unsafe_assumptions** (L110): The idle-transition timer is a single shared timer that resets on every `touch()` call for any session. If session A finishes at t=0 and session B finishes at t=4min, the timer resets to t=9min. Session A's idle broadcast is delayed 4 minutes beyond its actual DONE_TIMEOUT_MS threshold. This is safe because `compute()` evaluates all sessions at broadcast time (so A will correctly show as idle when the broadcast fires), but the UX lateness scales with inter-session activity gaps. `[fixable]`
- 🔵 **style** (L218): The 6-line block comment on the reconnect POST in `doConnect()` (lines 218–223) largely restates the 4-line comment already present at the `const url` declaration (lines 198–201). The duplication adds noise; a single-line reference like `// POST reconnect — see comment above` would suffice. `[fixable]`


_Solid fix: persistent sessions correctly transition done→idle, and SSE reconnect moves from fragile URL params to a reliable POST. One minor race (unawaited POST before flush) and a timer-precision tradeoff worth noting, but nothing blocking._

- 🔵 **missing_tests**: No test verifies behavior when the reconnect POST fails (e.g., network error or non-ok response). The `doPost` catch block silently swallows errors, which is intentional — a test documenting that the connection stays healthy after a failed reconnect POST would guard this invariant. `[fixable]`

server/session-overview.ts
```diff
@@ -91,9 +94,25 @@ export class SessionOverviewEmitter {
...
+    this.scheduleIdleTransition();
+  }
+
+  /**
+   * Schedule a broadcast at DONE_TIMEOUT_MS so that sessions entering "done"
+   * transition to "idle" even when no other events occur. Resets any existing
+   * timer — only the most recent touch matters since compute() evaluates all
+   * sessions at broadcast time.
+   */
+  private scheduleIdleTransition(): void {
```
dimakis, line 109: 🔵 **unsafe_assumptions**: scheduleIdleTransition resets to a single timer on every touch, keyed to the most recent touch only. If multiple sessions have staggered touch times (e.g., client-1 touched 4 min ago, client-2 touched just now), the timer fires at DONE_TIMEOUT_MS after client-2's touch (5 more minutes). Client-1's done→idle transition at the 5-minute mark is not proactively broadcast — it relies on compute() being triggered by some other event (e.g., the 60s uncommitted refresh). In practice this means up to 60s of stale 'done' state for sessions that hit the threshold between the last touch and the next refresh. Acceptable latency for a UI badge, but worth documenting. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,63 @@ describe('SseConnection', () => {
     expect(conn.isConnected()).toBe(false);
   });
 
-  it('includes sessions in reconnect URL', () => {
-    const conn = new SseConnection(createConfig());
+  it('sends reconnect POST on welcome when has tracked sessions', () => {
```
dimakis, line 293: 🔵 **missing_tests**: No test covers reconnect POST with multiple tracked sessions. The current test tracks only one session. A test with 2+ sessions would verify the serialization of the sessions array and confirm ordering is stable. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,6 +214,22 @@ export class SseConnection implements ChatConnection {
       }
       this._connectionId = msg.connectionId as string;
       this._connected = true;
+
+      // Always send reconnect on welcome if we have tracked sessions.
```
dimakis, line 218: 🔵 **style**: The 6-line comment block (lines 218-223) inside the welcome handler restates the problem description from the commit message and the comment at lines 198-201. The rationale is already captured once above doConnect(); repeating it in the handler adds noise without new information. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Clean, well-tested fix for two real bugs (SSE reconnect missing session reattach, done sessions never transitioning to idle). No correctness issues found — just minor suggestions around multi-session timer edge-case documentation and test coverage._

- 🔵 **unsafe_assumptions** (L110): scheduleIdleTransition resets to a single timer on every touch, keyed to the most recent touch only. If multiple sessions have staggered touch times (e.g., client-1 touched 4 min ago, client-2 touched just now), the timer fires at DONE_TIMEOUT_MS after client-2's touch (5 more minutes). Client-1's done→idle transition at the 5-minute mark is not proactively broadcast — it relies on compute() being triggered by some other event (e.g., the 60s uncommitted refresh). In practice this means up to 60s of stale 'done' state for sessions that hit the threshold between the last touch and the next refresh. Acceptable latency for a UI badge, but worth documenting. `[fixable]`


_Clean, well-tested fix for two real bugs (SSE reconnect missing session reattach, done sessions never transitioning to idle). No correctness issues found — just minor suggestions around multi-session timer edge-case documentation and test coverage._

- 🔵 **missing_tests** (L293): No test covers reconnect POST with multiple tracked sessions. The current test tracks only one session. A test with 2+ sessions would verify the serialization of the sessions array and confirm ordering is stable. `[fixable]`


_Clean, well-tested fix for two real bugs (SSE reconnect missing session reattach, done sessions never transitioning to idle). No correctness issues found — just minor suggestions around multi-session timer edge-case documentation and test coverage._

- 🔵 **style** (L218): The 6-line comment block (lines 218-223) inside the welcome handler restates the problem description from the commit message and the comment at lines 198-201. The rationale is already captured once above doConnect(); repeating it in the handler adds noise without new information. `[fixable]`

dimakis:
Requesting second Centaur review. Context:

- First review found 4 items, 3 of which are in SSE code that's already on main (will drop on rebase)
- The SSE reconnect race (finding #1) has been fixed separately in #390
- The remaining finding (#2, single idle timer) was acknowledged as safe-by-design
- Ready for second review of the actual session-overview changes

server/session-overview.ts
dimakis, line 109: 🔵 **unsafe_assumptions**: scheduleIdleTransition() uses a single timer that resets on every touch(). With multiple sessions touched at different times, earlier sessions don't get a precisely-timed idle broadcast — they rely on the 60-second uncommitted-refresh cycle. For example, session A touched at T=0 and session B touched at T=3min means session A won't get an idle broadcast until T=8min (from B's timer) or T=5min+60s (from the background refresh). Consider scheduling based on the earliest pending transition rather than always resetting to DONE_TIMEOUT_MS from now. `[fixable]`

server/session-overview.ts
```diff
@@ -91,9 +94,25 @@ export class SessionOverviewEmitter {
...
    */
   touch(clientId: string): void {
     this.lastEventTimes.set(clientId, Date.now());
+    this.scheduleIdleTransition();
+  }
+
+  /**
+   * Schedule a broadcast at DONE_TIMEOUT_MS so that sessions entering "done"
+   * transition to "idle" even when no other events occur. Resets any existing
+   * timer — only the most recent touch matters since compute() evaluates all
```
dimakis: 🔵 **style**: Comment says 'only the most recent touch matters since compute() evaluates all sessions at broadcast time' — this is true for correctness but misleading about timeliness. Earlier sessions' done→idle transitions are delayed until either this timer fires or the 60s background refresh runs. Consider noting the worst-case delay. `[fixable]`

dimakis (commented):
## Centaur Review

Found **4** issue(s).


_Clean, well-tested fix. The SSE reconnect change correctly moves session data from URL query params to POST (fixing EventSource auto-reconnect), and the session-overview change properly transitions persistent done sessions to idle. Only minor suggestions around timer precision with multiple staggered sessions._

- 🔵 **unsafe_assumptions** (L110): scheduleIdleTransition() uses a single timer that resets on every touch(). With multiple sessions touched at different times, earlier sessions don't get a precisely-timed idle broadcast — they rely on the 60-second uncommitted-refresh cycle. For example, session A touched at T=0 and session B touched at T=3min means session A won't get an idle broadcast until T=8min (from B's timer) or T=5min+60s (from the background refresh). Consider scheduling based on the earliest pending transition rather than always resetting to DONE_TIMEOUT_MS from now. `[fixable]`
- 🔵 **style** (L107): Comment says 'only the most recent touch matters since compute() evaluates all sessions at broadcast time' — this is true for correctness but misleading about timeliness. Earlier sessions' done→idle transitions are delayed until either this timer fires or the 60s background refresh runs. Consider noting the worst-case delay. `[fixable]`


_Clean, well-tested fix. The SSE reconnect change correctly moves session data from URL query params to POST (fixing EventSource auto-reconnect), and the session-overview change properly transitions persistent done sessions to idle. Only minor suggestions around timer precision with multiple staggered sessions._

- 🔵 **missing_tests**: No test covers the multi-session stagger scenario: session A touched at T=0, session B touched at T=3min — verify session A transitions to idle at approximately T=5min (not delayed until B's timer fires at T=8min). This would document the current behavior and catch regressions if the timer strategy changes. `[fixable]`
- 🔵 **missing_tests**: No test verifies that persistent sessions (from getAttentionSessions) schedule an idle transition timer. Currently persistent sessions only transition to idle when some other event triggers a broadcast (background refresh, another touch). A test exercising this path would make the timing contract explicit. `[fixable]`

PR #385, opened 2026-06-15
fix(infra): move Grafana to port 3002 + auto-start observability

dimakis:
## Summary
- Moved Grafana from port 3001 to 3002 — port 3001 had a persistent ghost reservation in podman's gvproxy networking layer that survived machine restarts and container pruning
- Added `scripts/ensure-observability.sh` that starts podman machine + all compose containers
- Updated `com.mitzo.podman-machine` launchd agent to call the new script — closes the gap where the machine could be running but containers were down for 7 weeks unnoticed
- Updated port references in CLAUDE.md and docs/onboarding.md

## Test plan
- [x] All three containers start and pass health checks
- [x] `curl http://localhost:3002/api/health` returns Grafana health
- [x] `curl http://localhost:3200/ready` returns Loki ready
- [x] `curl http://localhost:16686/api/services` returns Jaeger data
- [ ] Verify launchd agent runs on next reboot

### DIFF
--- a/scripts/ensure-observability.sh
+++ b/scripts/ensure-observability.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# Ensure the observability stack (Jaeger + Loki + Grafana) is running.
+# Called by the com.mitzo.podman-machine launchd agent after machine start,
+# and can be run manually: ./scripts/ensure-observability.sh
+set -euo pipefail
+
+PODMAN="${PODMAN:-$(command -v podman 2>/dev/null || echo /opt/homebrew/bin/podman)}"
+COMPOSE_FILE="$(dirname "$0")/../docker-compose.yml"
+
+# 1. Ensure the podman machine is running
+if ! "$PODMAN" machine inspect --format '{{.State}}' 2>/dev/null | grep -q "running"; then
+  echo "[ensure-observability] Starting podman machine..."
+  "$PODMAN" machine start 2>&1 || true
+  # Wait for the API socket to appear
+  for i in $(seq 1 30); do
+    "$PODMAN" info >/dev/null 2>&1 && break
+    sleep 1
+  done
+fi
+
+# 2. Bring up the compose stack
+echo "[ensure-observability] Starting observability containers..."
+"$PODMAN" compose -f "$COMPOSE_FILE" up -d 2>&1
+
+# 3. Verify
+for svc in jaeger loki grafana; do
+  container="mitzo_${svc}_1"
+  if "$PODMAN" ps --filter "name=$container" --format '{{.Names}}' 2>/dev/null | grep -q "$container"; then
+    echo "[ensure-observability] $container: running"
+  else
+    echo "[ensure-observability] $container: FAILED TO START" >&2
+  fi
+done
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -247,7 +247,7 @@ Auth via `ProtectedRoute` wrapper. Vite dev server proxies `/api` and `/ws` to b
 - **Tracing:** OpenTelemetry via `server/tracing.ts`, opt-in when `OTEL_EXPORTER_OTLP_ENDPOINT` is set. BatchSpanProcessor with OTLP HTTP exporter to Jaeger. Currently instrumented: `ws.switch_session`, `ws.send`, `ws.reconnect`. Deep instrumentation roadmap in `docs/design/otel-deep-instrumentation.md`.
 - **Infrastructure:** Three containers via `docker-compose.yml` (podman). Start with `npm run tracing:up`, stop with `npm run tracing:down`.
   - Jaeger: http://localhost:16686 (trace viewer, OTLP receiver on :4318)
-  - Grafana: http://localhost:3001 (log viewer + dashboards, no login required)
+  - Grafana: http://localhost:3002 (log viewer + dashboards, no login required)
   - Loki: http://localhost:3200 (log aggregation backend)
 - **Env vars:** `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318`, `LOKI_HOST=http://localhost:3200`, `LOG_LEVEL=info` (default).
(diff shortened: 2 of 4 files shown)


dimakis:
## Centaur Review

### `server/index.ts`
_Solid infra fix for the Grafana port conflict and a well-motivated SSE reconnect refactor. Main concern: the server-side legacy `?sessions=` query param handler is now dead code and should be cleaned up in this PR to avoid dual reconnect paths._

- 🟡 **regressions** (L391): The server-side `?sessions=` query param handler (lines 391-405) is now dead code — the client no longer sends session data as a URL param. This legacy path should be removed in this PR to avoid dual reconnect code paths and confusion about which mechanism is canonical. `[fixable]`


### `scripts/ensure-observability.sh`
_Solid infra fix for the Grafana port conflict and a well-motivated SSE reconnect refactor. Main concern: the server-side legacy `?sessions=` query param handler is now dead code and should be cleaned up in this PR to avoid dual reconnect paths._

- 🟡 **unsafe_assumptions** (L27): Container names `mitzo_<svc>_1` assume the compose project name is `mitzo`, which is inferred from the directory name. If the repo is cloned to a different directory name (or podman compose uses a different naming scheme like `mitzo-<svc>-1` with hyphens), the verification loop will report false failures. Consider using `podman compose -f $COMPOSE_FILE ps --format '{{.Names}}'` or `--services` to query actual container names, or set an explicit `name:` in docker-compose.yml. `[fixable]`
- 🔵 **style** (L15): Loop variable `i` in `for i in $(seq 1 30)` is unused — the loop body only checks `podman info`. This is a minor shellcheck warning (SC2034). Consider `for _ in $(seq 1 30)` or a while loop with a counter for clarity. `[fixable]`


_Solid infra fix for the Grafana port conflict and a well-motivated SSE reconnect refactor. Main concern: the server-side legacy `?sessions=` query param handler is now dead code and should be cleaned up in this PR to avoid dual reconnect paths._

- 🔵 **bugs** (L225): The reconnect POST (`doPost`) is fire-and-forget (not awaited) inside a synchronous event listener. If it fails (network blip, server restart), the client silently loses event replay with no retry. Consider logging the failure or scheduling a retry — currently the only recovery is a full reconnect cycle, which may not happen if the SSE stream stays open. `[fixable]`

scripts/ensure-observability.sh
```diff
@@ -0,0 +1,33 @@
...
+  done
+fi
+
+# 2. Bring up the compose stack
+echo "[ensure-observability] Starting observability containers..."
+"$PODMAN" compose -f "$COMPOSE_FILE" up -d 2>&1
+
+# 3. Verify
+for svc in jaeger loki grafana; do
+  container="mitzo_${svc}_1"
```
dimakis, line 27: 🟡 **unsafe_assumptions**: Container name pattern `mitzo_${svc}_1` uses underscore separators (docker-compose v1 convention). Podman compose and docker compose v2 use hyphens (`mitzo-${svc}-1`). Additionally, the `mitzo` project name is derived from the directory name at runtime — if the repo is cloned to a different folder, the prefix won't match. The verification loop will always report 'FAILED TO START' even when containers are healthy. Consider using `podman compose -f "$COMPOSE_FILE" ps --format json` or service-based filtering instead of hardcoded container names. `[fixable]`

scripts/ensure-observability.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# Ensure the observability stack (Jaeger + Loki + Grafana) is running.
+# Called by the com.mitzo.podman-machine launchd agent after machine start,
+# and can be run manually: ./scripts/ensure-observability.sh
+set -euo pipefail
+
+PODMAN="${PODMAN:-$(command -v podman 2>/dev/null || echo /opt/homebrew/bin/podman)}"
```
dimakis, line 7: 🔵 **unsafe_assumptions**: The npm scripts in package.json use `docker compose` while this script uses `podman compose`. If the user has only Docker (not Podman), the script will fail. Consider aligning with the existing convention or supporting both runtimes (e.g., `CONTAINER_CMD=${CONTAINER_CMD:-docker}`). `[fixable]`

scripts/ensure-observability.sh
```diff
@@ -0,0 +1,33 @@
...
+COMPOSE_FILE="$(dirname "$0")/../docker-compose.yml"
+
+# 1. Ensure the podman machine is running
+if ! "$PODMAN" machine inspect --format '{{.State}}' 2>/dev/null | grep -q "running"; then
+  echo "[ensure-observability] Starting podman machine..."
+  "$PODMAN" machine start 2>&1 || true
+  # Wait for the API socket to appear
+  for i in $(seq 1 30); do
```
dimakis, line 15: 🔵 **bugs**: The variable `i` in the polling loop (`for i in $(seq 1 30)`) is unused — the loop will spin for up to 30 seconds but silently exits if podman never becomes ready (no error/warning on timeout). Consider adding a timeout message after the loop. `[fixable]`

dimakis (commented):
## Centaur Review

### `scripts/ensure-observability.sh`
_Port change and docs are correct and complete. The new ensure-observability script has a container-naming assumption (`_` vs `-` separators) that will cause the verification step to always report failure on podman compose / docker compose v2._

- 🟡 **unsafe_assumptions** (L27): Container name pattern `mitzo_${svc}_1` uses underscore separators (docker-compose v1 convention). Podman compose and docker compose v2 use hyphens (`mitzo-${svc}-1`). Additionally, the `mitzo` project name is derived from the directory name at runtime — if the repo is cloned to a different folder, the prefix won't match. The verification loop will always report 'FAILED TO START' even when containers are healthy. Consider using `podman compose -f "$COMPOSE_FILE" ps --format json` or service-based filtering instead of hardcoded container names. `[fixable]`
- 🔵 **unsafe_assumptions** (L7): The npm scripts in package.json use `docker compose` while this script uses `podman compose`. If the user has only Docker (not Podman), the script will fail. Consider aligning with the existing convention or supporting both runtimes (e.g., `CONTAINER_CMD=${CONTAINER_CMD:-docker}`). `[fixable]`
- 🔵 **bugs** (L15): The variable `i` in the polling loop (`for i in $(seq 1 30)`) is unused — the loop will spin for up to 30 seconds but silently exits if podman never becomes ready (no error/warning on timeout). Consider adding a timeout message after the loop. `[fixable]`

PR #386, opened 2026-06-15
fix(chat): prevent duplicate user messages via idempotency guard

dimakis:
## Summary

- Adds `EventStore.hasUserMessage(sessionId, messageId)` — efficient SQLite check before storing user_message events
- Guards `sendToChat`, `interruptChat`, and the `startChat` resume path with the idempotency check — if a POST is retried (network retry, iOS lifecycle, SSE reconnect), the duplicate is silently dropped (no double echo, no double inputQueue push)
- Adds ref-based send guard in `ChatInput.handleSend` to prevent double-tap from firing two sends before React re-renders
- Fixes test isolation: existing tests used static session IDs that collided across test files sharing the file-based event store

## Test plan
- [x] `hasUserMessage` unit tests (4 cases: miss, hit, cross-session, non-user_message)
- [x] `sendToChat` dedup integration test (same clientMsgId sent twice → only one echo + one inputQueue push)
- [x] All existing event-store, send-to-chat, and chat tests pass

### DIFF
--- a/frontend/src/components/ChatInput.tsx
+++ b/frontend/src/components/ChatInput.tsx
@@ -144,12 +144,12 @@ export function ChatInput({
       clearDraft();
       setImages([]);
       if (!useExternal) setContextBlocks([]);
+      autoResize();
+      textareaRef.current?.focus();
     }
     requestAnimationFrame(() => {
       sendGuard.current = false;
     });
-    autoResize();
-    textareaRef.current?.focus();
   }
 
   function handleKeyDown(e: KeyboardEvent) {
--- a/frontend/src/components/ChatInput.tsx
+++ b/frontend/src/components/ChatInput.tsx
@@ -70,6 +70,7 @@ export function ChatInput({
   const fileInputRef = useRef<HTMLInputElement>(null);
   const initialApplied = useRef(false);
   const prevRunning = useRef(running);
+  const sendGuard = useRef(false);
 
   const autoResize = useCallback(() => {
     const el = textareaRef.current;
@@ -129,8 +130,10 @@ export function ChatInput({
   }
 
   function handleSend() {
+    if (sendGuard.current) return;
     const trimmed = text.trim();
     if (!trimmed && images.length === 0) return;
+    sendGuard.current = true;
     const sent = onSend(
       trimmed || 'What do you see in this image?',
       images.length > 0 ? images : undefined,
@@ -141,11 +144,12 @@ export function ChatInput({
       clearDraft();
       setImages([]);
       if (!useExternal) setContextBlocks([]);
-      requestAnimationFrame(() => {
-        autoResize();
-        textareaRef.current?.focus();
-      });
     }
+    requestAnimationFrame(() => {
+      sendGuard.current = false;
+      autoResize();
+      textareaRef.current?.focus();
+    });
   }
 
   function handleKeyDown(e: KeyboardEvent) {
--- a/frontend/src/components/ChatInput.tsx
+++ b/frontend/src/components/ChatInput.tsx
@@ -147,9 +147,9 @@ export function ChatInput({
     }
     requestAnimationFrame(() => {
       sendGuard.current = false;
-      autoResize();
-      textareaRef.current?.focus();
     });
+    autoResize();
+    textareaRef.current?.focus();
   }
 
   function handleKeyDown(e: KeyboardEvent) {
@@ -270,9 +270,11 @@ export function ChatInput({
   }
 
   function handleInterrupt() {
+    if (sendGuard.current) return;
     if (!onInterrupt) return;
     const trimmed = text.trim();
     if (!trimmed && images.length === 0) return;
+    sendGuard.current = true;
     onInterrupt(
       trimmed || 'What do you see in this image?',
       images.length > 0 ? images : undefined,
@@ -282,9 +284,10 @@ export function ChatInput({
     setImages([]);
     if (!useExternal) setContextBlocks([]);
     requestAnimationFrame(() => {
-      autoResize();
-      textareaRef.current?.focus();
+      sendGuard.current = false;
     });
+    autoResize();
+    textareaRef.current?.focus();
   }
 
   return (
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -126,6 +126,7 @@ export class EventStore {
     this.migrateAttentionTracking(db);
     this.migrateSessionState(db);
     this.migrateBootContext(db);
+    this.migrateUserMessageIndex(db);
 
     this.log.info('EventStore initialized', { dbPath });
 
@@ -302,6 +303,13 @@ export class EventStore {
       db.exec('ALTER TABLE sessions ADD COLUMN boot_context TEXT');
       this.log.info('migrated sessions table: added boot_context');
     }
+  }
+
+  private migrateUserMessageIndex(db: Database.Database): void {
+    db.exec(
+      `CREATE INDEX IF NOT EXISTS idx_events_user_msg_dedup
+       ON events (session_id, type, json_extract(payload, '$.messageId'))`,
+    );
   }
 
   close(): void {
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -95,6 +95,7 @@ export class EventStore {
   private log: EventStoreLogger;
   private stmts: {
     append: Database.Statement;
+    hasUserMessage: Database.Statement;
     eventsAfter: Database.Statement;
     eventsAfterLimited: Database.Statement;
     sessionEvents: Database.Statement;
@@ -130,6 +131,12 @@ export class EventStore {
 
     this.stmts = {
       append: db.prepare('INSERT INTO events (session_id, type, payload) VALUES (?, ?, ?)'),
+      hasUserMessage: db.prepare(
+        `SELECT 1 FROM events
+         WHERE session_id = ? AND type = 'user_message'
+           AND json_extract(payload, '$.messageId') = ?
+         LIMIT 1`,
+      ),
       eventsAfter: db.prepare(
         'SELECT seq, session_id, type, payload, created_at FROM events WHERE session_id = ? AND seq > ? ORDER BY seq',
       ),
@@ -307,6 +314,11 @@ export class EventStore {
   append(sessionId: string, type: string, payload: Record<string, unknown>): number {
     const result = this.stmts.append.run(sessionId, type, JSON.stringify(payload));
     return Number(result.lastInsertRowid);
+  }
+
+  /** Check if a user_message with the given messageId already exists for this session. */
+  hasUserMessage(sessionId: string, messageId: string): boolean {
+    return this.stmts.hasUserMessage.get(sessionId, messageId) != null;
   }
 
   getEventsAfter(sessionId: string, afterSeq: number, limit?: number): StoredEvent[] {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1007,24 +1007,27 @@ async function _startChatInner(
     // Store and echo it here so the frontend can replay it.
     if (options.resume) {
       const messageId = options.clientMsgId || `umsg-${Date.now()}-resume`;
-      eventStore.append(options.resume, 'user_message', {
-        v: 2,
-        type: 'user_message',
-        ts: Date.now(),
-        messageId,
-        text: fullPrompt,
-      });
-      eventStore.updateLastSpeaker(options.resume, 'user');
-      _onSessionChange?.(clientId, 'user_message');
-      const echo = {
-        type: 'user_message',
-        v: 2,
-        messageId,
-        text: fullPrompt,
-        sessionId: options.resume,
-      };
-      send(transport, echo);
-      broadcastToObservers(session.observers, echo);
+      // Idempotency guard: skip if this user_message was already stored (e.g. retried POST)
+      if (!eventStore.hasUserMessage(options.resume, messageId)) {
+        eventStore.append(options.resume, 'user_message', {
+          v: 2,
+          type: 'user_message',
+          ts: Date.now(),
+          messageId,
+          text: fullPrompt,
+        });
+        eventStore.updateLastSpeaker(options.resume, 'user');
+        _onSessionChange?.(clientId, 'user_message');
+        const echo = {
+          type: 'user_message',
+          v: 2,
+          messageId,
+          text: fullPrompt,
+          sessionId: options.resume,
+        };
+        send(transport, echo);
+        broadcastToObservers(session.observers, echo);
+      }
     }
 
     await runQueryLoop(
@@ -1136,6 +1139,10 @@ export function sendToChat(
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
     const messageId = clientMsgId || `umsg-${Date.now()}-send`;
     if (session.sessionId) {
+      // Idempotency guard: skip if this user_message was already stored (e.g. retried POST)
+      if (eventStore.hasUserMessage(session.sessionId, messageId)) {
+        return true;
+      }
       eventStore.append(session.sessionId, 'user_message', {
         v: 2,
         type: 'user_message',
@@ -1179,6 +1186,10 @@ export async function interruptChat(
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
     const messageId = clientMsgId || `umsg-${Date.now()}-interrupt`;
     if (session.sessionId) {
+      // Idempotency guard: skip if this user_message was already stored (e.g. retried POST)
+      if (eventStore.hasUserMessage(session.sessionId, messageId)) {
+        return true;
+      }
       eventStore.append(session.sessionId, 'user_message', {
         v: 2,
         type: 'user_message',
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1123,7 +1123,7 @@ function storeAndEchoIfNew(
   text: string,
   clientId: string,
   transport: SessionTransport,
-  observers?: Set<SessionTransport>,
+  observers: Set<SessionTransport>,
 ): boolean {
   if (eventStore.hasUserMessage(sessionId, messageId)) {
     return true;
@@ -1139,7 +1139,7 @@ function storeAndEchoIfNew(
   _onSessionChange?.(clientId, 'user_message');
   const echo = { type: 'user_message', v: 2, messageId, text, sessionId };
   send(transport, echo);
-  if (observers) broadcastToObservers(observers, echo);
+  broadcastToObservers(observers, echo);
   return false;
 }
 
@@ -1170,6 +1170,9 @@ export function sendToChat(
         /* errors logged internally */
       });
     } else {
+      // Pre-session-resolve: no eventStore to dedup against.
+      // The frontend deduplicates echoes by messageId, and server-generated
+      // fallback IDs include randomUUID, so duplicates are not possible in practice.
       const echo = { type: 'user_message', v: 2, messageId, text: fullPrompt };
       send(session.transport, echo);
       broadcastToObservers(session.observers, echo);
@@ -1194,11 +1197,11 @@ export async function interruptChat(
     if (model) session.model = model;
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
     const messageId = clientMsgId || `umsg-${Date.now()}-${randomUUID().slice(0, 8)}-interrupt`;
-    // Store and echo the user message, but always proceed with the interrupt
-    // even if the message is a duplicate — a retried interrupt must still stop
-    // the agent, just without re-storing the echo.
+    // Store and echo the user message. A retried interrupt must still stop
+    // the agent — only the echo/store is skipped on duplicate.
+    let isDup = false;
     if (session.sessionId) {
-      storeAndEchoIfNew(
+      isDup = storeAndEchoIfNew(
         session.sessionId,
         messageId,
         fullPrompt,
@@ -1207,6 +1210,7 @@ export async function interruptChat(
         session.observers,
       );
     } else {
+      // Pre-session-resolve: no eventStore to dedup against (see sendToChat).
       const echo = { type: 'user_message', v: 2, messageId, text: fullPrompt };
       send(session.transport, echo);
       broadcastToObservers(session.observers, echo);
@@ -1223,7 +1227,11 @@ export async function interruptChat(
       await Promise.allSettled(stops);
     }
     await session.queryInstance.interrupt();
-    session.inputQueue.push(makeUserMessage(fullPrompt, 'now'));
+    // Only push to inputQueue on first delivery — a retried interrupt should
+    // still call interrupt() (to halt the agent) but not double-queue the prompt.
+    if (!isDup) {
+      session.inputQueue.push(makeUserMessage(fullPrompt, 'now'));
+    }
     return true;
   });
 }
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1006,8 +1006,9 @@ async function _startChatInner(
     // in the event store — making user messages invisible after WS reconnect.
     // Store and echo it here so the frontend can replay it.
     if (options.resume) {
-      const messageId = options.clientMsgId || `umsg-${Date.now()}-resume`;
-      storeAndEchoUserMessage(
+      const messageId =
+        options.clientMsgId || `umsg-${Date.now()}-${randomUUID().slice(0, 8)}-resume`;
+      storeAndEchoIfNew(
         options.resume,
         messageId,
         fullPrompt,
@@ -1113,10 +1114,10 @@ async function tryAutoRename(sessionId: string, clientId: string): Promise<void>
 }
 
 /**
- * Deduplicated store-and-echo for user messages.
+ * Store and echo a user message only if it hasn't been stored before.
  * Returns true if the message was a duplicate (already stored), false if newly stored.
  */
-function storeAndEchoUserMessage(
+function storeAndEchoIfNew(
   sessionId: string,
   messageId: string,
   text: string,
@@ -1154,9 +1155,9 @@ export function sendToChat(
     const session = registry.get(clientId);
     if (!session?.inputQueue) return false;
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
-    const messageId = clientMsgId || `umsg-${Date.now()}-send`;
+    const messageId = clientMsgId || `umsg-${Date.now()}-${randomUUID().slice(0, 8)}-send`;
     if (session.sessionId) {
-      const isDup = storeAndEchoUserMessage(
+      const isDup = storeAndEchoIfNew(
         session.sessionId,
         messageId,
         fullPrompt,
@@ -1192,9 +1193,12 @@ export async function interruptChat(
     if (!session?.queryInstance || !session?.inputQueue) return false;
     if (model) session.model = model;
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
-    const messageId = clientMsgId || `umsg-${Date.now()}-interrupt`;
+    const messageId = clientMsgId || `umsg-${Date.now()}-${randomUUID().slice(0, 8)}-interrupt`;
+    // Store and echo the user message, but always proceed with the interrupt
+    // even if the message is a duplicate — a retried interrupt must still stop
+    // the agent, just without re-storing the echo.
     if (session.sessionId) {
-      const isDup = storeAndEchoUserMessage(
+      storeAndEchoIfNew(
         session.sessionId,
         messageId,
         fullPrompt,
@@ -1202,7 +1206,6 @@ export async function interruptChat(
         session.transport,
         session.observers,
       );
-      if (isDup) return true;
     } else {
       const echo = { type: 'user_message', v: 2, messageId, text: fullPrompt };
       send(session.transport, echo);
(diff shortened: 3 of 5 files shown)


frontend/src/components/ChatInput.tsx
```diff
@@ -129,8 +130,10 @@ export function ChatInput({
   }
 
   function handleSend() {
+    if (sendGuard.current) return;
```
dimakis, line 133: 🟡 **bugs**: `sendGuard` protects `handleSend` but `handleInterrupt` (line 272) has no equivalent guard. The same double-tap/double-Enter scenario that motivated this fix can fire `handleInterrupt` twice when `running && onInterrupt` is true, since `handleInterrupt` is reachable from both `handleKeyDown` (Enter) and the interrupt button's `onClick`. `[fixable]`

frontend/src/components/ChatInput.tsx
```diff
@@ -141,11 +144,12 @@ export function ChatInput({
       clearDraft();
       setImages([]);
       if (!useExternal) setContextBlocks([]);
-      requestAnimationFrame(() => {
-        autoResize();
-        textareaRef.current?.focus();
-      });
     }
+    requestAnimationFrame(() => {
```
dimakis, line 150: 🔵 **regressions**: Moving `autoResize()` and `focus()` out of the `if (sent)` block means they now run even when the send fails. Previously, a failed send left the textarea state unchanged (no resize, no re-focus). The new behavior is arguably better (guard resets correctly), but it's a subtle behavioral change — the textarea will re-focus after a failed send, which could be confusing if the user navigated away from the input.

packages/protocol/src/event-store.ts
```diff
@@ -130,6 +131,12 @@ export class EventStore {
 
     this.stmts = {
       append: db.prepare('INSERT INTO events (session_id, type, payload) VALUES (?, ?, ?)'),
+      hasUserMessage: db.prepare(
```
dimakis, line 135: 🔵 **unsafe_assumptions**: The `hasUserMessage` query uses `json_extract(payload, '$.messageId')` without a covering index. The existing index is `(session_id, seq)`. For typical session sizes (hundreds of events) this is fine, but if session event counts grow large, this will become a per-send bottleneck. Consider adding a computed index: `CREATE INDEX idx_events_user_msg ON events (session_id, type, json_extract(payload, '$.messageId')) WHERE type = 'user_message'`. `[fixable]`

server/chat.ts
```diff
@@ -1136,6 +1139,10 @@ export function sendToChat(
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
     const messageId = clientMsgId || `umsg-${Date.now()}-send`;
     if (session.sessionId) {
+      // Idempotency guard: skip if this user_message was already stored (e.g. retried POST)
```
dimakis: 🔵 **style**: The idempotency comment is duplicated verbatim in three places (lines 1010, 1142, 1189). Consider extracting the dedup-check-and-store logic into a small helper (e.g. `storeUserMessageIfNew(sessionId, messageId, text)`) to reduce repetition and ensure the three call sites stay in sync. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid idempotency fix with good test coverage for the core path. Main gaps: no dedup test for `interruptChat`, no `sendGuard` on `handleInterrupt`, and some opportunity to DRY up the three identical guard sites._

- 🟡 **missing_tests**: No deduplication test for `interruptChat`. `sendToChat` has `'skips duplicate when same clientMsgId is sent twice'` but there is no equivalent test for `interruptChat`, which has the same idempotency guard. Add a test that calls `interruptChat` twice with the same `clientMsgId` and asserts the interrupt/push/echo only happen once. `[fixable]`
- 🟡 **missing_tests**: No test for the `_startChatInner` resume-path idempotency guard. The resume path in `_startChatInner` now wraps the echo/append in `if (!eventStore.hasUserMessage(...))`, but there is no test that calls `startChat` twice with the same `clientMsgId` on a resume and verifies the user_message is only stored/echoed once. `[fixable]`


### `frontend/src/components/ChatInput.tsx`
_Solid idempotency fix with good test coverage for the core path. Main gaps: no dedup test for `interruptChat`, no `sendGuard` on `handleInterrupt`, and some opportunity to DRY up the three identical guard sites._

- 🟡 **bugs** (L133): `sendGuard` protects `handleSend` but `handleInterrupt` (line 272) has no equivalent guard. The same double-tap/double-Enter scenario that motivated this fix can fire `handleInterrupt` twice when `running && onInterrupt` is true, since `handleInterrupt` is reachable from both `handleKeyDown` (Enter) and the interrupt button's `onClick`. `[fixable]`
- 🔵 **regressions** (L148): Moving `autoResize()` and `focus()` out of the `if (sent)` block means they now run even when the send fails. Previously, a failed send left the textarea state unchanged (no resize, no re-focus). The new behavior is arguably better (guard resets correctly), but it's a subtle behavioral change — the textarea will re-focus after a failed send, which could be confusing if the user navigated away from the input.


_Solid idempotency fix with good test coverage for the core path. Main gaps: no dedup test for `interruptChat`, no `sendGuard` on `handleInterrupt`, and some opportunity to DRY up the three identical guard sites._

- 🔵 **unsafe_assumptions** (L134): The `hasUserMessage` query uses `json_extract(payload, '$.messageId')` without a covering index. The existing index is `(session_id, seq)`. For typical session sizes (hundreds of events) this is fine, but if session event counts grow large, this will become a per-send bottleneck. Consider adding a computed index: `CREATE INDEX idx_events_user_msg ON events (session_id, type, json_extract(payload, '$.messageId')) WHERE type = 'user_message'`. `[fixable]`


### `server/chat.ts`
_Solid idempotency fix with good test coverage for the core path. Main gaps: no dedup test for `interruptChat`, no `sendGuard` on `handleInterrupt`, and some opportunity to DRY up the three identical guard sites._

- 🔵 **style** (L1142): The idempotency comment is duplicated verbatim in three places (lines 1010, 1142, 1189). Consider extracting the dedup-check-and-store logic into a small helper (e.g. `storeUserMessageIfNew(sessionId, messageId, text)`) to reduce repetition and ensure the three call sites stay in sync. `[fixable]`

server/chat.ts
```diff
@@ -1179,25 +1186,16 @@ export async function interruptChat(
...
-        messageId,
-        text: fullPrompt,
-      });
-      eventStore.updateLastSpeaker(session.sessionId, 'user');
-      _onSessionChange?.(clientId, 'user_message');
+      const isDup = storeAndEchoUserMessage(
+        session.sessionId, messageId, fullPrompt, clientId,
+        session.transport, session.observers,
+      );
+      if (isDup) return true;
```
dimakis: 🔴 **bugs**: When `isDup` is true in `interruptChat`, the function returns early — skipping `session.queryInstance.interrupt()`, subagent `stopTask()` calls, and `inputQueue.push()`. A duplicate interrupt request (same `clientMsgId` retried due to network flake) will silently do nothing while the agent keeps running. The dedup should only suppress the echo/store, not the interrupt side-effect. Consider: if isDup, skip store+echo but still call interrupt+push. `[fixable]`

server/chat.ts
```diff
@@ -1136,28 +1152,19 @@ export function sendToChat(
...
       tryAutoRename(session.sessionId, clientId).catch(() => {
         /* errors logged internally */
       });
+    } else {
+      const echo = { type: 'user_message', v: 2, messageId, text: fullPrompt };
```
dimakis, line 1176: 🟡 **regressions**: The refactored `sendToChat` no longer includes `sessionId` in the echo when `session.sessionId` is falsy (the else branch on line 1164 omits it). The old code used `...(session.sessionId ? { sessionId: session.sessionId } : {})`, which also omitted it, so this is actually fine — but the `interruptChat` else branch (line 1195) similarly omits `sessionId`, matching the old behavior. No regression here on closer inspection.

server/chat.ts
```diff
@@ -1136,28 +1152,19 @@ export function sendToChat(
     const fullPrompt = assemblePrompt(prompt, session.cwd ?? '.', images, contextBlocks);
     const messageId = clientMsgId || `umsg-${Date.now()}-send`;
```
dimakis: 🟡 **unsafe_assumptions**: When `clientMsgId` is undefined, the fallback `umsg-${Date.now()}-send` uses millisecond timestamps. Two calls in the same millisecond (e.g., rapid retries from a load balancer or batched requests) would produce the same messageId, making the dedup check a false positive that silently drops the second message. Consider appending a random suffix or monotonic counter. `[fixable]`

server/chat.ts
```diff
@@ -1122,6 +1108,36 @@ async function tryAutoRename(sessionId: string, clientId: string): Promise<void>
   }
 }
 
+/**
+ * Deduplicated store-and-echo for user messages.
+ * Returns true if the message was a duplicate (already stored), false if newly stored.
+ */
+function storeAndEchoUserMessage(
```
dimakis: 🔵 **style**: The JSDoc says "Returns true if the message was a duplicate" but the return semantics are inverted from what callers might expect (`true` = duplicate = skip). Consider naming it `isDuplicate` at call sites (already done) and/or renaming the function to `storeAndEchoIfNew` for clarity, since "storeAndEcho" implies it always stores. `[fixable]`

frontend/src/components/ChatInput.tsx
```diff
@@ -141,11 +144,12 @@ export function ChatInput({
...
+      sendGuard.current = false;
+    });
+    autoResize();
```
dimakis: 🟡 **regressions**: `autoResize()` and `textareaRef.current?.focus()` were moved outside the `if (sent)` block and are now called unconditionally — even when `onSend` returns false (message rejected). Previously they only ran on success. This means a failed send still clears the textarea height and re-focuses, which is a minor UX regression. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_The core idempotency mechanism (EventStore + index + frontend guard) is solid, but `interruptChat` has a critical bug where duplicate detection suppresses the actual interrupt side-effect, not just the echo — a retried interrupt will silently fail to stop the agent._

- 🔴 **bugs** (L1193): When `isDup` is true in `interruptChat`, the function returns early — skipping `session.queryInstance.interrupt()`, subagent `stopTask()` calls, and `inputQueue.push()`. A duplicate interrupt request (same `clientMsgId` retried due to network flake) will silently do nothing while the agent keeps running. The dedup should only suppress the echo/store, not the interrupt side-effect. Consider: if isDup, skip store+echo but still call interrupt+push. `[fixable]`
- 🟡 **regressions** (L1164): The refactored `sendToChat` no longer includes `sessionId` in the echo when `session.sessionId` is falsy (the else branch on line 1164 omits it). The old code used `...(session.sessionId ? { sessionId: session.sessionId } : {})`, which also omitted it, so this is actually fine — but the `interruptChat` else branch (line 1195) similarly omits `sessionId`, matching the old behavior. No regression here on closer inspection.
- 🟡 **unsafe_assumptions** (L1153): When `clientMsgId` is undefined, the fallback `umsg-${Date.now()}-send` uses millisecond timestamps. Two calls in the same millisecond (e.g., rapid retries from a load balancer or batched requests) would produce the same messageId, making the dedup check a false positive that silently drops the second message. Consider appending a random suffix or monotonic counter. `[fixable]`
- 🔵 **style** (L1115): The JSDoc says "Returns true if the message was a duplicate" but the return semantics are inverted from what callers might expect (`true` = duplicate = skip). Consider naming it `isDuplicate` at call sites (already done) and/or renaming the function to `storeAndEchoIfNew` for clarity, since "storeAndEcho" implies it always stores. `[fixable]`


### `frontend/src/components/ChatInput.tsx`
_The core idempotency mechanism (EventStore + index + frontend guard) is solid, but `interruptChat` has a critical bug where duplicate detection suppresses the actual interrupt side-effect, not just the echo — a retried interrupt will silently fail to stop the agent._

- 🟡 **regressions** (L151): `autoResize()` and `textareaRef.current?.focus()` were moved outside the `if (sent)` block and are now called unconditionally — even when `onSend` returns false (message rejected). Previously they only ran on success. This means a failed send still clears the textarea height and re-focuses, which is a minor UX regression. `[fixable]`


_The core idempotency mechanism (EventStore + index + frontend guard) is solid, but `interruptChat` has a critical bug where duplicate detection suppresses the actual interrupt side-effect, not just the echo — a retried interrupt will silently fail to stop the agent._

- 🔵 **missing_tests**: No test covers the `sendToChat`/`interruptChat` fallback path when `session.sessionId` is falsy (the else branch). The refactor changed this branch's echo shape (no conditional `sessionId` spread), so a regression test for the no-sessionId case would be valuable. `[fixable]`

dimakis:
## Centaur Review

Found **4** issue(s).


### `frontend/src/components/ChatInput.tsx`
_Well-structured idempotency guard with correct server-side dedup via EventStore and client-side sendGuard. The core approach is sound. Minor concerns: interruptChat pushes to inputQueue even on duplicate (agent processes the prompt twice), autoResize moved out of rAF is cosmetically redundant but functionally covered by a useEffect, and the interrupt dedup test should assert pushSpy call count to document the intended behavior._

- 🔵 **regressions** (L147): autoResize() and focus() were moved out of requestAnimationFrame. The synchronous autoResize() measures el.scrollHeight before React has flushed the cleared text to the DOM, so it reads the old height. The existing useEffect on `text` (line 95-97: `useEffect(() => { autoResize(); }, [text, autoResize])`) corrects this on the next render, making the synchronous call redundant but harmless — worst case is a single frame at the old height before the effect fixes it. Same pattern applies to handleInterrupt at line 289. `[fixable]`


_Well-structured idempotency guard with correct server-side dedup via EventStore and client-side sendGuard. The core approach is sound. Minor concerns: interruptChat pushes to inputQueue even on duplicate (agent processes the prompt twice), autoResize moved out of rAF is cosmetically redundant but functionally covered by a useEffect, and the interrupt dedup test should assert pushSpy call count to document the intended behavior._

- 🔵 **missing_tests** (L276): The interrupt dedup test verifies interruptSpy is called twice and only one echo is sent, but doesn't assert pushSpy call count. Since interruptChat intentionally pushes to inputQueue even on duplicates (the agent gets the prompt twice), this behavior should be documented by asserting `expect(pushSpy).toHaveBeenCalledTimes(2)` — otherwise it's unclear whether the double-push is intended or an oversight. `[fixable]`
- 🔵 **style** (L33): Three older tests still use static sessionIds ('sess-123', 'sess-456', 'sess-789') while the new dedup tests use Date.now()-based IDs. Since the module-level EventStore persists across tests, static sessionIds risk cross-test pollution if a future test reuses the same ID with a matching clientMsgId. Consider making these dynamic for consistency with the new tests. `[fixable]`


### `server/chat.ts`
_Well-structured idempotency guard with correct server-side dedup via EventStore and client-side sendGuard. The core approach is sound. Minor concerns: interruptChat pushes to inputQueue even on duplicate (agent processes the prompt twice), autoResize moved out of rAF is cosmetically redundant but functionally covered by a useEffect, and the interrupt dedup test should assert pushSpy call count to document the intended behavior._

- 🔵 **unsafe_assumptions** (L1226): interruptChat always calls inputQueue.push(makeUserMessage(fullPrompt, 'now')) even when storeAndEchoIfNew detects a duplicate. This means a retried interrupt pushes the same prompt to the agent's input queue twice, causing the agent to process the same user message twice. The echo is correctly deduplicated, but the agent-side duplicate is not. Consider guarding the push behind `!isDup` (like sendToChat does) and only calling interrupt() unconditionally — that way a retry still stops the agent without double-queuing the message. `[fixable]`

dimakis:
## Centaur Review

### `server/chat.ts`
_Solid idempotency guard implementation with good test coverage. Main concern: interruptChat always pushes to inputQueue on duplicate, which could cause the agent to process the same message twice on retried interrupts — consider gating the push behind the dedup check as sendToChat does._

- 🟡 **bugs** (L1226): interruptChat always pushes to inputQueue even on duplicate clientMsgId. The echo is correctly deduplicated, but the agent will receive the same user message twice in its input queue on a retried interrupt. sendToChat correctly skips the push on dedup (line 1168), but interruptChat does not. If two interrupt calls arrive before the agent consumes the first push, the agent processes the message twice. `[fixable]`
- 🔵 **style** (L1120): storeAndEchoIfNew returns true for duplicate and false for new — inverted from typical boolean semantics (e.g., 'wasStored' returning true on success). The caller uses `if (isDup)` which reads clearly, but the function's own return value is counterintuitive. Consider renaming to `isDuplicateUserMessage` or inverting the return. `[fixable]`


_Solid idempotency guard implementation with good test coverage. Main concern: interruptChat always pushes to inputQueue on duplicate, which could cause the agent to process the same message twice on retried interrupts — consider gating the push behind the dedup check as sendToChat does._

- 🟡 **missing_tests** (L269): The interrupt dedup test verifies interruptSpy call count and echo dedup, but does not assert on pushSpy call count. This leaves the double-push to inputQueue on retried interrupts unspecified — whether it's intentional behavior or a bug is ambiguous without a test pinning it down. `[fixable]`


### `frontend/src/components/ChatInput.tsx`
_Solid idempotency guard implementation with good test coverage. Main concern: interruptChat always pushes to inputQueue on duplicate, which could cause the agent to process the same message twice on retried interrupts — consider gating the push behind the dedup check as sendToChat does._

- 🔵 **regressions** (L147): autoResize() and focus() were moved out of requestAnimationFrame in handleSend. autoResize reads scrollHeight from the DOM, but React hasn't flushed the setText('') update yet, so it briefly computes height from stale content. The useEffect on line 97 (text dependency) self-corrects on the next render, so this is cosmetic — but the original rAF placement was more correct. Same issue in handleInterrupt (line 289). `[fixable]`
- 🔵 **unsafe_assumptions** (L73): sendGuard only protects handleSend and handleInterrupt. The auto-send-from-queue path (line ~105, useEffect that calls onSend when running transitions false) and fireQueuedAsInterrupt (line ~261) bypass it. These paths now rely solely on the server-side dedup, which is fine, but the asymmetry could be confusing to future maintainers.

dimakis:
## Centaur Review

_Solid idempotency guard implementation — the server-side dedup via EventStore is well-designed with proper index migration, and the frontend sendGuard prevents UI double-fires. The interrupt dedup test should assert on inputQueue push behavior, and two frontend code paths (fireQueuedAsInterrupt, auto-send effect) don't use the new sendGuard pattern._

- 🟡 **missing_tests** (L266): The 'deduplicates echo but still interrupts on retried clientMsgId' test verifies that interrupt() fires twice and only one echo is sent, but does not assert on pushSpy. Since interruptChat always calls inputQueue.push even on duplicate (line 1226 of chat.ts), the test should explicitly assert whether pushSpy was called once or twice to document the intended behavior. `[fixable]`
- 🔵 **missing_tests** (L99): Test isolation is fragile: send-to-chat tests share a module-level EventStore (real SQLite file) with no per-test reset, relying on `Date.now()` in session IDs for uniqueness. Compare with event-store.test.ts which properly uses ':memory:' databases per test. If two test runs overlap or Date.now() collides, dedup assertions could produce false positives. `[fixable]`


### `frontend/src/components/ChatInput.tsx`
_Solid idempotency guard implementation — the server-side dedup via EventStore is well-designed with proper index migration, and the frontend sendGuard prevents UI double-fires. The interrupt dedup test should assert on inputQueue push behavior, and two frontend code paths (fireQueuedAsInterrupt, auto-send effect) don't use the new sendGuard pattern._

- 🔵 **unsafe_assumptions** (L261): fireQueuedAsInterrupt() calls onInterrupt without the sendGuard check, inconsistent with handleInterrupt() which does check it. While removeQueued() provides a natural guard against the exact same item being sent twice, rapid clicks on different queued items could overlap. Consider adding a sendGuard check for consistency with the pattern introduced in this PR. `[fixable]`
- 🔵 **unsafe_assumptions** (L105): The auto-send useEffect (line 101-113) calls onSend without the sendGuard check. The dequeue() call provides a natural guard since it removes the item, but this is inconsistent with handleSend which now uses sendGuard. If a rapid running→not-running→running→not-running state transition occurred, the effect could fire twice before the first send completes. `[fixable]`

server/chat.ts
```diff
@@ -1134,30 +1155,25 @@ export function sendToChat(
...
+        session.observers,
+      );
+      if (isDup) return true;
       tryAutoRename(session.sessionId, clientId).catch(() => {
         /* errors logged internally */
       });
+    } else {
+      const echo = { type: 'user_message', v: 2, messageId, text: fullPrompt };
+      send(session.transport, echo);
+      broadcastToObservers(session.observers, echo);
```
dimakis, line 1178: 🟡 **bugs**: In the `else` branch of `sendToChat` (no `sessionId`), `broadcastToObservers(session.observers, echo)` is called unconditionally — but there is no dedup protection here because `storeAndEchoIfNew` is only called when `session.sessionId` is truthy. If two identical sends arrive before the session ID is assigned, both will echo to observers. This is a narrow window (pre-session-resolve), but it's inconsistent with the dedup intent. The same pattern exists in `interruptChat` at line 1212. `[fixable]`

server/chat.ts
```diff
@@ -1122,6 +1113,36 @@ async function tryAutoRename(sessionId: string, clientId: string): Promise<void>
...
+ * Store and echo a user message only if it hasn't been stored before.
+ * Returns true if the message was a duplicate (already stored), false if newly stored.
+ */
+function storeAndEchoIfNew(
+  sessionId: string,
+  messageId: string,
+  text: string,
+  clientId: string,
+  transport: SessionTransport,
+  observers?: Set<SessionTransport>,
```
dimakis: 🔵 **style**: `storeAndEchoIfNew` declares `observers` as optional (`observers?: Set<SessionTransport>`) even though every caller always passes `session.observers`, which is always initialized as `new Set()` and is never undefined. Making the parameter required would remove the `if (observers)` guard and better express the actual contract. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/chat.ts`
_Solid two-layer dedup design (frontend sendGuard + backend messageId idempotency). Three minor items: pre-sessionId sends bypass dedup, interruptChat silently double-pushes to inputQueue on retry without test coverage of that behavior, and the `observers` parameter is unnecessarily optional._

- 🟡 **bugs** (L1175): In the `else` branch of `sendToChat` (no `sessionId`), `broadcastToObservers(session.observers, echo)` is called unconditionally — but there is no dedup protection here because `storeAndEchoIfNew` is only called when `session.sessionId` is truthy. If two identical sends arrive before the session ID is assigned, both will echo to observers. This is a narrow window (pre-session-resolve), but it's inconsistent with the dedup intent. The same pattern exists in `interruptChat` at line 1212. `[fixable]`
- 🔵 **style** (L1126): `storeAndEchoIfNew` declares `observers` as optional (`observers?: Set<SessionTransport>`) even though every caller always passes `session.observers`, which is always initialized as `new Set()` and is never undefined. Making the parameter required would remove the `if (observers)` guard and better express the actual contract. `[fixable]`


_Solid two-layer dedup design (frontend sendGuard + backend messageId idempotency). Three minor items: pre-sessionId sends bypass dedup, interruptChat silently double-pushes to inputQueue on retry without test coverage of that behavior, and the `observers` parameter is unnecessarily optional._

- 🔵 **missing_tests**: The `interruptChat` dedup test ('deduplicates echo but still interrupts on retried clientMsgId') verifies `interruptSpy` is called twice and echo count is 1, but does not assert `pushSpy` count. The implementation pushes to `inputQueue` twice (the agent sees the same prompt twice). If that's intentional, a comment or explicit assertion would clarify; if not, it's a subtle bug where a retried interrupt double-queues the message. `[fixable]`

dimakis:
## Centaur Review

### `server/chat.ts`
_Well-structured dual-layer dedup (frontend sendGuard + server-side EventStore idempotency). The core logic is correct and well-tested. Main concern is the unchecked return value of `storeAndEchoIfNew` in the resume path — verify that duplicate `_startChatInner` calls are externally prevented._

- 🟡 **regressions** (L1168): When `sendToChat` detects a duplicate (`isDup` is true), it returns `true` without pushing to `inputQueue` and without calling `tryAutoRename`. Returning `true` tells the caller the send succeeded, which is correct. However, unlike `interruptChat` (which explicitly documents why it still fires `interrupt()` on dedup), `sendToChat` silently drops the message on dedup with no logging. Consider adding a debug log so operators can trace dedup events — silent drops are hard to diagnose in production. `[fixable]`
- 🟡 **bugs** (L1010): In the resume path of `_startChatInner`, the return value of `storeAndEchoIfNew()` is ignored. If the user's resume request is a duplicate (same `clientMsgId`), the echo is correctly suppressed, but `runQueryLoop` still proceeds — starting a second query loop for the same session. This is safe only if external callers already prevent duplicate `_startChatInner` invocations for the same session. Verify the caller (`startChat`) guards against this; otherwise a retried resume could spawn a second concurrent query loop.
- 🔵 **unsafe_assumptions** (L1158): The fallback `messageId` for `sendToChat` uses `Date.now()` plus 8 chars of `randomUUID()`. While practically unique, `Date.now()` has millisecond granularity. Two calls within the same millisecond could theoretically generate IDs that differ only in the 8-char UUID suffix — which is fine for uniqueness but worth noting. The previous code used only `Date.now()`, so this is strictly an improvement.


### `frontend/src/components/ChatInput.tsx`
_Well-structured dual-layer dedup (frontend sendGuard + server-side EventStore idempotency). The core logic is correct and well-tested. Main concern is the unchecked return value of `storeAndEchoIfNew` in the resume path — verify that duplicate `_startChatInner` calls are externally prevented._

- 🔵 **style** (L147): The synchronous `autoResize()` call in `handleSend` (and `handleInterrupt` at line 289) runs before React re-renders the cleared text. It's effectively a no-op because the `useEffect` at line 96-98 triggers `autoResize()` after the text state update anyway. Consider removing the synchronous calls to reduce confusion — the useEffect already handles it correctly. `[fixable]`


_Well-structured dual-layer dedup (frontend sendGuard + server-side EventStore idempotency). The core logic is correct and well-tested. Main concern is the unchecked return value of `storeAndEchoIfNew` in the resume path — verify that duplicate `_startChatInner` calls are externally prevented._

- 🔵 **missing_tests**: The dedup tests cover `sendToChat` and `interruptChat`, but there's no test for the resume path dedup in `_startChatInner`. The resume path now calls `storeAndEchoIfNew` (line 1011) but its duplicate-handling behavior (ignoring the return value) is untested. The event-store test at 'prevents duplicate resume-path user messages' tests the store layer but not the chat.ts integration. `[fixable]`
- 🔵 **style** (L33): The first test ('sends a user_message event after persisting to event store') uses a hardcoded session ID `'sess-123'` while the new tests use `Date.now()` for uniqueness. Since all tests share a module-level singleton EventStore, this is safe today (server-generated fallback IDs include randomUUID), but inconsistent. Consider making it `sess-${Date.now()}` for consistency with the rest of the file. `[fixable]`

frontend/src/components/ChatInput.tsx
```diff
@@ -129,8 +130,10 @@ export function ChatInput({
   }
 
   function handleSend() {
+    if (sendGuard.current) return;
     const trimmed = text.trim();
     if (!trimmed && images.length === 0) return;
+    sendGuard.current = true;
```
dimakis, line 136: 🔵 **bugs**: sendGuard is set to `true` before `onSend()`/`onInterrupt()` but the `requestAnimationFrame` reset at line 150 is not reachable if the callback throws. A thrown exception would permanently lock the input until the component remounts. Wrapping the body in try/finally (with the rAF in `finally`) would be more robust, though in practice React error boundaries will recover. `[fixable]`

frontend/src/components/ChatInput.tsx
```diff
@@ -278,9 +284,10 @@ export function ChatInput({
     setImages([]);
     if (!useExternal) setContextBlocks([]);
     requestAnimationFrame(() => {
-      autoResize();
-      textareaRef.current?.focus();
+      sendGuard.current = false;
     });
+    autoResize();
```
dimakis, line 289: 🔵 **style**: In `handleInterrupt`, `autoResize()` and `focus()` were moved from inside `requestAnimationFrame` to synchronous execution. This is safe because `useEffect` at line 96 re-runs `autoResize` whenever `text` changes, so the effect self-corrects after the re-render. However, it's an incidental behavioral change — worth noting that it was intentional (keeping rAF solely for the guard reset).

dimakis (commented):
## Centaur Review

Found **3** issue(s).


### `frontend/src/components/ChatInput.tsx`
_Solid idempotency guard with server-side dedup via EventStore, well-tested dedup semantics for both send and interrupt paths, and correct interrupt-retry behavior (still halts the agent but skips double-queuing). Only minor robustness suggestions — no bugs or regressions._

- 🔵 **bugs** (L136): sendGuard is set to `true` before `onSend()`/`onInterrupt()` but the `requestAnimationFrame` reset at line 150 is not reachable if the callback throws. A thrown exception would permanently lock the input until the component remounts. Wrapping the body in try/finally (with the rAF in `finally`) would be more robust, though in practice React error boundaries will recover. `[fixable]`
- 🔵 **style** (L289): In `handleInterrupt`, `autoResize()` and `focus()` were moved from inside `requestAnimationFrame` to synchronous execution. This is safe because `useEffect` at line 96 re-runs `autoResize` whenever `text` changes, so the effect self-corrects after the re-render. However, it's an incidental behavioral change — worth noting that it was intentional (keeping rAF solely for the guard reset).


_Solid idempotency guard with server-side dedup via EventStore, well-tested dedup semantics for both send and interrupt paths, and correct interrupt-retry behavior (still halts the agent but skips double-queuing). Only minor robustness suggestions — no bugs or regressions._

- 🔵 **missing_tests**: No test covers the resume-path deduplication in `_startChatInner` (line 1011 of chat.ts). The resume path calls `storeAndEchoIfNew` and intentionally ignores the return value (a retried resume should still run the full query loop). The underlying `hasUserMessage` is well-tested in event-store.test.ts, so this is low-risk, but an integration test showing that a retried resume doesn't double-echo would close the loop. `[fixable]`

PR #387, opened 2026-06-15
feat(harness): multi-model reasoning engine — deliberation + fusion

dimakis:
## Summary

- **Provider adapter layer** — thin adapters for Claude (Anthropic Vertex) and Gemini (Google Vertex REST) with unified `ModelProvider` interface, per-model cost tracking, and budget enforcement
- **Deliberation orchestrator** — sequential adversarial protocol (propose → challenge → respond → converge) with mind-change tracking, ported from the mgmt Python PoC
- **Fusion orchestrator** — parallel fan-out to N models → structured judge analysis (consensus/contradictions/unique insights/blind spots) → synthesis, inspired by OpenRouter Fusion API benchmarks showing ~75% of gain comes from the synthesis framework itself
- **`/deliberate` and `/fuse` native commands** — async execution with WS event streaming, formatted markdown results
- **Frontend type + parser** — `reasoning_event` WS message type registered for future live streaming UI

### Architecture

Two complementary reasoning modes sharing common infrastructure:

| | Deliberation | Fusion |
|---|---|---|
| Pattern | Sequential adversarial | Parallel consensus |
| Strength | Depth — finds *why* something is wrong | Breadth — surfaces multi-perspective agreement |
| Default models | Opus proposes, Gemini challenges | Opus + Gemini + Sonnet panel, Opus judges |
| Cost | 2-3× | N+1× |

### Follow-up work

- `ReasoningThread` frontend component for live streaming UI
- Centaur schema `fusion:` block
- Agent-initiated reasoning (expose as tools)
- System memory persistence

## Test plan

- [x] 18 new tests passing (providers, deliberation, fusion)
- [x] Harness package builds clean (`tsc -b packages/harness`)
- [x] No regressions in existing harness/native-command/skill tests
- [ ] Manual test: `/deliberate` and `/fuse` commands in Mitzo chat
- [ ] Centaur review


server/native-commands.ts
```diff
@@ -30,7 +56,19 @@ export class NativeCommandRegistry {
   ): NativeCommandResult | undefined {
     const handler = this.commands.get(name);
     if (!handler) return undefined;
-    return handler(args, skillRegistry);
+    // Sync-only path (backwards compat)
+    return handler(args, skillRegistry, {}) as NativeCommandResult;
```
dimakis: 🔴 **bugs**: `execute()` casts the return of async handlers (`deliberateCommand`, `fuseCommand`) as `NativeCommandResult` via `as NativeCommandResult`. If called with 'deliberate' or 'fuse', it silently returns an unawaited `Promise` object instead of the result. While both current call sites use `executeAsync()`, this is a type-safety hole — the compiler thinks it returns `NativeCommandResult` but it returns `Promise<NativeCommandResult>`. Either delete `execute()` or add a runtime guard that throws if the handler returns a Promise. `[fixable]`

dimakis (commented):
## Centaur Review

Found **10** issue(s) (1 critical) (6 warning).


### `server/native-commands.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🔴 **bugs** (L60): `execute()` casts the return of async handlers (`deliberateCommand`, `fuseCommand`) as `NativeCommandResult` via `as NativeCommandResult`. If called with 'deliberate' or 'fuse', it silently returns an unawaited `Promise` object instead of the result. While both current call sites use `executeAsync()`, this is a type-safety hole — the compiler thinks it returns `NativeCommandResult` but it returns `Promise<NativeCommandResult>`. Either delete `execute()` or add a runtime guard that throws if the handler returns a Promise. `[fixable]`
- 🟡 **missing_tests**: No tests for the new `/deliberate` and `/fuse` commands, the `executeAsync()` method, or the `buildEventEmitter()` helper. The existing `execute()` method also lacks tests. At minimum, test: (1) empty-args usage help, (2) event emission via transport mock, (3) error handling when orchestrator throws. `[fixable]`
- 🟡 **regressions** (L60): The sync `execute()` method now passes an empty `{}` as the third arg to handlers. For the existing `skillsCommand` and `closeCommand`, this works because they ignore the third arg. But the type union `SyncHandler | AsyncHandler` means TypeScript won't catch if a future sync handler accidentally destructures `ctx` — the empty object will produce `undefined` fields silently. `[fixable]`


### `packages/harness/src/reasoning/deliberation.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🟡 **bugs** (L28): Instance state (`transcript`, `totalCost`) is never reset between `run()` calls. If an orchestrator instance is reused, the second call starts with accumulated cost (causing premature budget exhaustion) and a stale transcript. Add `this.transcript = []; this.totalCost = 0;` at the top of `run()`. Same issue exists in `FusionOrchestrator`. `[fixable]`


### `packages/harness/src/reasoning/fusion.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🟡 **bugs** (L30): Same instance-state accumulation bug as `DeliberationOrchestrator` — `transcript` and `totalCost` are not reset at the start of `run()`. Reusing a `FusionOrchestrator` instance will produce incorrect cost tracking and inflated transcripts. `[fixable]`


### `packages/harness/src/providers/google-vertex.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🟡 **unsafe_assumptions** (L133): `getAccessToken()` shells out to `gcloud auth print-access-token` via `execSync` on every API call. This blocks the event loop (5s timeout), requires the gcloud CLI to be installed, and has no token caching. In a deliberation with 4 Gemini calls, that's 4 synchronous subprocess spawns. Consider caching the token with a TTL, or using `google-auth-library` for non-blocking ADC token retrieval. `[fixable]`


### `packages/harness/__tests__/deliberation.test.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🟡 **style** (L122): Orphaned `import { beforeEach } from 'vitest';` at the end of the file (line 122). `beforeEach` is already imported at the top via the `describe`/`it`/`expect`/`vi` import on line 1. This duplicate import after all test code is a copy-paste artifact and will likely trigger linting errors. `[fixable]`
- 🔵 **missing_tests**: No tests for budget exhaustion (early termination), JSON parse failure in `respond()`, or orchestrator instance reuse. The happy path is covered but edge cases are not. `[fixable]`


_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🔵 **style** (L32): The base `AnthropicVertexProvider` class (lines 32-118) is dead code — never instantiated anywhere. Its `resolveModel()` method always returns `'claude-opus-4-6'` regardless of context, making it broken by design. Only `AnthropicVertexModelProvider` (line 124) is used by the provider factory. Consider removing the dead class to reduce confusion. `[fixable]`


### `packages/harness/__tests__/fusion.test.ts`
_Solid multi-model reasoning architecture with good type design, but has a critical type-safety hole in `execute()` returning unawaited Promises, orchestrator state accumulation bugs on reuse, and missing test coverage for error/edge paths and the new native commands._

- 🔵 **missing_tests**: No tests for: all panel members failing (should throw), budget exhaustion mid-pipeline, judge JSON parse failure (should return empty analysis), or orchestrator instance reuse. `[fixable]`

server/native-commands.ts
dimakis: 🟡 **regressions**: The sync `execute()` method now passes an empty `{}` as the third arg to handlers. For the existing `skillsCommand` and `closeCommand`, this works because they ignore the third arg. But the type union `SyncHandler | AsyncHandler` means TypeScript won't catch if a future sync handler accidentally destructures `ctx` — the empty object will produce `undefined` fields silently. `[fixable]`

packages/harness/src/reasoning/deliberation.ts
```diff
@@ -0,0 +1,298 @@
...
+  DeliberationResult,
+  DeliberationRound,
+  TranscriptEntry,
+} from './types.js';
+
+const log = createLogger('deliberation');
+
+export class DeliberationOrchestrator {
+  private proposerProvider: ModelProvider;
+  private challengerProvider: ModelProvider;
```
dimakis, line 28: 🟡 **bugs**: Instance state (`transcript`, `totalCost`) is never reset between `run()` calls. If an orchestrator instance is reused, the second call starts with accumulated cost (causing premature budget exhaustion) and a stale transcript. Add `this.transcript = []; this.totalCost = 0;` at the top of `run()`. Same issue exists in `FusionOrchestrator`. `[fixable]`

packages/harness/src/reasoning/fusion.ts
```diff
@@ -0,0 +1,377 @@
...
+import type {
+  FusionConfig,
+  FusionResult,
+  JudgeAnalysis,
+  TranscriptEntry,
+} from './types.js';
+
+const log = createLogger('fusion');
+
+export class FusionOrchestrator {
```
dimakis, line 25: 🟡 **bugs**: Same instance-state accumulation bug as `DeliberationOrchestrator` — `transcript` and `totalCost` are not reset at the start of `run()`. Reusing a `FusionOrchestrator` instance will produce incorrect cost tracking and inflated transcripts. `[fixable]`

packages/harness/src/providers/google-vertex.ts
```diff
@@ -0,0 +1,156 @@
...
+      role: m.role === 'assistant' ? 'model' : 'user',
+      parts: [{ text: m.content }],
+    }));
+  }
+
+  /** Get an access token from Application Default Credentials. */
+  private async getAccessToken(): Promise<string> {
+    // Use gcloud to get a token — this works with ADC and service accounts
+    const { execSync } = await import('node:child_process');
+    const token = execSync('gcloud auth print-access-token', {
```
dimakis: 🟡 **unsafe_assumptions**: `getAccessToken()` shells out to `gcloud auth print-access-token` via `execSync` on every API call. This blocks the event loop (5s timeout), requires the gcloud CLI to be installed, and has no token caching. In a deliberation with 4 Gemini calls, that's 4 synchronous subprocess spawns. Consider caching the token with a TTL, or using `google-auth-library` for non-blocking ADC token retrieval. `[fixable]`

packages/harness/__tests__/deliberation.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  it('has system prompts for both roles', () => {
+    expect(DEFAULT_DELIBERATION_CONFIG.proposer.systemPrompt).toBeTruthy();
+    expect(DEFAULT_DELIBERATION_CONFIG.challenger.systemPrompt).toBeTruthy();
+    expect(DEFAULT_DELIBERATION_CONFIG.proposer.name).toBe('architect');
+    expect(DEFAULT_DELIBERATION_CONFIG.challenger.name).toBe('critic');
+  });
+});
+
+import { beforeEach } from 'vitest';
```
dimakis: 🟡 **style**: Orphaned `import { beforeEach } from 'vitest';` at the end of the file (line 122). `beforeEach` is already imported at the top via the `describe`/`it`/`expect`/`vi` import on line 1. This duplicate import after all test code is a copy-paste artifact and will likely trigger linting errors. `[fixable]`

packages/harness/src/providers/anthropic-vertex.ts
```diff
@@ -0,0 +1,185 @@
...
+export interface AnthropicVertexProviderOptions {
+  /** GCP project ID. Falls back to ANTHROPIC_VERTEX_PROJECT_ID env var. */
+  projectId?: string;
+  /** GCP region. Falls back to CLOUD_ML_REGION env var, then 'us-east5'. */
+  region?: string;
+  /** Use standard Anthropic API instead of Vertex. */
+  useDirectApi?: boolean;
+}
+
+export class AnthropicVertexProvider implements ModelProvider {
```
dimakis: 🔵 **style**: The base `AnthropicVertexProvider` class (lines 32-118) is dead code — never instantiated anywhere. Its `resolveModel()` method always returns `'claude-opus-4-6'` regardless of context, making it broken by design. Only `AnthropicVertexModelProvider` (line 124) is used by the provider factory. Consider removing the dead class to reduce confusion. `[fixable]`

packages/harness/src/reasoning/fusion.ts
```diff
@@ -0,0 +1,381 @@
...
+        panelResponses,
+        totalCost: this.totalCost,
+        transcript: this.transcript,
+      };
+    }
+
+    // Phase 2: Judge analysis
+    const judgeAnalysis = await this.judge(panelResponses, task, context);
+
+    if (this.budgetExhausted()) {
```
dimakis, line 91: 🔴 **bugs**: Budget-exhausted check after judge phase logs a warning but does NOT skip the synthesize phase — execution falls through to `this.synthesize()` on line 97. The deliberation orchestrator correctly uses `break` to stop when budget is exhausted, but fusion just warns and proceeds to spend more. Should return early with the judge analysis as the final output, similar to the budget-exhausted fallback after fan-out on lines 73-86. `[fixable]`

server/native-commands.ts
```diff
@@ -91,3 +123,169 @@ function closeCommand(): NativeCommandResult {
...
+        'Fan-out to 3 models, judge analyzes consensus/contradictions, synthesizer writes final answer.\n\n' +
+        'Options:\n- `/fuse --self <task>` — self-fusion (same model x2, cheapest)\n\n' +
+        'Example: `/fuse What are the tradeoffs of event sourcing vs CRUD for our order system?`',
+    };
+  }
+
+  // Parse --self flag
+  let task = args.trim();
+  let fusionConfig = DEFAULT_FUSION_CONFIG;
+  if (task.startsWith('--self ')) {
```
dimakis: 🟡 **bugs**: `/fuse --self` is only recognized at the start of the task string. `/fuse --self` alone (no task) passes the empty-args check on line 194 because `args.trim()` is `'--self'` which is truthy, then `task.startsWith('--self ')` fails (no trailing space) so it's treated as a literal task `'--self'`, sent to the models as the prompt text. Should handle `--self` without a trailing space (or with no task after it) and show usage. `[fixable]`

server/native-commands.ts
```diff
@@ -23,14 +49,20 @@ export class NativeCommandRegistry {
     return new Set(this.commands.keys());
   }
 
-  execute(
+  /**
+   * Execute a native command. Always returns a Promise for uniform handling.
+   * Callers must await the result — async handlers (deliberate, fuse) return
+   * Promises that would otherwise be silently dropped.
+   */
+  async execute(
```
dimakis, line 57: 🟡 **missing_tests**: The `execute()` method signature changed from sync to async (`Promise<NativeCommandResult | undefined>`), and both callers in `index.ts` and `ws-handler-v2.ts` switched to fire-and-forget `.then().catch()`. There are no tests for the `NativeCommandRegistry` class at all — no test file exists. The sync-to-async contract change and fire-and-forget error handling in callers are untested. `[fixable]`

server/native-commands.ts
```diff
@@ -91,3 +123,169 @@ function closeCommand(): NativeCommandResult {
...
+
+  try {
+    const orchestrator = new DeliberationOrchestrator({
+      ...DEFAULT_DELIBERATION_CONFIG,
+      onEvent,
+    });
+
+    const result = await orchestrator.run(args.trim(), '');
+
+    const roundsSummary = result.rounds
```
dimakis, line 154: 🔵 **style**: The type annotations on `.map()` callbacks (e.g., `(r: { roundNum: number; positionChanged: boolean; changeDetail?: string })`) are redundant — TypeScript infers these from `result.rounds` which is typed as `DeliberationRound[]`. Same pattern on lines 223-224, 232, 235, 248. These explicit annotations add visual noise without type safety benefit. `[fixable]`

dimakis (commented):
## Centaur Review

Found **8** issue(s) (1 critical) (5 warning).


### `packages/harness/src/reasoning/fusion.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🔴 **bugs** (L92): Budget-exhausted check after judge phase logs a warning but does NOT skip the synthesize phase — execution falls through to `this.synthesize()` on line 97. The deliberation orchestrator correctly uses `break` to stop when budget is exhausted, but fusion just warns and proceeds to spend more. Should return early with the judge analysis as the final output, similar to the budget-exhausted fallback after fan-out on lines 73-86. `[fixable]`


### `server/native-commands.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🟡 **bugs** (L208): `/fuse --self` is only recognized at the start of the task string. `/fuse --self` alone (no task) passes the empty-args check on line 194 because `args.trim()` is `'--self'` which is truthy, then `task.startsWith('--self ')` fails (no trailing space) so it's treated as a literal task `'--self'`, sent to the models as the prompt text. Should handle `--self` without a trailing space (or with no task after it) and show usage. `[fixable]`
- 🟡 **missing_tests** (L57): The `execute()` method signature changed from sync to async (`Promise<NativeCommandResult | undefined>`), and both callers in `index.ts` and `ws-handler-v2.ts` switched to fire-and-forget `.then().catch()`. There are no tests for the `NativeCommandRegistry` class at all — no test file exists. The sync-to-async contract change and fire-and-forget error handling in callers are untested. `[fixable]`
- 🔵 **style** (L154): The type annotations on `.map()` callbacks (e.g., `(r: { roundNum: number; positionChanged: boolean; changeDetail?: string })`) are redundant — TypeScript infers these from `result.rounds` which is typed as `DeliberationRound[]`. Same pattern on lines 223-224, 232, 235, 248. These explicit annotations add visual noise without type safety benefit. `[fixable]`


### `packages/harness/__tests__/providers.test.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🟡 **missing_tests** (L52): The `createProvider` tests (lines 52-67) instantiate real SDK clients (`new Anthropic()`, `new GoogleVertexModelProvider()`) without mocking. Unlike `deliberation.test.ts` and `fusion.test.ts` which mock `createProvider`, this file imports the real implementation. These tests will fail in CI if `@anthropic-ai/sdk` constructor requires `ANTHROPIC_API_KEY` or throws on missing credentials. Consider mocking the SDK constructors or testing only the factory routing logic without instantiation. `[fixable]`


### `packages/harness/src/providers/google-vertex.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🟡 **unsafe_assumptions** (L143): `getAccessToken()` shells out to `gcloud auth print-access-token` via `execFile`. No error handling if `gcloud` is not installed (ENOENT) — the raw error propagates up as a cryptic child_process error. Should catch and rethrow with a user-friendly message like 'gcloud CLI not found — install Google Cloud SDK or set up Application Default Credentials'. `[fixable]`


### `packages/harness/src/reasoning/deliberation.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🟡 **unsafe_assumptions** (L171): The respond phase parses a JSON response from the proposer model using a greedy regex (`/\{[\s\S]*\}/`). If the model response contains multiple JSON objects (e.g., in example code blocks), this regex matches from the first `{` to the last `}` in the entire response, potentially capturing invalid JSON. A non-greedy match (`/\{[\s\S]*?\}/`) or structured output (Anthropic tool_use) would be more reliable. `[fixable]`


### `server/index.ts`
_Solid architecture for multi-model reasoning, but fusion has a budget-enforcement bug (synthesize runs even when budget is exhausted after judge), `/fuse --self` without a task silently sends '--self' as the prompt, and provider tests will likely fail in CI without SDK credentials._

- 🔵 **regressions** (L758): The native command execution block no longer has an explicit `return` or `break` after the fire-and-forget `.then().catch()`. In the original synchronous code, the block fell through to the `else if` branches which wouldn't match. This still works correctly because the `if (resolution.type === 'native')` block is exclusive with the `else if` branches. However, the `ws-handler-v2.ts` version (line 482) correctly has `return;` after the async block — this inconsistency could cause confusion. `[fixable]`

packages/harness/__tests__/providers.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    expect(calculateCost('unknown-model', usage)).toBe(0);
+  });
+
+  it('returns 0 for zero token usage', () => {
+    const usage: TokenUsage = { inputTokens: 0, outputTokens: 0 };
+    expect(calculateCost('claude-opus-4-6', usage)).toBe(0);
+  });
+});
+
+describe('createProvider', () => {
```
dimakis, line 52: 🟡 **missing_tests**: The `createProvider` tests (lines 52-67) instantiate real SDK clients (`new Anthropic()`, `new GoogleVertexModelProvider()`) without mocking. Unlike `deliberation.test.ts` and `fusion.test.ts` which mock `createProvider`, this file imports the real implementation. These tests will fail in CI if `@anthropic-ai/sdk` constructor requires `ANTHROPIC_API_KEY` or throws on missing credentials. Consider mocking the SDK constructors or testing only the factory routing logic without instantiation. `[fixable]`

packages/harness/src/providers/google-vertex.ts
```diff
@@ -0,0 +1,170 @@
...
+    const now = Date.now();
+    if (this.cachedToken && now < this.cachedToken.expiresAt) {
+      return this.cachedToken.token;
+    }
+
+    const { execFile } = await import('node:child_process');
+    const { promisify } = await import('node:util');
+    const execFileAsync = promisify(execFile);
+
+    const { stdout } = await execFileAsync('gcloud', ['auth', 'print-access-token'], {
```
dimakis: 🟡 **unsafe_assumptions**: `getAccessToken()` shells out to `gcloud auth print-access-token` via `execFile`. No error handling if `gcloud` is not installed (ENOENT) — the raw error propagates up as a cryptic child_process error. Should catch and rethrow with a user-friendly message like 'gcloud CLI not found — install Google Cloud SDK or set up Application Default Credentials'. `[fixable]`

packages/harness/src/reasoning/deliberation.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  private async respond(
+    proposal: string,
+    challenge: string,
+    task: string,
+    context: string,
+    roundNum: number,
+  ): Promise<{ response: string; changed: boolean; detail?: string; revisedProposal: string }> {
+    const prompt = `## Context\n${context}\n\n## Task\n${task}\n\n## Your Original Proposal\n${proposal}\n\n## Challenge Received\n${challenge}\n\n## Your Role\nRespond to the challenge. You MUST commit: did the challenge change your position?\n"I'll think about it" is not valid. Commit YES or NO.\n\nRespond with a JSON object:\n{\n  "response": "Your defense or acknowledgment",\n  "position_changed": false,\n  "change_detail": null,\n  "revised_proposal": null\n}\n\nIf you ARE changing your position:\n{\n  "response": "Fair point. Adjusting approach.",\n  "position_changed": true,\n  "change_detail": "Specific description of what changed",\n  "revised_proposal": "Full updated proposal incorporating the change"\n}`;
+
```
dimakis, line 171: 🟡 **unsafe_assumptions**: The respond phase parses a JSON response from the proposer model using a greedy regex (`/\{[\s\S]*\}/`). If the model response contains multiple JSON objects (e.g., in example code blocks), this regex matches from the first `{` to the last `}` in the entire response, potentially capturing invalid JSON. A non-greedy match (`/\{[\s\S]*?\}/`) or structured output (Anthropic tool_use) would be more reliable. `[fixable]`

server/index.ts
```diff
@@ -758,19 +758,24 @@ function handleChatWs(
             span.setAttribute('ws.send.resolution', resolution.type);
```
dimakis, line 758: 🔵 **regressions**: The native command execution block no longer has an explicit `return` or `break` after the fire-and-forget `.then().catch()`. In the original synchronous code, the block fell through to the `else if` branches which wouldn't match. This still works correctly because the `if (resolution.type === 'native')` block is exclusive with the `else if` branches. However, the `ws-handler-v2.ts` version (line 482) correctly has `return;` after the async block — this inconsistency could cause confusion. `[fixable]`

packages/harness/src/reasoning/deliberation.ts
```diff
@@ -0,0 +1,302 @@
...
+      challenger: `${this.config.challenger.name} (${this.config.challenger.model})`,
+      maxRounds: this.config.maxRounds,
+    });
+
+    onEvent?.({ type: 'reasoning_start', mode: 'deliberation', task });
+
+    // Phase 1: Propose
+    const proposal = await this.propose(task, context);
+
+    if (this.budgetExhausted()) {
```
dimakis, line 57: 🟡 **bugs**: Budget is checked AFTER the cost is incurred (post-hoc). `callModel` adds to `this.totalCost`, then `budgetExhausted()` checks on the next iteration. The budget can be exceeded by the cost of one full API call. This is by-design for a ceiling (not a pre-check), but the budget semantics should be documented — a $2 budget can result in $2 + (cost of last call) actual spend. `[fixable]`

packages/harness/src/providers/types.ts
```diff
@@ -0,0 +1,76 @@
...
+  outputPerMillion: number;
+}
+
+/** Known model cost rates (USD per million tokens). */
+export const MODEL_COSTS: Record<string, CostRate> = {
+  // Claude via Vertex AI
+  'claude-opus-4-6': { inputPerMillion: 15.0, outputPerMillion: 75.0 },
+  'claude-sonnet-4-6': { inputPerMillion: 3.0, outputPerMillion: 15.0 },
+  'claude-haiku-4-5': { inputPerMillion: 0.8, outputPerMillion: 4.0 },
+
```
dimakis, line 62: 🟡 **bugs**: MODEL_COSTS for 'claude-opus-4-6' lists $15/$75 per million tokens, but Opus 4 pricing is $15/$75 only for the original Opus 4 (not 4.6). Verify these rates match your actual Vertex AI billing — if they're wrong, budget enforcement and cost reporting will be inaccurate. The Gemini 2.5 Pro rates ($1.25/$10) also vary by whether you hit the >200k context tier ($2.50/$10). `[fixable]`

packages/harness/src/providers/google-vertex.ts
```diff
@@ -0,0 +1,181 @@
...
+    messages: ProviderMessage[],
+  ): Array<{ role: string; parts: Array<{ text: string }> }> {
+    return messages.map((m) => ({
+      role: m.role === 'assistant' ? 'model' : 'user',
+      parts: [{ text: m.content }],
+    }));
+  }
+
+  /** Cached token + expiry (4 min TTL to stay inside gcloud's 60 min window). */
+  private cachedToken: { token: string; expiresAt: number } | null = null;
```
dimakis, line 130: 🟡 **unsafe_assumptions**: Token caching has a race condition under concurrent calls. If two parallel `call()` invocations both see an expired cache simultaneously, both will spawn `gcloud auth print-access-token` subprocesses. Not a correctness bug (both get valid tokens), but wasteful. Consider a pending-promise lock: store the in-flight fetch promise and return it to concurrent callers. `[fixable]`

packages/harness/__tests__/fusion.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  it('returns best panel response when budget exhausted after fan-out', async () => {
+    const config: FusionConfig = {
+      ...DEFAULT_FUSION_CONFIG,
+      budgetUsd: 0.01, // tiny budget — exhausted after first panel call ($0.05)
+      onEvent: () => {},
+    };
+
+    const orchestrator = new FusionOrchestrator(config);
+    const result = await orchestrator.run('Test task', 'Test context');
+
```
dimakis, line 122: 🟡 **missing_tests**: The test 'returns empty analysis when judge JSON parse fails' doesn't actually test the error path. The mock always returns valid JSON at callCount=4, so the `catch` block in `judge()` (fusion.ts:258-267) is never exercised. The test comment acknowledges this ('we test the production fallback path by verifying the empty-analysis structure exists') but that only checks the shape, not the fallback logic. `[fixable]`

server/native-commands.ts
```diff
@@ -91,3 +123,180 @@ function closeCommand(): NativeCommandResult {
...
+  }
+}
+
+// --- /fuse command ---
+
+async function fuseCommand(
+  args: string,
+  _skillRegistry: SkillRegistry,
+  ctx: NativeCommandContext,
+): Promise<NativeCommandResult> {
```
dimakis, line 202: 🔵 **style**: The `/fuse` usage help string is duplicated verbatim — it appears both when `!args.trim()` (line 193) and when `!task` after stripping `--self` (line 211). Extract to a constant to avoid drift between the two copies. `[fixable]`

server/native-commands.ts
```diff
@@ -91,3 +123,180 @@ function closeCommand(): NativeCommandResult {
...
+
+  const onEvent = buildEventEmitter(ctx.transport, 'deliberation');
+
+  try {
+    const orchestrator = new DeliberationOrchestrator({
+      ...DEFAULT_DELIBERATION_CONFIG,
+      onEvent,
+    });
+
+    const result = await orchestrator.run(args.trim(), '');
```
dimakis, line 152: 🔵 **unsafe_assumptions**: Both `/deliberate` and `/fuse` pass empty string for the `context` parameter (`orchestrator.run(args.trim(), '')`). The prompts inside the orchestrators include `## Context\n${context}` — an empty context produces a vacuous section that consumes prompt space. Consider either omitting the context section when empty, or allowing users to supply context (e.g., via `--context` flag or active session context). `[fixable]`

packages/harness/src/reasoning/fusion.ts
```diff
@@ -0,0 +1,384 @@
...
+    onEvent?.({ type: 'reasoning_start', mode: 'fusion', task });
+
+    // Phase 1: Fan-out (parallel)
+    const panelResponses = await this.fanOut(task, context);
+
+    if (this.budgetExhausted()) {
+      log.warn('Budget exhausted after fan-out, returning best panel response');
+      onEvent?.({ type: 'reasoning_end', mode: 'fusion', totalCost: this.totalCost });
+      // Return the longest panel response as a reasonable fallback
+      const best = panelResponses.reduce((a, b) => (a.response.length > b.response.length ? a : b));
```
dimakis, line 72: 🔵 **style**: Budget-exhausted fallback picks the longest panel response (`panelResponses.reduce((a, b) => a.response.length > b.response.length ? a : b)`). Length is a weak proxy for quality — verbose padding would win. This heuristic appears twice (lines 72 and 94). Consider documenting the rationale or using a different signal.

dimakis (commented):
## Centaur Review

Found **10** issue(s) (5 warning).


### `packages/harness/src/reasoning/deliberation.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🟡 **bugs** (L57): Budget is checked AFTER the cost is incurred (post-hoc). `callModel` adds to `this.totalCost`, then `budgetExhausted()` checks on the next iteration. The budget can be exceeded by the cost of one full API call. This is by-design for a ceiling (not a pre-check), but the budget semantics should be documented — a $2 budget can result in $2 + (cost of last call) actual spend. `[fixable]`


### `packages/harness/src/providers/types.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🟡 **bugs** (L62): MODEL_COSTS for 'claude-opus-4-6' lists $15/$75 per million tokens, but Opus 4 pricing is $15/$75 only for the original Opus 4 (not 4.6). Verify these rates match your actual Vertex AI billing — if they're wrong, budget enforcement and cost reporting will be inaccurate. The Gemini 2.5 Pro rates ($1.25/$10) also vary by whether you hit the >200k context tier ($2.50/$10). `[fixable]`


### `packages/harness/src/providers/google-vertex.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🟡 **unsafe_assumptions** (L130): Token caching has a race condition under concurrent calls. If two parallel `call()` invocations both see an expired cache simultaneously, both will spawn `gcloud auth print-access-token` subprocesses. Not a correctness bug (both get valid tokens), but wasteful. Consider a pending-promise lock: store the in-flight fetch promise and return it to concurrent callers. `[fixable]`


### `packages/harness/package.json`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🟡 **regressions**: `@anthropic-ai/vertex-sdk` is imported at runtime in `src/providers/anthropic-vertex.ts` but only listed as a `devDependency` in the harness package.json. It should be a `dependency` or `peerDependency`. Works in the monorepo because the root package.json has it, but the package metadata is incorrect if the harness is ever consumed standalone. `[fixable]`


### `packages/harness/__tests__/fusion.test.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🟡 **missing_tests** (L122): The test 'returns empty analysis when judge JSON parse fails' doesn't actually test the error path. The mock always returns valid JSON at callCount=4, so the `catch` block in `judge()` (fusion.ts:258-267) is never exercised. The test comment acknowledges this ('we test the production fallback path by verifying the empty-analysis structure exists') but that only checks the shape, not the fallback logic. `[fixable]`
- 🔵 **missing_tests**: No test for the case where all panel members fail in fan-out. The code throws 'All panel members failed — no responses to judge' (fusion.ts:188), but this path is never exercised in tests. `[fixable]`


### `packages/harness/__tests__/providers.test.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🔵 **missing_tests**: Provider tests only cover the factory function and cost calculation. No tests for Anthropic provider initialization (Vertex vs Direct API switching based on CLAUDE_CODE_USE_VERTEX), Google provider's token caching/refresh, or error handling when gcloud is not found (the ENOENT path in google-vertex.ts:152). `[fixable]`


### `server/native-commands.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🔵 **style** (L193): The `/fuse` usage help string is duplicated verbatim — it appears both when `!args.trim()` (line 193) and when `!task` after stripping `--self` (line 211). Extract to a constant to avoid drift between the two copies. `[fixable]`
- 🔵 **unsafe_assumptions** (L152): Both `/deliberate` and `/fuse` pass empty string for the `context` parameter (`orchestrator.run(args.trim(), '')`). The prompts inside the orchestrators include `## Context\n${context}` — an empty context produces a vacuous section that consumes prompt space. Consider either omitting the context section when empty, or allowing users to supply context (e.g., via `--context` flag or active session context). `[fixable]`


### `packages/harness/src/reasoning/fusion.ts`
_Well-structured multi-model reasoning feature with good architectural separation. Key concerns: budget enforcement is post-hoc (can overshoot by one call), vertex-sdk is misclassified as devDependency, and several test paths (judge JSON parse failure, all-panels-fail) are declared but not actually exercised._

- 🔵 **style** (L72): Budget-exhausted fallback picks the longest panel response (`panelResponses.reduce((a, b) => a.response.length > b.response.length ? a : b)`). Length is a weak proxy for quality — verbose padding would win. This heuristic appears twice (lines 72 and 94). Consider documenting the rationale or using a different signal.

server/native-commands.ts
```diff
@@ -91,3 +123,180 @@ function closeCommand(): NativeCommandResult {
...
+      '',
+      '### Debate',
+      roundsSummary || '_No rounds completed_',
+      '',
+      '### Final Output',
+      result.final,
+    ].join('\n');
+
+    return { command: 'deliberate', content };
+  } catch (err) {
```
dimakis, line 178: 🟡 **bugs**: If the orchestrator's `run()` throws (e.g. provider API error mid-deliberation), `reasoning_start` has already been emitted but `reasoning_end` is never sent. The catch block returns an error message but doesn't emit `reasoning_end`, leaving the streaming UI in a 'reasoning in progress' state. Same issue in `fuseCommand` at line 253. Fix: emit `reasoning_end` in the catch block via the `onEvent` callback. `[fixable]`

server/native-commands.ts
```diff
@@ -91,3 +123,180 @@ function closeCommand(): NativeCommandResult {
...
+}
+
+// --- /fuse command ---
+
+async function fuseCommand(
+  args: string,
+  _skillRegistry: SkillRegistry,
+  ctx: NativeCommandContext,
+): Promise<NativeCommandResult> {
+  if (!args.trim()) {
```
dimakis, line 194: 🔵 **style**: The usage help string for `/fuse` is duplicated verbatim at lines 197-201 and 217-220. Extract to a constant (e.g. `FUSE_USAGE`) to avoid drift. `[fixable]`

dimakis (commented):
## Centaur Review

Found **9** issue(s) (4 warning).


### `server/native-commands.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🟡 **bugs** (L178): If the orchestrator's `run()` throws (e.g. provider API error mid-deliberation), `reasoning_start` has already been emitted but `reasoning_end` is never sent. The catch block returns an error message but doesn't emit `reasoning_end`, leaving the streaming UI in a 'reasoning in progress' state. Same issue in `fuseCommand` at line 253. Fix: emit `reasoning_end` in the catch block via the `onEvent` callback. `[fixable]`
- 🔵 **style** (L194): The usage help string for `/fuse` is duplicated verbatim at lines 197-201 and 217-220. Extract to a constant (e.g. `FUSE_USAGE`) to avoid drift. `[fixable]`


### `packages/harness/src/reasoning/deliberation.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🟡 **bugs** (L183): When `respond()` parses JSON with `position_changed: true` but the model omits `revised_proposal`, the fallback `data.revised_proposal ?? proposal` returns the original proposal unchanged. Yet `positionChanged` is reported as `true` and `mindChanges` increments. Downstream, `currentProposal` doesn't actually change despite the round claiming it did. Either validate that `revised_proposal` is present when `position_changed` is true, or set `changed: false` when it's missing. `[fixable]`
- 🔵 **style** (L1): Multi-line comments at the top of deliberation.ts (lines 1-13), fusion.ts (lines 1-17), and types.ts (lines 1-9) explain what the module does. The project's code style says 'default to writing no comments' and file-level docblocks tend to rot. The module names + exports are self-documenting. `[fixable]`


_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🟡 **regressions** (L224): The reconnect POST (`doPost('reconnect', ...)`) is fire-and-forget (not awaited), but `flushPendingSends()` runs immediately after on line 234. If there are queued sends, they'll race the reconnect POST to the server. The server may receive a `send` for a session that hasn't been reattached yet (the reconnect handler does watch + reattach + replay). Consider awaiting the reconnect POST before flushing. `[fixable]`


### `packages/harness/__tests__/deliberation.test.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🟡 **missing_tests**: No test for provider call failure. If `provider.call()` throws mid-deliberation (API timeout, auth error, rate limit), the error propagates unhandled from the orchestrator. A test should verify that errors from `callModel` propagate correctly and that the caller (the command handler) handles them. `[fixable]`


### `packages/harness/__tests__/fusion.test.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🔵 **missing_tests** (L120): The 'returns empty analysis when judge JSON parse fails' test never actually triggers the parse failure path — the mock returns valid JSON at `callCount=4`. The test just asserts the shape exists. Consider a dedicated mock that returns non-JSON from the judge to exercise the catch block in `fusion.ts:246`. `[fixable]`


### `packages/harness/src/providers/google-vertex.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🔵 **unsafe_assumptions** (L64): `projectId` and `region` are interpolated directly into the Vertex AI endpoint URL without validation. While these typically come from env vars (low injection risk), a malformed value with `/` or special characters could alter the URL path. Consider a regex check (e.g. `/^[a-z][a-z0-9-]*$/`) on both values in the constructor. `[fixable]`


### `packages/harness/src/reasoning/fusion.ts`
_Well-structured multi-model reasoning engine with good test coverage of the happy path. Main concerns: missing `reasoning_end` event on error paths (UI state leak), a subtle `position_changed` / `revised_proposal` inconsistency in deliberation, and a race between reconnect POST and pending send flush in the SSE client._

- 🔵 **unsafe_assumptions** (L79): Budget exhaustion during parallel `fanOut` is only checked after all panel calls complete (`Promise.allSettled`). If the first panel call already exceeds the budget, the remaining 2 panel calls still run to completion and incur cost. This is acceptable for small panels but worth documenting — budget is a soft ceiling, not a hard stop during parallel phases.

packages/harness/src/reasoning/deliberation.ts
```diff
@@ -0,0 +1,302 @@
...
+      this.config.proposer,
+      prompt,
+      `respond-${roundNum}`,
+      this.config.proposer.name,
+    );
+
+    try {
+      // Extract JSON from the response (may be wrapped in markdown code block)
+      const jsonMatch = rawResponse.match(/\{[\s\S]*\}/);
+      if (!jsonMatch) throw new Error('No JSON found');
```
dimakis, line 183: 🟡 **bugs**: When `respond()` parses JSON with `position_changed: true` but the model omits `revised_proposal`, the fallback `data.revised_proposal ?? proposal` returns the original proposal unchanged. Yet `positionChanged` is reported as `true` and `mindChanges` increments. Downstream, `currentProposal` doesn't actually change despite the round claiming it did. Either validate that `revised_proposal` is present when `position_changed` is true, or set `changed: false` when it's missing. `[fixable]`

packages/harness/src/reasoning/deliberation.ts
```diff
@@ -0,0 +1,302 @@
+/**
```
dimakis, line 1: 🔵 **style**: Multi-line comments at the top of deliberation.ts (lines 1-13), fusion.ts (lines 1-17), and types.ts (lines 1-9) explain what the module does. The project's code style says 'default to writing no comments' and file-level docblocks tend to rot. The module names + exports are self-documenting. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,6 +214,22 @@ export class SseConnection implements ChatConnection {
...
       this._connectionId = msg.connectionId as string;
       this._connected = true;
+
+      // Always send reconnect on welcome if we have tracked sessions.
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      if (this._isReconnect && this.seqBySession.size > 0) {
```
dimakis, line 224: 🟡 **regressions**: The reconnect POST (`doPost('reconnect', ...)`) is fire-and-forget (not awaited), but `flushPendingSends()` runs immediately after on line 234. If there are queued sends, they'll race the reconnect POST to the server. The server may receive a `send` for a session that hasn't been reattached yet (the reconnect handler does watch + reattach + replay). Consider awaiting the reconnect POST before flushing. `[fixable]`

packages/harness/__tests__/fusion.test.ts
```diff
@@ -0,0 +1,203 @@
...
+  });
+
+  it('returns best panel response when budget exhausted after fan-out', async () => {
+    const config: FusionConfig = {
+      ...DEFAULT_FUSION_CONFIG,
+      budgetUsd: 0.01, // tiny budget — exhausted after first panel call ($0.05)
+      onEvent: () => {},
+    };
+
+    const orchestrator = new FusionOrchestrator(config);
```
dimakis, line 120: 🔵 **missing_tests**: The 'returns empty analysis when judge JSON parse fails' test never actually triggers the parse failure path — the mock returns valid JSON at `callCount=4`. The test just asserts the shape exists. Consider a dedicated mock that returns non-JSON from the judge to exercise the catch block in `fusion.ts:246`. `[fixable]`

packages/harness/src/providers/google-vertex.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // Build Gemini API request
+    const systemInstruction = messages
+      .filter((m) => m.role === 'system')
+      .map((m) => m.content)
+      .join('\n\n');
+
+    const contents = this.convertMessages(messages.filter((m) => m.role !== 'system'));
+
+    const endpoint = `https://${this.region}-aiplatform.googleapis.com/v1/projects/${this.projectId}/locations/${this.region}/publishers/google/models/${apiModel}:generateContent`;
```
dimakis, line 64: 🔵 **unsafe_assumptions**: `projectId` and `region` are interpolated directly into the Vertex AI endpoint URL without validation. While these typically come from env vars (low injection risk), a malformed value with `/` or special characters could alter the URL path. Consider a regex check (e.g. `/^[a-z][a-z0-9-]*$/`) on both values in the constructor. `[fixable]`

packages/harness/src/reasoning/fusion.ts
```diff
@@ -0,0 +1,384 @@
...
+      return {
+        finalOutput: best.response,
+        judgeAnalysis: {
+          consensus: [],
+          contradictions: [],
+          partialCoverage: [],
+          uniqueInsights: [],
```
dimakis, line 79: 🔵 **unsafe_assumptions**: Budget exhaustion during parallel `fanOut` is only checked after all panel calls complete (`Promise.allSettled`). If the first panel call already exceeds the budget, the remaining 2 panel calls still run to completion and incur cost. This is acceptable for small panels but worth documenting — budget is a soft ceiling, not a hard stop during parallel phases.

PR #391, opened 2026-06-20
fix(worktree): prevent primary worktree deletion on cleanup

dimakis:
## Summary

- **Root cause:** When the base repo (mgmt) is also listed in `.mitzo.json` repos, `discoverSessionWorktrees` adds both `"primary"` and `"mgmt"` entries pointing to the same worktree directory. On session cleanup, `cleanupSessionWorktrees` skips `"primary"` but removes `"mgmt"` — destroying the primary worktree underneath the active SDK session.
- **Symptom:** `"Path does not exist"` errors mid-session, especially on iOS where frequent WS reconnects trigger zombie-abort → resume cycles that hit the cleanup path.
- **Fix (two layers):**
  - `discoverSessionWorktrees`: skip secondary repos whose path matches `primaryRepo`
  - `cleanupSessionWorktrees`: guard against removing any secondary whose worktree path matches the primary (defense in depth)

## Test plan

- [x] New test: `discoverSessionWorktrees` deduplicates secondary repos matching primaryRepo
- [x] New test: `cleanupSessionWorktrees` skips secondary whose path matches primary worktree
- [x] All 87 existing tests pass (worktree.test.ts + chat.test.ts)
- [ ] Deploy and verify no more "Path does not exist" errors on iOS sessions

### DIFF
--- a/server/__tests__/worktree.test.ts
+++ b/server/__tests__/worktree.test.ts
@@ -613,6 +613,19 @@ describe('discoverSessionWorktrees', () => {
     expect(result.has('primary')).toBe(true);
     expect(result.has('mgmt')).toBe(false);
   });
+
+  it('deduplicates even with trailing slash differences', () => {
+    const wtId = '2026-04-20-abc123def456';
+    mkdirSync(join(primaryRepo, '.claude', 'worktrees', wtId), { recursive: true });
+
+    const result = discoverSessionWorktrees(wtId, primaryRepo, {
+      mgmt: primaryRepo + '/',
+    });
+
+    expect(result.size).toBe(1);
+    expect(result.has('primary')).toBe(true);
+    expect(result.has('mgmt')).toBe(false);
+  });
 });
 
 describe('cleanupStaleWorktrees', () => {
--- a/server/__tests__/worktree.test.ts
+++ b/server/__tests__/worktree.test.ts
@@ -598,6 +598,21 @@ describe('discoverSessionWorktrees', () => {
     expect(result.has('primary')).toBe(true);
     expect(result.has('secondary')).toBe(false);
   });
+
+  it('deduplicates secondary repos that match primaryRepo', () => {
+    const wtId = '2026-04-20-abc123def456';
+    mkdirSync(join(primaryRepo, '.claude', 'worktrees', wtId), { recursive: true });
+
+    // Pass primaryRepo as a secondary too (simulates mgmt in .mitzo.json repos)
+    const result = discoverSessionWorktrees(wtId, primaryRepo, {
+      mgmt: primaryRepo,
+    });
+
+    // Should only have "primary", not "mgmt" — dedup prevents double-mapping
+    expect(result.size).toBe(1);
+    expect(result.has('primary')).toBe(true);
+    expect(result.has('mgmt')).toBe(false);
+  });
 });
 
 describe('cleanupStaleWorktrees', () => {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1260,10 +1260,21 @@ export function cleanupSessionWorktrees(
   session: import('./session-registry.js').ManagedSession,
 ): void {
   const config = getRepoConfig();
-  for (const [repoName, { wtId }] of session.worktreePaths) {
+  const primaryPath = session.worktreePaths.get('primary')?.path;
+  for (const [repoName, { wtId, path }] of session.worktreePaths) {
     if (repoName === 'primary') continue;
     const repoPath = config.repos[repoName];
     if (!repoPath) continue;
+    // Guard: never remove a secondary whose path matches the primary worktree.
+    // This can happen when the base repo is also listed in .mitzo.json repos
+    // (e.g. "mgmt"), causing discoverSessionWorktrees to add both entries.
+    if (primaryPath && path === primaryPath) {
+      log.info('skipping secondary cleanup — path matches primary worktree', {
+        repoName,
+        path,
+      });
+      continue;
+    }
     try {
       removeWorktree(wtId, repoPath);
     } catch (err: unknown) {
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1266,8 +1266,6 @@ export function cleanupSessionWorktrees(
     const repoPath = config.repos[repoName];
     if (!repoPath) continue;
     // Guard: never remove a secondary whose path matches the primary worktree.
-    // This can happen when the base repo is also listed in .mitzo.json repos
-    // (e.g. "mgmt"), causing discoverSessionWorktrees to add both entries.
     if (primaryPath && resolve(path) === resolve(primaryPath)) {
       log.info('skipping secondary cleanup — path matches primary worktree', {
         repoName,
--- a/server/chat.ts
+++ b/server/chat.ts
@@ -1268,7 +1268,7 @@ export function cleanupSessionWorktrees(
     // Guard: never remove a secondary whose path matches the primary worktree.
     // This can happen when the base repo is also listed in .mitzo.json repos
     // (e.g. "mgmt"), causing discoverSessionWorktrees to add both entries.
-    if (primaryPath && path === primaryPath) {
+    if (primaryPath && resolve(path) === resolve(primaryPath)) {
       log.info('skipping secondary cleanup — path matches primary worktree', {
         repoName,
         path,
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -237,6 +237,10 @@ export function discoverSessionWorktrees(
 
   const allRepos: Array<[string, string]> = [['primary', primaryRepo]];
   for (const [name, repoPath] of Object.entries(secondaryRepos)) {
+    // Skip secondary repos that resolve to the same path as the primary —
+    // otherwise both "primary" and "mgmt" map to the same worktree, and
+    // cleanupSessionWorktrees removes the primary thinking it's a secondary.
+    if (repoPath === primaryRepo) continue;
     allRepos.push([name, repoPath]);
   }
 
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -14,7 +14,7 @@ import {
   unlinkSync,
 } from 'fs';
 import { mkdir } from 'fs/promises';
-import { join, basename } from 'path';
+import { join, basename, resolve } from 'path';
 
 const execFileAsync = promisify(execFileCb);
 import {
@@ -240,7 +240,7 @@ export function discoverSessionWorktrees(
     // Skip secondary repos that resolve to the same path as the primary —
     // otherwise both "primary" and "mgmt" map to the same worktree, and
     // cleanupSessionWorktrees removes the primary thinking it's a secondary.
-    if (repoPath === primaryRepo) continue;
+    if (resolve(repoPath) === resolve(primaryRepo)) continue;
     allRepos.push([name, repoPath]);
   }
 
--- a/server/__tests__/chat.test.ts
+++ b/server/__tests__/chat.test.ts
@@ -211,6 +211,41 @@ describe('cleanupSessionWorktrees', () => {
 
     expect(session.worktreePaths.size).toBe(0);
   });
+
+  it('skips secondary whose path matches primary worktree', async () => {
+    const { loadRepoConfig } = await import('../repo-config.js');
+    (loadRepoConfig as ReturnType<typeof vi.fn>).mockReturnValue({
+      repos: { mgmt: '/repo', mitzo: '/tools/mitzo' },
+      isolation: true,
+    });
+
+    const realNow = Date.now;
+    Date.now = () => realNow() + 10_000;
+
+    const { cleanupSessionWorktrees } = await import('../chat.js');
+
+    // "mgmt" secondary points to the same path as "primary" —
+    // simulates discoverSessionWorktrees adding both entries
+    const session = {
+      worktreePaths: new Map([
+        ['primary', { path: '/repo/.claude/worktrees/abc', wtId: 'abc' }],
+        ['mgmt', { path: '/repo/.claude/worktrees/abc', wtId: 'abc' }],
+        ['mitzo', { path: '/tools/mitzo/.claude/worktrees/abc', wtId: 'abc' }],
+      ]),
+    } as unknown as ManagedSession;
+
+    cleanupSessionWorktrees(session);
+
+    // mgmt should NOT be removed (same path as primary), mitzo should be removed
+    expect(removeWorktreeMock).toHaveBeenCalledWith('abc', '/tools/mitzo');
+    expect(removeWorktreeMock).toHaveBeenCalledTimes(1);
+
+    expect(session.worktreePaths.has('primary')).toBe(true);
+    expect(session.worktreePaths.size).toBe(1);
+
+    Date.now = realNow;
+    vi.restoreAllMocks();
+  });
 });
 
 describe('createSessionWorktrees — lazy secondary creation', () => {


packages/client/src/sse-connection.ts
```diff
@@ -217,6 +214,21 @@ export class SseConnection implements ChatConnection {
...
       this._connectionId = msg.connectionId as string;
       this._connected = true;
+
+      // Always send reconnect on welcome if we have tracked sessions.
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      if (this._isReconnect && this.seqBySession.size > 0) {
```
dimakis, line 223: 🟡 **bugs**: Race condition: `doPost('reconnect', ...)` is fire-and-forget (not awaited), but `flushPendingSends()` on line 234 runs immediately after. If there are queued sends, they race the reconnect POST to the server. If a send arrives before the reconnect, the server hasn't reattached the session yet — events may be lost or misordered. The reconnect POST should be awaited before flushing pending sends. `[fixable]`

server/worktree.ts
```diff
@@ -237,6 +237,10 @@ export function discoverSessionWorktrees(
 
   const allRepos: Array<[string, string]> = [['primary', primaryRepo]];
   for (const [name, repoPath] of Object.entries(secondaryRepos)) {
+    // Skip secondary repos that resolve to the same path as the primary —
+    // otherwise both "primary" and "mgmt" map to the same worktree, and
+    // cleanupSessionWorktrees removes the primary thinking it's a secondary.
+    if (repoPath === primaryRepo) continue;
```
dimakis: 🟡 **unsafe_assumptions**: Path deduplication uses strict string equality (`repoPath === primaryRepo`) with no normalization. `loadRepoConfig` in `repo-config.ts` stores paths as raw strings from `.mitzo.json`. If the primary and secondary differ by trailing slash, symlink, or relative vs absolute form, the guard fails and both entries are added. Consider using `path.resolve()` or `fs.realpathSync()` before comparison. `[fixable]`

server/chat.ts
```diff
@@ -1227,10 +1227,21 @@ export function cleanupSessionWorktrees(
...
-  for (const [repoName, { wtId }] of session.worktreePaths) {
+  const primaryPath = session.worktreePaths.get('primary')?.path;
+  for (const [repoName, { wtId, path }] of session.worktreePaths) {
     if (repoName === 'primary') continue;
     const repoPath = config.repos[repoName];
     if (!repoPath) continue;
+    // Guard: never remove a secondary whose path matches the primary worktree.
+    // This can happen when the base repo is also listed in .mitzo.json repos
+    // (e.g. "mgmt"), causing discoverSessionWorktrees to add both entries.
+    if (primaryPath && path === primaryPath) {
```
dimakis: 🔵 **unsafe_assumptions**: Same string-equality issue as `discoverSessionWorktrees`: `path === primaryPath` compares worktree paths constructed via `join(repoPath, ...)`. If the repo paths weren't normalized at discovery time, this guard also fails. However, since both paths are constructed with `join()` from the same source, the cleanup guard will match whenever the discovery guard matches — so this is consistent, just not robust against symlinks or non-canonical paths. `[fixable]`

server/chat.ts
```diff
@@ -1227,10 +1227,21 @@ export function cleanupSessionWorktrees(
   session: import('./session-registry.js').ManagedSession,
 ): void {
   const config = getRepoConfig();
-  for (const [repoName, { wtId }] of session.worktreePaths) {
+  const primaryPath = session.worktreePaths.get('primary')?.path;
+  for (const [repoName, { wtId, path }] of session.worktreePaths) {
     if (repoName === 'primary') continue;
     const repoPath = config.repos[repoName];
```
dimakis, line 1266: 🔵 **style**: The `cleanupSessionWorktrees` guard is defense-in-depth against duplicates that `discoverSessionWorktrees` should have already filtered. The comment explains why, which is good. But this means there are two dedup mechanisms that must stay in sync. Consider extracting the path normalization into a shared helper so both call sites use identical logic. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,63 @@ describe('SseConnection', () => {
     expect(conn.isConnected()).toBe(false);
   });
 
-  it('includes sessions in reconnect URL', () => {
-    const conn = new SseConnection(createConfig());
+  it('sends reconnect POST on welcome when has tracked sessions', () => {
```
dimakis, line 293: 🔵 **missing_tests**: No test verifies the ordering contract: reconnect POST must complete before pending sends are flushed. A test where `mockFetch` delays the reconnect response and verifies sends aren't dispatched until after would catch the race condition identified above. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,63 @@ describe('SseConnection', () => {
...
+      expect.objectContaining({
+        method: 'POST',
+        body: JSON.stringify({
+          type: 'reconnect',
+          sessions: [{ sessionId: 'sess-1', lastSeq: 10 }],
+        }),
+      }),
+    );
+  });
+
```
dimakis, line 320: 🔵 **missing_tests**: No test covers what happens when the reconnect POST fails (fetch rejects or returns non-ok). The current code silently swallows the error, but the test suite should verify the client degrades gracefully (e.g., still flushes sends, doesn't throw). `[fixable]`

dimakis (commented):
## Centaur Review

_Worktree dedup fix is solid defense-in-depth but relies on string equality for path comparison without normalization. The SSE reconnect refactor has a race condition: the fire-and-forget reconnect POST may not arrive before flushed pending sends._

- 🟡 **bugs** (L224): Race condition: `doPost('reconnect', ...)` is fire-and-forget (not awaited), but `flushPendingSends()` on line 234 runs immediately after. If there are queued sends, they race the reconnect POST to the server. If a send arrives before the reconnect, the server hasn't reattached the session yet — events may be lost or misordered. The reconnect POST should be awaited before flushing pending sends. `[fixable]`


### `server/worktree.ts`
_Worktree dedup fix is solid defense-in-depth but relies on string equality for path comparison without normalization. The SSE reconnect refactor has a race condition: the fire-and-forget reconnect POST may not arrive before flushed pending sends._

- 🟡 **unsafe_assumptions** (L243): Path deduplication uses strict string equality (`repoPath === primaryRepo`) with no normalization. `loadRepoConfig` in `repo-config.ts` stores paths as raw strings from `.mitzo.json`. If the primary and secondary differ by trailing slash, symlink, or relative vs absolute form, the guard fails and both entries are added. Consider using `path.resolve()` or `fs.realpathSync()` before comparison. `[fixable]`


### `server/chat.ts`
_Worktree dedup fix is solid defense-in-depth but relies on string equality for path comparison without normalization. The SSE reconnect refactor has a race condition: the fire-and-forget reconnect POST may not arrive before flushed pending sends._

- 🔵 **unsafe_assumptions** (L1238): Same string-equality issue as `discoverSessionWorktrees`: `path === primaryPath` compares worktree paths constructed via `join(repoPath, ...)`. If the repo paths weren't normalized at discovery time, this guard also fails. However, since both paths are constructed with `join()` from the same source, the cleanup guard will match whenever the discovery guard matches — so this is consistent, just not robust against symlinks or non-canonical paths. `[fixable]`
- 🔵 **style** (L1233): The `cleanupSessionWorktrees` guard is defense-in-depth against duplicates that `discoverSessionWorktrees` should have already filtered. The comment explains why, which is good. But this means there are two dedup mechanisms that must stay in sync. Consider extracting the path normalization into a shared helper so both call sites use identical logic. `[fixable]`


_Worktree dedup fix is solid defense-in-depth but relies on string equality for path comparison without normalization. The SSE reconnect refactor has a race condition: the fire-and-forget reconnect POST may not arrive before flushed pending sends._

- 🔵 **missing_tests** (L293): No test verifies the ordering contract: reconnect POST must complete before pending sends are flushed. A test where `mockFetch` delays the reconnect response and verifies sends aren't dispatched until after would catch the race condition identified above. `[fixable]`
- 🔵 **missing_tests** (L320): No test covers what happens when the reconnect POST fails (fetch rejects or returns non-ok). The current code silently swallows the error, but the test suite should verify the client degrades gracefully (e.g., still flushes sends, doesn't throw). `[fixable]`

dimakis:
## Centaur Review

### `server/worktree.ts`
_Correct fix for a real bug (primary worktree deletion). Defense-in-depth at both discover and cleanup layers is good. Main concern is that string equality may miss symlink/trailing-slash path variants — consider normalizing paths._

- 🟡 **unsafe_assumptions** (L243): String comparison `repoPath === primaryRepo` doesn't handle symlinks or trailing-slash differences. If a user configures `.mitzo.json` repos with a symlinked path (e.g. `~/tools/mitzo` → `/Users/x/tools/mitzo`) while `REPO_PATH` uses the resolved path (or vice versa), the dedup fails silently and the original bug reappears. Consider using `realpathSync` or `path.resolve` for normalization. The same applies to the `path === primaryPath` guard in `chat.ts:1271`. `[fixable]`
- 🔵 **bugs** (L316): `createSessionWorktrees` (worktree.ts:303-329) iterates `secondaryRepos` without the same dedup guard added to `discoverSessionWorktrees`. If both the primary and a secondary resolve to the same repo, `createWorktree` is called twice — it doesn't crash (the second call reuses the existing worktree), but the returned `results` map will contain a duplicate entry that could propagate into `worktreePaths`. The fix is correct at the `discover` and `cleanup` layers, but adding the same `if (repoPath === primaryRepo) continue` here would make all three functions consistent. `[fixable]`


### `server/chat.ts`
_Correct fix for a real bug (primary worktree deletion). Defense-in-depth at both discover and cleanup layers is good. Main concern is that string equality may miss symlink/trailing-slash path variants — consider normalizing paths._

- 🔵 **style** (L1268): The 3-line comment on the guard duplicates the comment already present in `discoverSessionWorktrees` (worktree.ts:240-242). Since the `discover` dedup is the primary fix and this guard is a defense-in-depth fallback, a single line like `// Defense-in-depth: see discoverSessionWorktrees dedup` would reduce duplication. `[fixable]`

server/worktree.ts
dimakis: 🟡 **unsafe_assumptions**: String equality (`repoPath === primaryRepo`) is fragile — paths from different sources (.mitzo.json vs REPO_PATH env var) may differ by trailing slash, symlink, or relative component (e.g. `/repo` vs `/repo/` vs `/home/user/../repo`). Use `resolve(repoPath) === resolve(primaryRepo)` (from `path`) for canonical comparison. The same applies to `path === primaryPath` in chat.ts:1271. Since `repo-config.ts` stores paths as-is without normalization, the mismatch risk is real. `[fixable]`

server/__tests__/worktree.test.ts
```diff
@@ -598,6 +598,21 @@ describe('discoverSessionWorktrees', () => {
     expect(result.has('primary')).toBe(true);
     expect(result.has('secondary')).toBe(false);
   });
+
+  it('deduplicates secondary repos that match primaryRepo', () => {
```
dimakis, line 602: 🔵 **missing_tests**: The test only covers the exact-string-match dedup case. A test with a trailing-slash variant (e.g. `mgmt: primaryRepo + '/'`) would document the current limitation or validate a `resolve()`-based fix. `[fixable]`

server/chat.ts
```diff
@@ -1260,10 +1260,21 @@ export function cleanupSessionWorktrees(
...
     if (!repoPath) continue;
+    // Guard: never remove a secondary whose path matches the primary worktree.
```
dimakis, line 1268: 🔵 **style**: The 3-line comment block explaining the guard is justified (non-obvious invariant), but the log message at line 1272 duplicates the same explanation. Consider shortening the comment to just the first line since the log message already serves as documentation for runtime debugging. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_Solid defense-in-depth fix at two layers (prevention + safety net), but the string-equality path comparison is fragile — should use `path.resolve()` to handle trailing slashes, symlinks, and relative segments._

- 🟡 **unsafe_assumptions** (L243): String equality (`repoPath === primaryRepo`) is fragile — paths from different sources (.mitzo.json vs REPO_PATH env var) may differ by trailing slash, symlink, or relative component (e.g. `/repo` vs `/repo/` vs `/home/user/../repo`). Use `resolve(repoPath) === resolve(primaryRepo)` (from `path`) for canonical comparison. The same applies to `path === primaryPath` in chat.ts:1271. Since `repo-config.ts` stores paths as-is without normalization, the mismatch risk is real. `[fixable]`


_Solid defense-in-depth fix at two layers (prevention + safety net), but the string-equality path comparison is fragile — should use `path.resolve()` to handle trailing slashes, symlinks, and relative segments._

- 🔵 **missing_tests** (L602): The test only covers the exact-string-match dedup case. A test with a trailing-slash variant (e.g. `mgmt: primaryRepo + '/'`) would document the current limitation or validate a `resolve()`-based fix. `[fixable]`


### `server/chat.ts`
_Solid defense-in-depth fix at two layers (prevention + safety net), but the string-equality path comparison is fragile — should use `path.resolve()` to handle trailing slashes, symlinks, and relative segments._

- 🔵 **style** (L1268): The 3-line comment block explaining the guard is justified (non-obvious invariant), but the log message at line 1272 duplicates the same explanation. Consider shortening the comment to just the first line since the log message already serves as documentation for runtime debugging. `[fixable]`

dimakis:
## Centaur Review

### `server/worktree.ts`
_Correct fix for the primary-worktree-deletion bug with good test coverage, but `createSessionWorktrees` in `worktree.ts` has the same missing dedup guard that was fixed in `discoverSessionWorktrees` — worth closing that gap in this PR._

- 🟡 **bugs** (L316): The same dedup guard added to `discoverSessionWorktrees` (line 243) is missing from `createSessionWorktrees` (line 316). When the primary repo is also listed in `.mitzo.json` repos, this function will attempt to `createWorktree` for the same repo twice — the second call may fail (branch already exists) or silently create a redundant entry in the results map that reintroduces the double-mapping this PR aims to prevent. Adding `if (resolve(repoPath) === resolve(primaryRepo)) continue;` in the secondary loop would close the gap. `[fixable]`


_Correct fix for the primary-worktree-deletion bug with good test coverage, but `createSessionWorktrees` in `worktree.ts` has the same missing dedup guard that was fixed in `discoverSessionWorktrees` — worth closing that gap in this PR._

- 🔵 **missing_tests** (L602): The `discoverSessionWorktrees` dedup test passes identical string paths (`primaryRepo` for both). Consider adding a case where the paths differ only in normalization (e.g., trailing slash or `./` prefix) to verify that `resolve()` is actually doing useful work, rather than just string equality. `[fixable]`


### `server/chat.ts`
_Correct fix for the primary-worktree-deletion bug with good test coverage, but `createSessionWorktrees` in `worktree.ts` has the same missing dedup guard that was fixed in `discoverSessionWorktrees` — worth closing that gap in this PR._

- 🔵 **style** (L1268): The 3-line comment block (lines 1268-1270) explaining the guard repeats the same information already in the `discoverSessionWorktrees` comment. A single line like `// Dedup: secondary may alias primary (see discoverSessionWorktrees)` would suffice as defense-in-depth commentary. `[fixable]`

dimakis:
## Centaur Review

Found **2** issue(s).


### `server/worktree.ts`
_Solid targeted fix with good test coverage. One gap: the same dedup guard should also be applied to `createSessionWorktrees` in worktree.ts (called from the external hooks endpoint) for consistency._

- 🔵 **missing_tests** (L316): `createSessionWorktrees` in worktree.ts (called from app.ts external hooks endpoint) iterates secondaryRepos without the same `resolve()` dedup guard added to `discoverSessionWorktrees`. If a secondary repo resolves to the same path as the primary, the returned results map will contain a duplicate entry. The cleanup guard in chat.ts would prevent data loss, but adding the dedup here too would be defense-in-depth and keep the three codepaths consistent. `[fixable]`
- 🔵 **style** (L240): The 3-line comment on the dedup guard is useful for the first occurrence but duplicates context already captured in the commit message. A single line like `// Dedup: secondary that resolves to primaryRepo would shadow cleanup.` would suffice. `[fixable]`

PR #392, opened 2026-06-22
fix(client): await reconnect POST before flushing pending sends

dimakis:
## Summary
- Pending sends and `_open` were firing immediately after `doPost('reconnect', ...)` without awaiting, causing user messages queued during tool execution to race the server-side reconnect setup
- Messages sent while a tool was executing (e.g. "?" to interrupt) were silently dropped
- Uses `.finally()` to ensure flush only happens after reconnect POST completes

## Test plan
- [ ] Send a message while agent is executing a long-running Bash command
- [ ] Verify the message interrupts/gets processed instead of hanging
- [ ] Verify normal (non-reconnect) message flow still works

### DIFF
--- a/packages/client/src/__tests__/sse-connection.test.ts
+++ b/packages/client/src/__tests__/sse-connection.test.ts
@@ -475,6 +475,94 @@ describe('SseConnection', () => {
     expect(postEndpoints).toEqual(['reconnect', 'send']);
   });
 
+  it('stays disconnected when reconnect POST fails', async () => {
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        return Promise.resolve({ ok: false, status: 500 });
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    const listener = vi.fn();
+    conn.onMessage(listener);
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect
+    conn.checkAndReconnect(true);
+    conn.send({ type: 'send', prompt: 'should stay queued', clientMsgId: 'q-1' });
+    listener.mockClear();
+
+    // New welcome — reconnect POST will fail
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+
+    // Must stay disconnected — server never ran handleReconnect
+    expect(conn.isConnected()).toBe(false);
+    expect(listener).not.toHaveBeenCalledWith({ type: '_open' });
+  });
+
+  it('stays disconnected when reconnect POST throws network error', async () => {
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        return Promise.reject(new Error('network error'));
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    const listener = vi.fn();
+    conn.onMessage(listener);
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    conn.checkAndReconnect(true);
+    listener.mockClear();
+
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+
+    expect(conn.isConnected()).toBe(false);
+    expect(listener).not.toHaveBeenCalledWith({ type: '_open' });
+  });
+
+  it('recovers after failed reconnect when EventSource auto-reconnects', async () => {
+    let callCount = 0;
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        callCount++;
+        // First reconnect fails, second succeeds
+        if (callCount === 1) return Promise.resolve({ ok: false, status: 500 });
+        return Promise.resolve({ ok: true });
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    const listener = vi.fn();
+    conn.onMessage(listener);
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect
+    conn.checkAndReconnect(true);
+    listener.mockClear();
+
+    // First welcome — reconnect POST fails
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+    expect(conn.isConnected()).toBe(false);
+
+    // EventSource auto-reconnect fires a new welcome
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-ghi' });
+    await vi.runAllTimersAsync();
+
+    // Second attempt succeeds
+    expect(conn.isConnected()).toBe(true);
+    expect(listener).toHaveBeenCalledWith({ type: '_open' });
+  });
+
   it('does not send reconnect POST on first connection', () => {
     const mockFetch = vi.fn().mockResolvedValue({ ok: true });
     const conn = new SseConnection(createConfig({ fetch: mockFetch }));
--- a/packages/client/src/__tests__/sse-connection.test.ts
+++ b/packages/client/src/__tests__/sse-connection.test.ts
@@ -563,6 +563,84 @@ describe('SseConnection', () => {
     expect(listener).toHaveBeenCalledWith({ type: '_open' });
   });
 
+  it('dispatches SSE events to listener while reconnect POST is in-flight', async () => {
+    let resolveReconnect!: (v: { ok: boolean }) => void;
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        return new Promise((resolve) => {
+          resolveReconnect = resolve;
+        });
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    const listener = vi.fn();
+    conn.onMessage(listener);
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    conn.checkAndReconnect(true);
+    listener.mockClear();
+
+    // Welcome — reconnect POST in-flight, _connected = false
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    expect(conn.isConnected()).toBe(false);
+
+    // Server replays events via SSE while reconnect POST is processing.
+    // onmessage is independent of _connected — these must still dispatch.
+    lastES()._emit('block_delta', {
+      type: 'block_delta',
+      sessionId: 'sess-1',
+      seq: 11,
+      delta: 'replayed',
+    });
+
+    expect(listener).toHaveBeenCalledWith(
+      expect.objectContaining({ type: 'block_delta', delta: 'replayed' }),
+    );
+
+    resolveReconnect({ ok: true });
+    await vi.runAllTimersAsync();
+  });
+
+  it('queued sends survive POST failure and flush on successful retry', async () => {
+    let callCount = 0;
+    const postEndpoints: string[] = [];
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      const endpoint = url.replace('https://localhost:3100/api/chat/', '');
+      if (url.includes('/reconnect')) {
+        callCount++;
+        if (callCount === 1) return Promise.resolve({ ok: false, status: 500 });
+        postEndpoints.push(endpoint);
+        return Promise.resolve({ ok: true });
+      }
+      postEndpoints.push(endpoint);
+      return Promise.resolve({ ok: true });
+    });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect and queue a send
+    conn.checkAndReconnect(true);
+    conn.send({ type: 'send', prompt: 'must survive', clientMsgId: 'q-1' });
+    postEndpoints.length = 0;
+
+    // First welcome — reconnect fails, send stays queued
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+    expect(conn.isConnected()).toBe(false);
+    expect(postEndpoints).toEqual([]);
+
+    // Second welcome — reconnect succeeds, queued send flushes
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-ghi' });
+    await vi.runAllTimersAsync();
+    expect(conn.isConnected()).toBe(true);
+    expect(postEndpoints).toEqual(['reconnect', 'send']);
+  });
+
   it('does not send reconnect POST on first connection', () => {
     const mockFetch = vi.fn().mockResolvedValue({ ok: true });
     const conn = new SseConnection(createConfig({ fetch: mockFetch }));
--- a/packages/client/src/__tests__/sse-connection.test.ts
+++ b/packages/client/src/__tests__/sse-connection.test.ts
@@ -641,6 +641,75 @@ describe('SseConnection', () => {
     expect(postEndpoints).toEqual(['reconnect', 'send']);
   });
 
+  it('forces reconnect when reconnect POST fails (SSE stream stays up)', async () => {
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        return Promise.resolve({ ok: false, status: 500 });
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect
+    conn.checkAndReconnect(true);
+
+    // Welcome — reconnect POST will fail
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+
+    // Should have forced a new reconnect (new EventSource created)
+    // The failed POST triggers checkAndReconnect(true), which closes
+    // the old ES and creates a new one
+    expect(conn.isConnected()).toBe(false);
+    const esCount = MockEventSource.instances.length;
+    expect(esCount).toBeGreaterThanOrEqual(3); // initial + reconnect + forced retry
+
+    warnSpy.mockRestore();
+  });
+
+  it('recovers via checkAndReconnect(false) when POST fails but SSE stays up', async () => {
+    let reconnectCallCount = 0;
+    const mockFetch = vi.fn().mockImplementation((url: string) => {
+      if (url.includes('/reconnect')) {
+        reconnectCallCount++;
+        // Fail first two (initial + forced retry), succeed on third
+        if (reconnectCallCount <= 2) return Promise.resolve({ ok: false, status: 500 });
+        return Promise.resolve({ ok: true });
+      }
+      return Promise.resolve({ ok: true });
+    });
+    const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
+    const listener = vi.fn();
+    conn.onMessage(listener);
+    conn.connect();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect — POST fails, triggers checkAndReconnect(true)
+    conn.checkAndReconnect(true);
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+
+    // Forced retry also fails
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-ghi' });
+    await vi.runAllTimersAsync();
+    expect(conn.isConnected()).toBe(false);
+
+    // Third welcome — reconnect POST succeeds
+    listener.mockClear();
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-jkl' });
+    await vi.runAllTimersAsync();
+
+    expect(conn.isConnected()).toBe(true);
+    expect(listener).toHaveBeenCalledWith({ type: '_open' });
+    warnSpy.mockRestore();
+  });
+
   it('connects immediately on reconnect when seqBySession is empty', () => {
     const mockFetch = vi.fn().mockResolvedValue({ ok: true });
     const conn = new SseConnection(createConfig({ fetch: mockFetch }));
--- a/packages/client/src/__tests__/sse-connection.test.ts
+++ b/packages/client/src/__tests__/sse-connection.test.ts
@@ -384,7 +384,7 @@ describe('SseConnection', () => {
     expect(conn.isConnected()).toBe(false);
     listener.mockClear();
 
-    // Resolve the reconnect POST — .finally() should bail out
+    // Resolve the reconnect POST — staleness guard should bail out
     resolveReconnect({ ok: true });
     await vi.runAllTimersAsync();
 
--- a/packages/client/src/__tests__/sse-connection.test.ts
+++ b/packages/client/src/__tests__/sse-connection.test.ts
@@ -290,20 +290,63 @@ describe('SseConnection', () => {
     expect(conn.isConnected()).toBe(false);
   });
 
-  it('includes sessions in reconnect URL', () => {
-    const conn = new SseConnection(createConfig());
+  it('sends reconnect POST on welcome when has tracked sessions', () => {
+    const mockFetch = vi.fn().mockResolvedValue({ ok: true });
+    const conn = new SseConnection(createConfig({ fetch: mockFetch }));
     conn.connect();
     lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
 
     // Track a session
     conn.trackSeq('sess-1', 10);
 
+    // Force reconnect — creates new EventSource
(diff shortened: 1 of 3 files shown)


packages/client/src/sse-connection.ts
```diff
@@ -217,10 +214,29 @@ export class SseConnection implements ChatConnection {
       }
       this._connectionId = msg.connectionId as string;
       this._connected = true;
```
dimakis: 🔴 **bugs**: `_connected = true` is set before the reconnect POST completes. The store layer calls `connection.send()` gated only on `isConnected()` (i.e. `_connected`), not on `_open`. This opens a race window where external `send()` calls bypass the pending queue and fire before the reconnect POST reaches the server — defeating the fix this PR intends. `doPost()` only requires `_connectionId`, not `_connected`, so the reconnect POST itself would still work. Move `this._connected = true` into both branches: inside the `.finally()` callback for the reconnect path, and before `flushPendingSends()` in the else path. `[fixable]`

dimakis (commented):
## Centaur Review

_The fix has the right idea (POST-based reconnect, deferred flush), but `_connected = true` is set before the POST completes, opening a race window that defeats the stated goal — the store can `send()` during the gap because it gates on `isConnected()`, not `_open`._

- 🔴 **bugs** (L216): `_connected = true` is set before the reconnect POST completes. The store layer calls `connection.send()` gated only on `isConnected()` (i.e. `_connected`), not on `_open`. This opens a race window where external `send()` calls bypass the pending queue and fire before the reconnect POST reaches the server — defeating the fix this PR intends. `doPost()` only requires `_connectionId`, not `_connected`, so the reconnect POST itself would still work. Move `this._connected = true` into both branches: inside the `.finally()` callback for the reconnect path, and before `flushPendingSends()` in the else path. `[fixable]`


_The fix has the right idea (POST-based reconnect, deferred flush), but `_connected = true` is set before the POST completes, opening a race window that defeats the stated goal — the store can `send()` during the gap because it gates on `isConnected()`, not `_open`._

- 🟡 **missing_tests**: The PR title is "await reconnect POST before flushing pending sends" but no test verifies the ordering — i.e., that pending sends queued before reconnect are flushed only after the reconnect POST resolves. A test should queue a send, trigger reconnect + welcome, and assert that the queued send's fetch call happens after the reconnect POST's fetch call (e.g., by checking mockFetch call order or using a deferred promise for the reconnect POST). `[fixable]`
- 🔵 **missing_tests**: No test covers the `.finally()` error path: if the reconnect POST fails (fetch rejects), `flushPendingSends()` and `_open` should still fire. This is the whole reason `.finally()` was chosen over `.then()`, but there's no test proving it works. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +213,35 @@ export class SseConnection implements ChatConnection {
...
+      // lets external send() bypass the pending queue before the server has
+      // reattached the session.
+      if (this._isReconnect && this.seqBySession.size > 0) {
+        this.doPost('reconnect', {
+          type: 'reconnect',
+          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          })),
+        }).finally(() => {
```
dimakis: 🟡 **bugs**: If `disconnect()` is called while the reconnect POST is in-flight, `.finally()` will still fire and set `_connected = true`, call `flushPendingSends()`, and emit `_open` — even though the connection has been intentionally torn down. `disconnect()` sets `_connected = false` and closes the EventSource, but the dangling `.finally()` callback will overwrite that state. Consider adding a guard (e.g., check `this.es !== null` or use an abort signal) inside the `.finally()` to bail out if the connection was closed during the POST. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +213,35 @@ export class SseConnection implements ChatConnection {
...
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      //
+      // _connected is deferred until reconnect completes — setting it earlier
+      // lets external send() bypass the pending queue before the server has
+      // reattached the session.
+      if (this._isReconnect && this.seqBySession.size > 0) {
```
dimakis, line 225: 🔵 **unsafe_assumptions**: If two welcome events arrive in rapid succession (e.g., EventSource auto-reconnect races with an explicit reconnect), two overlapping `doPost('reconnect', ...)` calls run concurrently, each with its own `.finally()`. Both will set `_connected = true` and call `flushPendingSends()` — the second flush is harmless (empty queue) but the double `_open` emission could confuse listeners. This is an edge case but worth noting since the old code (synchronous set) was immune to it. `[fixable]`

dimakis, line 225: 🔵 **regressions**: Contrast with the WS-based `connection.ts` (line 221): the WS path sets `_connected = true` synchronously on welcome, sends reconnect, then flushes. This PR intentionally diverges — SSE defers `_connected` until POST completes. This is correct for SSE (where reconnect is a separate HTTP round-trip), but the behavioral difference between the two transport implementations should be noted if they're expected to be interchangeable drop-in replacements. Not a bug, but a contract divergence.

dimakis (commented):
## Centaur Review

_Sound fix for the SSE reconnect race condition. The `.finally()` approach correctly ensures flush waits for the POST, but a disconnect-during-POST race can leave the connection in an incorrect `_connected = true` state — that's the one actionable bug. Test coverage should be strengthened to pin the deferred-state behavior that is the PR's core change._

- 🟡 **missing_tests**: The core behavior change — deferring `_connected` and `flushPendingSends` until the reconnect POST resolves — is not directly tested. There should be a test that verifies: (1) `isConnected()` returns false between welcome and reconnect POST completion, (2) `send()` queues messages during that window instead of posting directly, and (3) `_open` listener fires only after the POST resolves. The existing 'sends reconnect POST on welcome' test verifies the POST is issued but not the deferred state transition that is the PR's stated fix. `[fixable]`
- 🔵 **missing_tests**: No test covers the reconnect POST failure path. Since `doPost` swallows errors, `.finally()` always runs, but a test should verify that `_connected` still becomes true and `flushPendingSends` still fires even when the reconnect POST returns a non-ok response or the fetch rejects. This is the happy consequence of using `.finally()` over `.then()`, and it's worth pinning. `[fixable]`


_Sound fix for the SSE reconnect race condition. The `.finally()` approach correctly ensures flush waits for the POST, but a disconnect-during-POST race can leave the connection in an incorrect `_connected = true` state — that's the one actionable bug. Test coverage should be strengthened to pin the deferred-state behavior that is the PR's core change._

- 🟡 **bugs** (L234): If `disconnect()` is called while the reconnect POST is in-flight, `.finally()` will still fire and set `_connected = true`, call `flushPendingSends()`, and emit `_open` — even though the connection has been intentionally torn down. `disconnect()` sets `_connected = false` and closes the EventSource, but the dangling `.finally()` callback will overwrite that state. Consider adding a guard (e.g., check `this.es !== null` or use an abort signal) inside the `.finally()` to bail out if the connection was closed during the POST. `[fixable]`
- 🔵 **unsafe_assumptions** (L227): If two welcome events arrive in rapid succession (e.g., EventSource auto-reconnect races with an explicit reconnect), two overlapping `doPost('reconnect', ...)` calls run concurrently, each with its own `.finally()`. Both will set `_connected = true` and call `flushPendingSends()` — the second flush is harmless (empty queue) but the double `_open` emission could confuse listeners. This is an edge case but worth noting since the old code (synchronous set) was immune to it. `[fixable]`
- 🔵 **regressions** (L227): Contrast with the WS-based `connection.ts` (line 221): the WS path sets `_connected = true` synchronously on welcome, sends reconnect, then flushes. This PR intentionally diverges — SSE defers `_connected` until POST completes. This is correct for SSE (where reconnect is a separate HTTP round-trip), but the behavioral difference between the two transport implementations should be noted if they're expected to be interchangeable drop-in replacements. Not a bug, but a contract divergence.

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +213,39 @@ export class SseConnection implements ChatConnection {
...
+      // reattached the session.
+      const welcomeConnectionId = msg.connectionId as string;
+      if (this._isReconnect && this.seqBySession.size > 0) {
+        this.doPost('reconnect', {
+          type: 'reconnect',
+          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          })),
+        }).finally(() => {
```
dimakis: 🔵 **unsafe_assumptions**: .finally() marks the connection as live regardless of whether the reconnect POST succeeded or failed (doPost swallows all errors). If the POST fails (network blip, server 500), the server hasn't re-watched or re-attached sessions, but the client will immediately flush pending sends into the void. Consider using .then() and only setting _connected on success, or at minimum adding a test that documents this as intentional fire-and-forget behavior. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Well-designed fix that correctly moves reconnect data from URL params to POST and defers _connected with solid race-condition guards — the main gap is handling/documenting behavior when the reconnect POST itself fails._

- 🔵 **missing_tests**: No test for reconnect POST failure. Since doPost() swallows errors and .finally() always runs, a failed reconnect POST still sets _connected=true and flushes pending sends — even though the server never ran handleReconnect (no watch, no reattach, no replay). A test should verify the behavior when the fetch rejects or returns { ok: false }, so the chosen semantics are documented and intentional. `[fixable]`
- 🔵 **missing_tests**: No test for the EventSource auto-reconnect path (onerror fires while reconnect POST is in-flight). When onerror fires, _connected is already false so the handler is a no-op, but after auto-reconnect a new welcome arrives and triggers a second concurrent reconnect POST. The stale-welcome guard (connectionId check) should handle it, but this scenario is untested. `[fixable]`


_Well-designed fix that correctly moves reconnect data from URL params to POST and defers _connected with solid race-condition guards — the main gap is handling/documenting behavior when the reconnect POST itself fails._

- 🔵 **unsafe_assumptions** (L235): .finally() marks the connection as live regardless of whether the reconnect POST succeeded or failed (doPost swallows all errors). If the POST fails (network blip, server 500), the server hasn't re-watched or re-attached sessions, but the client will immediately flush pending sends into the void. Consider using .then() and only setting _connected on success, or at minimum adding a test that documents this as intentional fire-and-forget behavior. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +213,26 @@ export class SseConnection implements ChatConnection {
...
+      // Native EventSource auto-reconnect reuses the original URL (without
+      // ?sessions= param), so the server never runs handleReconnect — no
+      // watch, no reattach, no event replay. This POST ensures every
+      // reconnect (auto or explicit) triggers the full server-side
+      // reconnect flow: watch + reattach + event replay.
+      //
+      // _connected is deferred until reconnect completes — setting it earlier
+      // lets external send() bypass the pending queue before the server has
+      // reattached the session.
+      const welcomeConnectionId = msg.connectionId as string;
```
dimakis: 🔵 **style**: Redundant cast: `welcomeConnectionId` is assigned `msg.connectionId as string`, but `this._connectionId` was already set to the identical value on line 215. Consider `const welcomeConnectionId = this._connectionId;` to make it explicit this is a snapshot of the instance field for the staleness guard, not a re-parse of the message. `[fixable]`

dimakis (commented):
## Centaur Review

Found **4** issue(s).


_Well-designed fix that correctly moves reconnect from URL query params to POST, with thorough race-condition guards (disconnect, stale welcome, failure) and strong test coverage. No bugs found — suggestions are minor style/test improvements._

- 🔵 **style** (L227): Redundant cast: `welcomeConnectionId` is assigned `msg.connectionId as string`, but `this._connectionId` was already set to the identical value on line 215. Consider `const welcomeConnectionId = this._connectionId;` to make it explicit this is a snapshot of the instance field for the staleness guard, not a re-parse of the message. `[fixable]`
- 🔵 **regressions**: Server-side dead code: `server/index.ts:391-405` still parses `?sessions=` from the EventSource URL query params and calls `handleReconnect`. Since the client no longer sends this param, that code path is unreachable from SSE clients. Not a bug (harmless), but worth a follow-up cleanup to remove the dead path and avoid confusion about which reconnect path is canonical. `[fixable]`


_Well-designed fix that correctly moves reconnect from URL query params to POST, with thorough race-condition guards (disconnect, stale welcome, failure) and strong test coverage. No bugs found — suggestions are minor style/test improvements._

- 🔵 **missing_tests**: No test verifies that SSE `onmessage` events are still dispatched to the listener while the reconnect POST is in-flight (`_connected = false`). The `onmessage` handler is independent of `_connected`, so this works, but an explicit test would document this important invariant — replay events from `handleReconnect` arrive via SSE before the POST response. `[fixable]`
- 🔵 **missing_tests**: The 'stays disconnected when reconnect POST fails' test queues a send (`'should stay queued'`) but never verifies the queued message survives the failure and is flushed on a successful retry. The 'recovers after failed reconnect' test proves reconnection works but doesn't check pending sends. Adding an assertion that the queued send fires after recovery would close the loop. `[fixable]`

dimakis:
## Centaur Review

Found **3** issue(s).


_Well-implemented migration from query-param to POST-based reconnect with thorough race-condition guards and comprehensive test coverage; only minor suggestions around dead code and passive-only retry on failure._

- 🔵 **unsafe_assumptions** (L300): When the reconnect POST returns `!res.ok`, the client stays disconnected but the EventSource connection remains open and healthy. EventSource only auto-reconnects on transport errors — if the SSE stream stays alive (e.g. stable localhost), there's no automatic trigger for a retry. Recovery depends on a visibility change (checkAndReconnect) or an eventual ES transport error. Consider scheduling a delayed retry or forced reconnect on !res.ok so the client doesn't sit in limbo on stable desktop connections. `[fixable]`
- 🔵 **style** (L365): The `'reconnect'` → `'reconnect'` mapping in `messageTypeToEndpoint` is now dead code — reconnect is handled internally by `doReconnectPost`, never via `send()`. Keeping it means an external `send({ type: 'reconnect', ... })` call would bypass the staleness guards in `doReconnectPost`. Consider removing it to prevent misuse. `[fixable]`
- 🔵 **style** (L307): The `catch {}` block silently swallows all errors. While this matches `doPost`'s pattern, reconnect failures are higher stakes (entire session recovery depends on it). A `console.warn` would make debugging production reconnect issues significantly easier without changing behavior. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -253,6 +268,48 @@ export class SseConnection implements ChatConnection {
...
+      if (!this.es || this._connectionId !== welcomeConnectionId) return;
+
+      if (res.ok) {
+        this._connected = true;
+        this.flushPendingSends();
+        this.listener?.({ type: '_open' });
+      }
+      // On !res.ok: stay disconnected. EventSource auto-reconnect will
+      // trigger a new welcome and retry the reconnect POST.
+    } catch {
```
dimakis: 🔵 **unsafe_assumptions**: The `catch {}` in `doReconnectPost` silently swallows all errors with no logging or diagnostic output. While the recovery path (stay disconnected, let EventSource auto-reconnect) is correct, completely silent failure makes debugging production reconnect issues harder. Consider emitting a diagnostic event or at minimum logging to console.warn. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Solid fix for a real race condition — the staleness guards are well-designed and test coverage of the new async reconnect flow is thorough across normal, failure, and concurrent scenarios._

- 🔵 **missing_tests**: No test covers the path where `_isReconnect=true` but `seqBySession` is empty (e.g., all sessions cleared before reconnect). The code correctly falls through to the immediate-connect branch, but this behavior is untested and could regress. `[fixable]`
- 🔵 **style**: Two test names reference `.finally()` ('discards dangling .finally() if disconnect()…', 'discards stale .finally() when a newer welcome arrives') but the implementation uses an `if` guard after `await`, not `.finally()`. The names appear vestigial from an earlier design iteration and are mildly misleading. `[fixable]`


_Solid fix for a real race condition — the staleness guards are well-designed and test coverage of the new async reconnect flow is thorough across normal, failure, and concurrent scenarios._

- 🔵 **unsafe_assumptions** (L307): The `catch {}` in `doReconnectPost` silently swallows all errors with no logging or diagnostic output. While the recovery path (stay disconnected, let EventSource auto-reconnect) is correct, completely silent failure makes debugging production reconnect issues harder. Consider emitting a diagnostic event or at minimum logging to console.warn. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -253,6 +268,49 @@ export class SseConnection implements ChatConnection {
...
+          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          })),
+        }),
+      });
+
+      // Guard: bail if disconnect() was called or a newer welcome
+      // arrived while the POST was in-flight.
+      if (!this.es || this._connectionId !== welcomeConnectionId) return;
```
dimakis: 🟡 **bugs**: On `!res.ok` (or network error), the client stays `_connected=false` and the comment says "EventSource auto-reconnect will trigger a new welcome and retry." But EventSource only auto-reconnects on *connection* errors — if the SSE stream stays healthy while the reconnect POST returns 500, no auto-reconnect fires and the client is stuck in a disconnected state. Recovery still happens via browser lifecycle (`visibilitychange` triggers `checkAndReconnect(false)`, which proceeds when `_connected=false`), but it's user-action-dependent, not automatic. Consider forcing a reconnect on POST failure (`this.checkAndReconnect(true)`) or adding a short retry timer. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +213,29 @@ export class SseConnection implements ChatConnection {
         return;
       }
       this._connectionId = msg.connectionId as string;
-      this._connected = true;
-      this._isReconnect = true;
 
-      this.flushPendingSends();
-      this.listener?.({ type: '_open' });
+      // Always send reconnect on welcome if we have tracked sessions.
+      // Native EventSource auto-reconnect reuses the original URL (without
```
dimakis: 🔵 **style**: The 9-line block comment in the welcome handler is verbose for a codebase that defaults to no comments (per CLAUDE.md). The method name `doReconnectPost` and the staleness guard are self-documenting. Consider trimming to 2-3 lines covering only the non-obvious insight: why `_connected` is deferred and why `welcomeConnectionId` is snapshotted. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,411 @@ describe('SseConnection', () => {
...
+    listener.mockClear();
+
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+    await vi.runAllTimersAsync();
+
+    expect(conn.isConnected()).toBe(false);
+    expect(listener).not.toHaveBeenCalledWith({ type: '_open' });
+  });
+
+  it('recovers after failed reconnect when EventSource auto-reconnects', async () => {
```
dimakis, line 530: 🔵 **missing_tests**: The 'recovers after failed reconnect when EventSource auto-reconnects' test simulates recovery by manually emitting a new welcome on the same EventSource — which models the EventSource-error-and-reconnect path. There's no test for the more likely recovery path: POST fails, SSE stays up, then `checkAndReconnect(false)` is called (e.g. via `visibilitychange`) while `_connected=false`. This is the actual recovery mechanism when the SSE stream doesn't drop. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid fix for a real reliability bug — the POST-on-welcome approach with staleness guards is well-designed and thoroughly tested. One edge case: if the reconnect POST fails but the SSE stream stays healthy, the client stays disconnected with no automatic retry (recovery requires a browser lifecycle event)._

- 🟡 **bugs** (L298): On `!res.ok` (or network error), the client stays `_connected=false` and the comment says "EventSource auto-reconnect will trigger a new welcome and retry." But EventSource only auto-reconnects on *connection* errors — if the SSE stream stays healthy while the reconnect POST returns 500, no auto-reconnect fires and the client is stuck in a disconnected state. Recovery still happens via browser lifecycle (`visibilitychange` triggers `checkAndReconnect(false)`, which proceeds when `_connected=false`), but it's user-action-dependent, not automatic. Consider forcing a reconnect on POST failure (`this.checkAndReconnect(true)`) or adding a short retry timer. `[fixable]`
- 🔵 **style** (L218): The 9-line block comment in the welcome handler is verbose for a codebase that defaults to no comments (per CLAUDE.md). The method name `doReconnectPost` and the staleness guard are self-documenting. Consider trimming to 2-3 lines covering only the non-obvious insight: why `_connected` is deferred and why `welcomeConnectionId` is snapshotted. `[fixable]`


_Solid fix for a real reliability bug — the POST-on-welcome approach with staleness guards is well-designed and thoroughly tested. One edge case: if the reconnect POST fails but the SSE stream stays healthy, the client stays disconnected with no automatic retry (recovery requires a browser lifecycle event)._

- 🔵 **missing_tests** (L530): The 'recovers after failed reconnect when EventSource auto-reconnects' test simulates recovery by manually emitting a new welcome on the same EventSource — which models the EventSource-error-and-reconnect path. There's no test for the more likely recovery path: POST fails, SSE stays up, then `checkAndReconnect(false)` is called (e.g. via `visibilitychange`) while `_connected=false`. This is the actual recovery mechanism when the SSE stream doesn't drop. `[fixable]`

dimakis:
## Centaur Review

_Well-designed fix with excellent test coverage. The migration from query-param reconnect to POST-based reconnect correctly solves the EventSource URL-reuse bug. Two minor concerns: no backoff on persistent POST failures (could create a tight retry loop), and spurious _close events when the connection was never opened._

- 🟡 **unsafe_assumptions** (L296): No backoff on reconnect POST failure retry. When `doReconnectPost` fails (HTTP error or network error), it immediately calls `checkAndReconnect(true)`, which tears down the EventSource and creates a new one. If `/api/chat/reconnect` is persistently returning 500, this produces a tight loop: connect → welcome → POST → 500 → close → reconnect → welcome → POST → 500 → ... The `reconnectDelayMs` config exists but is never applied here. Natural EventSource connection latency provides some rate limiting, but there is no exponential backoff or max-retry cap. `[fixable]`
- 🔵 **style** (L184): `checkAndReconnect(true)` unconditionally emits `{ type: '_close' }` to the listener even when `_connected` was never `true`. On the `doReconnectPost` failure path (lines 296, 301), this produces a spurious `_close` for a connection that was never opened. The listener likely tolerates this, but gating with `if (this._connected)` (mirroring the guard in `es.onerror` at line 251) would be more precise. `[fixable]`

dimakis (commented):
## Centaur Review

_Sound architectural shift from query-param to POST-based reconnect with excellent test coverage. One race condition remains: a stale `doReconnectPost` can set `_connected = true` if `checkAndReconnect` fires mid-flight, fixable with a defensive `_connected = false` in the welcome handler's reconnect branch._

- 🟡 **bugs** (L224): Race condition: if `checkAndReconnect()` fires while `doReconnectPost(A)` is in-flight, it closes ES1 and creates ES2 via `doConnect()`. When the stale POST resolves, the guard `this._connectionId !== welcomeConnectionId` passes because `_connectionId` is still 'A' (ES2's welcome hasn't arrived yet), and `!this.es` is false (ES2 exists). This sets `_connected = true` prematurely. When ES2's welcome arrives, the reconnect branch does NOT reset `_connected` to false, so `send()` bypasses the queue during `doReconnectPost(B)` — the exact race this PR aims to prevent. Fix: add `this._connected = false;` before `this.doReconnectPost(welcomeConnectionId)` as a defensive reset. `[fixable]`


_Sound architectural shift from query-param to POST-based reconnect with excellent test coverage. One race condition remains: a stale `doReconnectPost` can set `_connected = true` if `checkAndReconnect` fires mid-flight, fixable with a defensive `_connected = false` in the welcome handler's reconnect branch._

- 🔵 **missing_tests**: No test for `checkAndReconnect` called while `doReconnectPost` is in-flight (the race condition above). A test scenario: start doReconnectPost with a pending promise, call `checkAndReconnect(true)` before resolving it, then verify the stale POST's resolution does NOT set `_connected = true`. `[fixable]`
- 🔵 **missing_tests**: No test for the `wasConnected` guard in `checkAndReconnect` — i.e. verifying that `_close` is NOT emitted when `checkAndReconnect` is called while already disconnected (e.g. double-reconnect without intermediate connection). `[fixable]`
- 🔵 **style** (L498): Test name says 'recovers via checkAndReconnect(false)' but the test body calls `conn.checkAndReconnect(true)`. The recovery actually happens via `scheduleReconnect` → `doConnect()`, not `checkAndReconnect(false)`. Consider renaming to 'recovers via scheduleReconnect after repeated POST failures'. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -216,11 +216,19 @@ export class SseConnection implements ChatConnection {
...
-      this._connected = true;
-      this._isReconnect = true;
 
-      this.flushPendingSends();
-      this.listener?.({ type: '_open' });
+      // _connected deferred until doReconnectPost succeeds — prevents
+      // external send() from bypassing the pending queue mid-reconnect.
+      // welcomeConnectionId is a staleness guard for concurrent welcomes.
+      const welcomeConnectionId = this._connectionId;
+      if (this._isReconnect && this.seqBySession.size > 0) {
```
dimakis, line 225: 🟡 **bugs**: Race condition: if `checkAndReconnect()` fires while `doReconnectPost(A)` is in-flight, it closes ES1 and creates ES2 via `doConnect()`. When the stale POST resolves, the guard `this._connectionId !== welcomeConnectionId` passes because `_connectionId` is still 'A' (ES2's welcome hasn't arrived yet), and `!this.es` is false (ES2 exists). This sets `_connected = true` prematurely. When ES2's welcome arrives, the reconnect branch does NOT reset `_connected` to false, so `send()` bypasses the queue during `doReconnectPost(B)` — the exact race this PR aims to prevent. Fix: add `this._connected = false;` before `this.doReconnectPost(welcomeConnectionId)` as a defensive reset. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,478 @@ describe('SseConnection', () => {
...
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-abc' });
+    conn.trackSeq('sess-1', 10);
+
+    // Force reconnect
+    conn.checkAndReconnect(true);
+    conn.send({ type: 'send', prompt: 'should stay queued', clientMsgId: 'q-1' });
+    listener.mockClear();
+
+    // New welcome — reconnect POST will fail
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
```
dimakis, line 498: 🔵 **style**: Test name says 'recovers via checkAndReconnect(false)' but the test body calls `conn.checkAndReconnect(true)`. The recovery actually happens via `scheduleReconnect` → `doConnect()`, not `checkAndReconnect(false)`. Consider renaming to 'recovers via scheduleReconnect after repeated POST failures'. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -290,20 +290,523 @@ describe('SseConnection', () => {
...
+
+    // New welcome — reconnect POST in-flight
+    lastES()._emit('welcome', { type: 'welcome', protocolVersion: 2, connectionId: 'conn-def' });
+
+    // Disconnect while POST is in-flight
+    conn.disconnect();
+    expect(conn.isConnected()).toBe(false);
+    listener.mockClear();
+
+    // Resolve the reconnect POST — .finally() should bail out
```
dimakis: 🔵 **style**: Test comment says '// Resolve the reconnect POST — .finally() should bail out' but the implementation uses try/catch with staleness guards, not .finally(). The comment is misleading — consider rewording to 'staleness guard should bail out'. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


_Solid fix for a real race condition — the old query-param reconnect was unreliable because EventSource auto-reconnect reuses the original URL. The new POST-based approach with deferred _connected, staleness guards, and scheduleReconnect is well-designed and thoroughly tested across 16 new test cases covering all edge cases (stale POSTs, failures, disconnect mid-flight, rapid welcome races)._

- 🔵 **style** (L387): Test comment says '// Resolve the reconnect POST — .finally() should bail out' but the implementation uses try/catch with staleness guards, not .finally(). The comment is misleading — consider rewording to 'staleness guard should bail out'. `[fixable]`

PR #397, opened 2026-06-25
fix(worktree): skip rescue for noise-only worktrees

dimakis:
## Summary

- Worktree rescue was creating draft PRs for sessions that had no meaningful work — just the `.mitzo-session` marker file. This caused 15+ empty "Rescued" PRs to pile up in mgmt.
- Added a noise-file check after staging: if all staged files are in `RESCUE_NOISE_FILES` (currently just `.mitzo-session`), the rescue bails out early — no commit, no push, no PR.
- The set is extensible if other noise markers appear in the future.

## Test plan

- [x] New test: skips rescue when only `.mitzo-session` is staged
- [x] New test: skips rescue when no files are staged
- [x] New test: proceeds with rescue when meaningful files exist alongside noise
- [x] All 15 existing + new rescue-worktree tests pass

### DIFF
--- a/server/__tests__/rescue-worktree.test.ts
+++ b/server/__tests__/rescue-worktree.test.ts
@@ -107,6 +107,8 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add -u
     mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → meaningful file
+    mockExecFileSync.mockReturnValueOnce('src/feature.ts\n' as never);
     // Step 2: git commit
     mockExecFileSync.mockReturnValueOnce('' as never);
     // Step 3: git push
@@ -186,6 +188,8 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add -u succeeds
     mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → meaningful file
+    mockExecFileSync.mockReturnValueOnce('src/feature.ts\n' as never);
     // Step 2: git commit fails
     mockExecFileSync.mockImplementationOnce(() => {
       throw new Error('nothing to commit');
@@ -202,6 +206,8 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add -u
     mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → meaningful file
+    mockExecFileSync.mockReturnValueOnce('src/feature.ts\n' as never);
     // Step 2: git commit
     mockExecFileSync.mockReturnValueOnce('' as never);
     // Step 3: git push fails
@@ -220,6 +226,8 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add
     mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → meaningful file
+    mockExecFileSync.mockReturnValueOnce('src/feature.ts\n' as never);
     // Step 2: git commit
     mockExecFileSync.mockReturnValueOnce('' as never);
     // Step 3: git push
@@ -252,14 +260,68 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add -u succeeds (but stages nothing — only untracked files)
     mockExecFileSync.mockReturnValueOnce('' as never);
-    // Step 2: git commit fails — nothing staged
-    mockExecFileSync.mockImplementationOnce(() => {
-      throw new Error('nothing to commit, working tree clean');
-    });
+    // Step 1b: git diff --cached --name-only → empty (nothing staged)
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git reset HEAD (unstage)
+    mockExecFileSync.mockReturnValueOnce('' as never);
 
     const result = rescueDirtyWorktree('/tmp/worktree', 'session/test-123', 'test-123');
 
     expect(result.success).toBe(false);
-    expect(result.error).toContain('nothing to commit');
+    expect(result.error).toContain('no meaningful changes');
+  });
+
+  it('skips rescue when only noise files (.mitzo-session) are staged', () => {
+    // Step 0: getRepoRemote
+    mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
+    // Step 1: git add -A
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → only .mitzo-session
+    mockExecFileSync.mockReturnValueOnce('.mitzo-session\n' as never);
+    // Step 1b: git reset HEAD (unstage)
+    mockExecFileSync.mockReturnValueOnce('' as never);
+
+    const result = rescueDirtyWorktree('/tmp/worktree', 'session/test-123', 'test-123');
+
+    expect(result.success).toBe(false);
+    expect(result.error).toContain('no meaningful changes');
+    // Should NOT have called commit, push, or pr create
+    expect(mockExecFileSync).toHaveBeenCalledTimes(4); // remote + add + diff + reset
+  });
+
+  it('proceeds with rescue when meaningful files are staged alongside noise files', () => {
+    // Step 0: getRepoRemote
+    mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
+    // Step 1: git add -A
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → noise + real file
+    mockExecFileSync.mockReturnValueOnce('.mitzo-session\nsrc/feature.ts\n' as never);
+    // Step 2: git commit
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 3: git push
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 4: gh pr create
+    mockExecFileSync.mockReturnValueOnce('https://github.com/dimakis/mgmt/pull/99\n' as never);
+
+    const result = rescueDirtyWorktree('/tmp/worktree', 'session/test-123', 'test-123');
+
+    expect(result.success).toBe(true);
+    expect(result.prUrl).toBe('https://github.com/dimakis/mgmt/pull/99');
+  });
+
+  it('skips rescue when no files are staged', () => {
+    // Step 0: getRepoRemote
+    mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
+    // Step 1: git add -A
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → empty
+    mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git reset HEAD (unstage)
+    mockExecFileSync.mockReturnValueOnce('' as never);
+
+    const result = rescueDirtyWorktree('/tmp/worktree', 'session/test-123', 'test-123');
+
+    expect(result.success).toBe(false);
+    expect(result.error).toContain('no meaningful changes');
   });
 });
--- a/server/worktree.ts
+++ b/server/worktree.ts
@@ -29,6 +29,13 @@ import { createLogger } from './logger.js';
 const log = createLogger('worktree');
 
 /**
+ * Files that are not meaningful work — rescue should skip if these are the only changes.
+ * Matched via exact path from `git diff --cached --name-only` (root-relative).
+ * Only root-level files are supported; subdirectory paths would need `subdir/file` entries.
+ */
+const RESCUE_NOISE_FILES = new Set(['.mitzo-session']);
+
+/**
  * Detect the default branch of a repo (e.g. 'main' or 'master').
  * Prefers origin/HEAD (remote truth) since session worktrees should branch
  * from the canonical default, not whatever is locally checked out.
@@ -461,6 +468,28 @@ export function rescueDirtyWorktree(
     execFileSync('git', ['-C', worktreePath, 'add', '-u'], gitOpts);
   } catch (err: unknown) {
     return { success: false, error: err instanceof Error ? err.message : String(err) };
+  }
+
+  // 1b. Check if staged changes contain meaningful work (not just noise markers)
+  try {
+    const staged = execFileSync('git', ['-C', worktreePath, 'diff', '--cached', '--name-only'], {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: WORKTREE_GIT_TIMEOUT_MS,
+    }).trim();
+    const files = staged ? staged.split('\n') : [];
+    if (files.length === 0 || files.every((f) => RESCUE_NOISE_FILES.has(f))) {
+      log.info('rescue skipped — only noise files changed', { wtId, files });
+      // Unstage so the worktree can be cleaned up without leaving staged changes
+      try {
+        execFileSync('git', ['-C', worktreePath, 'reset', 'HEAD'], gitOpts);
+      } catch {
+        // Non-fatal — worktree is about to be removed anyway
+      }
+      return { success: false, error: 'no meaningful changes to rescue' };
+    }
+  } catch {
+    // If we can't inspect staged files, proceed with the rescue anyway — safer than skipping
   }
 
   try {


dimakis:
## Centaur Review

_Solid multi-pronged fix for replay storms and message loss. The worktree noise-file filtering and shouldSync/replay-cap changes are clean and well-tested. The SSE POST retry mechanism works but should guard against permanent HTTP failures to prevent infinite re-queue cycles._

- 🟡 **unsafe_assumptions** (L336): Re-queues on ALL non-ok responses (`!res.ok`), including permanent failures like 400 Bad Request or 403 Forbidden. These will never succeed but will cycle through `pendingSends` on every reconnect indefinitely. Consider limiting re-queue to transient statuses (404, 502, 503) or adding a max-retry counter to avoid infinite retry loops for permanently rejected messages. `[fixable]`
- 🔵 **unsafe_assumptions** (L352): `flushPendingSends` calls `doPost` without `await`. If a flushed POST fails and a new reconnect completes before the failed `doPost`'s catch handler runs, the re-queued message won't be included in the next flush (it pushes to `pendingSends` after the new flush already cleared and iterated). This is a narrow timing window but could cause a message to sit in `pendingSends` until a third reconnect. `[fixable]`


_Solid multi-pronged fix for replay storms and message loss. The worktree noise-file filtering and shouldSync/replay-cap changes are clean and well-tested. The SSE POST retry mechanism works but should guard against permanent HTTP failures to prevent infinite re-queue cycles._

- 🔵 **missing_tests** (L617): New `getHeadSeq` method has no unit test in the protocol package's event-store test suite. It is only mocked in the connection-registry tests. A dedicated test covering the three cases (session with events, empty session, non-existent session) would be valuable given this is a SQL query backing a replay-storm prevention mechanism. `[fixable]`


### `server/index.ts`
_Solid multi-pronged fix for replay storms and message loss. The worktree noise-file filtering and shouldSync/replay-cap changes are clean and well-tested. The SSE POST retry mechanism works but should guard against permanent HTTP failures to prevent infinite re-queue cycles._

- 🔵 **style** (L375): Comment says SSE is the 'Default transport' and WS is 'opt-in via localStorage', but this is a comment-only change in a PR titled 'fix(worktree): skip rescue for noise-only worktrees'. Consider splitting transport documentation updates into their own commit/PR to keep the changeset focused.

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_Solid defensive PR addressing replay storms and noise worktrees. Two substantive issues: `basename()` in noise-file check is overly broad (should be strict equality), and silently capping replay depth without a client-side gap signal may cause subtle data loss on reconnect._

- 🟡 **bugs** (L470): `basename(f)` matches `.mitzo-session` at any directory depth (e.g. `subdir/.mitzo-session`). Since `.mitzo-session` is only created at the worktree root, a simple equality check `f === '.mitzo-session'` would be more precise and avoid false positives if a file with the same name ever appears in a subdirectory. `[fixable]`


_Solid defensive PR addressing replay storms and noise worktrees. Two substantive issues: `basename()` in noise-file check is overly broad (should be strict equality), and silently capping replay depth without a client-side gap signal may cause subtle data loss on reconnect._

- 🟡 **unsafe_assumptions** (L248): When the replay gap is capped, events between the old and new cursor are silently dropped with no client-side notification. The comment says clients can lazy-load via REST API, but there is no gap signal (e.g. `events_skipped` message) sent to the client, so the client has no way to know it missed events unless it independently tracks seq continuity. Consider sending a gap notification so the client can trigger a full restore. `[fixable]`


### `server/chat.ts`
_Solid defensive PR addressing replay storms and noise worktrees. Two substantive issues: `basename()` in noise-file check is overly broad (should be strict equality), and silently capping replay depth without a client-side gap signal may cause subtle data loss on reconnect._

- 🔵 **missing_tests** (L1269): The new try/catch around `queryInstance.interrupt()` has no test coverage. A test where `interrupt` rejects (e.g. `vi.fn().mockRejectedValue(new Error('not ready'))`) should verify that the function still returns true and the prompt is still pushed to `inputQueue`. `[fixable]`


_Solid defensive PR addressing replay storms and noise worktrees. Two substantive issues: `basename()` in noise-file check is overly broad (should be strict equality), and silently capping replay depth without a client-side gap signal may cause subtle data loss on reconnect._

- 🔵 **missing_tests** (L507): The 'caps replay depth' test doesn't verify the log.warn call or that the cursor is actually persisted (subsequent sync rounds should start from the new cursor, not re-cap). A follow-up assertion after a second timer advance would confirm the cursor update sticks. `[fixable]`


### `server/index.ts`
_Solid defensive PR addressing replay storms and noise worktrees. Two substantive issues: `basename()` in noise-file check is overly broad (should be strict equality), and silently capping replay depth without a client-side gap signal may cause subtle data loss on reconnect._

- 🔵 **style** (L372): The comment update ('SSE Chat Transport (primary)' and 'Default transport') is documentation about the transport migration status — useful context but may go stale. Minor nit, no action needed.

server/worktree.ts
```diff
@@ -456,6 +459,28 @@ export function rescueDirtyWorktree(
...
 
+  // 1b. Check if staged changes contain meaningful work (not just noise markers)
+  try {
+    const staged = execFileSync('git', ['-C', worktreePath, 'diff', '--cached', '--name-only'], {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: WORKTREE_GIT_TIMEOUT_MS,
+    }).trim();
+    const files = staged ? staged.split('\n') : [];
+    if (files.length === 0 || files.every((f) => RESCUE_NOISE_FILES.has(basename(f)))) {
```
dimakis: 🟡 **bugs**: `basename(f)` matches `.mitzo-session` at any directory depth (e.g. `subdir/.mitzo-session`). Since `.mitzo-session` is only created at the worktree root, a simple equality check `f === '.mitzo-session'` would be more precise and avoid false positives if a file with the same name ever appears in a subdirectory. `[fixable]`

packages/harness/src/connection-registry.ts
```diff
@@ -221,12 +228,31 @@ export class ConnectionRegistry {
...
+          // Prevents replay storms after server restart or long disconnects.
+          if (this.eventStore.getHeadSeq) {
+            const head = this.eventStore.getHeadSeq(sessionId);
+            if (head - cursor > MAX_REPLAY_GAP) {
+              const newCursor = head - MAX_REPLAY_GAP;
+              log.warn('periodic sync: cursor too far behind, skipping ahead', {
+                connectionId,
+                sessionId,
+                oldCursor: cursor,
+                newCursor,
```
dimakis: 🟡 **unsafe_assumptions**: When the replay gap is capped, events between the old and new cursor are silently dropped with no client-side notification. The comment says clients can lazy-load via REST API, but there is no gap signal (e.g. `events_skipped` message) sent to the client, so the client has no way to know it missed events unless it independently tracks seq continuity. Consider sending a gap notification so the client can trigger a full restore. `[fixable]`

server/chat.ts
```diff
@@ -1266,7 +1266,17 @@ export async function interruptChat(
       );
       await Promise.allSettled(stops);
     }
-    await session.queryInstance.interrupt();
+    try {
```
dimakis: 🔵 **missing_tests**: The new try/catch around `queryInstance.interrupt()` has no test coverage. A test where `interrupt` rejects (e.g. `vi.fn().mockRejectedValue(new Error('not ready'))`) should verify that the function still returns true and the prompt is still pushed to `inputQueue`. `[fixable]`

packages/harness/__tests__/connection-registry.test.ts
```diff
@@ -504,6 +507,56 @@ describe('ConnectionRegistry', () => {
```
dimakis, line 504: 🔵 **missing_tests**: The 'caps replay depth' test doesn't verify the log.warn call or that the cursor is actually persisted (subsequent sync rounds should start from the new cursor, not re-cap). A follow-up assertion after a second timer advance would confirm the cursor update sticks. `[fixable]`

server/index.ts
```diff
@@ -362,9 +369,10 @@ const v2Ctx: V2HandlerContext = {
   nativeCommands,
 };
 
-// ─── SSE Chat Transport ──────────────────────────────────────────────────────
+// ─── SSE Chat Transport (primary) ────────────────────────────────────────────
```
dimakis: 🔵 **style**: The comment update ('SSE Chat Transport (primary)' and 'Default transport') is documentation about the transport migration status — useful context but may go stale. Minor nit, no action needed.

packages/client/src/sse-connection.ts
```diff
@@ -325,17 +325,27 @@ export class SseConnection implements ChatConnection {
...
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
           'X-Connection-ID': this._connectionId,
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok && endpoint === 'send') {
+        // Message delivery failed (e.g. 404 from stale connectionId).
+        // Re-queue so it flushes on the next successful reconnect.
```
dimakis: 🟡 **bugs**: Re-queuing in `doPost` bypasses the `MAX_PENDING_SENDS` cap that guards `send()`. If a `send` POST fails persistently (e.g. 400 from malformed payload), it will be re-queued on every reconnect indefinitely with no retry limit, attempt counter, or backoff. The queue won't grow unbounded within a single flush (since `flushPendingSends` swaps the array), but across reconnect cycles a poisoned message retries forever. `[fixable]`

packages/client/src/__tests__/sse-connection.test.ts
```diff
@@ -872,4 +872,103 @@ describe('SseConnection', () => {
 
     expect(mockFetch).not.toHaveBeenCalled();
   });
+
+  // ─── POST retry on failure ───────────────────────────────────────────────
+
+  it('re-queues send POST on HTTP error (e.g. 404 from stale connectionId)', async () => {
```
dimakis: 🔵 **missing_tests**: No test covers the case where a re-queued `send` fails again on the next reconnect cycle. This is the scenario that would expose the unbounded retry issue — a message that always returns 400 should eventually be discarded or capped. `[fixable]`

server/chat.ts
dimakis: 🔵 **missing_tests**: The new try/catch around `queryInstance.interrupt()` has no test covering the error path. The existing interrupt tests in `send-to-chat.test.ts` mock `interrupt()` to succeed. A test where `interrupt()` rejects would verify the error is caught and logged without crashing the session. `[fixable]`

server/index.ts
```diff
@@ -99,11 +99,18 @@ const connRegistry = new ConnectionRegistry();
...
+// getHeadSeq caps replay depth so post-crash reconnects don't flood clients.
 connRegistry.setEventStore({
   getEventsAfter: (sessionId, afterSeq, limit) =>
     eventStore.getEventsAfter(sessionId, afterSeq, limit),
-  isSessionActive: (sessionId) => eventStore.getSession(sessionId)?.isActive ?? false,
+  shouldSync: (sessionId) => {
+    const state = eventStore.getSessionState(sessionId);
+    // Only sync sessions that are actively running or starting.
+    // ENDED/SUSPENDED/DETACHED/CLOSING don't need delivery retries.
+    return state === 'ACTIVE' || state === 'STARTING' || state === 'CREATED';
```
dimakis: 🔵 **style**: The `shouldSync` allowlist (`ACTIVE || STARTING || CREATED`) is correct but fragile — adding a new syncable state requires remembering to update this check. Consider a denylist (`state !== 'ENDED' && state !== 'SUSPENDED' && ...`) that matches the comment, or import the states as a const set, so new states default to synced rather than silently dropped. `[fixable]`

server/worktree.ts
```diff
@@ -456,6 +459,28 @@ export function rescueDirtyWorktree(
     return { success: false, error: err instanceof Error ? err.message : String(err) };
   }
 
+  // 1b. Check if staged changes contain meaningful work (not just noise markers)
```
dimakis, line 473: 🔵 **unsafe_assumptions**: The `RESCUE_NOISE_FILES` set only contains `.mitzo-session`, which is correct since that file is only created at the worktree root. But the set is declared at module scope — if future noise files are added at subdirectory paths, `git diff --cached --name-only` would return them as `subdir/file` and exact-match would miss them. A comment noting the root-only assumption would help future maintainers. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid defensive fixes for replay storms and noise-only worktree rescue. Main concern: the SSE `send` POST re-queue has no retry limit or backoff, risking infinite retries for persistently failing messages._

- 🟡 **bugs** (L338): Re-queuing in `doPost` bypasses the `MAX_PENDING_SENDS` cap that guards `send()`. If a `send` POST fails persistently (e.g. 400 from malformed payload), it will be re-queued on every reconnect indefinitely with no retry limit, attempt counter, or backoff. The queue won't grow unbounded within a single flush (since `flushPendingSends` swaps the array), but across reconnect cycles a poisoned message retries forever. `[fixable]`


_Solid defensive fixes for replay storms and noise-only worktree rescue. Main concern: the SSE `send` POST re-queue has no retry limit or backoff, risking infinite retries for persistently failing messages._

- 🔵 **missing_tests** (L878): No test covers the case where a re-queued `send` fails again on the next reconnect cycle. This is the scenario that would expose the unbounded retry issue — a message that always returns 400 should eventually be discarded or capped. `[fixable]`


### `server/chat.ts`
_Solid defensive fixes for replay storms and noise-only worktree rescue. Main concern: the SSE `send` POST re-queue has no retry limit or backoff, risking infinite retries for persistently failing messages._

- 🔵 **missing_tests** (L1269): The new try/catch around `queryInstance.interrupt()` has no test covering the error path. The existing interrupt tests in `send-to-chat.test.ts` mock `interrupt()` to succeed. A test where `interrupt()` rejects would verify the error is caught and logged without crashing the session. `[fixable]`


### `server/index.ts`
_Solid defensive fixes for replay storms and noise-only worktree rescue. Main concern: the SSE `send` POST re-queue has no retry limit or backoff, risking infinite retries for persistently failing messages._

- 🔵 **style** (L111): The `shouldSync` allowlist (`ACTIVE || STARTING || CREATED`) is correct but fragile — adding a new syncable state requires remembering to update this check. Consider a denylist (`state !== 'ENDED' && state !== 'SUSPENDED' && ...`) that matches the comment, or import the states as a const set, so new states default to synced rather than silently dropped. `[fixable]`


### `server/worktree.ts`
_Solid defensive fixes for replay storms and noise-only worktree rescue. Main concern: the SSE `send` POST re-queue has no retry limit or backoff, risking infinite retries for persistently failing messages._

- 🔵 **unsafe_assumptions** (L462): The `RESCUE_NOISE_FILES` set only contains `.mitzo-session`, which is correct since that file is only created at the worktree root. But the set is declared at module scope — if future noise files are added at subdirectory paths, `git diff --cached --name-only` would return them as `subdir/file` and exact-match would miss them. A comment noting the root-only assumption would help future maintainers. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_The core transport fixes (SSE retry, replay cap, interrupt guard) are solid and well-tested. The main issue is that noise-only worktrees become permanently stuck in cleanup — `cleanupStaleWorktrees` needs to handle the 'no meaningful changes' rescue result by removing the worktree instead of skipping it._

- 🟡 **bugs** (L470): Noise-only worktrees become permanently stuck. `cleanupStaleWorktrees` calls `hasUncommittedWork` (which reports dirty for `.mitzo-session`), then `rescueDirtyWorktree` returns `{ success: false, error: 'no meaningful changes to rescue' }`. The cleanup treats this identically to a real rescue failure — it skips the worktree and posts to inbox. On subsequent cycles, `alreadyNotified` is true so rescue is never retried, and the worktree is never removed. Pre-PR, these worktrees were rescued (commit+push+PR); post-PR, they accumulate on disk forever. Fix: `cleanupStaleWorktrees` should detect the 'no meaningful changes' case and remove the worktree directly, since there's nothing worth saving. `[fixable]`


_The core transport fixes (SSE retry, replay cap, interrupt guard) are solid and well-tested. The main issue is that noise-only worktrees become permanently stuck in cleanup — `cleanupStaleWorktrees` needs to handle the 'no meaningful changes' rescue result by removing the worktree instead of skipping it._

- 🟡 **missing_tests** (L535): The 'caps replay depth when cursor is far behind head' test (line 511) is missing `registry.stopPeriodicSync()` at the end, unlike every other periodic sync test in this describe block. This leaks a `setInterval` timer that could cause flaky test behavior if timer state bleeds between tests. `[fixable]`


_The core transport fixes (SSE retry, replay cap, interrupt guard) are solid and well-tested. The main issue is that noise-only worktrees become permanently stuck in cleanup — `cleanupStaleWorktrees` needs to handle the 'no meaningful changes' rescue result by removing the worktree instead of skipping it._

- 🔵 **missing_tests** (L618): The new `getHeadSeq` method has no direct unit test in `packages/protocol/__tests__/event-store.test.ts`. It's only indirectly tested via mocked `getHeadSeq` in connection-registry tests. A bug in the SQL query (`SELECT MAX(seq)`) or null handling would not be caught. Add a test covering: session with events returns max seq, session with no events returns 0, unknown session returns 0. `[fixable]`


### `server/index.ts`
_The core transport fixes (SSE retry, replay cap, interrupt guard) are solid and well-tested. The main issue is that noise-only worktrees become permanently stuck in cleanup — `cleanupStaleWorktrees` needs to handle the 'no meaningful changes' rescue result by removing the worktree instead of skipping it._

- 🔵 **style** (L107): The `shouldSync` callback returns `false` when `getSessionState` returns `null` (session not in DB). This is likely correct, but worth a brief comment since the implicit null→false behavior is non-obvious and could silently skip sync for sessions whose state record was lost (e.g. after DB corruption). `[fixable]`

server/worktree.ts
```diff
@@ -456,6 +459,28 @@ export function rescueDirtyWorktree(
...
+  try {
+    const staged = execFileSync('git', ['-C', worktreePath, 'diff', '--cached', '--name-only'], {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: WORKTREE_GIT_TIMEOUT_MS,
+    }).trim();
+    const files = staged ? staged.split('\n') : [];
+    if (files.length === 0 || files.every((f) => RESCUE_NOISE_FILES.has(f))) {
```
dimakis, line 481: 🟡 **bugs**: Noise-only worktrees become permanently stuck. `cleanupStaleWorktrees` calls `hasUncommittedWork` (which reports dirty for `.mitzo-session`), then `rescueDirtyWorktree` returns `{ success: false, error: 'no meaningful changes to rescue' }`. The cleanup treats this identically to a real rescue failure — it skips the worktree and posts to inbox. On subsequent cycles, `alreadyNotified` is true so rescue is never retried, and the worktree is never removed. Pre-PR, these worktrees were rescued (commit+push+PR); post-PR, they accumulate on disk forever. Fix: `cleanupStaleWorktrees` should detect the 'no meaningful changes' case and remove the worktree directly, since there's nothing worth saving. `[fixable]`

packages/harness/__tests__/connection-registry.test.ts
```diff
@@ -504,6 +507,56 @@ describe('ConnectionRegistry', () => {
...
+      registry.watch('conn-1', 'sess-a');
+      // Cursor at 0 — gap of 10000 >> MAX_REPLAY_GAP (200)
+      registry.startPeriodicSync();
+
+      await vi.advanceTimersByTimeAsync(5000);
+
+      // Should have fetched from 9800 (head - 200), not from 0
+      const calls = (store.getEventsAfter as ReturnType<typeof vi.fn>).mock.calls;
+      expect(calls[0][1]).toBe(9800); // afterSeq should be head - MAX_REPLAY_GAP
+    });
```
dimakis: 🟡 **missing_tests**: The 'caps replay depth when cursor is far behind head' test (line 511) is missing `registry.stopPeriodicSync()` at the end, unlike every other periodic sync test in this describe block. This leaks a `setInterval` timer that could cause flaky test behavior if timer state bleeds between tests. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +614,12 @@ export class EventStore {
     return (row?.state as SessionState) ?? null;
   }
 
+  /** Return the highest seq number for a session, or 0 if no events exist. */
+  getHeadSeq(sessionId: string): number {
```
dimakis: 🔵 **missing_tests**: The new `getHeadSeq` method has no direct unit test in `packages/protocol/__tests__/event-store.test.ts`. It's only indirectly tested via mocked `getHeadSeq` in connection-registry tests. A bug in the SQL query (`SELECT MAX(seq)`) or null handling would not be caught. Add a test covering: session with events returns max seq, session with no events returns 0, unknown session returns 0. `[fixable]`

server/index.ts
```diff
@@ -99,11 +99,18 @@ const connRegistry = new ConnectionRegistry();
...
 
 // Wire up EventStore for periodic sync (enables delivery guarantee).
-// Provide isSessionActive so periodic sync skips ended sessions.
+// shouldSync skips ENDED/SUSPENDED/DETACHED sessions to prevent replay storms.
+// getHeadSeq caps replay depth so post-crash reconnects don't flood clients.
 connRegistry.setEventStore({
   getEventsAfter: (sessionId, afterSeq, limit) =>
     eventStore.getEventsAfter(sessionId, afterSeq, limit),
-  isSessionActive: (sessionId) => eventStore.getSession(sessionId)?.isActive ?? false,
+  shouldSync: (sessionId) => {
```
dimakis: 🔵 **style**: The `shouldSync` callback returns `false` when `getSessionState` returns `null` (session not in DB). This is likely correct, but worth a brief comment since the implicit null→false behavior is non-obvious and could silently skip sync for sessions whose state record was lost (e.g. after DB corruption). `[fixable]`

server/index.ts
dimakis: 🟡 **bugs**: shouldSync returns false when getSessionState returns null (unknown/missing session). This means a session that hasn't been registered in the EventStore yet (e.g. race between watch and session creation) will be silently skipped by periodic sync. The previous isSessionActive had the same behavior (defaulted to false via `?? false`), so this isn't a regression — but worth noting as a latent issue. Consider treating null state as sync-eligible. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -325,17 +325,27 @@ export class SseConnection implements ChatConnection {
...
+      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
           'X-Connection-ID': this._connectionId,
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok && endpoint === 'send') {
+        // Message delivery failed (e.g. 404 from stale connectionId).
```
dimakis: 🟡 **unsafe_assumptions**: doPost re-queues failed sends into pendingSends, but flushPendingSends does not await doPost calls (line 357). If the flush fires and all POSTs fail (e.g. server is up but returns 500), messages silently accumulate back into pendingSends without any backoff or retry-count cap. The MAX_PENDING_SENDS=100 limit in send() does not apply to re-queued items pushed from doPost — only to items pushed via the public send() method — so a persistent server-side error could cause the same messages to cycle indefinitely through flush→fail→requeue. `[fixable]`

packages/protocol/src/event-store.ts
dimakis: 🔵 **missing_tests**: getHeadSeq is a new public method on EventStore but has no unit test. Should verify: returns 0 for unknown session, returns correct max seq for a session with events, and returns correct value after new events are appended. `[fixable]`

packages/harness/src/connection-registry.ts
```diff
@@ -221,12 +228,31 @@ export class ConnectionRegistry {
...
           }
 
-          const cursor = connCursors.get(sessionId) ?? 0;
+          let cursor = connCursors.get(sessionId) ?? 0;
+
+          // Cap replay depth: if cursor is far behind head, skip ahead.
+          // Prevents replay storms after server restart or long disconnects.
+          if (this.eventStore.getHeadSeq) {
+            const head = this.eventStore.getHeadSeq(sessionId);
+            if (head - cursor > MAX_REPLAY_GAP) {
```
dimakis: 🔵 **bugs**: The replay gap check `head - cursor > MAX_REPLAY_GAP` performs integer subtraction without clamping. If getHeadSeq returns 0 (no events) and cursor is 0, the gap is 0 which is fine. But if a buggy adapter returns a negative head, the subtraction could produce unexpected results. Minor — the real EventStore.getHeadSeq returns `row?.head ?? 0` so this can't happen in practice. `[fixable]`

server/worktree.ts
```diff
@@ -456,6 +463,28 @@ export function rescueDirtyWorktree(
     return { success: false, error: err instanceof Error ? err.message : String(err) };
   }
 
+  // 1b. Check if staged changes contain meaningful work (not just noise markers)
```
dimakis, line 473: 🔵 **style**: The noise-file check step is labeled '1b' in test comments but has no step comment in the actual code (unlike steps 1, 2, 3, 4 which have numbered comments). Adding a consistent step label would improve readability. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🟡 **bugs** (L111): shouldSync returns false when getSessionState returns null (unknown/missing session). This means a session that hasn't been registered in the EventStore yet (e.g. race between watch and session creation) will be silently skipped by periodic sync. The previous isSessionActive had the same behavior (defaulted to false via `?? false`), so this isn't a regression — but worth noting as a latent issue. Consider treating null state as sync-eligible. `[fixable]`


_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🟡 **unsafe_assumptions** (L337): doPost re-queues failed sends into pendingSends, but flushPendingSends does not await doPost calls (line 357). If the flush fires and all POSTs fail (e.g. server is up but returns 500), messages silently accumulate back into pendingSends without any backoff or retry-count cap. The MAX_PENDING_SENDS=100 limit in send() does not apply to re-queued items pushed from doPost — only to items pushed via the public send() method — so a persistent server-side error could cause the same messages to cycle indefinitely through flush→fail→requeue. `[fixable]`


_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🔵 **missing_tests**: No test verifies behavior when a re-queued send fails again on the second flush. This would exercise the flush→fail→requeue→flush cycle and confirm no infinite loop or unbounded growth. `[fixable]`


_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🔵 **missing_tests** (L618): getHeadSeq is a new public method on EventStore but has no unit test. Should verify: returns 0 for unknown session, returns correct max seq for a session with events, and returns correct value after new events are appended. `[fixable]`


_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🔵 **bugs** (L242): The replay gap check `head - cursor > MAX_REPLAY_GAP` performs integer subtraction without clamping. If getHeadSeq returns 0 (no events) and cursor is 0, the gap is 0 which is fine. But if a buggy adapter returns a negative head, the subtraction could produce unexpected results. Minor — the real EventStore.getHeadSeq returns `row?.head ?? 0` so this can't happen in practice. `[fixable]`


### `server/worktree.ts`
_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🔵 **style** (L466): The noise-file check step is labeled '1b' in test comments but has no step comment in the actual code (unlike steps 1, 2, 3, 4 which have numbered comments). Adding a consistent step label would improve readability. `[fixable]`


_Solid defensive PR addressing three real failure modes (replay storms, noise-file rescues, interrupt crashes). The re-queue logic in SseConnection lacks backoff/retry-count enforcement on the doPost path which could cause spin under persistent server errors, but the bounded queue prevents memory issues. No critical bugs found._

- 🔵 **missing_tests**: No test covers the case where the `git diff --cached --name-only` command itself fails (the outer catch block at worktree.ts line 484). The comment says 'proceed with rescue anyway' — a test confirming this fallthrough behavior would be valuable. `[fixable]`

server/worktree.ts
```diff
@@ -463,6 +470,28 @@ export function rescueDirtyWorktree(
...
+    const files = staged ? staged.split('\n') : [];
+    if (files.length === 0 || files.every((f) => RESCUE_NOISE_FILES.has(f))) {
+      log.info('rescue skipped — only noise files changed', { wtId, files });
+      // Unstage so the worktree can be cleaned up without leaving staged changes
+      try {
+        execFileSync('git', ['-C', worktreePath, 'reset', 'HEAD'], gitOpts);
+      } catch {
+        // Non-fatal — worktree is about to be removed anyway
+      }
+      return { success: false, error: 'no meaningful changes to rescue' };
```
dimakis, line 489: 🟡 **bugs**: When rescue returns `{ success: false, error: 'no meaningful changes to rescue' }`, both callers (`cleanupStaleWorktrees` at line ~682 and `cleanupSessionWorktrees` in chat.ts at line ~1319) treat this identically to a real rescue failure: the worktree is preserved on disk, and in the stale-GC path an inbox notification is posted. Since `hasUncommittedWork` still detects `.mitzo-session` as dirty (it's not in `.gitignore`), stale GC will re-encounter the same worktree every cycle, re-attempt rescue, get the same 'no meaningful changes' result, and preserve it indefinitely. Noise-only worktrees will accumulate on disk. Consider either (a) returning a distinct signal (e.g. `{ success: false, noMeaningfulChanges: true }`) so callers can safely remove the worktree, or (b) filtering noise files in `hasUncommittedWork` so the worktree is seen as clean. `[fixable]`

server/__tests__/rescue-worktree.test.ts
```diff
@@ -252,14 +260,68 @@ describe('rescueDirtyWorktree', () => {
...
 
     expect(result.success).toBe(false);
-    expect(result.error).toContain('nothing to commit');
+    expect(result.error).toContain('no meaningful changes');
+  });
+
+  it('skips rescue when only noise files (.mitzo-session) are staged', () => {
+    // Step 0: getRepoRemote
+    mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
+    // Step 1: git add -A
```
dimakis, line 277: 🔵 **style**: Three new test cases have comments saying 'Step 1: git add -A' (lines 277, 296, 317) but the implementation at worktree.ts:468 uses `git add -u`. The distinction matters (tracked-only vs all files). Inaccurate test comments can mislead future maintainers about what the code actually does. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/worktree.ts`
_The noise-file filter correctly prevents useless rescue PRs, but callers still treat the result as a rescue failure — noise-only worktrees will accumulate on disk because stale GC re-encounters them every cycle._

- 🟡 **bugs** (L489): When rescue returns `{ success: false, error: 'no meaningful changes to rescue' }`, both callers (`cleanupStaleWorktrees` at line ~682 and `cleanupSessionWorktrees` in chat.ts at line ~1319) treat this identically to a real rescue failure: the worktree is preserved on disk, and in the stale-GC path an inbox notification is posted. Since `hasUncommittedWork` still detects `.mitzo-session` as dirty (it's not in `.gitignore`), stale GC will re-encounter the same worktree every cycle, re-attempt rescue, get the same 'no meaningful changes' result, and preserve it indefinitely. Noise-only worktrees will accumulate on disk. Consider either (a) returning a distinct signal (e.g. `{ success: false, noMeaningfulChanges: true }`) so callers can safely remove the worktree, or (b) filtering noise files in `hasUncommittedWork` so the worktree is seen as clean. `[fixable]`


_The noise-file filter correctly prevents useless rescue PRs, but callers still treat the result as a rescue failure — noise-only worktrees will accumulate on disk because stale GC re-encounters them every cycle._

- 🔵 **style** (L277): Three new test cases have comments saying 'Step 1: git add -A' (lines 277, 296, 317) but the implementation at worktree.ts:468 uses `git add -u`. The distinction matters (tracked-only vs all files). Inaccurate test comments can mislead future maintainers about what the code actually does. `[fixable]`
- 🔵 **missing_tests**: No test covers the outer `catch` at worktree.ts:491 where `git diff --cached --name-only` itself throws. The code deliberately proceeds with rescue in that case ('safer than skipping'), but this fail-open behavior is untested. `[fixable]`

PR #399, opened 2026-06-25
feat(signals): register Centaur callbacks on centaur_review watch

dimakis:
## Summary
- `SignalProcessor` now registers a callback URL with Centaur's signal bridge when watching a `centaur_review` gate
- Deregisters on unwatch (best-effort)
- Polling fallback continues alongside push — belt and suspenders
- Gracefully handles Centaur being unavailable (fire-and-forget registration)

Part of multi-agent orchestration Phase 2 (Centaur signal bridge). Companion PR: centaur side.

## Test plan
- [x] `vitest run server/__tests__/signal-processor.test.ts` — 11 tests pass
- [ ] Manual integration test with Centaur

### DIFF
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -325,7 +325,7 @@ export class SseConnection implements ChatConnection {
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
     if (!this._connectionId) return;
     try {
-      await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
+      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
@@ -333,9 +333,19 @@ export class SseConnection implements ChatConnection {
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok && endpoint === 'send') {
+        // Message delivery failed (e.g. 404 from stale connectionId).
+        // Re-queue so it flushes on the next successful reconnect.
+        console.warn(`[SseConnection] POST /${endpoint} returned ${res.status}, re-queuing`);
+        this.pendingSends.push({ endpoint, body });
+      }
     } catch {
-      // POST failures are non-fatal — the server may be temporarily
-      // unreachable. The SSE stream will reconnect and replay missed events.
+      if (endpoint === 'send') {
+        // Network error — re-queue for retry after reconnect.
+        console.warn(`[SseConnection] POST /${endpoint} failed, re-queuing`);
+        this.pendingSends.push({ endpoint, body });
+      }
+      // Non-send POST failures (stop, interrupt, etc.) are non-fatal.
     }
   }
 
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -337,13 +337,13 @@ export class SseConnection implements ChatConnection {
         // Message delivery failed (e.g. 404 from stale connectionId).
         // Re-queue so it flushes on the next successful reconnect.
         console.warn(`[SseConnection] POST /${endpoint} returned ${res.status}, re-queuing`);
-        this.pendingSends.push({ endpoint, body });
+        this.enqueuePending(endpoint, body);
       }
     } catch {
       if (endpoint === 'send') {
         // Network error — re-queue for retry after reconnect.
         console.warn(`[SseConnection] POST /${endpoint} failed, re-queuing`);
-        this.pendingSends.push({ endpoint, body });
+        this.enqueuePending(endpoint, body);
       }
       // Non-send POST failures (stop, interrupt, etc.) are non-fatal.
     }
@@ -356,6 +356,14 @@ export class SseConnection implements ChatConnection {
     for (const { endpoint, body } of toFlush) {
       this.doPost(endpoint, body);
     }
+  }
+
+  /** Enqueue a pending send with the same MAX_PENDING_SENDS cap as send(). */
+  private enqueuePending(endpoint: string, body: Record<string, unknown>): void {
+    if (this.pendingSends.length >= MAX_PENDING_SENDS) {
+      this.pendingSends.shift();
+    }
+    this.pendingSends.push({ endpoint, body });
   }
 
   /**
--- a/packages/harness/src/connection-registry.ts
+++ b/packages/harness/src/connection-registry.ts
@@ -38,15 +38,22 @@ export interface EventStoreAdapter {
     seq: number;
     payload: Record<string, unknown>;
   }>;
-  /** Optional: check if a session is still active. When provided, periodic sync
-   *  skips ended sessions to avoid unnecessary EventStore queries. */
-  isSessionActive?(sessionId: string): boolean;
+  /** Check if a session should receive periodic sync delivery.
+   *  When provided, periodic sync skips sessions where this returns false
+   *  (e.g. ENDED, SUSPENDED, DETACHED). */
+  shouldSync?(sessionId: string): boolean;
+  /** Return the latest seq number for a session, used to cap replay depth. */
+  getHeadSeq?(sessionId: string): number;
 }
 
 // Periodic sync fires every 5s to retry missed events
 const SYNC_INTERVAL_MS = 5000;
 // Limit events per sync round per connection to avoid overwhelming slow clients
 const SYNC_BATCH_LIMIT = 50;
+// Maximum gap between cursor and head before we skip ahead.
+// Prevents replay storms after server restart or long disconnects.
+// Client can lazy-load older events via the /api/sessions/:id/events API.
+const MAX_REPLAY_GAP = 200;
 
 export class ConnectionRegistry {
   private connections = new Map<string, Connection>();
@@ -221,12 +228,31 @@ export class ConnectionRegistry {
         if (!connCursors) continue;
 
         for (const sessionId of conn.watchedSessions) {
-          // Skip ended sessions to avoid unnecessary EventStore queries
-          if (this.eventStore.isSessionActive && !this.eventStore.isSessionActive(sessionId)) {
+          // Skip sessions that don't need sync (ENDED, SUSPENDED, DETACHED)
+          if (this.eventStore.shouldSync && !this.eventStore.shouldSync(sessionId)) {
             continue;
           }
 
-          const cursor = connCursors.get(sessionId) ?? 0;
+          let cursor = connCursors.get(sessionId) ?? 0;
+
+          // Cap replay depth: if cursor is far behind head, skip ahead.
+          // Prevents replay storms after server restart or long disconnects.
+          if (this.eventStore.getHeadSeq) {
+            const head = this.eventStore.getHeadSeq(sessionId);
+            if (head - cursor > MAX_REPLAY_GAP) {
+              const newCursor = head - MAX_REPLAY_GAP;
+              log.warn('periodic sync: cursor too far behind, skipping ahead', {
+                connectionId,
+                sessionId,
+                oldCursor: cursor,
+                newCursor,
+                head,
+                skipped: newCursor - cursor,
+              });
+              connCursors.set(sessionId, newCursor);
+              cursor = newCursor;
+            }
+          }
 
           // Fetch missed events from EventStore
           let missedEvents: Array<{ seq: number; payload: Record<string, unknown> }>;
--- a/server/signal-processor.ts
+++ b/server/signal-processor.ts
@@ -34,10 +34,19 @@ export class SignalProcessor {
   private watches = new Map<string, WatchEntry>();
   private store: TaskStore;
   private onSignalResolved: (taskId: string) => void;
+  private centaurBaseUrl: string;
+  private mitzoBaseUrl: string;
 
-  constructor(store: TaskStore, onSignalResolved: (taskId: string) => void) {
+  constructor(
+    store: TaskStore,
+    onSignalResolved: (taskId: string) => void,
+    centaurBaseUrl = 'http://localhost:8642',
+    mitzoBaseUrl = 'http://localhost:3100',
+  ) {
     this.store = store;
     this.onSignalResolved = onSignalResolved;
+    this.centaurBaseUrl = centaurBaseUrl;
+    this.mitzoBaseUrl = mitzoBaseUrl;
   }
 
   watch(taskId: string, gateConfig: GateConfig): void {
@@ -57,12 +66,23 @@ export class SignalProcessor {
     }
 
     this.watches.set(taskId, { taskId, gateConfig, intervalId });
+
+    // Register callback with Centaur for push-based resolution
+    if (gateConfig.type === 'centaur_review') {
+      this.registerCentaurCallback(taskId, gateConfig as GateConfig & { pr_url: string });
+    }
   }
 
   unwatch(taskId: string): void {
     const entry = this.watches.get(taskId);
     if (!entry) return;
     if (entry.intervalId) clearInterval(entry.intervalId);
+
+    // Deregister callback with Centaur
+    if (entry.gateConfig.type === 'centaur_review') {
+      this.deregisterCentaurCallback(taskId);
+    }
+
     this.watches.delete(taskId);
     log.info('unwatched task', { taskId });
   }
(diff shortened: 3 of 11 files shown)


packages/client/src/sse-connection.ts
```diff
@@ -325,17 +325,27 @@ export class SseConnection implements ChatConnection {
...
-      await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
+      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
           'X-Connection-ID': this._connectionId,
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok && endpoint === 'send') {
```
dimakis, line 336: 🟡 **bugs**: No retry limit on re-queued sends. When `doPost` fails for a 'send' endpoint, it calls `enqueuePending`, and `flushPendingSends` calls `doPost` again. If the server persistently rejects a message (e.g. 400/422 validation error), it cycles through flush → fail → re-queue indefinitely across reconnects. Consider adding a retry counter to break the loop after N attempts. `[fixable]`

server/signal-processor.ts
```diff
@@ -57,12 +66,23 @@ export class SignalProcessor {
     }
 
     this.watches.set(taskId, { taskId, gateConfig, intervalId });
+
+    // Register callback with Centaur for push-based resolution
+    if (gateConfig.type === 'centaur_review') {
```
dimakis, line 72: 🟡 **bugs**: Race condition: `registerCentaurCallback` is async but not awaited in `watch()`. If `unwatch(taskId)` is called before the registration POST completes, the DELETE arrives first and the POST creates a dangling registration in Centaur. Consider tracking the registration promise and awaiting it in `unwatch`/`deregisterCentaurCallback`. `[fixable]`

server/signal-processor.ts
```diff
@@ -154,6 +174,48 @@ export class SignalProcessor {
...
+      if (res.ok) {
+        log.info('registered centaur callback', { taskId, pr_url: config.pr_url });
+      } else {
+        log.warn('centaur callback registration failed', { taskId, status: res.status });
+      }
+    } catch (err) {
+      // Centaur might not be running — polling fallback covers this case
+      log.warn('centaur callback registration error', {
+        taskId,
+        error: (err as Error).message,
```
dimakis: 🔵 **style**: Uses `(err as Error).message` which violates codebase convention of `err instanceof Error ? err.message : String(err)` (see chat.ts:1271, etc.). If a non-Error is thrown, `.message` will be `undefined` instead of a useful string. `[fixable]`

server/signal-processor.ts
```diff
@@ -34,10 +34,19 @@ export class SignalProcessor {
...
   private onSignalResolved: (taskId: string) => void;
-
-  constructor(store: TaskStore, onSignalResolved: (taskId: string) => void) {
+  private centaurBaseUrl: string;
+  private mitzoBaseUrl: string;
+
+  constructor(
+    store: TaskStore,
+    onSignalResolved: (taskId: string) => void,
+    centaurBaseUrl = 'http://localhost:8642',
```
dimakis, line 44: 🔵 **unsafe_assumptions**: Constructor defaults `centaurBaseUrl='http://localhost:8642'` and `mitzoBaseUrl='http://localhost:3100'`, but the production instantiation at `server/index.ts:177` passes neither. These hardcoded localhost URLs will break if Centaur or Mitzo run on different hosts/ports. Consider wiring these from env vars (e.g. `CENTAUR_BASE_URL`, `PORT`) at the call site. `[fixable]`

packages/harness/src/connection-registry.ts
```diff
@@ -221,12 +228,31 @@ export class ConnectionRegistry {
...
           }
 
-          const cursor = connCursors.get(sessionId) ?? 0;
+          let cursor = connCursors.get(sessionId) ?? 0;
+
+          // Cap replay depth: if cursor is far behind head, skip ahead.
+          // Prevents replay storms after server restart or long disconnects.
+          if (this.eventStore.getHeadSeq) {
+            const head = this.eventStore.getHeadSeq(sessionId);
+            if (head - cursor > MAX_REPLAY_GAP) {
```
dimakis, line 242: 🟡 **unsafe_assumptions**: When the cursor is skipped ahead via `MAX_REPLAY_GAP`, the client is never notified that events were dropped. The comment says 'Client can lazy-load older events via the /api/sessions/:id/events API' but the client has no way to know it needs to. Consider sending a `gap_detected` event to the client so it can trigger a full restore. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid defensive hardening (POST retry, replay-gap cap, interrupt crash guard) with good test coverage, but the send retry path lacks a retry limit and the Centaur callback registration has a race between async register and sync unwatch._

- 🟡 **bugs** (L336): No retry limit on re-queued sends. When `doPost` fails for a 'send' endpoint, it calls `enqueuePending`, and `flushPendingSends` calls `doPost` again. If the server persistently rejects a message (e.g. 400/422 validation error), it cycles through flush → fail → re-queue indefinitely across reconnects. Consider adding a retry counter to break the loop after N attempts. `[fixable]`


_Solid defensive hardening (POST retry, replay-gap cap, interrupt crash guard) with good test coverage, but the send retry path lacks a retry limit and the Centaur callback registration has a race between async register and sync unwatch._

- 🟡 **bugs** (L71): Race condition: `registerCentaurCallback` is async but not awaited in `watch()`. If `unwatch(taskId)` is called before the registration POST completes, the DELETE arrives first and the POST creates a dangling registration in Centaur. Consider tracking the registration promise and awaiting it in `unwatch`/`deregisterCentaurCallback`. `[fixable]`
- 🔵 **style** (L202): Uses `(err as Error).message` which violates codebase convention of `err instanceof Error ? err.message : String(err)` (see chat.ts:1271, etc.). If a non-Error is thrown, `.message` will be `undefined` instead of a useful string. `[fixable]`
- 🔵 **unsafe_assumptions** (L43): Constructor defaults `centaurBaseUrl='http://localhost:8642'` and `mitzoBaseUrl='http://localhost:3100'`, but the production instantiation at `server/index.ts:177` passes neither. These hardcoded localhost URLs will break if Centaur or Mitzo run on different hosts/ports. Consider wiring these from env vars (e.g. `CENTAUR_BASE_URL`, `PORT`) at the call site. `[fixable]`


_Solid defensive hardening (POST retry, replay-gap cap, interrupt crash guard) with good test coverage, but the send retry path lacks a retry limit and the Centaur callback registration has a race between async register and sync unwatch._

- 🟡 **unsafe_assumptions** (L242): When the cursor is skipped ahead via `MAX_REPLAY_GAP`, the client is never notified that events were dropped. The comment says 'Client can lazy-load older events via the /api/sessions/:id/events API' but the client has no way to know it needs to. Consider sending a `gap_detected` event to the client so it can trigger a full restore. `[fixable]`


_Solid defensive hardening (POST retry, replay-gap cap, interrupt crash guard) with good test coverage, but the send retry path lacks a retry limit and the Centaur callback registration has a race between async register and sync unwatch._

- 🔵 **missing_tests**: No test for a persistently failing send (e.g. server returns 400 on every retry). The existing tests only cover transient failures that succeed on the next reconnect. A test showing the message is eventually dropped or capped would document the intended behavior. `[fixable]`


_Solid defensive hardening (POST retry, replay-gap cap, interrupt crash guard) with good test coverage, but the send retry path lacks a retry limit and the Centaur callback registration has a race between async register and sync unwatch._

- 🔵 **missing_tests**: No test for `getHeadSeq` returning 0 (no events for session). The arithmetic `head - cursor > MAX_REPLAY_GAP` evaluates to `0 - 0 > 200` which is false, so behavior is correct but the edge case is undocumented by tests. `[fixable]`

dimakis:
## Centaur Review

_Solid PR with good test coverage for SSE retry, replay caps, and interrupt crash resilience. Main gap: unwatchAll() doesn't deregister Centaur callbacks (bug), and checkCentaurReview still hardcodes the Centaur URL instead of using the newly injected base URL._

- 🟡 **bugs** (L96): unwatchAll() does not call deregisterCentaurCallback() for centaur_review entries. When the orchestrator stops (or process shuts down), Centaur retains stale registered callbacks that will POST to dead endpoints. unwatchAll() should iterate entries and call deregisterCentaurCallback for centaur_review tasks, matching the logic in unwatch(). `[fixable]`
- 🟡 **bugs** (L89): unwatch() is synchronous but calls async deregisterCentaurCallback() without awaiting. deregisterCentaurCallback internally awaits pendingRegistrations (line 214-217), but since unwatch() doesn't await, that await is lost. If unwatch() is called right after watch(), the DELETE can race ahead of the POST, and the registration becomes a dangling entry in Centaur. Consider making unwatch() async, or at minimum documenting the fire-and-forget semantics. `[fixable]`
- 🟡 **unsafe_assumptions** (L337): checkCentaurReview() (standalone function at line 337) hardcodes 'http://localhost:8642' instead of using the injected centaurBaseUrl. The constructor now accepts centaurBaseUrl for register/deregister, but polling still bypasses it. If CENTAUR_URL is set to a non-localhost value, polling will hit the wrong endpoint while register/deregister hit the correct one. `[fixable]`
- 🔵 **unsafe_assumptions** (L187): registerCentaurCallback calls fetch() without an AbortSignal timeout. If Centaur hangs (DNS, network), the promise never resolves and the pendingRegistrations entry leaks indefinitely. Consider adding AbortSignal.timeout(5000) to the fetch options for both register and deregister calls. `[fixable]`


_Solid PR with good test coverage for SSE retry, replay caps, and interrupt crash resilience. Main gap: unwatchAll() doesn't deregister Centaur callbacks (bug), and checkCentaurReview still hardcodes the Centaur URL instead of using the newly injected base URL._

- 🔵 **missing_tests**: No test verifies that unwatchAll() deregisters centaur_review callbacks (which it currently doesn't — see related bug finding). Once the bug is fixed, a test should confirm that unwatchAll() sends DELETE requests for all centaur_review entries. `[fixable]`


_Solid PR with good test coverage for SSE retry, replay caps, and interrupt crash resilience. Main gap: unwatchAll() doesn't deregister Centaur callbacks (bug), and checkCentaurReview still hardcodes the Centaur URL instead of using the newly injected base URL._

- 🔵 **missing_tests** (L511): The MAX_REPLAY_GAP test 'caps replay depth when cursor is far behind head' is missing a call to registry.stopPeriodicSync() at the end, unlike the companion test on line 557. This leaks the periodic timer. Cosmetic since tests use fake timers, but inconsistent. `[fixable]`


### `server/index.ts`
_Solid PR with good test coverage for SSE retry, replay caps, and interrupt crash resilience. Main gap: unwatchAll() doesn't deregister Centaur callbacks (bug), and checkCentaurReview still hardcodes the Centaur URL instead of using the newly injected base URL._

- 🔵 **style** (L190): MITZO_URL default uses string interpolation with PORT: `http://localhost:${PORT}`. If PORT parsing fails (NaN from invalid env var), this silently creates 'http://localhost:NaN'. Low risk since PORT is validated by the HTTP server bind, but the fallback is fragile. `[fixable]`

dimakis:
## Centaur Review

_Solid reliability improvements (SSE retry, replay cap, interrupt guard). The main actionable issue is that `unwatchAll()` doesn't deregister Centaur callbacks, inconsistent with `unwatch()` — a straightforward fix._

- 🟡 **bugs** (L96): `unwatchAll()` does not call `deregisterCentaurCallback()` for centaur_review watches, leaving dangling callback registrations in Centaur. This is inconsistent with `unwatch()` (line 88-90) which does deregister. Called during graceful shutdown (`server/index.ts:962`), so registered Centaur callbacks will persist across server restarts. Fix: iterate entries and call `deregisterCentaurCallback(taskId)` for centaur_review gates before clearing the map. `[fixable]`
- 🔵 **unsafe_assumptions** (L89): `unwatch()` is synchronous (`void`) but fire-and-forgets the async `deregisterCentaurCallback()`. If `unwatch` is followed by a `watch` for the same taskId before deregister completes, the DELETE could arrive after the new registration, creating a dangling registration. This is unlikely in practice (orchestrator tick is single-threaded), but worth noting. Consider returning the promise or documenting the constraint. `[fixable]`
- 🔵 **style** (L75): The type assertion `gateConfig as GateConfig & { pr_url: string }` is used in both `watch()` (line 75) and `checkCentaurReview()` (line ~260). Consider adding a `CentaurReviewGateConfig` discriminated type to avoid repeated unsafe casts. `[fixable]`


_Solid reliability improvements (SSE retry, replay cap, interrupt guard). The main actionable issue is that `unwatchAll()` doesn't deregister Centaur callbacks, inconsistent with `unwatch()` — a straightforward fix._

- 🟡 **missing_tests**: No test verifies that `unwatchAll()` deregisters Centaur callbacks — the existing `unwatchAll` test (line 146) only uses `human_approval` gates. This gap masks the bug where `unwatchAll` doesn't call `deregisterCentaurCallback`. `[fixable]`


### `server/index.ts`
_Solid reliability improvements (SSE retry, replay cap, interrupt guard). The main actionable issue is that `unwatchAll()` doesn't deregister Centaur callbacks, inconsistent with `unwatch()` — a straightforward fix._

- 🔵 **unsafe_assumptions** (L188): `process.env.CENTAUR_URL || 'http://localhost:8642'` — an empty string env var (`CENTAUR_URL=''`) passes the `||` check and falls through to the default. This is probably fine, but `?? 'http://localhost:8642'` would be more precise if you want to distinguish empty from unset. `[fixable]`


_Solid reliability improvements (SSE retry, replay cap, interrupt guard). The main actionable issue is that `unwatchAll()` doesn't deregister Centaur callbacks, inconsistent with `unwatch()` — a straightforward fix._

- 🔵 **regressions** (L240): When replay is capped (cursor skipped ahead), the client receives no notification that events were dropped. The log.warn is server-side only. The client may have stale state with no way to know it needs a full refresh. Consider emitting a `replay_gap` event to the client so it can trigger a full session restore via the REST API. `[fixable]`


_Solid reliability improvements (SSE retry, replay cap, interrupt guard). The main actionable issue is that `unwatchAll()` doesn't deregister Centaur callbacks, inconsistent with `unwatch()` — a straightforward fix._

- 🔵 **bugs** (L335): When `doPost` re-queues a failed send, the `send()` method already returned `true` to the caller. If the re-queued message also fails on the next reconnect, it will be re-queued again indefinitely (no retry limit). This is bounded by MAX_PENDING_SENDS eviction, but a message could cycle between queue and failure forever if the server consistently rejects it (e.g. invalid payload). Consider adding a retry counter. `[fixable]`

dimakis:
Superseded by a clean PR from feat/centaur-signal-callbacks — squashed commits, removed mixed SSE fixes.

dimakis:
## Centaur Review

_`unwatchAll` bypasses the new Centaur deregistration path, and the `checkCentaurReview` polling function still hardcodes the Centaur URL despite the constructor now accepting it as a parameter. The core replay-storm prevention (MAX_REPLAY_GAP, shouldSync, POST retry) is well-tested and sound._

- 🟡 **bugs** (L70): `unwatchAll()` clears intervals and the watches map but does not call `deregisterCentaurCallback()` for `centaur_review` entries. It is called during server shutdown (`server/index.ts:985`) and orchestrator stop, leaving dangling callbacks registered with Centaur that would POST back to a dead Mitzo server. Fix: iterate watches and call `unwatch(taskId)` per entry (which does deregister) instead of manually clearing intervals and the map. `[fixable]`
- 🟡 **unsafe_assumptions** (L266): The PR makes `centaurBaseUrl` configurable via the constructor (used by `registerCentaurCallback` and `deregisterCentaurCallback`), but the standalone `checkCentaurReview()` polling function still hardcodes `http://localhost:8642`. Registration uses the configured URL; polling uses the hardcoded one. If `CENTAUR_URL` is set to anything other than `http://localhost:8642`, registration and polling will target different hosts. `[fixable]`
- 🔵 **style** (L180): `registerCentaurCallback` takes `config: { pr_url: string }` but it's called with `gateConfig as GateConfig & { pr_url: string }`. Since `GateConfig` for `centaur_review` may use `repo` + `pr` instead of `pr_url` (per the `/api/signals/resolve` endpoint matching logic in app.ts), the cast could mask a missing `pr_url` at runtime. Consider guarding: `if (!config.pr_url) return;`. `[fixable]`


_`unwatchAll` bypasses the new Centaur deregistration path, and the `checkCentaurReview` polling function still hardcodes the Centaur URL despite the constructor now accepting it as a parameter. The core replay-storm prevention (MAX_REPLAY_GAP, shouldSync, POST retry) is well-tested and sound._

- 🔵 **missing_tests**: No test verifies that `unwatchAll()` deregisters Centaur callbacks for `centaur_review` entries (once the `unwatchAll` bug above is fixed). The existing `unwatchAll clears everything` test at line 146 should be extended to cover this. `[fixable]`


### `server/index.ts`
_`unwatchAll` bypasses the new Centaur deregistration path, and the `checkCentaurReview` polling function still hardcodes the Centaur URL despite the constructor now accepting it as a parameter. The core replay-storm prevention (MAX_REPLAY_GAP, shouldSync, POST retry) is well-tested and sound._

- 🔵 **regressions** (L107): The old `isSessionActive` (backed by the `is_active` column) included DETACHED and SUSPENDED sessions in periodic sync. The new `shouldSync` excludes them. This is a deliberate behavioral change and correct for the stated goal (no connected client to deliver to), but worth noting: if a client reconnects to a DETACHED session, events emitted during the detached window won't have been pushed via periodic sync. The MAX_REPLAY_GAP cap compounds this — the client may miss events that fall outside the 200-event window. Verify the `/api/sessions/:id/events` lazy-load path covers this gap.

PR #404, opened 2026-06-27
fix(ui): opaque session banner + header stacking context

dimakis:
## Summary

- Session banner used undefined `--bg-primary` CSS variable (resolved to `transparent`), causing scrolled content to bleed through
- Header toolbar had no stacking context, so scroll content could render above it
- Top padding on `.chat-messages` created a transparent gap between header and sticky banner

## Changes

- Replace `var(--bg-primary)` with `var(--surface)` on `.session-banner`
- Add `position: relative` + `z-index: 20` to `.chat-header`
- Move top padding from `.chat-messages` to bottom-only, so sticky banner sits flush
- Horizontal negative margin on banner to span full scroll container width

## Test plan

- [ ] Open a session in mobile Safari — no content bleeds through header or banner
- [ ] Scroll up/down — banner sticks flush, no transparent gap
- [ ] Check desktop layout is unaffected
- [ ] Verify keyboard-open state still compacts header correctly

### DIFF
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -2429,7 +2429,7 @@ textarea:focus {
   position: sticky;
   top: 0;
   z-index: 10;
-  margin: 0 calc(-1 * var(--space-3));
+  margin: -1px calc(-1 * var(--space-3)) 0;
   padding: 0 var(--space-3);
   border-bottom: 1px solid var(--border);
   background: var(--surface);
--- a/frontend/src/styles/global.css
+++ b/frontend/src/styles/global.css
@@ -1475,6 +1475,8 @@ textarea:focus {
   background: var(--surface);
   border-bottom: 1px solid var(--border);
   flex-shrink: 0;
+  position: relative;
+  z-index: 20;
   transition:
     max-height 0.2s,
     padding 0.2s,
@@ -1666,7 +1668,7 @@ textarea:focus {
   flex: 1;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
-  padding: var(--space-4) var(--space-3);
+  padding: 0 var(--space-3) var(--space-4);
   display: flex;
   flex-direction: column;
   gap: var(--space-3);
@@ -2427,8 +2429,10 @@ textarea:focus {
   position: sticky;
   top: 0;
   z-index: 10;
+  margin: 0 calc(-1 * var(--space-3));
+  padding: 0 var(--space-3);
   border-bottom: 1px solid var(--border);
-  background: var(--bg-primary);
+  background: var(--surface);
 }
 
 .session-banner-header {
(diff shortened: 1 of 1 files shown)


dimakis (commented):
## Centaur Review

LGTM — no issues found.

frontend/src/styles/global.css
```diff
@@ -1666,7 +1669,7 @@ textarea:focus {
   flex: 1;
   overflow-y: auto;
   -webkit-overflow-scrolling: touch;
-  padding: var(--space-4) var(--space-3);
+  padding: 0 var(--space-3) var(--space-4);
```
dimakis, line 1672: 🟡 **regressions**: Removing top padding from `.chat-messages` (was `var(--space-4)`, now `0`) means that when no session banner is rendered (SessionBanner returns null when bootContext and sessionContext are both absent), the first message sits flush against the header. Combined with the new 4px box-shadow on `.chat-header` (z-index 20), the top ~4px of the first message will be visually obscured. Consider adding a small top margin/padding to `.chat-messages` children or using a CSS `:first-child` rule to restore spacing when the banner is absent. `[fixable]`

frontend/src/styles/global.css
```diff
@@ -2427,8 +2430,10 @@ textarea:focus {
   position: sticky;
   top: 0;
   z-index: 10;
+  margin: -1px calc(-1 * var(--space-3)) 0;
```
dimakis, line 2433: 🔵 **style**: The negative horizontal margin `calc(-1 * var(--space-3))` makes the banner full-bleed on mobile where `.chat-messages` uses `var(--space-3)` horizontal padding, but on desktop the container padding is `var(--space-6)` (2rem) while the banner only pulls back 0.75rem. The banner won't reach the container edges on desktop. This may be fine visually, but if full-bleed is intended across breakpoints, the negative margin should use a CSS variable or be overridden in the desktop media query. `[fixable]`

dimakis (commented):
## Centaur Review

_Good fix for the undefined `--bg-primary` variable (banner was transparent) and proper stacking context, but removing top padding from `.chat-messages` introduces a minor visual regression when no session banner is present — first message content gets clipped by the header's box-shadow on mobile._

- 🟡 **regressions** (L1672): Removing top padding from `.chat-messages` (was `var(--space-4)`, now `0`) means that when no session banner is rendered (SessionBanner returns null when bootContext and sessionContext are both absent), the first message sits flush against the header. Combined with the new 4px box-shadow on `.chat-header` (z-index 20), the top ~4px of the first message will be visually obscured. Consider adding a small top margin/padding to `.chat-messages` children or using a CSS `:first-child` rule to restore spacing when the banner is absent. `[fixable]`
- 🔵 **style** (L2433): The negative horizontal margin `calc(-1 * var(--space-3))` makes the banner full-bleed on mobile where `.chat-messages` uses `var(--space-3)` horizontal padding, but on desktop the container padding is `var(--space-6)` (2rem) while the banner only pulls back 0.75rem. The banner won't reach the container edges on desktop. This may be fine visually, but if full-bleed is intended across breakpoints, the negative margin should use a CSS variable or be overridden in the desktop media query. `[fixable]`

PR #405, opened 2026-06-27
fix(infra): restore Jaeger trace persistence with v2 config

dimakis:
## Summary

- Jaeger v2 (`latest` tag) silently ignores v1 env vars (`SPAN_STORAGE_TYPE`, `BADGER_*`) and falls back to **in-memory storage** — all traces lost on container restart
- `.jaeger-data/` was empty (0 bytes) despite `BADGER_EPHEMERAL=false`
- Replace env var config with proper v2 OTel Collector YAML config (`infra/jaeger-v2-config.yaml`) that explicitly configures Badger persistent storage
- Verified: traces survive container restarts

## Context

Part of a larger observability persistence fix. Investigation also found that the event store only persists `user_message` events (lifecycle events like `session_end`, `block_end` never reach `store.append()`), and all 434 sessions have zero for duration/turns/cost metrics. Those fixes will follow in separate PRs.

## Test plan

- [x] Jaeger starts with Badger storage (`Initializing storage 'badger_main'` in logs)
- [x] `.jaeger-data/data/` and `.jaeger-data/key/` populated with Badger files
- [x] Test span sent via OTLP HTTP, visible in Jaeger UI
- [x] Traces survive `podman restart mitzo_jaeger_1`

### DIFF
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -1572,6 +1572,44 @@ describe('runQueryLoop', () => {
       expect(events.some((e) => e.type === 'session_end')).toBe(true);
     });
 
+    it('records fallback usage for sessions that abort without SDK result', async () => {
+      const store = new EventStore(':memory:');
+      const connRegistry = new ConnectionRegistry();
+      const v2Transport = fakeTransport();
+      connRegistry.register(clientId, v2Transport);
+      connRegistry.watch(clientId, 'sess-usage-fallback');
+      connRegistry.setActive(clientId, 'sess-usage-fallback');
+
+      const session = registry.get(clientId)!;
+      session.sessionId = 'sess-usage-fallback';
+
+      // Seed the session row so recordUsage UPDATE has a target
+      store.upsertSession({ sessionId: 'sess-usage-fallback', cwd: '/tmp' });
+
+      // Stream that yields an assistant event (resolves sessionId) then errors
+      // before an SDK result event arrives — simulates WS disconnect mid-session.
+      async function* abortStream() {
+        yield {
+          type: 'assistant',
+          session_id: 'sess-usage-fallback',
+          message: { id: 'msg-1', role: 'assistant' },
+        };
+        throw new Error('connection lost');
+      }
+
+      await runQueryLoop(abortStream(), clientId, registry, abortController, store, undefined, {
+        connRegistry,
+      });
+
+      // Fallback usage should be recorded with best-effort data
+      const sessionMeta = store.getSession('sess-usage-fallback');
+      expect(sessionMeta).toBeDefined();
+      // durationMs may be 0 in fast tests — verify it was written (not default -1)
+      expect(sessionMeta!.durationMs).toBeGreaterThanOrEqual(0);
+      // Session should be marked ENDED with error reason (abort path)
+      expect(sessionMeta!.state).toBe('ENDED');
+    });
+
     it('delivers events to a NEW connection after WS reconnect (old connection gone)', async () => {
       const connRegistry = new ConnectionRegistry();
       const oldTransport = fakeTransport();
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -1608,6 +1608,10 @@ describe('runQueryLoop', () => {
       expect(sessionMeta!.durationMs).toBeGreaterThanOrEqual(0);
       // Session should be marked ENDED with error reason (abort path)
       expect(sessionMeta!.state).toBe('ENDED');
+      // Fallback records zeros for partial sessions (no token events in stream)
+      expect(sessionMeta!.outputTokens).toBe(0);
+      expect(sessionMeta!.numTurns).toBe(0);
+      expect(sessionMeta!.totalCostUsd).toBe(0);
     });
 
     it('delivers events to a NEW connection after WS reconnect (old connection gone)', async () => {
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -138,6 +138,16 @@ function sendOrBuffer(
       enriched = { ...enriched, seq };
     }
   }
+  // DEBUG: trace event persistence decisions (remove after investigation)
+  if (data.type && data.type !== 'token_update' && data.type !== 'progress') {
+    log.info('sendOrBuffer', {
+      type: data.type,
+      v: data.v,
+      hasSessionId: !!sessionId,
+      hasStore: !!store,
+      stored: !!(data.v === 2 && sessionId && store),
+    });
+  }
 
   // Suspend buffer: if the session is proactively suspended (iOS backgrounding),
   // buffer events instead of delivering via the driver transport (which is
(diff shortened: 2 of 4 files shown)


dimakis (commented):
## Centaur Review

### `infra/jaeger-v2-config.yaml`
_Clean v1→v2 Jaeger migration; only gap is the dropped 14-day trace TTL which will cause unbounded disk growth._

- 🟡 **regressions**: The old config had `BADGER_SPAN_STORE_TTL: 336h` (14 days) for trace retention. The new v2 config has no equivalent `span_store_ttl` setting under the Badger backend. Without it, Badger will retain traces indefinitely, causing unbounded disk growth in `.jaeger-data/`. Consider adding `span_store_ttl: 336h` under the `badger` key to restore the previous retention behavior (Loki already has its own `retention_period: 720h` for parity). `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

server/query-loop.ts
```diff
@@ -138,6 +138,16 @@ function sendOrBuffer(
       enriched = { ...enriched, seq };
     }
   }
+  // DEBUG: trace event persistence decisions (remove after investigation)
```
dimakis: 🟡 **style**: Debug logging left in production code. The comment explicitly says "remove after investigation" — this should not be merged. It logs on every non-token/progress event in sendOrBuffer, which is the hot path for all session events. This will be noisy in production logs and pollute Loki. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Jaeger v2 config migration and fallback usage recording look correct, but a debug log block explicitly marked for removal ("remove after investigation") is still in the diff and should be stripped before merge._

- 🟡 **style** (L141): Debug logging left in production code. The comment explicitly says "remove after investigation" — this should not be merged. It logs on every non-token/progress event in sendOrBuffer, which is the hot path for all session events. This will be noisy in production logs and pollute Loki. `[fixable]`


_Jaeger v2 config migration and fallback usage recording look correct, but a debug log block explicitly marked for removal ("remove after investigation") is still in the diff and should be stripped before merge._

- 🔵 **missing_tests** (L1604): The test only asserts durationMs and state, but doesn't verify the actual usage values written by the fallback path (outputTokens, numTurns, totalCostUsd). Since the stream yields no token events, cumulativeOutputTokens and turnIndex are both 0 — asserting those would document that the fallback records zeros for partial sessions, which is the stated intent. `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -1572,6 +1572,44 @@ describe('runQueryLoop', () => {
...
+          message: { id: 'msg-1', role: 'assistant' },
+        };
+        throw new Error('connection lost');
+      }
+
+      await runQueryLoop(abortStream(), clientId, registry, abortController, store, undefined, {
+        connRegistry,
+      });
+
+      // Fallback usage should be recorded with best-effort data
```
dimakis, line 1604: 🔵 **missing_tests**: The test only asserts durationMs and state, but doesn't verify the actual usage values written by the fallback path (outputTokens, numTurns, totalCostUsd). Since the stream yields no token events, cumulativeOutputTokens and turnIndex are both 0 — asserting those would document that the fallback records zeros for partial sessions, which is the stated intent. `[fixable]`

server/query-loop.ts
```diff
@@ -1378,6 +1379,22 @@ async function _runQueryLoopInner(
...
+      // (disconnect, abort, timeout). The normal path records full SDK-provided
+      // metrics; this fallback uses live counters so partial sessions aren't zero.
+      if (!doneSent && store && resolvedSessionId) {
+        const fallbackDurationMs = Date.now() - sessionStartedAt;
+        store.recordUsage(resolvedSessionId, {
+          inputTokens: 0, // per-type breakdown unavailable without SDK result
+          outputTokens: cumulativeOutputTokens,
+          cacheReadTokens: 0,
+          cacheCreationTokens: 0,
+          totalCostUsd: finalSession?.cumulativeCostUsd ?? 0,
```
dimakis, line 1392: 🟡 **bugs**: After `registry.remove(clientId)` at line 1380, `finalSession` is still a valid reference (Map.delete doesn't invalidate it), but the comment at lines 1357-1360 only documents the `cumulativeCostUsd` read for the span attributes block. The new fallback code also reads `finalSession?.cumulativeCostUsd` — safe due to the same reference semantics, but the existing comment should be updated to mention both read sites to prevent future refactoring from introducing a use-after-remove bug. `[fixable]`

server/query-loop.ts
```diff
@@ -1378,6 +1379,22 @@ async function _runQueryLoopInner(
...
+          numTurns: turnIndex,
```
dimakis, line 1393: 🔵 **unsafe_assumptions**: `turnIndex` is used as `numTurns` in the fallback, but `turnIndex` increments on each parent `message_start` (line 801), not on completed turns. For an aborted session, the last turn may be incomplete — this records an in-progress turn as completed. Minor inaccuracy, documented behavior is 'best-effort', but worth noting.

server/__tests__/query-loop.test.ts
```diff
@@ -1572,6 +1572,48 @@ describe('runQueryLoop', () => {
...
+
+      const session = registry.get(clientId)!;
+      session.sessionId = 'sess-usage-fallback';
+
+      // Seed the session row so recordUsage UPDATE has a target
+      store.upsertSession({ sessionId: 'sess-usage-fallback', cwd: '/tmp' });
+
+      // Stream that yields an assistant event (resolves sessionId) then errors
+      // before an SDK result event arrives — simulates WS disconnect mid-session.
+      async function* abortStream() {
```
dimakis, line 1591: 🔵 **missing_tests**: The test only covers the abort-with-error path (stream throws). Consider adding a test for the abort-controller signal path (external abort without error) — the fallback block would run with `caughtError` being falsy, and `setSessionState` would record reason `'completed'` instead of `'error'`, which may be misleading for aborted sessions. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Clean infrastructure migration (Jaeger v1 env-vars → v2 config YAML) with a well-guarded fallback usage recorder for aborted sessions. No bugs — findings are minor documentation and edge-case test gaps._

- 🟡 **bugs** (L1392): After `registry.remove(clientId)` at line 1380, `finalSession` is still a valid reference (Map.delete doesn't invalidate it), but the comment at lines 1357-1360 only documents the `cumulativeCostUsd` read for the span attributes block. The new fallback code also reads `finalSession?.cumulativeCostUsd` — safe due to the same reference semantics, but the existing comment should be updated to mention both read sites to prevent future refactoring from introducing a use-after-remove bug. `[fixable]`
- 🔵 **unsafe_assumptions** (L1393): `turnIndex` is used as `numTurns` in the fallback, but `turnIndex` increments on each parent `message_start` (line 801), not on completed turns. For an aborted session, the last turn may be incomplete — this records an in-progress turn as completed. Minor inaccuracy, documented behavior is 'best-effort', but worth noting.


_Clean infrastructure migration (Jaeger v1 env-vars → v2 config YAML) with a well-guarded fallback usage recorder for aborted sessions. No bugs — findings are minor documentation and edge-case test gaps._

- 🔵 **missing_tests** (L1591): The test only covers the abort-with-error path (stream throws). Consider adding a test for the abort-controller signal path (external abort without error) — the fallback block would run with `caughtError` being falsy, and `setSessionState` would record reason `'completed'` instead of `'error'`, which may be misleading for aborted sessions. `[fixable]`

PR #407, opened 2026-06-27
fix(client): sync running state on iOS foreground return

dimakis:
## Summary

- Fixes stale stop button when iOS Safari backgrounds during agent processing and the `session_end` SSE event is silently dropped
- On foreground return (`_foreground` event), queries `/api/sessions/:id/meta` and sets `running: false` when `isActive` is false
- Independent of the reconnect path, so it works regardless of `doReconnectPost` issues being fixed in #fix/reconnect-on-stale-post

## Root cause

The `_foreground` handler called `fetchAndRestoreMessages()` which restored messages but never validated `running` state. When SSE `session_end` was lost during iOS background (EventSource appeared alive, so no reconnect triggered), `running` stayed `true` indefinitely, showing the stop/interrupt/queue buttons instead of the send button.

## Test plan

- [x] New test: `syncs running state from session meta on foreground when session_end was missed`
- [x] New test: `does not change running state when session is still active on foreground`
- [x] All 345 client tests pass
- [ ] Manual: background iOS Safari during agent processing, return after agent finishes, verify send button appears

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -226,7 +226,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       .getSessionMeta(sessionId)
       .then((meta) => {
         if (!meta) return;
-        const { messages: msgState } = store.getState();
+        // Guard: if the user switched sessions while the fetch was in flight,
+        // don't apply stale isActive=false to the new session's running state.
+        const { sessions, messages: msgState } = store.getState();
+        if (sessions.active !== sessionId) return;
         if (msgState.running && !meta.isActive) {
           store.setState((s) => ({
             messages: messagesReducer(s.messages, { type: 'SET_RUNNING', running: false }),
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -216,6 +216,28 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       });
   }
 
+  /**
+   * Sync running state from server — fixes stale stop button after iOS
+   * backgrounding drops the session_end SSE event. Independent of the
+   * reconnect path so it works even when doReconnectPost has issues.
+   */
+  function syncRunningState(sessionId: string) {
+    api
+      .getSessionMeta(sessionId)
+      .then((meta) => {
+        if (!meta) return;
+        const { messages: msgState } = store.getState();
+        if (msgState.running && !meta.isActive) {
+          store.setState((s) => ({
+            messages: messagesReducer(s.messages, { type: 'SET_RUNNING', running: false }),
+          }));
+        }
+      })
+      .catch(() => {
+        // Non-fatal — running state will correct on next event or user action
+      });
+  }
+
   function clearPendingSendTimer() {
     if (parserState.pendingSendTimer) {
       clearTimeout(parserState.pendingSendTimer);
@@ -701,7 +723,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
     // the REST API if we have an active session but no messages in the store.
     if (msg.type === '_foreground') {
       const { sessions } = store.getState();
-      if (sessions.active) fetchAndRestoreMessages(sessions.active);
+      if (sessions.active) {
+        fetchAndRestoreMessages(sessions.active);
+        syncRunningState(sessions.active);
+      }
       return;
     }
 
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -190,6 +190,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
   };
 
   let recoveryInFlight = false;
+  let runningSyncInFlight = false;
 
   function fetchAndRestoreMessages(sessionId: string) {
     if (recoveryInFlight) return;
@@ -216,12 +217,10 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       });
   }
 
-  /**
-   * Sync running state from server — fixes stale stop button after iOS
-   * backgrounding drops the session_end SSE event. Independent of the
-   * reconnect path so it works even when doReconnectPost has issues.
-   */
+  /** Sync running state from server when session_end may have been missed. */
   function syncRunningState(sessionId: string) {
+    if (runningSyncInFlight) return;
+    runningSyncInFlight = true;
     api
       .getSessionMeta(sessionId)
       .then((meta) => {
@@ -238,6 +237,9 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       })
       .catch(() => {
         // Non-fatal — running state will correct on next event or user action
+      })
+      .finally(() => {
+        runningSyncInFlight = false;
       });
   }
 
--- a/packages/client/__tests__/store.test.ts
+++ b/packages/client/__tests__/store.test.ts
@@ -1408,4 +1408,30 @@ describe('foreground recovery', () => {
     // running should still be true — stale meta response was discarded
     expect(store.getState().messages.running).toBe(true);
   });
+
+  it('preserves running state when meta fetch fails (network error)', async () => {
+    const transport = mockTransport();
+    const store = createReadyStore(transport);
+
+    store.setState((s) => ({
+      sessions: { ...s.sessions, active: 'sess-1' },
+      messages: { ...s.messages, running: true },
+    }));
+
+    (transport.fetch as ReturnType<typeof vi.fn>).mockImplementation((url: string) => {
+      if (typeof url === 'string' && url.includes('/meta')) {
+        return Promise.reject(new Error('network error'));
+      }
+      return Promise.resolve({
+        ok: true,
+        json: () => Promise.resolve([]),
+      });
+    });
+
+    lastWs.simulateMessage({ type: '_foreground' });
+    await new Promise((r) => setTimeout(r, 50));
+
+    // running unchanged — fetch failure is non-fatal
+    expect(store.getState().messages.running).toBe(true);
+  });
 });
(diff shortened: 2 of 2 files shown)


dimakis (commented):
## Centaur Review

_Clean, well-tested fix for a real iOS backgrounding bug. One minor race condition: syncRunningState should verify the active session hasn't changed before applying the state update._

- 🟡 **unsafe_assumptions** (L230): Race condition: `syncRunningState` captures `sessionId` at call time but doesn't verify `sessions.active` still matches when the promise resolves. If the user switches to a new session (which resets then re-sets `running: true`) while the meta fetch for the old session is in flight, `!meta.isActive` for the old session will incorrectly clear `running` for the new session. Guard with `if (store.getState().sessions.active !== sessionId) return;` before applying the state change. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +216,28 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+   * backgrounding drops the session_end SSE event. Independent of the
+   * reconnect path so it works even when doReconnectPost has issues.
+   */
+  function syncRunningState(sessionId: string) {
+    api
+      .getSessionMeta(sessionId)
+      .then((meta) => {
+        if (!meta) return;
+        const { messages: msgState } = store.getState();
+        if (msgState.running && !meta.isActive) {
```
dimakis, line 232: 🟡 **unsafe_assumptions**: Race condition: `syncRunningState` captures `sessionId` at call time but doesn't verify `sessions.active` still matches when the promise resolves. If the user switches to a new session (which resets then re-sets `running: true`) while the meta fetch for the old session is in flight, `!meta.isActive` for the old session will incorrectly clear `running` for the new session. Guard with `if (store.getState().sessions.active !== sessionId) return;` before applying the state change. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +216,31 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       });
   }
 
+  /**
```
dimakis: 🔵 **style**: The docstring is 4 lines — longer than the project's 'one short line max' comment convention. The session-switch guard and iOS context are already clear from the code and the PR description; this could be a single line or omitted entirely. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +216,31 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+   * Sync running state from server — fixes stale stop button after iOS
+   * backgrounding drops the session_end SSE event. Independent of the
+   * reconnect path so it works even when doReconnectPost has issues.
+   */
+  function syncRunningState(sessionId: string) {
```
dimakis, line 221: 🔵 **unsafe_assumptions**: `syncRunningState` has no deduplication guard (unlike `fetchAndRestoreMessages` which uses `recoveryInFlight`). Rapid foreground events (e.g., quick app-switch back and forth) could fire multiple concurrent meta fetches. In practice the session-switch guard and idempotent `SET_RUNNING(false)` make this harmless, but it's an asymmetry worth noting. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Clean, well-tested fix for a real iOS issue. The session-switch race guard is correct. Only minor style and coverage suggestions — no bugs or regressions._

- 🔵 **style** (L219): The docstring is 4 lines — longer than the project's 'one short line max' comment convention. The session-switch guard and iOS context are already clear from the code and the PR description; this could be a single line or omitted entirely. `[fixable]`
- 🔵 **unsafe_assumptions** (L224): `syncRunningState` has no deduplication guard (unlike `fetchAndRestoreMessages` which uses `recoveryInFlight`). Rapid foreground events (e.g., quick app-switch back and forth) could fire multiple concurrent meta fetches. In practice the session-switch guard and idempotent `SET_RUNNING(false)` make this harmless, but it's an asymmetry worth noting. `[fixable]`


_Clean, well-tested fix for a real iOS issue. The session-switch race guard is correct. Only minor style and coverage suggestions — no bugs or regressions._

- 🔵 **missing_tests**: No test covers the case where `getSessionMeta` rejects (network error, 404). The implementation silently catches, which is correct, but a test asserting that running state is unchanged on fetch failure would document this contract and prevent regressions. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +217,32 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
       });
   }
 
+  /** Sync running state from server when session_end may have been missed. */
+  function syncRunningState(sessionId: string) {
+    if (runningSyncInFlight) return;
```
dimakis, line 222: 🔵 **unsafe_assumptions**: The `runningSyncInFlight` guard silently drops subsequent foreground events while a meta fetch is in flight. If a rapid background→foreground→background→foreground cycle occurs and the first fetch takes long, the second foreground recovery won't sync running state. This is consistent with how `recoveryInFlight` works for message recovery, so it's a known tradeoff — but worth noting that the guard could be replaced with an AbortController pattern that cancels the stale request and starts a new one, rather than dropping it. Not a bug in practice since iOS foreground events are infrequent. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +217,32 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+        // Guard: if the user switched sessions while the fetch was in flight,
+        // don't apply stale isActive=false to the new session's running state.
+        const { sessions, messages: msgState } = store.getState();
+        if (sessions.active !== sessionId) return;
+        if (msgState.running && !meta.isActive) {
+          store.setState((s) => ({
+            messages: messagesReducer(s.messages, { type: 'SET_RUNNING', running: false }),
+          }));
+        }
+      })
```
dimakis, line 237: 🔵 **style**: Empty `catch(() => {})` swallows errors silently. The parallel `fetchAndRestoreMessages` logs a `console.warn` on failure. Consider adding a matching `console.warn('[mitzo] running state sync failed', err)` for consistency and debuggability, especially since this is a new network call that could fail for various reasons on iOS. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix for a real iOS edge case. The session-switch guard, the one-directional correction (only clears running, never sets it), and the four test scenarios covering happy path, still-active, stale-session, and network-error all demonstrate solid defensive thinking. Two minor suggestions on consistency._

- 🔵 **unsafe_assumptions** (L222): The `runningSyncInFlight` guard silently drops subsequent foreground events while a meta fetch is in flight. If a rapid background→foreground→background→foreground cycle occurs and the first fetch takes long, the second foreground recovery won't sync running state. This is consistent with how `recoveryInFlight` works for message recovery, so it's a known tradeoff — but worth noting that the guard could be replaced with an AbortController pattern that cancels the stale request and starts a new one, rather than dropping it. Not a bug in practice since iOS foreground events are infrequent. `[fixable]`
- 🔵 **style** (L237): Empty `catch(() => {})` swallows errors silently. The parallel `fetchAndRestoreMessages` logs a `console.warn` on failure. Consider adding a matching `console.warn('[mitzo] running state sync failed', err)` for consistency and debuggability, especially since this is a new network call that could fail for various reasons on iOS. `[fixable]`

packages/client/src/store.ts
```diff
@@ -216,6 +217,32 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
+    runningSyncInFlight = true;
+    api
+      .getSessionMeta(sessionId)
+      .then((meta) => {
+        if (!meta) return;
+        // Guard: if the user switched sessions while the fetch was in flight,
+        // don't apply stale isActive=false to the new session's running state.
+        const { sessions, messages: msgState } = store.getState();
+        if (sessions.active !== sessionId) return;
+        if (msgState.running && !meta.isActive) {
```
dimakis, line 232: 🔵 **unsafe_assumptions**: Minor TOCTOU: `msgState.running` is read from a snapshot on line 230, but `setState` on line 233 uses a fresh `s.messages`. If `running` became `false` between the check and the dispatch (e.g., a normal `session_end` arrived in the gap), the reducer redundantly sets `running: false` — harmless but technically unnecessary. Could collapse the guard and dispatch into a single `setState` updater that checks `s.messages.running && !meta.isActive` atomically. Not a bug, just a minor cleanliness improvement. `[fixable]`

dimakis, line 232: 🔵 **style**: The condition `!meta.isActive` treats `undefined` as falsy. If the server ever returns a response without the `isActive` field (e.g., older server version), this would incorrectly clear `running`. Consider an explicit `meta.isActive === false` check for defensive clarity, consistent with the explicit boolean semantics of the `SessionMetaResponse` type. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Clean, well-tested fix with good race-condition guards. Two minor suggestions around atomicity and defensive boolean checks — no blocking issues._

- 🔵 **unsafe_assumptions** (L232): Minor TOCTOU: `msgState.running` is read from a snapshot on line 230, but `setState` on line 233 uses a fresh `s.messages`. If `running` became `false` between the check and the dispatch (e.g., a normal `session_end` arrived in the gap), the reducer redundantly sets `running: false` — harmless but technically unnecessary. Could collapse the guard and dispatch into a single `setState` updater that checks `s.messages.running && !meta.isActive` atomically. Not a bug, just a minor cleanliness improvement. `[fixable]`
- 🔵 **style** (L232): The condition `!meta.isActive` treats `undefined` as falsy. If the server ever returns a response without the `isActive` field (e.g., older server version), this would incorrectly clear `running`. Consider an explicit `meta.isActive === false` check for defensive clarity, consistent with the explicit boolean semantics of the `SessionMetaResponse` type. `[fixable]`

PR #410, opened 2026-06-27
feat(observability): session context inventory — span attrs + context consumed banner

dimakis:
## Summary

Two independent workstreams for session context inventory:

**WS1: Enriched Jaeger traces**
- New `getToolInputSpanAttrs()` extracts structured, searchable span attributes from tool inputs
- Tool spans carry `tool.input.file_path`, `tool.input.command`, `tool.input.pattern`, `tool.input.path`, `tool.input.query`, `tool.input.url`, `tool.input.description`, `tool.input.subagent_type`, `tool.input.skill`
- Values truncated to 256 chars, indexed by Jaeger for search/filter
- Raw input event preserved for full detail in span view

**WS2: Context consumed in session banner**
- Messages reducer accumulates context consumption from BLOCK_END events (files read, searches, web fetches)
- Deduped by type+key, tracks repeat counts
- Subagent tool calls included
- Rebuilds on RESTORE (reconnect/session switch), resets on CLEAR
- Renders as collapsible "Context Consumed" section in SessionBanner
- Collapsed summary shows unique context count alongside boot context stats

## Test plan
- [x] 15 new tool-summary-attrs tests (all tool types, edge cases, truncation)
- [x] 7 new messages-slice tests (accumulation, dedup, multi-type, subagents, RESTORE, CLEAR)
- [ ] Deploy and verify span attrs appear in Jaeger UI
- [ ] Verify context consumed section renders in session banner

### DIFF
--- a/packages/protocol/src/tool-summary.ts
+++ b/packages/protocol/src/tool-summary.ts
@@ -71,7 +71,7 @@ export function getToolInputSpanAttrs(
 ): Record<string, string> {
   const attrs: Record<string, string> = {};
   const set = (key: string, val: unknown) => {
-    const s = String(val || '');
+    const s = String(val ?? '');
     if (s) attrs[`tool.input.${key}`] = s.slice(0, SPAN_ATTR_MAX);
   };
 
--- a/packages/protocol/src/tool-summary.ts
+++ b/packages/protocol/src/tool-summary.ts
@@ -58,6 +58,66 @@ export function getRawInput(
   }
 }
 
+const SPAN_ATTR_MAX = 256;
+
+/**
+ * Extract structured span attributes from tool input for Jaeger indexing.
+ * Returns flat `tool.input.*` key-value pairs that are searchable/filterable
+ * in Jaeger UI. Values are truncated to keep span metadata lightweight.
+ */
+export function getToolInputSpanAttrs(
+  toolName: string,
+  input: Record<string, unknown>,
+): Record<string, string> {
+  const attrs: Record<string, string> = {};
+  const set = (key: string, val: unknown) => {
+    const s = String(val || '');
+    if (s) attrs[`tool.input.${key}`] = s.slice(0, SPAN_ATTR_MAX);
+  };
+
+  switch (toolName) {
+    case 'Read':
+      set('file_path', input.file_path);
+      break;
+    case 'Write':
+      set('file_path', input.file_path);
+      break;
+    case 'Edit':
+    case 'StrReplace':
+      set('file_path', input.file_path);
+      break;
+    case 'Bash':
+    case 'Shell':
+      set('command', input.command);
+      break;
+    case 'Glob':
+      set('pattern', input.pattern);
+      if (input.path) set('path', input.path);
+      break;
+    case 'Grep':
+      set('pattern', input.pattern);
+      if (input.path) set('path', input.path);
+      break;
+    case 'WebSearch':
+      set('query', input.search_term);
+      break;
+    case 'WebFetch':
+      set('url', input.url);
+      break;
+    case 'Agent':
+      set('description', input.description);
+      if (input.subagent_type) set('subagent_type', input.subagent_type);
+      break;
+    case 'Skill':
+      set('skill', input.skill);
+      break;
+    default:
+      break;
+  }
+
+  return attrs;
+}
+
 export function summarizeToolInput(toolName: string, input: Record<string, unknown>): string {
   switch (toolName) {
     case 'Read':
--- a/packages/protocol/src/tool-summary.ts
+++ b/packages/protocol/src/tool-summary.ts
@@ -77,11 +77,7 @@ export function getToolInputSpanAttrs(
 
   switch (toolName) {
     case 'Read':
-      set('file_path', input.file_path);
-      break;
     case 'Write':
-      set('file_path', input.file_path);
-      break;
     case 'Edit':
     case 'StrReplace':
       set('file_path', input.file_path);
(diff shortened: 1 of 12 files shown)


packages/protocol/src/tool-summary.ts
```diff
@@ -58,6 +58,66 @@ export function getRawInput(
...
+  input: Record<string, unknown>,
+): Record<string, string> {
+  const attrs: Record<string, string> = {};
+  const set = (key: string, val: unknown) => {
+    const s = String(val || '');
+    if (s) attrs[`tool.input.${key}`] = s.slice(0, SPAN_ATTR_MAX);
+  };
+
+  switch (toolName) {
+    case 'Read':
```
dimakis, line 79: 🔵 **style**: The `Read`, `Write`, `Edit`, and `StrReplace` cases all extract the same `file_path` field — they can be collapsed into a single fallthrough block (`case 'Read': case 'Write': case 'Edit': case 'StrReplace':`), matching how `Edit`/`StrReplace` are already grouped. Same applies to `Glob`/`Grep` (both extract `pattern` + optional `path`). `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


### `packages/protocol/src/tool-summary.ts`
_Clean, well-tested feature addition. The new function follows existing patterns in tool-summary.ts, tests are thorough (all tools, edge cases, truncation, unknowns), integration in query-loop.ts is correct, and exports are properly wired. Only a minor style nit on collapsible switch cases._

- 🔵 **style** (L79): The `Read`, `Write`, `Edit`, and `StrReplace` cases all extract the same `file_path` field — they can be collapsed into a single fallthrough block (`case 'Read': case 'Write': case 'Edit': case 'StrReplace':`), matching how `Edit`/`StrReplace` are already grouped. Same applies to `Glob`/`Grep` (both extract `pattern` + optional `path`). `[fixable]`

dimakis:
## Centaur Review

_Solid feature — well-tested reducer logic and clean span attribute extraction, but DesktopChatView is missing the `contextConsumed` prop, leaving desktop users without the new UI section._

- 🟡 **regressions** (L244): DesktopChatView renders ChatArea but does not extract or pass `contextConsumed` from the store, unlike ChatView (which does at line 56/286). Desktop users will never see the context consumed section in SessionBanner. Likely needs `const contextConsumed = useMitzoStore((s) => s.messages.contextConsumed);` and `contextConsumed={contextConsumed}` on the ChatArea call. `[fixable]`


_Solid feature — well-tested reducer logic and clean span attribute extraction, but DesktopChatView is missing the `contextConsumed` prop, leaving desktop users without the new UI section._

- 🔵 **style** (L535): CLEAR handler explicitly sets `contextConsumed: []` but this is already the value in `INITIAL_MESSAGES_STATE` (which is spread on the same line). The override is redundant. `[fixable]`
- 🔵 **style** (L37): In `extractContextEntry`, the `input ?` ternary checks on lines 37/40/41/42 are redundant — line 34 already returns null when `!input`. Each case can simply return the object directly. `[fixable]`


### `packages/protocol/src/tool-summary.ts`
_Solid feature — well-tested reducer logic and clean span attribute extraction, but DesktopChatView is missing the `contextConsumed` prop, leaving desktop users without the new UI section._

- 🔵 **style** (L81): Read/Write cases can be merged with Edit/StrReplace since all three just `set('file_path', input.file_path)`. This would reduce the switch from 5 case arms to 3 (e.g. `case 'Read': case 'Write': case 'Edit': case 'StrReplace':`). `[fixable]`
- 🔵 **bugs** (L73): `String(val || '')` coerces falsy non-string values (0, false) to empty string, silently dropping them. In practice this is harmless since all currently extracted fields are strings, but `String(val ?? '')` would be more precise and future-proof. This matches the idiom used elsewhere in the codebase. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

dimakis:
## Centaur Review

### `server/query-loop.ts`
_Solid feature — well-tested reducer logic and clean span attribute extraction. Main gap: subagent tool spans in query-loop don't get the new `getToolInputSpanAttrs` attributes despite parsed input being available, creating an observability blind spot for subagent tool calls in Jaeger._

- 🟡 **bugs** (L1071): Subagent tool spans don't get `getToolInputSpanAttrs` attributes, unlike main tool spans. The parsed `toolInput` is already available at line 1049 — the same `getToolInputSpanAttrs` call and attribute-setting loop should be added here for consistency. Without it, subagent tool calls in Jaeger are missing the searchable `tool.input.*` attributes and the `tool.input` event, making them harder to find and debug. `[fixable]`


_Solid feature — well-tested reducer logic and clean span attribute extraction. Main gap: subagent tool spans in query-loop don't get the new `getToolInputSpanAttrs` attributes despite parsed input being available, creating an observability blind spot for subagent tool calls in Jaeger._

- 🔵 **missing_tests**: The `contextConsumed` tests cover Read, Grep, and WebFetch but never test WebSearch. The `extractContextEntry` function maps WebSearch → `web_search`, and ContextConsumedSection has a label for it, but there's no test confirming this mapping works end-to-end through the reducer. `[fixable]`


### `frontend/src/components/SessionBanner.tsx`
_Solid feature — well-tested reducer logic and clean span attribute extraction. Main gap: subagent tool spans in query-loop don't get the new `getToolInputSpanAttrs` attributes despite parsed input being available, creating an observability blind spot for subagent tool calls in Jaeger._

- 🔵 **style** (L142): When `bootContext` and `sessionContext` are both null but `contextConsumed` has entries, the banner renders with the label "Session Context" and a warning-colored dot (`session-banner-dot--warn`). This is misleading — the user sees a yellow warning indicator for what is normal runtime context. Consider showing a neutral label like "Context Inventory" and a neutral dot style when only `contextConsumed` is present. `[fixable]`
- 🔵 **style** (L97): In `ContextConsumedSection`, items use array index as the React key (`key={idx}`). Since each `ContextEntry` has a unique `key` field within its type group, using `${type}:${item.key}` would be a more stable key — though the current approach is harmless since the list is append-only. `[fixable]`

PR #411, opened 2026-06-27
fix: query-loop review — update comment + external abort test

dimakis:
## Summary
- Updated `finalSession` comment in `query-loop.ts` to document both read sites after `registry.remove()`
- Added test for external abort-controller cancellation path (fallback usage records `completed` reason)

Rescued from stale local main — was committed directly instead of via branch+PR.

## Test plan
- [ ] `npm test` passes

### DIFF
--- a/server/__tests__/query-loop.test.ts
+++ b/server/__tests__/query-loop.test.ts
@@ -1614,6 +1614,49 @@ describe('runQueryLoop', () => {
       expect(sessionMeta!.totalCostUsd).toBe(0);
     });
 
+    it('records fallback usage with completed reason on external abort', async () => {
+      const store = new EventStore(':memory:');
+      const connRegistry = new ConnectionRegistry();
+      const v2Transport = fakeTransport();
+      connRegistry.register(clientId, v2Transport);
+      connRegistry.watch(clientId, 'sess-usage-abort');
+      connRegistry.setActive(clientId, 'sess-usage-abort');
+
+      const session = registry.get(clientId)!;
+      session.sessionId = 'sess-usage-abort';
+
+      store.upsertSession({ sessionId: 'sess-usage-abort', cwd: '/tmp' });
+
+      // Stream that yields an assistant event then hangs — abort controller
+      // cancels the session externally (no error thrown).
+      async function* hangingStream() {
+        yield {
+          type: 'assistant',
+          session_id: 'sess-usage-abort',
+          message: { id: 'msg-1', role: 'assistant' },
+        };
+        // Simulate a hanging stream that gets cancelled
+        await new Promise((_, reject) => {
+          abortController.signal.addEventListener('abort', () => reject(new Error('aborted')));
+        });
+      }
+
+      // Abort after a tick to let the first event process
+      setTimeout(() => abortController.abort(), 10);
+
+      await runQueryLoop(hangingStream(), clientId, registry, abortController, store, undefined, {
+        connRegistry,
+      });
+
+      const sessionMeta = store.getSession('sess-usage-abort');
+      expect(sessionMeta).toBeDefined();
+      expect(sessionMeta!.state).toBe('ENDED');
+      expect(sessionMeta!.durationMs).toBeGreaterThanOrEqual(0);
+      expect(sessionMeta!.outputTokens).toBe(0);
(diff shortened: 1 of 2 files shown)


dimakis (commented):
## Centaur Review

_Comment update is fine. New test has a misleading title — the abort path enters the catch block (reason='error', not 'completed'), and the test doesn't assert on the reason field to catch the discrepancy._

- 🟡 **bugs** (L1617): Test title says 'records fallback usage with completed reason on external abort' but the hangingStream rejects with an Error on abort, which enters the catch block (setting caughtError=true), so the actual reason recorded at query-loop.ts:1404 is 'error', not 'completed'. The test doesn't assert on the reason field, so it passes despite the misleading name. Either the title should say 'error reason' or the test should assert `sessionMeta!.reason` to verify the intended behavior. `[fixable]`
- 🔵 **missing_tests** (L1651): The test omits an assertion on the session's reason field (the existing sibling test at line 1610 also omits it). Since the test title specifically claims 'completed reason', adding `expect(sessionMeta!.reason).toBe(...)` would make the test actually verify what it claims to test. `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -1614,6 +1614,49 @@ describe('runQueryLoop', () => {
       expect(sessionMeta!.totalCostUsd).toBe(0);
     });
 
+    it('records fallback usage with completed reason on external abort', async () => {
```
dimakis: 🟡 **bugs**: Test title says 'records fallback usage with completed reason on external abort' but the hangingStream rejects with an Error on abort, which enters the catch block (setting caughtError=true), so the actual reason recorded at query-loop.ts:1404 is 'error', not 'completed'. The test doesn't assert on the reason field, so it passes despite the misleading name. Either the title should say 'error reason' or the test should assert `sessionMeta!.reason` to verify the intended behavior. `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -1614,6 +1614,49 @@ describe('runQueryLoop', () => {
...
+      }
+
+      // Abort after a tick to let the first event process
+      setTimeout(() => abortController.abort(), 10);
+
+      await runQueryLoop(hangingStream(), clientId, registry, abortController, store, undefined, {
+        connRegistry,
+      });
+
+      const sessionMeta = store.getSession('sess-usage-abort');
```
dimakis, line 1651: 🔵 **missing_tests**: The test omits an assertion on the session's reason field (the existing sibling test at line 1610 also omits it). Since the test title specifically claims 'completed reason', adding `expect(sessionMeta!.reason).toBe(...)` would make the test actually verify what it claims to test. `[fixable]`

dimakis (commented):
## Centaur Review

Found **1** issue(s).


_Clean PR — the new test covers the external-abort path well, and the comment update accurately describes the two post-remove readers. One minor title/assertion alignment nit._

- 🔵 **missing_tests** (L1617): Test title says 'with error reason' but the reason field is never asserted. The `reason` passed to `setSessionState` is only logged, not persisted on `SessionMeta`, so it can't be checked via `getSession()`. Consider either removing 'error reason' from the title (e.g., 'records fallback usage on external abort') or verifying the reason indirectly (e.g., spy on the store's `setSessionState` to confirm it receives `reason: 'error'`). `[fixable]`

server/__tests__/query-loop.test.ts
```diff
@@ -1614,6 +1614,49 @@ describe('runQueryLoop', () => {
       expect(sessionMeta!.totalCostUsd).toBe(0);
     });
 
+    it('records fallback usage with error reason on external abort', async () => {
```
dimakis: 🔵 **missing_tests**: Test title says 'with error reason' but the reason field is never asserted. The `reason` passed to `setSessionState` is only logged, not persisted on `SessionMeta`, so it can't be checked via `getSession()`. Consider either removing 'error reason' from the title (e.g., 'records fallback usage on external abort') or verifying the reason indirectly (e.g., spy on the store's `setSessionState` to confirm it receives `reason: 'error'`). `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #414, opened 2026-06-27
feat(task-board): idempotent task creation via external_ref

dimakis:
## Summary

- Adds `external_ref` column to task store with a partial unique index (WHERE NOT NULL)
- Middleware on `POST /api/tasks` checks for existing task by `externalRef` before creating — returns 200 + `deduplicated: true` instead of a duplicate 201
- DB constraint is the safety net for concurrent requests; middleware provides the clean UX
- Migration 2 added for existing databases

Companion PR: dimakis/mgmt — wires `external_ref` into PR Shepherd goal creation and `mitzo_client.create_task()`

## Test plan

- [x] TaskStore: create with externalRef, getByExternalRef, rejects duplicates, allows null refs
- [x] Routes: POST deduplicates by externalRef, no dedup without ref
- [x] All 70 existing tests pass

### DIFF
--- a/server/__tests__/task-routes.test.ts
+++ b/server/__tests__/task-routes.test.ts
@@ -158,6 +158,41 @@ describe('task routes', () => {
     expect(res.body.task.annotations).toEqual(['a']);
   });
 
+  // --- POST /api/tasks idempotency ---
+
+  it('POST /api/tasks — deduplicates by externalRef', async () => {
+    const ref = `dedup-test-${Date.now()}`;
+    const first = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'First create', externalRef: ref })
+      .set('Cookie', authCookie);
+    expect(first.status).toBe(201);
+    expect(first.body.task.externalRef).toBe(ref);
+
+    const second = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'Duplicate create', externalRef: ref })
+      .set('Cookie', authCookie);
+    expect(second.status).toBe(200);
+    expect(second.body.deduplicated).toBe(true);
+    expect(second.body.task.id).toBe(first.body.task.id);
+    expect(second.body.task.title).toBe('First create');
+  });
+
+  it('POST /api/tasks — no dedup without externalRef', async () => {
+    const first = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'No ref A' })
+      .set('Cookie', authCookie);
+    const second = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'No ref B' })
+      .set('Cookie', authCookie);
+    expect(first.status).toBe(201);
+    expect(second.status).toBe(201);
+    expect(first.body.task.id).not.toBe(second.body.task.id);
+  });
+
   // --- GET /api/tasks/:id ---
 
   it('GET /api/tasks/:id — returns a task', async () => {
--- a/server/__tests__/task-routes.test.ts
+++ b/server/__tests__/task-routes.test.ts
@@ -179,6 +179,14 @@ describe('task routes', () => {
     expect(second.body.task.title).toBe('First create');
   });
 
+  it('POST /api/tasks — rejects empty externalRef', async () => {
+    const res = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'Empty ref', externalRef: '' })
+      .set('Cookie', authCookie);
+    expect(res.status).toBe(400);
+  });
+
   it('POST /api/tasks — no dedup without externalRef', async () => {
     const first = await request(app)
       .post('/api/tasks')
--- a/server/__tests__/task-store.test.ts
+++ b/server/__tests__/task-store.test.ts
@@ -509,6 +509,41 @@ describe('TaskStore', () => {
     store2.close();
   });
 
+  // --- externalRef ---
+
+  it('creates a task with externalRef', () => {
+    const task = store.create({ title: 'Ref task', externalRef: 'pr_shepherd:org/repo#42' });
+    expect(task.externalRef).toBe('pr_shepherd:org/repo#42');
+  });
+
+  it('defaults externalRef to null', () => {
+    const task = store.create({ title: 'No ref' });
+    expect(task.externalRef).toBeNull();
+  });
+
+  it('getByExternalRef finds a task', () => {
+    store.create({ title: 'Findable', externalRef: 'test:find-me' });
+    const found = store.getByExternalRef('test:find-me');
+    expect(found).not.toBeNull();
+    expect(found!.title).toBe('Findable');
+  });
+
+  it('getByExternalRef returns null for unknown ref', () => {
+    expect(store.getByExternalRef('nonexistent')).toBeNull();
+  });
+
+  it('rejects duplicate externalRef', () => {
+    store.create({ title: 'First', externalRef: 'unique:ref' });
+    expect(() => store.create({ title: 'Second', externalRef: 'unique:ref' })).toThrow();
+  });
+
+  it('allows multiple tasks with null externalRef', () => {
+    store.create({ title: 'A' });
+    store.create({ title: 'B' });
+    const roots = store.listRoots();
+    expect(roots.length).toBeGreaterThanOrEqual(2);
+  });
+
   it('migrates existing database to add workflow columns', () => {
     // Create a DB without workflow columns, then reopen to trigger migration
     const dbPath = join(TEST_DIR, 'migrate.db');
--- a/server/__tests__/task-store.test.ts
+++ b/server/__tests__/task-store.test.ts
@@ -561,6 +561,24 @@ describe('TaskStore', () => {
     store2.close();
   });
 
+  it('migrates existing database to add external_ref column', () => {
+    const dbPath = join(TEST_DIR, 'migrate-extref.db');
+    const store1 = new TaskStore(dbPath);
+    const t = store1.create({ title: 'Pre-extref task' });
+    store1.close();
+
+    // Reopen — migration 2 should add external_ref
+    const store2 = new TaskStore(dbPath);
+    const task = store2.get(t.id);
+    expect(task!.externalRef).toBeNull();
+
+    // Verify the column works after migration
+    const refTask = store2.create({ title: 'With ref', externalRef: 'migrated:ref' });
+    expect(refTask.externalRef).toBe('migrated:ref');
+    expect(store2.getByExternalRef('migrated:ref')!.id).toBe(refTask.id);
+    store2.close();
+  });
+
   it('preserves workflow fields in tree assembly', () => {
     const goal = store.create({ title: 'Goal', stageType: 'agent_work' });
     store.create({
--- a/server/api-schemas.ts
+++ b/server/api-schemas.ts
@@ -123,6 +123,7 @@ export const TaskCreateBody = z.object({
   gateConfig: z.record(z.string(), z.unknown()).optional(),
   maxRetries: z.number().min(0).optional(),
   templateId: z.string().optional(),
+  externalRef: z.string().optional(),
 });
 
 export const TaskUpdateBody = z.object({
--- a/server/api-schemas.ts
+++ b/server/api-schemas.ts
@@ -123,7 +123,7 @@ export const TaskCreateBody = z.object({
   gateConfig: z.record(z.string(), z.unknown()).optional(),
   maxRetries: z.number().min(0).optional(),
   templateId: z.string().optional(),
-  externalRef: z.string().optional(),
+  externalRef: z.string().min(1).optional(),
 });
 
 export const TaskUpdateBody = z.object({
--- a/server/app.ts
+++ b/server/app.ts
@@ -635,6 +635,14 @@ app.post('/api/tasks', (req, res) => {
     return;
   }
   try {
+    // Idempotency guard: if externalRef provided and already exists, return existing task
+    if (body.data.externalRef) {
+      const existing = taskStore.getByExternalRef(body.data.externalRef);
+      if (existing) {
+        res.status(200).json({ task: existing, deduplicated: true });
+        return;
+      }
+    }
     const task = taskStore.create(body.data as TaskCreateInput);
     res.status(201).json({ task });
     // Broadcast full tree so child tasks appear correctly in all clients
--- a/server/task-store.ts
+++ b/server/task-store.ts
@@ -305,9 +305,9 @@ export class TaskStore {
 
   /** Look up a task by its external reference (e.g. "pr_shepherd:dimakis/centaur#42"). */
   getByExternalRef(ref: string): Task | null {
-    const row = this.getDb()
-      .prepare('SELECT * FROM tasks WHERE external_ref = ?')
-      .get(ref) as TaskRow | undefined;
+    const row = this.getDb().prepare('SELECT * FROM tasks WHERE external_ref = ?').get(ref) as
+      | TaskRow
+      | undefined;
     return row ? rowToTask(row) : null;
   }
 
--- a/server/task-store.ts
+++ b/server/task-store.ts
@@ -49,6 +49,7 @@ export interface Task {
   retryCount: number;
   maxRetries: number;
   templateId: string | null;
+  externalRef: string | null;
   children: Task[];
 }
 
@@ -63,6 +64,7 @@ export interface TaskCreateInput {
   gateConfig?: GateConfig;
   maxRetries?: number;
   templateId?: string;
+  externalRef?: string;
 }
 
 export interface TaskUpdateInput {
@@ -106,6 +108,7 @@ interface TaskRow {
   retry_count: number;
   max_retries: number;
   template_id: string | null;
+  external_ref: string | null;
 }
 
 const SCHEMA = `
@@ -137,11 +140,13 @@ const SCHEMA = `
     artifacts       TEXT DEFAULT NULL,
     retry_count     INTEGER NOT NULL DEFAULT 0,
     max_retries     INTEGER NOT NULL DEFAULT 0,
-    template_id     TEXT DEFAULT NULL
+    template_id     TEXT DEFAULT NULL,
+    external_ref    TEXT DEFAULT NULL
   );
   CREATE INDEX IF NOT EXISTS idx_tasks_parent ON tasks(parent_id);
   CREATE INDEX IF NOT EXISTS idx_tasks_status ON tasks(status);
   CREATE INDEX IF NOT EXISTS idx_tasks_session ON tasks(session_id);
+  CREATE UNIQUE INDEX IF NOT EXISTS idx_tasks_external_ref ON tasks(external_ref) WHERE external_ref IS NOT NULL;
 `;
 
 const MIGRATIONS = [
@@ -156,6 +161,14 @@ const MIGRATIONS = [
       ALTER TABLE tasks ADD COLUMN retry_count INTEGER NOT NULL DEFAULT 0;
       ALTER TABLE tasks ADD COLUMN max_retries INTEGER NOT NULL DEFAULT 0;
       ALTER TABLE tasks ADD COLUMN template_id TEXT DEFAULT NULL;
+    `,
+  },
+  // Migration 2: Add external_ref for idempotent task creation
+  {
+    check: "SELECT COUNT(*) as cnt FROM pragma_table_info('tasks') WHERE name = 'external_ref'",
+    sql: `
+      ALTER TABLE tasks ADD COLUMN external_ref TEXT DEFAULT NULL;
+      CREATE UNIQUE INDEX IF NOT EXISTS idx_tasks_external_ref ON tasks(external_ref) WHERE external_ref IS NOT NULL;
     `,
   },
 ];
@@ -194,6 +207,7 @@ function rowToTask(row: TaskRow): Task {
     retryCount: row.retry_count ?? 0,
     maxRetries: row.max_retries ?? 0,
     templateId: row.template_id ?? null,
+    externalRef: row.external_ref ?? null,
     children: [],
   };
 }
@@ -259,8 +273,8 @@ export class TaskStore {
     const gateConfig = input.gateConfig ? JSON.stringify(input.gateConfig) : null;
 
     db.prepare(
-      `INSERT INTO tasks (id, parent_id, title, description, status, session_policy, priority, depth, annotations, stage_type, gate_config, max_retries, template_id, created_at, updated_at)
-       VALUES (?, ?, ?, ?, 'pending', ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+      `INSERT INTO tasks (id, parent_id, title, description, status, session_policy, priority, depth, annotations, stage_type, gate_config, max_retries, template_id, external_ref, created_at, updated_at)
+       VALUES (?, ?, ?, ?, 'pending', ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
     ).run(
       id,
       input.parentId ?? null,
@@ -274,6 +288,7 @@ export class TaskStore {
       gateConfig,
       input.maxRetries ?? 0,
       input.templateId ?? null,
+      input.externalRef ?? null,
       now,
       now,
     );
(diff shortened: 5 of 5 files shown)


server/api-schemas.ts
```diff
@@ -123,6 +123,7 @@ export const TaskCreateBody = z.object({
   gateConfig: z.record(z.string(), z.unknown()).optional(),
   maxRetries: z.number().min(0).optional(),
   templateId: z.string().optional(),
+  externalRef: z.string().optional(),
```
dimakis: 🔵 **unsafe_assumptions**: `externalRef` is validated as `z.string().optional()` — this accepts empty strings. An empty string is non-NULL in SQLite, so it would hit the UNIQUE constraint on a second task with `externalRef: ""`. Consider adding `.min(1)` (like `title` on line 116) to prevent ambiguous empty-string refs. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean, well-tested feature. The main gap is that frontend Task types in `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts` were not updated to include `externalRef`, creating a server/client type mismatch._

- 🟡 **regressions**: The server-side `Task` interface now includes `externalRef: string | null`, but the frontend Task type (`frontend/src/types/task.ts` line 48) and the client package Task type (`packages/client/src/slices/tasks.ts` line 39) were not updated. The server API returns `externalRef` in all task responses, creating a type mismatch. Any future frontend code referencing `task.externalRef` will get a TypeScript error despite the field being present at runtime. `[fixable]`


### `server/api-schemas.ts`
_Clean, well-tested feature. The main gap is that frontend Task types in `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts` were not updated to include `externalRef`, creating a server/client type mismatch._

- 🔵 **unsafe_assumptions** (L126): `externalRef` is validated as `z.string().optional()` — this accepts empty strings. An empty string is non-NULL in SQLite, so it would hit the UNIQUE constraint on a second task with `externalRef: ""`. Consider adding `.min(1)` (like `title` on line 116) to prevent ambiguous empty-string refs. `[fixable]`


### `server/__tests__/task-store.test.ts`
_Clean, well-tested feature. The main gap is that frontend Task types in `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts` were not updated to include `externalRef`, creating a server/client type mismatch._

- 🔵 **missing_tests**: No migration test for the new `external_ref` column. The existing migration test (line 512) verifies workflow columns survive reopen, but doesn't assert `externalRef` is populated correctly after migration 2 runs. Following the same pattern, a test should create a DB, close it, reopen (triggering migration 2), and verify `task.externalRef` is `null` for pre-existing rows. `[fixable]`

dimakis:
## Centaur Review

_Solid implementation with good test coverage and clean migration. The main gap is that both frontend `Task` type definitions (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) are missing the new `externalRef` field, creating a type contract mismatch with the server._

- 🟡 **regressions** (L49): Frontend `Task` interface is missing the new `externalRef: string | null` field. The server now returns this field in every task JSON response, but TypeScript on the frontend doesn't declare it — any future code that accesses `task.externalRef` will fail type-checking. `[fixable]`


### `packages/client/src/slices/tasks.ts`
_Solid implementation with good test coverage and clean migration. The main gap is that both frontend `Task` type definitions (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) are missing the new `externalRef` field, creating a type contract mismatch with the server._

- 🟡 **regressions** (L39): Same issue in `@mitzo/client`: the `Task` interface here also lacks `externalRef: string | null`. Both frontend type definitions need to stay in sync with the server's `Task` type. `[fixable]`


### `server/__tests__/task-store.test.ts`
_Solid implementation with good test coverage and clean migration. The main gap is that both frontend `Task` type definitions (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) are missing the new `externalRef` field, creating a type contract mismatch with the server._

- 🔵 **missing_tests** (L547): The existing migration test ('migrates existing database to add workflow columns') doesn't assert that `externalRef` defaults to null after migration 2 runs. Adding `expect(task!.externalRef).toBeNull()` would verify the new migration works on pre-existing databases. `[fixable]`


### `server/__tests__/task-routes.test.ts`
_Solid implementation with good test coverage and clean migration. The main gap is that both frontend `Task` type definitions (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) are missing the new `externalRef` field, creating a type contract mismatch with the server._

- 🔵 **missing_tests** (L163): No test for the case where a deduplicated task is returned but the caller sent different fields (e.g. different `parentId`, `priority`, or `description`). The current idempotency guard silently ignores mismatched fields and returns the original — this is fine if intentional, but a test documenting this 'first-write-wins' behavior would prevent future confusion. `[fixable]`

server/__tests__/task-routes.test.ts
```diff
@@ -158,6 +158,41 @@ describe('task routes', () => {
...
+      .send({ title: 'No ref A' })
+      .set('Cookie', authCookie);
+    const second = await request(app)
+      .post('/api/tasks')
+      .send({ title: 'No ref B' })
+      .set('Cookie', authCookie);
+    expect(first.status).toBe(201);
+    expect(second.status).toBe(201);
+    expect(first.body.task.id).not.toBe(second.body.task.id);
+  });
```
dimakis, line 202: 🔵 **missing_tests**: No test that `externalRef: ""` (empty string) is rejected with 400. Commit 0ece048 specifically added `.min(1)` to fix this — a regression test would prevent that validator from being accidentally weakened. `[fixable]`

server/__tests__/task-store.test.ts
```diff
@@ -509,6 +509,41 @@ describe('TaskStore', () => {
...
+    store.create({ title: 'First', externalRef: 'unique:ref' });
+    expect(() => store.create({ title: 'Second', externalRef: 'unique:ref' })).toThrow();
+  });
+
+  it('allows multiple tasks with null externalRef', () => {
+    store.create({ title: 'A' });
+    store.create({ title: 'B' });
+    const roots = store.listRoots();
+    expect(roots.length).toBeGreaterThanOrEqual(2);
+  });
```
dimakis, line 545: 🔵 **missing_tests**: No migration test for the new `external_ref` column. The existing `migrates existing database to add workflow columns` test covers migration 1 by creating a DB without the column and reopening it. A parallel test for migration 2 would ensure the ALTER TABLE + CREATE UNIQUE INDEX migration runs correctly on pre-existing databases. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `server/__tests__/task-routes.test.ts`
_Clean, well-structured PR. Schema, migration, idempotency guard, and test coverage are all solid. Two minor test gaps — an empty-string rejection regression test and a migration-specific test — would harden the feature further._

- 🔵 **missing_tests** (L194): No test that `externalRef: ""` (empty string) is rejected with 400. Commit 0ece048 specifically added `.min(1)` to fix this — a regression test would prevent that validator from being accidentally weakened. `[fixable]`


### `server/__tests__/task-store.test.ts`
_Clean, well-structured PR. Schema, migration, idempotency guard, and test coverage are all solid. Two minor test gaps — an empty-string rejection regression test and a migration-specific test — would harden the feature further._

- 🔵 **missing_tests** (L545): No migration test for the new `external_ref` column. The existing `migrates existing database to add workflow columns` test covers migration 1 by creating a DB without the column and reopening it. A parallel test for migration 2 would ensure the ALTER TABLE + CREATE UNIQUE INDEX migration runs correctly on pre-existing databases. `[fixable]`

dimakis:
## Centaur Review

_Solid, well-tested feature. The main gap is that the frontend Task types (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) were not updated to include `externalRef`, which will cause type mismatches if the field is accessed on the client side._

- 🟡 **regressions** (L48): The server `Task` interface now includes `externalRef: string | null`, but the frontend type at line 48 (and `packages/client/src/slices/tasks.ts` line 38) still end at `templateId`. Both need `externalRef: string | null` added to stay in sync — without it, any frontend code that tries to read `task.externalRef` (e.g. for display on the task board) won't type-check. `[fixable]`


### `server/app.ts`
_Solid, well-tested feature. The main gap is that the frontend Task types (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) were not updated to include `externalRef`, which will cause type mismatches if the field is accessed on the client side._

- 🔵 **unsafe_assumptions** (L639): TOCTOU gap: between the `getByExternalRef` check (line 640) and the `create` call (line 646), a concurrent request with the same `externalRef` could insert first. The UNIQUE index protects data integrity (the second insert throws), but the caller gets an opaque SQLite constraint error (400) instead of the clean `{ deduplicated: true, task }` response. For a single-user app this is low-risk, but wrapping the check+insert in a transaction or catching the constraint violation with a retry/lookup would make the idempotency guarantee airtight. `[fixable]`


### `server/__tests__/task-routes.test.ts`
_Solid, well-tested feature. The main gap is that the frontend Task types (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) were not updated to include `externalRef`, which will cause type mismatches if the field is accessed on the client side._

- 🔵 **missing_tests** (L163): The dedup route test only verifies `id` and `title` of the returned task. Consider also asserting that the second request's different body fields (e.g. `title: 'Duplicate create'`) do NOT overwrite the original — i.e. the existing task is returned unmodified. A test for `deduplicated` on the first (non-dedup) response being absent (`expect(first.body.deduplicated).toBeUndefined()`) would also strengthen the contract. `[fixable]`


### `server/__tests__/task-store.test.ts`
_Solid, well-tested feature. The main gap is that the frontend Task types (`frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) were not updated to include `externalRef`, which will cause type mismatches if the field is accessed on the client side._

- 🔵 **missing_tests** (L564): The migration test creates a fresh DB (which already gets the full schema including `external_ref`), so the migration check finds the column exists and skips the ALTER TABLE path entirely. It tests idempotent reopen, not actual migration. To test the real migration path, the test would need to create a DB with a schema that lacks `external_ref` (e.g. drop the column manually before reopening). This matches the pattern of the existing migration-1 test, so it's a pre-existing limitation, not introduced by this PR. `[fixable]`

dimakis:
## Centaur Review

_Clean, well-tested idempotency feature. The main issue is that the frontend `Task` type (in both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) was not updated to include the new `externalRef` field, creating a type-level inconsistency with the server API response._

- 🟡 **regressions** (L48): The server `Task` interface now includes `externalRef: string | null`, but the frontend `Task` type is missing it. The API response will include the field; the frontend type should stay in sync. Same issue in `packages/client/src/slices/tasks.ts` (line 39). `[fixable]`


### `server/__tests__/task-store.test.ts`
_Clean, well-tested idempotency feature. The main issue is that the frontend `Task` type (in both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) was not updated to include the new `externalRef` field, creating a type-level inconsistency with the server API response._

- 🔵 **missing_tests**: No test verifies that `getByExternalRef` leverages the unique index efficiently (e.g. with many tasks). More importantly, there's no test that `externalRef` uniqueness is scoped globally — i.e. two tasks under different parents with the same `externalRef` still conflict. The `rejects duplicate externalRef` test only covers the same-parent case implicitly. A cross-parent uniqueness test would document the design intent. `[fixable]`


### `server/app.ts`
_Clean, well-tested idempotency feature. The main issue is that the frontend `Task` type (in both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts`) was not updated to include the new `externalRef` field, creating a type-level inconsistency with the server API response._

- 🔵 **style** (L646): `body.data as TaskCreateInput` — this cast is used because `TaskCreateBody` (Zod) and `TaskCreateInput` (interface) are nearly identical but technically separate types. Now that both include `externalRef`, the cast still works, but consider sharing the type (e.g. using `z.infer<typeof TaskCreateBody>` as `TaskCreateInput`) to avoid future drift. `[fixable]`

dimakis:
## Centaur Review

### `server/app.ts`
_Clean feature addition with good test coverage. Main concern is a TOCTOU race in the idempotency guard that can surface a raw SQLite error on concurrent duplicate requests, and the frontend Task type is missing the new externalRef field._

- 🟡 **unsafe_assumptions** (L638): TOCTOU race: getByExternalRef + create are not atomic. Two concurrent requests with the same externalRef can both pass the check, then the second create hits the UNIQUE constraint and returns a raw 400 error instead of a graceful 200 dedup response. Since this is single-process Node.js with synchronous SQLite, the window is narrow but real (e.g. two requests arrive before the event loop yields). Consider catching the UNIQUE constraint error and falling back to getByExternalRef, or wrapping both in a single better-sqlite3 transaction. `[fixable]`


_Clean feature addition with good test coverage. Main concern is a TOCTOU race in the idempotency guard that can surface a raw SQLite error on concurrent duplicate requests, and the frontend Task type is missing the new externalRef field._

- 🟡 **regressions** (L48): The server Task interface now includes `externalRef: string | null`, but the frontend Task type in types/task.ts is missing this field. The server serializes externalRef in JSON responses, so the frontend type is out of sync. This won't cause a runtime error (TypeScript types are erased), but it means frontend code can't safely access task.externalRef without casting. Add `externalRef: string | null` to the frontend Task interface for type-level consistency. `[fixable]`


### `server/__tests__/task-routes.test.ts`
_Clean feature addition with good test coverage. Main concern is a TOCTOU race in the idempotency guard that can surface a raw SQLite error on concurrent duplicate requests, and the frontend Task type is missing the new externalRef field._

- 🔵 **missing_tests** (L163): No test for the TOCTOU/UNIQUE-constraint race scenario: what happens when two requests with the same externalRef race past the getByExternalRef check and the second hits the SQLite UNIQUE constraint? Currently this would return a 400 with a raw SQLite error message. A test would document and enforce the desired behavior (either graceful dedup or a clear error).


### `server/__tests__/task-store.test.ts`
_Clean feature addition with good test coverage. Main concern is a TOCTOU race in the idempotency guard that can surface a raw SQLite error on concurrent duplicate requests, and the frontend Task type is missing the new externalRef field._

- 🔵 **missing_tests** (L514): No test that getByExternalRef returns the task with children populated (since rowToTask always sets children: []). If a caller expects getByExternalRef to return a full tree like getTree does, this could be surprising. Consider a test documenting that getByExternalRef returns a flat task without assembled children.

dimakis:
## Centaur Review

_Clean, well-tested feature. The main gap is the frontend Task type not being updated to include externalRef, which creates a type divergence between server and client._

- 🟡 **regressions** (L48): The server-side `Task` interface now includes `externalRef: string | null`, but the frontend `Task` type in `frontend/src/types/task.ts` was not updated. While no frontend code currently reads this field, the type divergence means any future frontend code (e.g., displaying externalRef in TaskNode or TaskSidebar) would lack type safety. The field should be added to keep the two interfaces in sync. `[fixable]`


### `server/app.ts`
_Clean, well-tested feature. The main gap is the frontend Task type not being updated to include externalRef, which creates a type divergence between server and client._

- 🔵 **bugs** (L639): The check-then-insert for externalRef deduplication is a TOCTOU pattern. However, since this project uses `better-sqlite3` (synchronous) in a single-threaded Node.js process, concurrent interleaving is not possible — the race condition cannot actually occur at runtime. The UNIQUE index provides a safety net regardless. No change needed, but worth a comment if the server ever moves to async DB calls.
- 🔵 **style** (L642): The deduplicated response returns the existing task without broadcasting via `onTaskBroadcast`. This is correct (no new task was created), but the response also omits the `deduplicated` field from any schema — it's an ad-hoc addition to the response shape. Consider documenting or typing this response variant for frontend consumers. `[fixable]`


### `server/__tests__/task-routes.test.ts`
_Clean, well-tested feature. The main gap is the frontend Task type not being updated to include externalRef, which creates a type divergence between server and client._

- 🔵 **missing_tests**: No test verifies that the UNIQUE constraint error from `taskStore.create()` (when two tasks share an externalRef) is caught and returns a 400 response. The route test only covers the application-level dedup check (200 path). A test that bypasses the app-level check (e.g., by directly calling `store.create` twice) exists in task-store.test.ts, but the HTTP error path for a constraint violation is untested. `[fixable]`

server/app.ts
```diff
@@ -635,6 +635,14 @@ app.post('/api/tasks', (req, res) => {
     return;
   }
   try {
+    // Idempotency guard: if externalRef provided and already exists, return existing task
+    if (body.data.externalRef) {
```
dimakis, line 639: 🟡 **bugs**: TOCTOU race condition: the check-then-insert pattern (`getByExternalRef` → `create`) is not atomic. Two concurrent requests with the same `externalRef` can both pass the check, and the second `create()` will throw a UNIQUE constraint violation, returning a generic 400 error instead of the expected 200 deduplicated response. Fix by catching the UNIQUE constraint error in the `catch` block and retrying the lookup, or by using `INSERT ... ON CONFLICT` in the store layer. `[fixable]`

server/app.ts
```diff
@@ -635,6 +635,14 @@ app.post('/api/tasks', (req, res) => {
...
+      const existing = taskStore.getByExternalRef(body.data.externalRef);
+      if (existing) {
+        res.status(200).json({ task: existing, deduplicated: true });
```
dimakis, line 642: 🔵 **unsafe_assumptions**: The deduplicated response returns a flat `Task` object (with `children: []` since `rowToTask` always sets `children: []`), while the normal creation path also returns a flat task. This is consistent, but note that if the existing task has children (e.g., it was created and subtasks were added), the dedup response will show `children: []` because `getByExternalRef` uses `rowToTask` which doesn't assemble the tree. This may surprise callers expecting the full subtree. Consider whether `get()` (which also returns flat) or `getTree()` filtered to the task is more appropriate. `[fixable]`

server/task-store.ts
```diff
@@ -288,6 +303,14 @@ export class TaskStore {
     return row ? rowToTask(row) : null;
   }
 
+  /** Look up a task by its external reference (e.g. "pr_shepherd:dimakis/centaur#42"). */
```
dimakis, line 306: 🔵 **style**: The JSDoc example `pr_shepherd:dimakis/centaur#42` contains a GitHub username (`dimakis`). While this is just a doc comment example, the CLAUDE.md states 'No hardcoded machine-specific paths or configuration.' Consider using a generic example like `pr_shepherd:org/repo#42` (which is what the test uses). `[fixable]`

dimakis (commented):
## Centaur Review

### `server/app.ts`
_Clean, well-tested idempotency feature with good schema design (partial unique index, immutable ref). The main concern is a TOCTOU race in the check-then-insert pattern — under concurrent requests the UNIQUE constraint will fire as a raw 400 error instead of a graceful 200 dedup response. Low risk for single-user usage but worth hardening._

- 🟡 **bugs** (L639): TOCTOU race condition: the check-then-insert pattern (`getByExternalRef` → `create`) is not atomic. Two concurrent requests with the same `externalRef` can both pass the check, and the second `create()` will throw a UNIQUE constraint violation, returning a generic 400 error instead of the expected 200 deduplicated response. Fix by catching the UNIQUE constraint error in the `catch` block and retrying the lookup, or by using `INSERT ... ON CONFLICT` in the store layer. `[fixable]`
- 🔵 **unsafe_assumptions** (L642): The deduplicated response returns a flat `Task` object (with `children: []` since `rowToTask` always sets `children: []`), while the normal creation path also returns a flat task. This is consistent, but note that if the existing task has children (e.g., it was created and subtasks were added), the dedup response will show `children: []` because `getByExternalRef` uses `rowToTask` which doesn't assemble the tree. This may surprise callers expecting the full subtree. Consider whether `get()` (which also returns flat) or `getTree()` filtered to the task is more appropriate. `[fixable]`


### `server/__tests__/task-routes.test.ts`
_Clean, well-tested idempotency feature with good schema design (partial unique index, immutable ref). The main concern is a TOCTOU race in the check-then-insert pattern — under concurrent requests the UNIQUE constraint will fire as a raw 400 error instead of a graceful 200 dedup response. Low risk for single-user usage but worth hardening._

- 🔵 **missing_tests**: No test for the UNIQUE constraint error path. If the race condition from the TOCTOU gap occurs (or if `taskStore.create()` is called directly with a duplicate ref bypassing the route), the catch block returns a raw SQLite error message to the client. A test that calls `store.create()` with a duplicate ref via the API (simulating the race) would validate error handling. `[fixable]`
- 🔵 **missing_tests**: No test verifying that the deduplicated response still returns the original task's full shape (e.g., `description`, `priority`, `annotations`). The current test only checks `id` and `title`. If `rowToTask` ever changes field mapping, this wouldn't catch regressions in the dedup path. `[fixable]`


### `server/task-store.ts`
_Clean, well-tested idempotency feature with good schema design (partial unique index, immutable ref). The main concern is a TOCTOU race in the check-then-insert pattern — under concurrent requests the UNIQUE constraint will fire as a raw 400 error instead of a graceful 200 dedup response. Low risk for single-user usage but worth hardening._

- 🔵 **style** (L306): The JSDoc example `pr_shepherd:dimakis/centaur#42` contains a GitHub username (`dimakis`). While this is just a doc comment example, the CLAUDE.md states 'No hardcoded machine-specific paths or configuration.' Consider using a generic example like `pr_shepherd:org/repo#42` (which is what the test uses). `[fixable]`

dimakis:
## Centaur Review

_Clean, well-tested feature with correct migration and idempotency logic; the only gap is the frontend/client Task types missing the new `externalRef` field._

- 🟡 **regressions** (L48): The server `Task` type now includes `externalRef: string | null`, but the frontend `Task` interface (and the identical one in `packages/client/src/slices/tasks.ts`) was not updated. The field will silently arrive in JSON responses and be ignored by TypeScript. No runtime breakage, but frontend code can't access `task.externalRef` without type assertions, and the types are now out of sync with the API contract. `[fixable]`

server/__tests__/task-store.test.ts
```diff
@@ -526,6 +561,24 @@ describe('TaskStore', () => {
     store2.close();
   });
 
+  it('migrates existing database to add external_ref column', () => {
```
dimakis, line 564: 🔵 **missing_tests**: The migration test doesn't actually exercise the ALTER TABLE path. Because the test creates a brand-new database, `SCHEMA` (which already includes `external_ref`) creates the column at table-creation time, so migration 2 is a no-op on reopen. This is the same pattern as the existing migration 1 test, but it means the ALTER TABLE + CREATE UNIQUE INDEX migration is never tested. A true migration test would need to create the DB with the old schema (without `external_ref`) then reopen with the new code. `[fixable]`

server/app.ts
dimakis, line 639: 🔵 **unsafe_assumptions**: The check-then-create pattern (`getByExternalRef` → `create`) has a TOCTOU gap. If two requests with the same `externalRef` arrive concurrently, both could pass the check. The UNIQUE index provides a safety net (the second INSERT throws, caught by the `catch` block and returned as a 400), but the 400 error message would be a raw SQLite constraint violation — not a clean "deduplicated" response. In practice this is very unlikely with Node.js single-threaded + synchronous better-sqlite3, but wrapping both the check and insert in a SQLite transaction would make it airtight. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean, well-tested feature. Main gap is that the frontend and @mitzo/client Task types need `externalRef` added to stay in sync with the backend._

- 🟡 **regressions**: Frontend `Task` interface is missing `externalRef: string | null`. The backend now serializes this field in every task response, but both `frontend/src/types/task.ts` and `packages/client/src/slices/tasks.ts` are out of sync. While TypeScript types are erased at runtime (so no crash), any frontend code that needs to read `externalRef` won't typecheck, and the types no longer match the API contract. `[fixable]`


### `server/__tests__/task-store.test.ts`
_Clean, well-tested feature. Main gap is that the frontend and @mitzo/client Task types need `externalRef` added to stay in sync with the backend._

- 🔵 **missing_tests** (L564): The migration test doesn't actually exercise the ALTER TABLE path. Because the test creates a brand-new database, `SCHEMA` (which already includes `external_ref`) creates the column at table-creation time, so migration 2 is a no-op on reopen. This is the same pattern as the existing migration 1 test, but it means the ALTER TABLE + CREATE UNIQUE INDEX migration is never tested. A true migration test would need to create the DB with the old schema (without `external_ref`) then reopen with the new code. `[fixable]`


### `server/app.ts`
_Clean, well-tested feature. Main gap is that the frontend and @mitzo/client Task types need `externalRef` added to stay in sync with the backend._

- 🔵 **unsafe_assumptions** (L639): The check-then-create pattern (`getByExternalRef` → `create`) has a TOCTOU gap. If two requests with the same `externalRef` arrive concurrently, both could pass the check. The UNIQUE index provides a safety net (the second INSERT throws, caught by the `catch` block and returned as a 400), but the 400 error message would be a raw SQLite constraint violation — not a clean "deduplicated" response. In practice this is very unlikely with Node.js single-threaded + synchronous better-sqlite3, but wrapping both the check and insert in a SQLite transaction would make it airtight. `[fixable]`

PR #415, opened 2026-06-27
fix(orchestrator): default session_policy to spawn

dimakis:
## Summary

- Flips the default `sessionPolicy` from `reuse` to `spawn` in `TaskOrchestrator.tick()`
- Tasks without an explicit policy now get dedicated headless sessions instead of injecting into the user's active conversation
- The `spawn` infrastructure was already built in PR #377 but was opt-in — this makes it the default
- Explicit `sessionPolicy: 'reuse'` still works for tasks that need it

## Test plan

- [x] All 45 existing orchestrator tests pass (tests that relied on reuse either set `sessionPolicy: 'reuse'` explicitly or don't provide `spawnSession` in deps, so they fall through correctly)
- [ ] Verify task board no longer injects prompts into interactive sessions
- [ ] Verify spawned tasks complete in their own headless sessions

### DIFF
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -727,7 +727,11 @@ describe('TaskOrchestrator', () => {
       });
 
       const goal2 = store.create({ title: 'Goal 2' });
-      const reuse = store.create({ title: 'Reuse task', parentId: goal2.id });
+      const reuse = store.create({
+        title: 'Reuse task',
+        parentId: goal2.id,
+        sessionPolicy: 'reuse',
+      });
 
       // Start goal1 (spawns), stop, start goal2
       orch.start(goal1.id);
@@ -815,6 +819,23 @@ describe('TaskOrchestrator', () => {
       await vi.waitFor(() => {
         expect(spawnSession).toHaveBeenCalledTimes(2);
       });
+    });
+
+    it('auto policy (store default) spawns instead of reusing', async () => {
+      const spawnSession = vi.fn().mockResolvedValue('spawned-client-1');
+      const deps = createTestDeps(store);
+      deps.spawnSession = spawnSession;
+      const orch = new TaskOrchestrator(deps);
+
+      const goal = store.create({ title: 'Goal' });
+      // No explicit sessionPolicy — store defaults to 'auto'
+      const task = store.create({ title: 'Auto task', parentId: goal.id });
+
+      orch.start(goal.id);
+      await vi.waitFor(() => expect(spawnSession).toHaveBeenCalled());
+
+      expect(spawnSession).toHaveBeenCalledWith(task.id, expect.any(String), goal.id);
+      expect(deps.setTaskContext).not.toHaveBeenCalled();
     });
 
     it('reuse policy tasks use pinned session as before', () => {
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -835,6 +835,7 @@ describe('TaskOrchestrator', () => {
       await vi.waitFor(() => expect(spawnSession).toHaveBeenCalled());
 
       expect(spawnSession).toHaveBeenCalledWith(task.id, expect.any(String), goal.id);
+      expect(store.get(task.id)?.status).toBe('active');
       expect(deps.setTaskContext).not.toHaveBeenCalled();
     });
 
--- a/server/task-orchestrator.ts
+++ b/server/task-orchestrator.ts
@@ -342,7 +342,10 @@ export class TaskOrchestrator {
 
       case 'agent_work':
       default: {
-        const policy = next.sessionPolicy ?? 'spawn';
+        // Default to spawn: 'auto' and undefined both resolve to 'spawn'
+        // so tasks get dedicated sessions unless explicitly set to 'reuse'.
+        const raw = next.sessionPolicy ?? 'spawn';
+        const policy = raw === 'auto' ? 'spawn' : raw;
 
         if (policy === 'spawn' && this.deps.spawnSession) {
           // Spawn a dedicated headless session for this task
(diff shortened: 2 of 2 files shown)


server/task-orchestrator.ts
```diff
@@ -342,7 +342,7 @@ export class TaskOrchestrator {
 
       case 'agent_work':
       default: {
-        const policy = next.sessionPolicy ?? 'reuse';
+        const policy = next.sessionPolicy ?? 'spawn';
```
dimakis: 🔵 **regressions**: The `??` fallback only triggers when `next.sessionPolicy` is null/undefined — but `TaskStore.create()` defaults `sessionPolicy` to `'auto'` (task-store.ts:269), and `'auto'` is neither `'spawn'` nor `'reuse'`, so it falls through to the reuse (else) branch regardless of this change. In practice this makes the change a no-op for tasks created through the normal store path. If the intent is for tasks without an explicit policy to spawn by default, the store's default at line 269 (`input.sessionPolicy ?? 'auto'`) may also need updating to `'spawn'`, or the orchestrator should handle `'auto'` explicitly (e.g., treating it as `'spawn'`). `[fixable]`

dimakis (commented):
## Centaur Review

_The change is logically correct in intent but effectively a no-op — `TaskStore.create()` defaults `sessionPolicy` to `'auto'`, which bypasses the `??` fallback. Either the store default or the `'auto'` handling in the orchestrator also needs updating to achieve the desired spawn-by-default behavior._

- 🟡 **missing_tests**: No test verifies the new default. All spawn tests explicitly set `sessionPolicy: 'spawn'`, and all non-spawn tests either omit `sessionPolicy` (which the TaskStore defaults to `'auto'`, bypassing the `??` fallback entirely) or explicitly set `'reuse'`. A test should cover the edge case where `sessionPolicy` is null/undefined to confirm the new `'spawn'` default fires correctly. `[fixable]`


_The change is logically correct in intent but effectively a no-op — `TaskStore.create()` defaults `sessionPolicy` to `'auto'`, which bypasses the `??` fallback. Either the store default or the `'auto'` handling in the orchestrator also needs updating to achieve the desired spawn-by-default behavior._

- 🔵 **regressions** (L345): The `??` fallback only triggers when `next.sessionPolicy` is null/undefined — but `TaskStore.create()` defaults `sessionPolicy` to `'auto'` (task-store.ts:269), and `'auto'` is neither `'spawn'` nor `'reuse'`, so it falls through to the reuse (else) branch regardless of this change. In practice this makes the change a no-op for tasks created through the normal store path. If the intent is for tasks without an explicit policy to spawn by default, the store's default at line 269 (`input.sessionPolicy ?? 'auto'`) may also need updating to `'spawn'`, or the orchestrator should handle `'auto'` explicitly (e.g., treating it as `'spawn'`). `[fixable]`

server/task-orchestrator.ts
```diff
@@ -342,7 +342,10 @@ export class TaskOrchestrator {
 
       case 'agent_work':
       default: {
-        const policy = next.sessionPolicy ?? 'reuse';
+        // Default to spawn: 'auto' and undefined both resolve to 'spawn'
+        // so tasks get dedicated sessions unless explicitly set to 'reuse'.
+        const raw = next.sessionPolicy ?? 'spawn';
```
dimakis: 🔵 **style**: The `?? 'spawn'` fallback on line 347 is unreachable: `TaskStore.create()` always defaults `sessionPolicy` to `'auto'` (task-store.ts:269), so the field is never `undefined` on a stored task. The null-coalesce is dead code — only the `raw === 'auto'` branch matters. Not a bug, but it could mislead future readers into thinking `undefined` is a real case. Consider simplifying to `const policy = next.sessionPolicy === 'auto' ? 'spawn' : next.sessionPolicy;` or just removing the fallback. `[fixable]`

server/__tests__/task-orchestrator.test.ts
```diff
@@ -817,6 +821,23 @@ describe('TaskOrchestrator', () => {
...
+
+      const goal = store.create({ title: 'Goal' });
+      // No explicit sessionPolicy — store defaults to 'auto'
+      const task = store.create({ title: 'Auto task', parentId: goal.id });
+
+      orch.start(goal.id);
+      await vi.waitFor(() => expect(spawnSession).toHaveBeenCalled());
+
+      expect(spawnSession).toHaveBeenCalledWith(task.id, expect.any(String), goal.id);
+      expect(deps.setTaskContext).not.toHaveBeenCalled();
```
dimakis, line 839: 🔵 **missing_tests**: The new test asserts `setTaskContext` is not called (proving it spawns rather than reuses), but does not verify the task's status is set to `'active'` after spawn dispatch. Other spawn tests in this file also check status transitions. Consider adding `expect(store.get(task.id)?.status).toBe('active')` for parity. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


_Correct fix — resolves `auto` (the store default) to `spawn` instead of silently falling through to `reuse`. Minor style nit on dead-code fallback and one missing status assertion in the new test._

- 🔵 **style** (L347): The `?? 'spawn'` fallback on line 347 is unreachable: `TaskStore.create()` always defaults `sessionPolicy` to `'auto'` (task-store.ts:269), so the field is never `undefined` on a stored task. The null-coalesce is dead code — only the `raw === 'auto'` branch matters. Not a bug, but it could mislead future readers into thinking `undefined` is a real case. Consider simplifying to `const policy = next.sessionPolicy === 'auto' ? 'spawn' : next.sessionPolicy;` or just removing the fallback. `[fixable]`


_Correct fix — resolves `auto` (the store default) to `spawn` instead of silently falling through to `reuse`. Minor style nit on dead-code fallback and one missing status assertion in the new test._

- 🔵 **missing_tests** (L838): The new test asserts `setTaskContext` is not called (proving it spawns rather than reuses), but does not verify the task's status is set to `'active'` after spawn dispatch. Other spawn tests in this file also check status transitions. Consider adding `expect(store.get(task.id)?.status).toBe('active')` for parity. `[fixable]`

dimakis (commented):
## Centaur Review

LGTM — no issues found.

PR #420, opened 2026-06-28
fix(telos): promote Telos items to tasks without workloadStore lookup

dimakis:
## Summary
- Telos items (from Python script via `/api/todos`) were never ingested into `workloadStore`, so "Promote to Tasks" always returned 404 "Item not found"
- Frontend now sends item data (title, contextHints, sources) as fallback fields in the promote POST body
- Backend falls back to body data when `workloadStore.get()` misses, creating the task directly

## Test plan
- [ ] Open a Telos item detail view, tap "Promote to Tasks" -- should create a task and navigate to Task Board
- [ ] Verify workloadStore items still promote correctly (backwards compatible)
- [ ] Verify task title, description, and source annotations carry over from Telos data

### DIFF
--- a/server/api-schemas.ts
+++ b/server/api-schemas.ts
@@ -234,7 +234,6 @@ export const WorkloadItemUpdateBody = z.object({
 
 export const WorkloadPromoteBody = z.object({
   description: z.string().optional(),
-  // Fallback fields for Telos items not in workloadStore
   title: z.string().optional(),
   contextHints: z
     .object({
@@ -254,8 +253,8 @@ export const WorkloadPromoteBody = z.object({
         type: z.string(),
         url: z.string(),
         title: z.string(),
-        author: z.string(),
-        snippet: z.string(),
+        author: z.string().optional(),
+        snippet: z.string().optional(),
       }),
     )
     .optional(),
--- a/server/api-schemas.ts
+++ b/server/api-schemas.ts
@@ -234,6 +234,31 @@ export const WorkloadItemUpdateBody = z.object({
 
 export const WorkloadPromoteBody = z.object({
   description: z.string().optional(),
+  // Fallback fields for Telos items not in workloadStore
+  title: z.string().optional(),
+  contextHints: z
+    .object({
+      repos: z.array(z.string()).optional(),
+      paths: z.array(z.string()).optional(),
+      issues: z.array(z.string()).optional(),
+      docIds: z.array(z.string()).optional(),
+      people: z.array(z.string()).optional(),
+      jiraKeys: z.array(z.string()).optional(),
+      keywords: z.array(z.string()).optional(),
+      taskHint: z.string().optional(),
+    })
+    .optional(),
+  sources: z
+    .array(
+      z.object({
+        type: z.string(),
+        url: z.string(),
+        title: z.string(),
+        author: z.string(),
+        snippet: z.string(),
+      }),
+    )
+    .optional(),
 });
 
 // -- Session creation schemas --
--- a/server/app.ts
+++ b/server/app.ts
@@ -1925,7 +1925,7 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
   }
 
   const hints = item?.contextHints ?? body.data.contextHints;
-  const taskHint = hints && 'taskHint' in hints ? (hints.taskHint as string) : undefined;
+  const taskHint = hints?.taskHint ?? undefined;
 
   // Build description from item context
   const descParts: string[] = [];
@@ -1955,8 +1955,8 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
     workloadStore.setGoalId(item.id, task.id);
   }
 
-  const updatedItem = item ? workloadStore.get(item.id) : null;
-  res.status(201).json({ task, item: updatedItem });
+  const updatedItem = item ? workloadStore.get(item.id) : undefined;
+  res.status(201).json(updatedItem ? { task, item: updatedItem } : { task });
   onTaskBroadcast?.({ type: 'task_state', tasks: taskStore.getTree() });
   if (updatedItem) {
     onWorkloadBroadcast?.({ type: 'workload_item_updated', item: updatedItem });
--- a/server/app.ts
+++ b/server/app.ts
@@ -1916,34 +1916,51 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
   }
 
   const item = workloadStore.get(req.params.id);
-  if (!item) {
-    res.status(404).json({ error: 'Item not found' });
+
+  // Resolve title and context from workloadStore item or fallback body data (Telos items)
+  const title = item?.title ?? body.data.title;
+  if (!title) {
+    res.status(404).json({ error: 'Item not found and no title provided' });
     return;
   }
+
+  const hints = item?.contextHints ?? body.data.contextHints;
+  const taskHint = hints && 'taskHint' in hints ? (hints.taskHint as string) : undefined;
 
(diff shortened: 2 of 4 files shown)


dimakis (commented):
## Centaur Review

### `server/app.ts`
_Solid fallback approach for Telos items, but the new code path lacks test coverage and the `sources` schema may be too strict for real-world data._

- 🟡 **missing_tests**: No tests cover the new fallback path (promoting a Telos item not in workloadStore). The existing promote tests all create items via /api/workload/signals first. A test should verify: (1) promoting with just body fields (title + contextHints + sources) creates a task, (2) response has `item: null` in that case, (3) no workload broadcast fires when item is absent. `[fixable]`
- 🔵 **unsafe_assumptions** (L1930): The `taskHint` extraction uses `'taskHint' in hints` with a cast `(hints.taskHint as string)`. Since `hints` can come from the Zod-parsed body where `taskHint` is `z.string().optional()`, this could theoretically be `undefined` even when the key is present. The falsy check on `if (taskHint)` prevents it from being pushed, so it's safe at runtime, but the `as string` cast is misleading. Consider `hints?.taskHint ?? undefined` instead. `[fixable]`
- 🔵 **regressions** (L1955): When `item` is null (fallback path), the response is `{ task, item: null }`. The existing test at workload-routes.test.ts:508 asserts `res.body.item.status === 'acknowledged'` and line 509 asserts `res.body.item.goalId`. These tests only run the workloadStore path so they still pass, but any consumer expecting `item` to always be non-null in the promote response will break. The API contract change should be documented or the response shape should differ (e.g., omit `item` entirely when null). `[fixable]`


### `server/api-schemas.ts`
_Solid fallback approach for Telos items, but the new code path lacks test coverage and the `sources` schema may be too strict for real-world data._

- 🟡 **bugs** (L253): All fields in the `sources` array objects (`type`, `url`, `title`, `author`, `snippet`) are required strings. If any Telos source has a missing or empty `author`/`snippet` (common for programmatically-generated sources), Zod will reject the entire request with a 400. Consider making `author` and `snippet` optional (`.optional()`) to match real-world data shapes. `[fixable]`
- 🔵 **style** (L236): The comment `// Fallback fields for Telos items not in workloadStore` explains the 'why' but couples the schema to a specific caller. The schema should be self-describing — consider removing the comment since the field names and `.optional()` already convey that they're supplementary. `[fixable]`

server/app.ts
```diff
@@ -1916,34 +1916,51 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
...
+  const title = item?.title ?? body.data.title;
+  if (!title) {
+    res.status(404).json({ error: 'Item not found and no title provided' });
     return;
   }
 
+  const hints = item?.contextHints ?? body.data.contextHints;
+  const taskHint = hints && 'taskHint' in hints ? (hints.taskHint as string) : undefined;
+
   // Build description from item context
```
dimakis, line 1930: 🔵 **unsafe_assumptions**: The `taskHint` extraction uses `'taskHint' in hints` with a cast `(hints.taskHint as string)`. Since `hints` can come from the Zod-parsed body where `taskHint` is `z.string().optional()`, this could theoretically be `undefined` even when the key is present. The falsy check on `if (taskHint)` prevents it from being pushed, so it's safe at runtime, but the `as string` cast is misleading. Consider `hints?.taskHint ?? undefined` instead. `[fixable]`

server/app.ts
```diff
@@ -1916,34 +1916,51 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
...
     description: descParts.join('\n\n') || undefined,
-    annotations: item.sources.map((s) => `Source: [${s.sourceType}] ${s.title} — ${s.url}`),
+    annotations,
   });
 
-  // Link item to goal
-  workloadStore.setGoalId(item.id, task.id);
+  // Link item to goal only if it exists in workloadStore
+  if (item) {
+    workloadStore.setGoalId(item.id, task.id);
```
dimakis, line 1955: 🔵 **regressions**: When `item` is null (fallback path), the response is `{ task, item: null }`. The existing test at workload-routes.test.ts:508 asserts `res.body.item.status === 'acknowledged'` and line 509 asserts `res.body.item.goalId`. These tests only run the workloadStore path so they still pass, but any consumer expecting `item` to always be non-null in the promote response will break. The API contract change should be documented or the response shape should differ (e.g., omit `item` entirely when null). `[fixable]`

server/api-schemas.ts
```diff
@@ -234,6 +234,31 @@ export const WorkloadItemUpdateBody = z.object({
...
+      docIds: z.array(z.string()).optional(),
+      people: z.array(z.string()).optional(),
+      jiraKeys: z.array(z.string()).optional(),
+      keywords: z.array(z.string()).optional(),
+      taskHint: z.string().optional(),
+    })
+    .optional(),
+  sources: z
+    .array(
+      z.object({
```
dimakis, line 252: 🟡 **bugs**: All fields in the `sources` array objects (`type`, `url`, `title`, `author`, `snippet`) are required strings. If any Telos source has a missing or empty `author`/`snippet` (common for programmatically-generated sources), Zod will reject the entire request with a 400. Consider making `author` and `snippet` optional (`.optional()`) to match real-world data shapes. `[fixable]`

server/api-schemas.ts
```diff
@@ -234,6 +234,31 @@ export const WorkloadItemUpdateBody = z.object({
 
 export const WorkloadPromoteBody = z.object({
   description: z.string().optional(),
```
dimakis, line 236: 🔵 **style**: The comment `// Fallback fields for Telos items not in workloadStore` explains the 'why' but couples the schema to a specific caller. The schema should be self-describing — consider removing the comment since the field names and `.optional()` already convey that they're supplementary. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid fallback implementation with correct null-safety and conditional broadcasting; main concern is the broadcast test that passes vacuously because it intercepts a non-existent app property instead of the module-level closure._

- 🟡 **bugs**: The 'fallback does not broadcast workload update' test intercepts `(app as any)._workloadBroadcast`, but no such property exists on the Express app. The actual broadcast callback is the module-level closure `onWorkloadBroadcast` in app.ts (line 235), only settable via `setWorkloadBroadcast()`. The test passes vacuously: the mock is never installed, `onWorkloadBroadcast` is null in the test environment, so the `broadcasts` array stays empty regardless of what the route handler does. This test does not validate the behavior it claims. `[fixable]`
- 🔵 **missing_tests**: No test covers the precedence behavior when a workloadStore item exists AND the body also provides `title`/`contextHints`/`sources`. The code uses nullish coalescing (`item?.title ?? body.data.title`), so the item values win — but this isn't verified. A test promoting a real workloadStore item while also sending body fields would lock in the intended precedence. `[fixable]`


### `server/app.ts`
_Solid fallback implementation with correct null-safety and conditional broadcasting; main concern is the broadcast test that passes vacuously because it intercepts a non-existent app property instead of the module-level closure._

- 🔵 **regressions**: The 404 error message changed from `"Item not found"` to `"Item not found and no title provided"`. Any client that matches on this exact string will break. Low risk since clients typically check status codes, but the updated test description ('returns 404 for missing item with no body title') also subtly documents this contract change. `[fixable]`
- 🔵 **style**: The annotation-building ternary duplicates the `.map()` pattern with slightly different field names (`s.sourceType` vs `s.type`). Normalizing source objects before building annotations (e.g., mapping body sources to have `sourceType`) would reduce the surface area for field-name drift between the server `TodoSource` model and the API schema's source shape. `[fixable]`

PR #421, opened 2026-06-29
Rescued: 2026-06-25-90eaf3eb17db

dimakis:
Auto-rescued uncommitted work from session `2026-06-25-90eaf3eb17db`.

This PR was created automatically by Mitzo's worktree cleanup.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/dsaridak/tools/mitzo/node_modules
--- a/server/__tests__/rescue-worktree.test.ts
+++ b/server/__tests__/rescue-worktree.test.ts
@@ -107,6 +107,8 @@ describe('rescueDirtyWorktree', () => {
     mockExecFileSync.mockReturnValueOnce('git@github.com:dimakis/mgmt.git\n' as never);
     // Step 1: git add -u
     mockExecFileSync.mockReturnValueOnce('' as never);
+    // Step 1b: git diff --cached --name-only → meaningful file
+    mockExecFileSync.mockReturnValueOnce('src/feature.ts\n' as never);
     // Step 2: git commit
     mockExecFileSync.mockReturnValueOnce('' as never);
     // Step 3: git push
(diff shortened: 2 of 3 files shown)


dimakis:
## Centaur Review

### `node_modules`
_The noise-file filtering logic is sound, but the PR accidentally commits a `node_modules` symlink with a hardcoded machine path (critical), and the caller in `cleanupStaleWorktrees` doesn't distinguish 'no meaningful changes' from real rescue failures, so noise-only worktrees still accumulate on disk._

- 🔴 **bugs** (L1): A `node_modules` symlink pointing to `/Users/dsaridak/tools/mitzo/node_modules` was committed. `.gitignore` has `node_modules/` (trailing slash matches directories, not symlinks-to-directories). This leaks a machine-specific path into the repo and will break `npm install` for anyone who clones — git will create a file/symlink named `node_modules` before npm can create its directory. Must be removed from the branch and the `.gitignore` updated to also match `node_modules` (no trailing slash) or the symlink pattern. `[fixable]`


### `server/worktree.ts`
_The noise-file filtering logic is sound, but the PR accidentally commits a `node_modules` symlink with a hardcoded machine path (critical), and the caller in `cleanupStaleWorktrees` doesn't distinguish 'no meaningful changes' from real rescue failures, so noise-only worktrees still accumulate on disk._

- 🟡 **regressions** (L489): The caller in `cleanupStaleWorktrees` (line ~704) treats every `rescue.success === false` identically — it posts to the inbox and skips cleanup. When `rescueDirtyWorktree` returns `{ success: false, error: 'no meaningful changes' }`, the worktree stays on disk and generates an inbox entry, which is exactly the noise this PR is trying to suppress. The caller should distinguish 'no meaningful changes' from a real rescue failure so it can clean up the worktree instead of preserving it. `[fixable]`
- 🔵 **regressions** (L447): The docstring on `rescueDirtyWorktree` still says 'the commit will fail with "nothing to commit" and the worktree is preserved on disk' for the untracked-only case. With the new noise filter, the function now returns early before attempting commit, with a different error message ('no meaningful changes to rescue'). The docstring should be updated to reflect the new behavior. `[fixable]`


_The noise-file filtering logic is sound, but the PR accidentally commits a `node_modules` symlink with a hardcoded machine path (critical), and the caller in `cleanupStaleWorktrees` doesn't distinguish 'no meaningful changes' from real rescue failures, so noise-only worktrees still accumulate on disk._

- 🔵 **style** (L277): Test comments for the new noise-filter tests say 'git add -A' (e.g., lines 277, 296, 316) but the actual implementation uses `git add -u`. The comments are misleading — they should match the real command. `[fixable]`

PR #422, opened 2026-07-01
fix(harness): improve auto-rename prompt quality

dimakis:
## Summary

- Rewrites the Haiku system prompt with few-shot examples and strict constraints to produce specific, action-oriented session titles instead of conversational responses
- Adds `sanitizeSessionName()` that strips quotes, truncates at newlines, and rejects known bad patterns (apologies, generic starters, conversational filler)
- Separates prompts with `---` delimiters instead of bare newlines for clearer multi-prompt input
- Logs rejected/empty titles for observability

## Problem

Haiku was producing garbage session titles in production:
- `"I apologize, but I cannot access the specific Jira ticket li"` — full conversational response
- `"\"OpenShift OpenCode Agent Integration\"\n\nWould you like me to"` — embedded quotes + trailing conversation
- `"Start a Conversation"` — generic filler when it couldn't extract intent

## Test plan

- [x] 31 tests pass including new sanitization + integration tests
- [x] Full suite shows no regressions (pre-existing failures unrelated)
- [ ] Deploy and observe auto-rename log quality over a day of sessions

### DIFF
--- a/packages/harness/src/auto-rename.ts
+++ b/packages/harness/src/auto-rename.ts
@@ -30,7 +30,7 @@ const REJECTED_PATTERNS = [
   /^(hello|hi|hey|greetings)/i,
   /^(sure|certainly|of\s+course)/i,
   /^(here|let\s+me)/i,
-  /would you like/i,
+  /^would you like/i,
 ];
 
 /** Common stop words to filter out when extracting key terms. */
--- a/packages/harness/src/auto-rename.ts
+++ b/packages/harness/src/auto-rename.ts
@@ -240,7 +240,7 @@ export function sanitizeSessionName(raw: string): string {
   let name = raw.split('\n')[0].trim();
 
   // Strip surrounding quotes (single, double, smart quotes)
-  name = name.replace(/^[\s"'"'\u201C\u201D\u2018\u2019]+|[\s"'"'\u201C\u201D\u2018\u2019]+$/g, '');
+  name = name.replace(/^[\s"'\u201C\u201D\u2018\u2019]+|[\s"'\u201C\u201D\u2018\u2019]+$/g, '');
 
   // Reject conversational responses — Haiku went off-script
   for (const pattern of REJECTED_PATTERNS) {
--- a/packages/harness/src/auto-rename.ts
+++ b/packages/harness/src/auto-rename.ts
@@ -12,7 +12,7 @@ export const AUTO_RENAME_INTERVAL = 2;
 export const AUTO_RENAME_MODEL = 'claude-haiku-4-5-20251001';
 
 /** Model name on Vertex AI (uses different naming convention). */
-const VERTEX_MODEL = 'claude-3-5-haiku@20241022';
+const VERTEX_MODEL = 'claude-haiku-4-5@20251001';
 
 /** Max total characters of concatenated prompts sent to the LLM. */
 const MAX_PROMPT_INPUT_CHARS = 2000;
--- a/packages/harness/src/auto-rename.ts
+++ b/packages/harness/src/auto-rename.ts
@@ -23,6 +23,16 @@ const MAX_RECENT_PROMPTS = 8;
 /** Maximum length of the generated session name. */
 const MAX_NAME_LENGTH = 60;
 
+/** Patterns that indicate Haiku went off-script — fall back to keyword extraction. */
+const REJECTED_PATTERNS = [
+  /^I\s+(apologize|apologise|cannot|can't|don't|am\s+sorry)/i,
+  /^(start|begin|new)\s+(a\s+|new\s+)?(conversation|chat|session)/i,
+  /^(hello|hi|hey|greetings)/i,
+  /^(sure|certainly|of\s+course)/i,
+  /^(here|let\s+me)/i,
+  /would you like/i,
+];
+
 /** Common stop words to filter out when extracting key terms. */
 const STOP_WORDS = new Set([
   'a',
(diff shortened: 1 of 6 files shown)


packages/harness/src/auto-rename.ts
```diff
@@ -23,6 +23,16 @@ const MAX_RECENT_PROMPTS = 8;
...
 const MAX_NAME_LENGTH = 60;
 
+/** Patterns that indicate Haiku went off-script — fall back to keyword extraction. */
+const REJECTED_PATTERNS = [
+  /^I\s+(apologize|apologise|cannot|can't|don't|am\s+sorry)/i,
+  /^(start|begin|new)\s+(a\s+|new\s+)?(conversation|chat|session)/i,
+  /^(hello|hi|hey|greetings)/i,
+  /^(sure|certainly|of\s+course)/i,
+  /^(here|let\s+me)/i,
+  /would you like/i,
```
dimakis: 🔵 **unsafe_assumptions**: The `/would you like/i` pattern is the only REJECTED_PATTERN not anchored to `^`. This means a legitimate title like "Fix Would You Like Dialog" would be falsely rejected. Anchoring it to end-of-string (`/would you like.*$/i`) or keeping it unanchored but with more context (`/would you like (me |to )/i`) would reduce false positives while still catching Haiku run-on. `[fixable]`

packages/harness/src/providers/anthropic-vertex.ts
```diff
@@ -13,11 +13,13 @@ import { calculateCost } from './types.js';
 
 const log = createLogger('provider:anthropic');
 
-/** Vertex AI model name mapping (Vertex uses different naming). */
+/** Vertex AI model name mapping. Unversioned names auto-resolve to latest. */
 const VERTEX_MODEL_MAP: Record<string, string> = {
```
dimakis: 🔵 **style**: VERTEX_MODEL_MAP is now an identity mapping — every key maps to itself. The fallback `VERTEX_MODEL_MAP[this.model] ?? this.model` would return the same value with or without the map. It documents "supported models" but a `Set` or inline comment would communicate that intent more clearly and avoid confusion about whether the map is supposed to transform anything. `[fixable]`

dimakis (commented):
## Centaur Review

Found **2** issue(s).


### `packages/harness/src/auto-rename.ts`
_Clean quality-of-life improvement to auto-rename. Sanitization logic is well-tested and the prompt upgrade is solid. Two minor suggestions: tighten the unanchored `would you like` rejection pattern, and simplify the now-identity Vertex model map._

- 🔵 **unsafe_assumptions** (L33): The `/would you like/i` pattern is the only REJECTED_PATTERN not anchored to `^`. This means a legitimate title like "Fix Would You Like Dialog" would be falsely rejected. Anchoring it to end-of-string (`/would you like.*$/i`) or keeping it unanchored but with more context (`/would you like (me |to )/i`) would reduce false positives while still catching Haiku run-on. `[fixable]`


_Clean quality-of-life improvement to auto-rename. Sanitization logic is well-tested and the prompt upgrade is solid. Two minor suggestions: tighten the unanchored `would you like` rejection pattern, and simplify the now-identity Vertex model map._

- 🔵 **style** (L17): VERTEX_MODEL_MAP is now an identity mapping — every key maps to itself. The fallback `VERTEX_MODEL_MAP[this.model] ?? this.model` would return the same value with or without the map. It documents "supported models" but a `Set` or inline comment would communicate that intent more clearly and avoid confusion about whether the map is supposed to transform anything. `[fixable]`


---
_Reviewed at 55ff47cf68fdec98aaa8f7c2521923706f3e4b54_

packages/harness/src/providers/anthropic-vertex.ts
```diff
@@ -58,7 +51,7 @@ export class AnthropicVertexModelProvider implements ModelProvider {
     const systemMessages = messages.filter((m) => m.role === 'system');
     const conversationMessages = messages.filter((m) => m.role !== 'system');
 
-    const apiModel = this.isVertex ? (VERTEX_MODEL_MAP[this.model] ?? this.model) : this.model;
+    const apiModel = this.model;
```
dimakis, line 54: 🟡 **regressions**: The VERTEX_MODEL_MAP removal means callers must now pass Vertex-format model names (e.g. 'claude-haiku-4-5@20251001') directly. If any caller passes standard API names like 'claude-opus-4-6' to a Vertex client, the API call will fail — previously the map translated these. Verify that all call sites that construct AnthropicVertexModelProvider with Vertex enabled pass Vertex-compatible model names.

packages/harness/src/auto-rename.ts
```diff
@@ -220,6 +230,53 @@ export function generateSessionNameFallback(prompts: string[]): string {
...
+ * Sanitize a Haiku-generated session name.
+ * Strips quotes, truncates at newlines, and rejects conversational responses.
+ * Returns empty string if the name is rejected.
+ */
+export function sanitizeSessionName(raw: string): string {
+  // Truncate at first newline — Haiku sometimes generates multi-line output
+  let name = raw.split('\n')[0].trim();
+
+  // Strip surrounding quotes (single, double, smart quotes)
+  name = name.replace(/^[\s"'\u201C\u201D\u2018\u2019]+|[\s"'\u201C\u201D\u2018\u2019]+$/g, '');
```
dimakis, line 243: 🔵 **unsafe_assumptions**: The quote-stripping regex uses a character class that includes both whitespace (\s) and quote chars with `+` quantifier: `^[\s"'\u201C\u201D\u2018\u2019]+`. This means a title like `"  Spaced Title  "` correctly produces `Spaced Title`, but a title that legitimately starts with an apostrophe (e.g. `'Twas`) would lose it. Low-risk for session names but worth noting.

server/__tests__/auto-rename.test.ts
```diff
@@ -156,21 +156,12 @@ describe('generateSessionName', () => {
...
-            role: 'user',
-            content: 'Fix the auth bug\nUpdate login page',
-          },
-        ],
-      },
-      { timeout: 5000 },
-    );
+    const call = mockCreate.mock.calls[0];
+    expect(call[0].model).toBe(AUTO_RENAME_MODEL);
+    expect(call[0].max_tokens).toBe(20);
```
dimakis, line 161: 🔵 **style**: The server test relaxed its assertion from exact argument matching (toHaveBeenCalledWith) to individual field checks (call[0].model, call[0].messages[0].content). This is fine for resilience against prompt changes, but the system prompt (NAMING_SYSTEM_PROMPT) is now completely unverified in the server test suite. Consider adding a minimal assertion like `expect(call[0].system).toContain('3-6 word title')` so the server tests catch accidental system prompt deletions. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid improvement to auto-rename robustness with good sanitization logic and test coverage. Main concern is whether the Vertex model map removal could break callers that pass standard (non-Vertex) model names to Vertex clients._

- 🟡 **regressions** (L54): The VERTEX_MODEL_MAP removal means callers must now pass Vertex-format model names (e.g. 'claude-haiku-4-5@20251001') directly. If any caller passes standard API names like 'claude-opus-4-6' to a Vertex client, the API call will fail — previously the map translated these. Verify that all call sites that construct AnthropicVertexModelProvider with Vertex enabled pass Vertex-compatible model names.


### `server/auto-rename.ts`
_Solid improvement to auto-rename robustness with good sanitization logic and test coverage. Main concern is whether the Vertex model map removal could break callers that pass standard (non-Vertex) model names to Vertex clients._

- 🟡 **missing_tests**: sanitizeSessionName is not re-exported from server/auto-rename.ts (nor from packages/harness/src/index.ts). The server test file imports from ../auto-rename.js and cannot test sanitizeSessionName. Either add it to the re-export list in both files, or accept that sanitizeSessionName is only tested via the harness test suite. If it's intentionally internal, the export in auto-rename.ts should be test-only — but then the harness test imports it as a public export. `[fixable]`


### `packages/harness/__tests__/auto-rename.test.ts`
_Solid improvement to auto-rename robustness with good sanitization logic and test coverage. Main concern is whether the Vertex model map removal could break callers that pass standard (non-Vertex) model names to Vertex clients._

- 🔵 **missing_tests**: No test for sanitizeSessionName with smart/curly quotes (U+201C, U+201D, U+2018, U+2019) even though the regex explicitly handles them. Adding a test like sanitizeSessionName('\u201CHello\u201D') === 'Hello' would verify the Unicode quote stripping works. `[fixable]`


### `packages/harness/src/auto-rename.ts`
_Solid improvement to auto-rename robustness with good sanitization logic and test coverage. Main concern is whether the Vertex model map removal could break callers that pass standard (non-Vertex) model names to Vertex clients._

- 🔵 **unsafe_assumptions** (L243): The quote-stripping regex uses a character class that includes both whitespace (\s) and quote chars with `+` quantifier: `^[\s"'\u201C\u201D\u2018\u2019]+`. This means a title like `"  Spaced Title  "` correctly produces `Spaced Title`, but a title that legitimately starts with an apostrophe (e.g. `'Twas`) would lose it. Low-risk for session names but worth noting.


### `server/__tests__/auto-rename.test.ts`
_Solid improvement to auto-rename robustness with good sanitization logic and test coverage. Main concern is whether the Vertex model map removal could break callers that pass standard (non-Vertex) model names to Vertex clients._

- 🔵 **style** (L161): The server test relaxed its assertion from exact argument matching (toHaveBeenCalledWith) to individual field checks (call[0].model, call[0].messages[0].content). This is fine for resilience against prompt changes, but the system prompt (NAMING_SYSTEM_PROMPT) is now completely unverified in the server test suite. Consider adding a minimal assertion like `expect(call[0].system).toContain('3-6 word title')` so the server tests catch accidental system prompt deletions. `[fixable]`


---
_Reviewed at 474210de1c7dfd9421b049f3025da96a654aa026_

PR #425, opened 2026-07-02
feat(observability): MLflow session experiment tracking

dimakis:
## Summary
- Each Mitzo session becomes an MLflow run in the `mitzo-sessions` experiment
- Run tags capture session metadata (mode, model, branch, session ID)
- Final metrics logged at session end: tokens (input/output/cache), cost, turns, duration, compactions
- All MLflow calls are fire-and-forget no-ops when `MLFLOW_TRACKING_URI` is unset
- Handles both normal completion and fallback paths (disconnect/abort/timeout)

## Files
- `server/mlflow.ts` — new REST client (experiment management, run create/end)
- `server/query-loop.ts` — wired into session ID resolution + finally block
- `server/__tests__/mlflow.test.ts` — 9 tests covering create, end, caching, errors, disabled mode
- `.env.example` — added `MLFLOW_TRACKING_URI`

## Design
Spec: `mgmt/local_features/mlflow-session-tracking/spec.md`

Every commit to an agent definition or boot context recipe becomes an independent variable in a continuous experiment, with token spend/cost/turns as dependent variables. Phase 1 tracks session-level metrics; Phase 2 adds per-turn stepping and context consumed artifacts.

## Test plan
- [x] `npx vitest run server/__tests__/mlflow.test.ts` — 9/9 passing
- [ ] Deploy with `MLFLOW_TRACKING_URI=http://localhost:5050`, run a session, verify run appears in MLflow UI

### DIFF
--- a/server/mlflow.ts
+++ b/server/mlflow.ts
@@ -1,0 +1,225 @@
+/**
+ * MLflow session tracking — lightweight REST client.
+ *
+ * Each Mitzo session becomes an MLflow run in the "mitzo-sessions" experiment.
+ * All functions are no-ops when MLFLOW_TRACKING_URI is not set, so callers
+ * never need to guard.
+ *
+ * Endpoints used:
+ *   POST /api/2.0/mlflow/experiments/search
+ *   POST /api/2.0/mlflow/experiments/create
+ *   POST /api/2.0/mlflow/runs/create
+ *   POST /api/2.0/mlflow/runs/log-batch
+ *   POST /api/2.0/mlflow/runs/update
+ */
+
+import { createLogger } from './logger.js';
+
+const log = createLogger('mlflow');
+
+const TRACKING_URI = process.env.MLFLOW_TRACKING_URI ?? '';
+const EXPERIMENT_NAME = 'mitzo-sessions';
+
+let cachedExperimentId: string | null = null;
+
+export function isEnabled(): boolean {
+  return TRACKING_URI.length > 0;
+}
+
+// ---------------------------------------------------------------------------
+// Internal HTTP helper
+// ---------------------------------------------------------------------------
+
+async function post<T = unknown>(path: string, body: Record<string, unknown>): Promise<T> {
+  const url = `${TRACKING_URI}${path}`;
+  const res = await fetch(url, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+    signal: AbortSignal.timeout(5000),
+  });
+  if (!res.ok) {
+    const text = await res.text().catch(() => '');
+    throw new Error(`MLflow ${path} failed: ${res.status} ${text}`);
+  }
+  return res.json() as Promise<T>;
+}
+
+// ---------------------------------------------------------------------------
+// Experiment management
+// ---------------------------------------------------------------------------
+
+async function ensureExperiment(): Promise<string> {
+  if (cachedExperimentId) return cachedExperimentId;
+
+  // Search for existing experiment
+  interface SearchResult {
+    experiments?: Array<{ experiment_id: string; name: string }>;
+  }
+  const search = await post<SearchResult>('/api/2.0/mlflow/experiments/search', {
+    max_results: 1,
+    filter: `name = '${EXPERIMENT_NAME}'`,
+  });
+  if (search.experiments?.length) {
+    cachedExperimentId = search.experiments[0].experiment_id;
+    return cachedExperimentId;
+  }
+
+  // Create new experiment
+  interface CreateResult {
+    experiment_id: string;
+  }
+  const created = await post<CreateResult>('/api/2.0/mlflow/experiments/create', {
+    name: EXPERIMENT_NAME,
+  });
+  cachedExperimentId = created.experiment_id;
+  log.info('created MLflow experiment', { name: EXPERIMENT_NAME, id: cachedExperimentId });
+  return cachedExperimentId;
+}
+
+// ---------------------------------------------------------------------------
+// Public API
+// ---------------------------------------------------------------------------
+
+export interface SessionParams {
+  sessionId: string;
+  mode?: string;
+  model?: string;
+  cwd?: string;
+  branch?: string;
+  clientType?: string;
+  agentDefHash?: string;
+  recipeHash?: string;
+}
+
+export interface SessionMetrics {
+  inputTokens: number;
+  outputTokens: number;
+  cacheReadTokens: number;
+  cacheCreationTokens: number;
+  totalCostUsd: number;
+  numTurns: number;
+  durationMs: number;
+  durationApiMs: number;
+  numCompactions?: number;
+}
+
+/**
+ * Create an MLflow run for a session. Returns the run_id, or null if disabled.
+ */
+export async function createRun(params: SessionParams): Promise<string | null> {
+  if (!isEnabled()) return null;
+
+  try {
+    const experimentId = await ensureExperiment();
+
+    interface RunResult {
+      run: { info: { run_id: string } };
+    }
+
+    const tags = [
+      { key: 'session.id', value: params.sessionId },
+      ...(params.mode ? [{ key: 'session.mode', value: params.mode }] : []),
+      ...(params.model ? [{ key: 'session.model', value: params.model }] : []),
+      ...(params.clientType ? [{ key: 'session.client_type', value: params.clientType }] : []),
+      ...(params.agentDefHash
+        ? [{ key: 'context.agent_def_hash', value: params.agentDefHash }]
+        : []),
+      ...(params.recipeHash ? [{ key: 'context.recipe_hash', value: params.recipeHash }] : []),
+    ];
+
+    const result = await post<RunResult>('/api/2.0/mlflow/runs/create', {
+      experiment_id: experimentId,
+      run_name: params.sessionId,
+      start_time: Date.now(),
+      tags,
+    });
+
+    const runId = result.run.info.run_id;
+
+    // Log params (immutable session metadata)
+    const runParams = [
+      { key: 'cwd', value: params.cwd ?? '' },
+      { key: 'mode', value: params.mode ?? '' },
+      { key: 'model', value: params.model ?? '' },
+      { key: 'branch', value: params.branch ?? '' },
+    ];
+
+    await post('/api/2.0/mlflow/runs/log-batch', {
+      run_id: runId,
+      params: runParams,
+    });
+
+    log.info('created MLflow run', { runId, sessionId: params.sessionId });
+    return runId;
+  } catch (err) {
+    log.warn('failed to create MLflow run', {
+      error: err instanceof Error ? err.message : String(err),
+    });
+    return null;
+  }
+}
+
+/**
+ * End an MLflow run, logging final session metrics.
+ */
+export async function endRun(
+  runId: string | null,
+  metrics: SessionMetrics,
+  status: 'FINISHED' | 'FAILED' = 'FINISHED',
+): Promise<void> {
+  if (!isEnabled() || !runId) return;
+
+  try {
+    const now = Date.now();
+
+    // Log all metrics in a single batch
+    const mlMetrics = [
+      { key: 'tokens.input', value: metrics.inputTokens, timestamp: now, step: 0 },
+      { key: 'tokens.output', value: metrics.outputTokens, timestamp: now, step: 0 },
+      { key: 'tokens.cache_read', value: metrics.cacheReadTokens, timestamp: now, step: 0 },
+      {
+        key: 'tokens.cache_creation',
+        value: metrics.cacheCreationTokens,
+        timestamp: now,
+        step: 0,
+      },
+      { key: 'cost_usd', value: metrics.totalCostUsd, timestamp: now, step: 0 },
+      { key: 'num_turns', value: metrics.numTurns, timestamp: now, step: 0 },
+      { key: 'duration_ms', value: metrics.durationMs, timestamp: now, step: 0 },
+      { key: 'duration_api_ms', value: metrics.durationApiMs, timestamp: now, step: 0 },
+    ];
+
+    if (metrics.numCompactions !== undefined) {
+      mlMetrics.push({
+        key: 'num_compactions',
+        value: metrics.numCompactions,
+        timestamp: now,
+        step: 0,
+      });
+    }
+
+    await post('/api/2.0/mlflow/runs/log-batch', {
+      run_id: runId,
+      metrics: mlMetrics,
+    });
+
+    await post('/api/2.0/mlflow/runs/update', {
+      run_id: runId,
+      status,
+      end_time: now,
+    });
+
+    log.info('ended MLflow run', { runId, status });
+  } catch (err) {
+    log.warn('failed to end MLflow run', {
+      runId,
+      error: err instanceof Error ? err.message : String(err),
+    });
+  }
+}
+
+// For testing — reset cached experiment ID
+export function _resetCache(): void {
+  cachedExperimentId = null;
+}
--- a/server/query-loop.ts
+++ b/server/query-loop.ts
@@ -18,6 +18,7 @@ import { sendTurnCompleteNotification as apnsTurnComplete } from './apns.js';
 import { extractSnippet } from './notification-helpers.js';
 import { NOTIFY_SNIPPET_MAX_CHARS } from './constants.js';
 import { createGoal, reportUsage, deriveGoalTitle } from './goal-client.js';
+import * as mlflow from './mlflow.js';
 import { tracer } from './tracing.js';
 import { context, trace, SpanStatusCode, type Span } from '@opentelemetry/api';
 import { ProgressTracker } from './progress-tracker.js';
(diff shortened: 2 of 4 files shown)


dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Clean, well-tested observability module. The main concern is a race between fire-and-forget `createRun` and synchronous `mlflowRunId` check in `endRun` — store and await the promise like the adjacent `goalCreationPromise` pattern._

- 🟡 **bugs** (L540): Race condition: `createRun` is fire-and-forget (promise not stored), but `endRun` in the finally block checks `mlflowRunId` synchronously. If the session ends before `createRun` resolves (fast abort, immediate error, slow MLflow server), `mlflowRunId` is still `null` and `endRun` never fires — orphaning the MLflow run in RUNNING state. The `goalCreationPromise` pattern 10 lines above shows the fix: store the promise and `await` it before `endRun`. This is especially relevant since `createRun` makes 3–4 HTTP requests with 5s timeouts each. `[fixable]`
- 🔵 **bugs** (L1437): `totalCostUsd: lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||` which treats `0` as falsy. A zero-cost session (legitimate SDK result with $0.00) falls through to the alternative source. Should use `??` for consistency with the existing fallback at line 1411 which uses `??`. In practice both sources converge to the same value when cost is 0, so this won't corrupt data, but it's semantically incorrect. `[fixable]`
- 🔵 **unsafe_assumptions** (L553): `.catch(() => {})` silently swallows errors from `createRun` at the call site. While `mlflow.ts` logs internally, a reader of query-loop.ts has no indication errors are handled. The `goalCreationPromise` pattern above uses `.catch((err) => { log.warn(...); return null; })` — consider matching that pattern for consistency and debuggability. `[fixable]`


### `server/mlflow.ts`
_Clean, well-tested observability module. The main concern is a race between fire-and-forget `createRun` and synchronous `mlflowRunId` check in `endRun` — store and await the promise like the adjacent `goalCreationPromise` pattern._

- 🔵 **style** (L1): 14-line module-level doc comment violates the project convention: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." The endpoints list is especially verbose — callers don't need to know the REST paths. Consider collapsing to a single-line summary or removing entirely since the module name and exports are self-documenting. `[fixable]`


### `server/__tests__/mlflow.test.ts`
_Clean, well-tested observability module. The main concern is a race between fire-and-forget `createRun` and synchronous `mlflowRunId` check in `endRun` — store and await the promise like the adjacent `goalCreationPromise` pattern._

- 🔵 **missing_tests**: No test for the `isEnabled() === false` path of `endRun` — the test covers `runId === null` (line 230) but not the case where `isEnabled()` is false AND `runId` is non-null (the early return at the top of `endRun` checks both). Minor gap since the implementation clearly short-circuits, but it would complete the disabled-state coverage. `[fixable]`

server/query-loop.ts
```diff
@@ -533,6 +535,22 @@ async function _runQueryLoopInner(
                 return null;
               });
             }
+
+            // Create MLflow run for this session (fire-and-forget)
+            if (mlflow.isEnabled()) {
```
dimakis, line 540: 🟡 **bugs**: Race condition: `createRun` is fire-and-forget (promise not stored), but `endRun` in the finally block checks `mlflowRunId` synchronously. If the session ends before `createRun` resolves (fast abort, immediate error, slow MLflow server), `mlflowRunId` is still `null` and `endRun` never fires — orphaning the MLflow run in RUNNING state. The `goalCreationPromise` pattern 10 lines above shows the fix: store the promise and `await` it before `endRun`. This is especially relevant since `createRun` makes 3–4 HTTP requests with 5s timeouts each. `[fixable]`

server/query-loop.ts
```diff
@@ -1403,6 +1421,29 @@ async function _runQueryLoopInner(
...
+        mlflow
+          .endRun(
+            mlflowRunId,
+            {
+              inputTokens: lastReportedUsage.inputTokens,
+              outputTokens: doneSent ? lastReportedUsage.outputTokens : cumulativeOutputTokens,
+              cacheReadTokens: lastReportedUsage.cacheReadTokens,
+              cacheCreationTokens: lastReportedUsage.cacheCreationTokens,
+              totalCostUsd:
+                lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0),
```
dimakis, line 1437: 🔵 **bugs**: `totalCostUsd: lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||` which treats `0` as falsy. A zero-cost session (legitimate SDK result with $0.00) falls through to the alternative source. Should use `??` for consistency with the existing fallback at line 1411 which uses `??`. In practice both sources converge to the same value when cost is 0, so this won't corrupt data, but it's semantically incorrect. `[fixable]`

server/query-loop.ts
```diff
@@ -533,6 +535,22 @@ async function _runQueryLoopInner(
...
+                  mode: currentSession.mode,
+                  model: currentSession.model,
+                  cwd: currentSession.cwd,
+                  branch: currentSession.branch,
+                })
+                .then((id) => {
+                  mlflowRunId = id;
+                })
+                .catch(() => {});
+            }
```
dimakis, line 553: 🔵 **unsafe_assumptions**: `.catch(() => {})` silently swallows errors from `createRun` at the call site. While `mlflow.ts` logs internally, a reader of query-loop.ts has no indication errors are handled. The `goalCreationPromise` pattern above uses `.catch((err) => { log.warn(...); return null; })` — consider matching that pattern for consistency and debuggability. `[fixable]`

server/mlflow.ts
```diff
@@ -0,0 +1,225 @@
+/**
```
dimakis, line 1: 🔵 **style**: 14-line module-level doc comment violates the project convention: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max." The endpoints list is especially verbose — callers don't need to know the REST paths. Consider collapsing to a single-line summary or removing entirely since the module name and exports are self-documenting. `[fixable]`

server/query-loop.ts
```diff
@@ -533,6 +535,22 @@ async function _runQueryLoopInner(
                 return null;
```
dimakis, line 535: 🟡 **bugs**: Race condition: `createRun` is fire-and-forget (`.then()` assigns `mlflowRunId`), but `endRun` in the `finally` block guards on `if (mlflowRunId)`. If the session ends before `createRun` resolves (fast error, slow MLflow — up to 4 network hops × 5s timeout each), `mlflowRunId` is still `null`, so `endRun` is silently skipped and the MLflow run is orphaned in RUNNING state. The existing `goalCreationPromise` pattern solves this correctly: store the promise, then `await` it before consuming the result. Fix: `let mlflowRunPromise: Promise<string | null> | undefined;` → store the promise → `await mlflowRunPromise` in the finally block before calling `endRun`. `[fixable]`

server/query-loop.ts
```diff
@@ -1403,6 +1421,29 @@ async function _runQueryLoopInner(
...
+          .endRun(
+            mlflowRunId,
+            {
+              inputTokens: lastReportedUsage.inputTokens,
+              outputTokens: doneSent ? lastReportedUsage.outputTokens : cumulativeOutputTokens,
+              cacheReadTokens: lastReportedUsage.cacheReadTokens,
+              cacheCreationTokens: lastReportedUsage.cacheCreationTokens,
+              totalCostUsd:
+                lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0),
+              numTurns: doneSent ? lastReportedUsage.numTurns : turnIndex,
```
dimakis, line 1438: 🔵 **unsafe_assumptions**: `totalCostUsd: lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||`, which treats a legitimate zero cost as falsy and falls through to the session accumulator. In practice both paths yield zero so the result is correct, but `??` would express the intent more precisely ("use fallback only when the value is null/undefined, not when it's zero"). `[fixable]`

server/query-loop.ts
```diff
@@ -533,6 +535,22 @@ async function _runQueryLoopInner(
...
+            // Create MLflow run for this session (fire-and-forget)
+            if (mlflow.isEnabled()) {
+              mlflow
+                .createRun({
+                  sessionId: resolvedSessionId,
+                  mode: currentSession.mode,
+                  model: currentSession.model,
+                  cwd: currentSession.cwd,
+                  branch: currentSession.branch,
+                })
```
dimakis, line 548: 🔵 **style**: The `.catch(() => {})` on `createRun` silently swallows errors. While `createRun` has internal try/catch that logs warnings and returns null, the outer catch would only fire for truly unexpected failures (e.g., in the `.then()` callback). The adjacent `goalCreationPromise` logs in its catch. Minor inconsistency — consider `.catch(() => { /* logged internally */ })` or matching the goal pattern. `[fixable]`

server/mlflow.ts
dimakis, line 1: 🔵 **style**: The 14-line module-level JSDoc comment lists endpoints and describes behavior. CLAUDE.md says "default to writing no comments" and "never write multi-line comment blocks." The endpoint list is useful as a quick reference, but could be trimmed to a one-liner or moved to the CLAUDE.md architecture section where it's already partially documented. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Clean MLflow integration following existing patterns (goal tracking, notification fire-and-forget). One real race condition: the fire-and-forget `createRun` can leave orphaned MLflow runs if the session ends before the promise resolves — store the promise and await it like `goalCreationPromise`._

- 🟡 **bugs** (L535): Race condition: `createRun` is fire-and-forget (`.then()` assigns `mlflowRunId`), but `endRun` in the `finally` block guards on `if (mlflowRunId)`. If the session ends before `createRun` resolves (fast error, slow MLflow — up to 4 network hops × 5s timeout each), `mlflowRunId` is still `null`, so `endRun` is silently skipped and the MLflow run is orphaned in RUNNING state. The existing `goalCreationPromise` pattern solves this correctly: store the promise, then `await` it before consuming the result. Fix: `let mlflowRunPromise: Promise<string | null> | undefined;` → store the promise → `await mlflowRunPromise` in the finally block before calling `endRun`. `[fixable]`
- 🔵 **unsafe_assumptions** (L1438): `totalCostUsd: lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||`, which treats a legitimate zero cost as falsy and falls through to the session accumulator. In practice both paths yield zero so the result is correct, but `??` would express the intent more precisely ("use fallback only when the value is null/undefined, not when it's zero"). `[fixable]`
- 🔵 **style** (L548): The `.catch(() => {})` on `createRun` silently swallows errors. While `createRun` has internal try/catch that logs warnings and returns null, the outer catch would only fire for truly unexpected failures (e.g., in the `.then()` callback). The adjacent `goalCreationPromise` logs in its catch. Minor inconsistency — consider `.catch(() => { /* logged internally */ })` or matching the goal pattern. `[fixable]`


### `server/mlflow.ts`
_Clean MLflow integration following existing patterns (goal tracking, notification fire-and-forget). One real race condition: the fire-and-forget `createRun` can leave orphaned MLflow runs if the session ends before the promise resolves — store the promise and await it like `goalCreationPromise`._

- 🔵 **style** (L1): The 14-line module-level JSDoc comment lists endpoints and describes behavior. CLAUDE.md says "default to writing no comments" and "never write multi-line comment blocks." The endpoint list is useful as a quick reference, but could be trimmed to a one-liner or moved to the CLAUDE.md architecture section where it's already partially documented. `[fixable]`

server/query-loop.ts
```diff
@@ -533,6 +535,22 @@ async function _runQueryLoopInner(
                 return null;
               });
             }
+
+            // Create MLflow run for this session (fire-and-forget)
```
dimakis, line 539: 🟡 **bugs**: Race condition: `createRun` is fire-and-forget — `mlflowRunId` is set inside a `.then()` callback, but `endRun` at line 1426 checks `if (mlflowRunId)` synchronously. If the session ends before `createRun` resolves (fast error, immediate disconnect), `mlflowRunId` is still `null`, the `endRun` is skipped, and the MLflow run is orphaned in RUNNING state forever. The codebase already has a pattern for this: `goalCreationPromise` (line 529) stores the promise and `await`s it at session end (line 602). Store the `createRun` promise similarly and `await` it before the `endRun` block. `[fixable]`

server/query-loop.ts
```diff
@@ -1403,6 +1421,29 @@ async function _runQueryLoopInner(
...
+        const fallbackDurationMs = Date.now() - sessionStartedAt;
+        mlflow
+          .endRun(
+            mlflowRunId,
+            {
+              inputTokens: lastReportedUsage.inputTokens,
+              outputTokens: doneSent ? lastReportedUsage.outputTokens : cumulativeOutputTokens,
+              cacheReadTokens: lastReportedUsage.cacheReadTokens,
+              cacheCreationTokens: lastReportedUsage.cacheCreationTokens,
+              totalCostUsd:
```
dimakis, line 1436: 🔵 **bugs**: `lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||`, which treats a legitimate `0` cost the same as "no data reported." A zero-cost session (e.g. cached response) would fall through to the `cumulativeCostUsd` fallback. Use `??` instead of `||` for numeric fallback to distinguish zero from absent. `[fixable]`

server/mlflow.ts
```diff
@@ -0,0 +1,225 @@
...
+ *   POST /api/2.0/mlflow/runs/create
+ *   POST /api/2.0/mlflow/runs/log-batch
+ *   POST /api/2.0/mlflow/runs/update
+ */
+
+import { createLogger } from './logger.js';
+
+const log = createLogger('mlflow');
+
+const TRACKING_URI = process.env.MLFLOW_TRACKING_URI ?? '';
```
dimakis, line 20: 🔵 **unsafe_assumptions**: `TRACKING_URI` is captured once at module load time. This is consistent with other env-based config in the codebase (e.g. `OTEL_EXPORTER_OTLP_ENDPOINT`), so it's fine — but worth noting that `isEnabled()` can't change at runtime without a server restart.

server/mlflow.ts
dimakis, line 1: 🔵 **style**: The module-level doc block is 14 lines long. Per project conventions: "default to writing no comments" and "never write multi-paragraph docstrings or multi-line comment blocks — one short line max." Consider trimming to one line or removing entirely since the function names and types are self-documenting. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Clean, well-tested MLflow integration. The main concern is a race condition where `createRun` is fire-and-forget but `endRun` depends on its result — short sessions can orphan MLflow runs. Follow the existing `goalCreationPromise` pattern to store and await the promise._

- 🟡 **bugs** (L539): Race condition: `createRun` is fire-and-forget — `mlflowRunId` is set inside a `.then()` callback, but `endRun` at line 1426 checks `if (mlflowRunId)` synchronously. If the session ends before `createRun` resolves (fast error, immediate disconnect), `mlflowRunId` is still `null`, the `endRun` is skipped, and the MLflow run is orphaned in RUNNING state forever. The codebase already has a pattern for this: `goalCreationPromise` (line 529) stores the promise and `await`s it at session end (line 602). Store the `createRun` promise similarly and `await` it before the `endRun` block. `[fixable]`
- 🔵 **bugs** (L1436): `lastReportedUsage.totalCostUsd || (finalSession?.cumulativeCostUsd ?? 0)` uses `||`, which treats a legitimate `0` cost the same as "no data reported." A zero-cost session (e.g. cached response) would fall through to the `cumulativeCostUsd` fallback. Use `??` instead of `||` for numeric fallback to distinguish zero from absent. `[fixable]`


### `server/mlflow.ts`
_Clean, well-tested MLflow integration. The main concern is a race condition where `createRun` is fire-and-forget but `endRun` depends on its result — short sessions can orphan MLflow runs. Follow the existing `goalCreationPromise` pattern to store and await the promise._

- 🔵 **unsafe_assumptions** (L20): `TRACKING_URI` is captured once at module load time. This is consistent with other env-based config in the codebase (e.g. `OTEL_EXPORTER_OTLP_ENDPOINT`), so it's fine — but worth noting that `isEnabled()` can't change at runtime without a server restart.
- 🔵 **style** (L1): The module-level doc block is 14 lines long. Per project conventions: "default to writing no comments" and "never write multi-paragraph docstrings or multi-line comment blocks — one short line max." Consider trimming to one line or removing entirely since the function names and types are self-documenting. `[fixable]`


### `server/__tests__/mlflow.test.ts`
_Clean, well-tested MLflow integration. The main concern is a race condition where `createRun` is fire-and-forget but `endRun` depends on its result — short sessions can orphan MLflow runs. Follow the existing `goalCreationPromise` pattern to store and await the promise._

- 🔵 **missing_tests**: No test for `endRun` when the MLflow API call fails (e.g., `log-batch` returns 500). The `createRun` error path is tested, but `endRun`'s try/catch (which swallows errors and logs a warning) is not exercised. Adding a test would confirm the error is gracefully handled and doesn't throw. `[fixable]`

PR #428, opened 2026-07-02
feat: compaction indicator in token bar

dimakis:
## Summary

- Detect SDK compaction content blocks (`type: 'compaction'`) and emit `compaction_status` events to the frontend
- TokenBar replaces normal display with a pulsing amber **COMPACTING** label during active compaction
- Crash-safe: `compaction_delta` events are explicitly skipped, compaction blocks don't create phantom snapshot state
- Safety reset: if compaction flag is still active when the query ends, it resets with a warning log

## Files changed

- `server/query-loop.ts` — detect compaction blocks, emit status events, safety reset
- `frontend/src/components/TokenBar.tsx` — compacting state renders amber indicator
- `frontend/src/styles/global.css` — pulsing amber animation
- `frontend/src/types/ws-messages.ts` — `CompactionStatusMsg` type
- `packages/client/src/slices/tokens.ts` — `compacting: boolean` in `TokensState`
- `packages/client/src/protocol-parser.ts` — parse `compaction_status`
- `server/__tests__/token-update.test.ts` — full compaction block lifecycle coverage
- `packages/client/__tests__/protocol-parser.test.ts` — compaction_status parsing

## Test plan

- [x] Server: token-update tests pass (7/7) — includes compaction block start/delta/stop + status events
- [x] Client: protocol-parser tests pass (57/57) — includes compaction_status flag setting
- [ ] Manual: trigger compaction in a long session and verify amber indicator appears

### DIFF
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -523,7 +523,7 @@ export function parseServerMessage(
     }
 
     case 'compaction_status':
-      result.tokensUpdate = { compacting: msg.active as boolean };
+      result.tokensUpdate = { compacting: Boolean(msg.active) };
       break;
 
     // Progress tracking messages
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -522,6 +522,10 @@ export function parseServerMessage(
       break;
     }
 
+    case 'compaction_status':
+      result.tokensUpdate = { compacting: msg.active as boolean };
+      break;
+
     // Progress tracking messages
     case 'progress_start':
       result.progressUpdate = {
--- a/server/__tests__/token-update.test.ts
+++ b/server/__tests__/token-update.test.ts
@@ -455,6 +455,53 @@ describe('token_update emission', () => {
     expect(compactionStatuses[1]).toMatchObject({ type: 'compaction_status', active: false });
   });
 
+  it('resets compacting flag on result when success signal is missing', async () => {
+    const events: Record<string, unknown>[] = [
+      {
+        type: 'stream_event',
+        parent_tool_use_id: null,
+        event: {
+          type: 'message_start',
+          message: { id: 'msg-crash', usage: { input_tokens: 180000 } },
+        },
+      },
+      {
+        type: 'stream_event',
+        event: { type: 'content_block_start', index: 0, content_block: { type: 'text' } },
+      },
+      { type: 'stream_event', event: { type: 'content_block_stop', index: 0 } },
+      { type: 'assistant', message: { content: [] }, session_id: 'sess-crash' },
+      // Compaction starts but SDK never sends compact_result: 'success'
+      {
+        type: 'stream_event',
+        event: {
+          type: 'content_block_start',
+          index: 1,
+          content_block: { type: 'compaction' },
+        },
+      },
+      { type: 'stream_event', event: { type: 'content_block_stop', index: 1 } },
+      // Result arrives directly — no compact_result system message
+      {
+        type: 'result',
+        session_id: 'sess-crash',
+        usage: { input_tokens: 180000, output_tokens: 500 },
+        total_cost_usd: 0.01,
+        num_turns: 1,
+        duration_ms: 3000,
+        duration_api_ms: 2000,
+      },
+    ];
+
+    await runQueryLoop(eventStream(events), clientId, registry, abortController);
+
+    const compactionStatuses = transport.sent.filter((m) => m.type === 'compaction_status');
+    // Should have active:true from block_start and active:false from safety reset
+    expect(compactionStatuses).toHaveLength(2);
+    expect(compactionStatuses[0]).toMatchObject({ active: true });
+    expect(compactionStatuses[1]).toMatchObject({ active: false });
+  });
+
   it('handles missing usage on message_start gracefully', async () => {
     const events: Record<string, unknown>[] = [
       {
--- a/server/__tests__/token-update.test.ts
+++ b/server/__tests__/token-update.test.ts
@@ -389,6 +389,24 @@ describe('token_update emission', () => {
       },
       { type: 'stream_event', event: { type: 'content_block_stop', index: 0 } },
       { type: 'assistant', message: { content: [] }, session_id: 'sess-compact' },
+      // SDK compaction content block — start, delta, stop
+      {
+        type: 'stream_event',
+        event: {
+          type: 'content_block_start',
+          index: 1,
+          content_block: { type: 'compaction' },
+        },
+      },
+      {
+        type: 'stream_event',
+        event: {
+          type: 'content_block_delta',
+          index: 1,
+          delta: { type: 'compaction_delta', content: 'Summary of prior context...' },
+        },
+      },
+      { type: 'stream_event', event: { type: 'content_block_stop', index: 1 } },
       // SDK system status: compaction completed
       {
         type: 'system',
@@ -429,6 +447,12 @@ describe('token_update emission', () => {
     // The final token_update should include numCompactions: 1
     const last = tokenUpdates[tokenUpdates.length - 1];
     expect(last).toMatchObject({ numCompactions: 1 });
+
+    // compaction_status events should bracket the compaction
+    const compactionStatuses = transport.sent.filter((m) => m.type === 'compaction_status');
+    expect(compactionStatuses).toHaveLength(2);
+    expect(compactionStatuses[0]).toMatchObject({ type: 'compaction_status', active: true });
+    expect(compactionStatuses[1]).toMatchObject({ type: 'compaction_status', active: false });
   });
 
   it('handles missing usage on message_start gracefully', async () => {
--- a/server/__tests__/workload-routes.test.ts
+++ b/server/__tests__/workload-routes.test.ts
@@ -509,13 +509,56 @@ describe('workload routes', () => {
     expect(res.body.item.goalId).toBe(res.body.task.id);
   });
 
-  it('POST /api/workload/items/:id/promote — returns 404 for missing item', async () => {
+  it('POST /api/workload/items/:id/promote — returns 404 for missing item with no body title', async () => {
     const res = await request(app)
       .post('/api/workload/items/nonexistent-id/promote')
       .set('Cookie', authCookie)
       .send({});
 
     expect(res.status).toBe(404);
+  });
+
+  it('POST /api/workload/items/:id/promote — creates task from body fallback (Telos item)', async () => {
+    const res = await request(app)
+      .post('/api/workload/items/telos-item-123/promote')
+      .set('Cookie', authCookie)
+      .send({
+        title: 'Telos task title',
+        description: 'Extra context',
+        contextHints: {
+          repos: ['mitzo'],
+          taskHint: 'Check the API layer',
+        },
+        sources: [{ type: 'telos', url: 'https://example.com', title: 'Source doc' }],
+      });
+
+    expect(res.status).toBe(201);
+    expect(res.body.task).toMatchObject({
+      title: 'Telos task title',
+      status: 'pending',
+    });
+    expect(res.body.task.description).toContain('Extra context');
+    expect(res.body.task.description).toContain('Check the API layer');
+    expect(res.body.task.annotations).toEqual(
+      expect.arrayContaining([expect.stringContaining('Source doc')]),
+    );
+    expect(res.body.item).toBeUndefined();
+  });
+
+  it('POST /api/workload/items/:id/promote — fallback does not broadcast workload update', async () => {
+    const broadcasts: unknown[] = [];
+    const origBroadcast = (app as any)._workloadBroadcast;
+    (app as any)._workloadBroadcast = (msg: unknown) => broadcasts.push(msg);
+
+    await request(app)
+      .post('/api/workload/items/telos-no-broadcast/promote')
+      .set('Cookie', authCookie)
+      .send({ title: 'No broadcast test' });
+
+    const workloadBroadcasts = broadcasts.filter((b: any) => b.type === 'workload_item_updated');
+    expect(workloadBroadcasts).toHaveLength(0);
+
+    (app as any)._workloadBroadcast = origBroadcast;
   });
 
   it('POST /api/workload/items/:id/promote — broadcasts workload_item_updated', async () => {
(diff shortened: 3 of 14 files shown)


server/__tests__/workload-routes.test.ts
```diff
@@ -518,6 +518,49 @@ describe('workload routes', () => {
...
+    });
+    expect(res.body.task.description).toContain('Extra context');
+    expect(res.body.task.description).toContain('Check the API layer');
+    expect(res.body.task.annotations).toEqual(
+      expect.arrayContaining([expect.stringContaining('Source doc')]),
+    );
+    expect(res.body.item).toBeUndefined();
+  });
+
+  it('POST /api/workload/items/:id/promote — fallback does not broadcast workload update', async () => {
```
dimakis, line 548: 🟡 **missing_tests**: This test is ineffective — it monkey-patches `(app as any)._workloadBroadcast`, but that property does not exist on the app object. The actual broadcast callback is the module-level closure `onWorkloadBroadcast` (set via `setWorkloadBroadcast()` in index.ts, never called in the test harness). Since `onWorkloadBroadcast` is always `null` in tests, the assertion `expect(workloadBroadcasts).toHaveLength(0)` passes vacuously for every promote call, not just the fallback path. The test doesn't verify what it claims. `[fixable]`

server/__tests__/token-update.test.ts
```diff
@@ -429,6 +447,12 @@ describe('token_update emission', () => {
     // The final token_update should include numCompactions: 1
     const last = tokenUpdates[tokenUpdates.length - 1];
     expect(last).toMatchObject({ numCompactions: 1 });
```
dimakis, line 449: 🔵 **missing_tests**: The safety reset at query-loop.ts:573-581 (compaction flag reset on result when the SDK crashes/aborts before sending the system status success signal) has no test coverage. Only the happy path (content_block_start → system status success) is tested. A test with compaction started but result arriving without a preceding system status success would exercise the warn+reset path. `[fixable]`

server/query-loop.ts
```diff
@@ -956,6 +967,12 @@ async function _runQueryLoopInner(
                   blockType: 'text',
                 }),
               );
+            } else if (blockType === 'compaction') {
```
dimakis, line 970: 🔵 **style**: Compaction content_block_start still runs the generic registration at lines 900-902 (blockIdByIndex.set + openBlockCount++) before reaching the compaction branch. The blockId and index mapping are unused — at content_block_stop, the code finds the blockId but then no snapshot block exists, so block_end is silently skipped. The openBlockCount stays balanced (incremented then decremented), so this is harmless, but a brief comment noting this intentional fall-through would prevent future confusion. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -522,6 +522,10 @@ export function parseServerMessage(
       break;
     }
 
+    case 'compaction_status':
```
dimakis, line 525: 🔵 **style**: `msg.active as boolean` is a bare type assertion on an untyped message field. Other cases in this switch use similar casts (e.g., `msg.agentContext as number`), so this is consistent with the codebase, but a `Boolean(msg.active)` coercion would be more defensive against unexpected payloads. `[fixable]`

dimakis (commented):
## Centaur Review

_Well-structured feature with good test coverage for the happy path. The main concern is the broadcast-interception test for the promote fallback path, which passes vacuously due to monkey-patching a non-existent property; the compaction safety-reset path also lacks test coverage._

- 🟡 **missing_tests** (L548): This test is ineffective — it monkey-patches `(app as any)._workloadBroadcast`, but that property does not exist on the app object. The actual broadcast callback is the module-level closure `onWorkloadBroadcast` (set via `setWorkloadBroadcast()` in index.ts, never called in the test harness). Since `onWorkloadBroadcast` is always `null` in tests, the assertion `expect(workloadBroadcasts).toHaveLength(0)` passes vacuously for every promote call, not just the fallback path. The test doesn't verify what it claims. `[fixable]`


### `server/__tests__/token-update.test.ts`
_Well-structured feature with good test coverage for the happy path. The main concern is the broadcast-interception test for the promote fallback path, which passes vacuously due to monkey-patching a non-existent property; the compaction safety-reset path also lacks test coverage._

- 🔵 **missing_tests** (L449): The safety reset at query-loop.ts:573-581 (compaction flag reset on result when the SDK crashes/aborts before sending the system status success signal) has no test coverage. Only the happy path (content_block_start → system status success) is tested. A test with compaction started but result arriving without a preceding system status success would exercise the warn+reset path. `[fixable]`


### `server/query-loop.ts`
_Well-structured feature with good test coverage for the happy path. The main concern is the broadcast-interception test for the promote fallback path, which passes vacuously due to monkey-patching a non-existent property; the compaction safety-reset path also lacks test coverage._

- 🔵 **style** (L970): Compaction content_block_start still runs the generic registration at lines 900-902 (blockIdByIndex.set + openBlockCount++) before reaching the compaction branch. The blockId and index mapping are unused — at content_block_stop, the code finds the blockId but then no snapshot block exists, so block_end is silently skipped. The openBlockCount stays balanced (incremented then decremented), so this is harmless, but a brief comment noting this intentional fall-through would prevent future confusion. `[fixable]`


_Well-structured feature with good test coverage for the happy path. The main concern is the broadcast-interception test for the promote fallback path, which passes vacuously due to monkey-patching a non-existent property; the compaction safety-reset path also lacks test coverage._

- 🔵 **style** (L525): `msg.active as boolean` is a bare type assertion on an untyped message field. Other cases in this switch use similar casts (e.g., `msg.agentContext as number`), so this is consistent with the codebase, but a `Boolean(msg.active)` coercion would be more defensive against unexpected payloads. `[fixable]`

server/__tests__/workload-routes.test.ts
dimakis, line 548: 🟡 **missing_tests**: The 'fallback does not broadcast workload update' test intercepts `(app as any)._workloadBroadcast`, but that property doesn't exist on the Express app. The actual broadcast function is the module-level `onWorkloadBroadcast` closure, set via `setWorkloadBroadcast()`. Since `onWorkloadBroadcast` is never set in this test suite, it's always `null`, so the test passes vacuously — it would pass even if the `if (updatedItem)` guard were removed. The assertion verifies the correct outcome but doesn't actually test the code path. `[fixable]`

server/__tests__/workload-routes.test.ts
```diff
@@ -509,7 +509,7 @@ describe('workload routes', () => {
     expect(res.body.item.goalId).toBe(res.body.task.id);
   });
 
-  it('POST /api/workload/items/:id/promote — returns 404 for missing item', async () => {
+  it('POST /api/workload/items/:id/promote — returns 404 for missing item with no body title', async () => {
```
dimakis, line 512: 🔵 **style**: The renamed test description `'returns 404 for missing item with no body title'` is slightly awkward — it reads as though the body has 'no body title' rather than 'no title in body'. Consider `'returns 404 when item not found and body has no title'`. `[fixable]`

server/app.ts
```diff
@@ -1916,34 +1916,51 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
...
+  // Resolve title and context from workloadStore item or fallback body data (Telos items)
+  const title = item?.title ?? body.data.title;
+  if (!title) {
+    res.status(404).json({ error: 'Item not found and no title provided' });
     return;
   }
 
+  const hints = item?.contextHints ?? body.data.contextHints;
+  const taskHint = hints?.taskHint ?? undefined;
+
```
dimakis, line 1929: 🔵 **style**: When the item isn't found AND no title is provided, a 404 is returned with `'Item not found and no title provided'`. A 404 is semantically correct for a missing resource, but when the item is intentionally absent (Telos fallback path) and the real issue is the missing `title` field, a 400 would more accurately convey that the request body is incomplete. Minor — current behavior is workable since the only caller always provides a title. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean two-feature PR (compaction indicator + promote fallback). The compaction lifecycle is well-structured with proper safety nets. One test (`fallback does not broadcast`) has a broken interception mechanism that makes it pass vacuously rather than verifying the guard logic._

- 🟡 **missing_tests** (L548): The 'fallback does not broadcast workload update' test intercepts `(app as any)._workloadBroadcast`, but that property doesn't exist on the Express app. The actual broadcast function is the module-level `onWorkloadBroadcast` closure, set via `setWorkloadBroadcast()`. Since `onWorkloadBroadcast` is never set in this test suite, it's always `null`, so the test passes vacuously — it would pass even if the `if (updatedItem)` guard were removed. The assertion verifies the correct outcome but doesn't actually test the code path. `[fixable]`
- 🔵 **style** (L512): The renamed test description `'returns 404 for missing item with no body title'` is slightly awkward — it reads as though the body has 'no body title' rather than 'no title in body'. Consider `'returns 404 when item not found and body has no title'`. `[fixable]`


### `server/app.ts`
_Clean two-feature PR (compaction indicator + promote fallback). The compaction lifecycle is well-structured with proper safety nets. One test (`fallback does not broadcast`) has a broken interception mechanism that makes it pass vacuously rather than verifying the guard logic._

- 🔵 **style** (L1929): When the item isn't found AND no title is provided, a 404 is returned with `'Item not found and no title provided'`. A 404 is semantically correct for a missing resource, but when the item is intentionally absent (Telos fallback path) and the real issue is the missing `title` field, a 400 would more accurately convey that the request body is incomplete. Minor — current behavior is workable since the only caller always provides a title. `[fixable]`


### `server/__tests__/token-update.test.ts`
_Clean two-feature PR (compaction indicator + promote fallback). The compaction lifecycle is well-structured with proper safety nets. One test (`fallback does not broadcast`) has a broken interception mechanism that makes it pass vacuously rather than verifying the guard logic._

- 🔵 **missing_tests** (L447): No test covers the safety-net path where the `result` event arrives while `compacting` is still true (i.e., SDK crashed or skipped the `compact_result: 'success'` system message). This would verify the warn log and the fallback `compaction_status: active: false` emission. The logic is straightforward but it's a defensive branch worth covering. `[fixable]`

server/__tests__/token-update.test.ts
```diff
@@ -429,6 +447,12 @@ describe('token_update emission', () => {
     // The final token_update should include numCompactions: 1
```
dimakis, line 447: 🔵 **missing_tests**: No test covers the safety-net path where the `result` event arrives while `compacting` is still true (i.e., SDK crashed or skipped the `compact_result: 'success'` system message). This would verify the warn log and the fallback `compaction_status: active: false` emission. The logic is straightforward but it's a defensive branch worth covering. `[fixable]`

server/__tests__/workload-routes.test.ts
```diff
@@ -518,6 +518,49 @@ describe('workload routes', () => {
...
+    expect(res.body.task.description).toContain('Check the API layer');
+    expect(res.body.task.annotations).toEqual(
+      expect.arrayContaining([expect.stringContaining('Source doc')]),
+    );
+    expect(res.body.item).toBeUndefined();
+  });
+
+  it('POST /api/workload/items/:id/promote — fallback does not broadcast workload update', async () => {
+    const broadcasts: unknown[] = [];
+    const origBroadcast = (app as any)._workloadBroadcast;
```
dimakis, line 550: 🟡 **bugs**: Test 'fallback does not broadcast workload update' monkey-patches `(app as any)._workloadBroadcast`, but the actual broadcast callback is a module-scoped closure variable `onWorkloadBroadcast` set via `setWorkloadBroadcast()`. The test creates a property on the `app` object that nothing reads, so the assertion passes vacuously — the broadcast is never intercepted. Furthermore, `setWorkloadBroadcast` is never called in the test setup, so `onWorkloadBroadcast` is `null` and no broadcast fires at all. This test proves nothing about the fallback path's broadcast behavior. `[fixable]`

server/__tests__/workload-routes.test.ts
dimakis, line 512: 🔵 **style**: The test description was renamed to 'returns 404 for missing item with no body title' but the test body only sends `{ description: '' }` — consider also adding a test that sends neither `title` nor `description` (empty body) to cover the pure 404 case. `[fixable]`

server/app.ts
```diff
@@ -1916,34 +1916,51 @@ app.post('/api/workload/items/:id/promote', (req, res) => {
   }
 
   const item = workloadStore.get(req.params.id);
-  if (!item) {
-    res.status(404).json({ error: 'Item not found' });
+
+  // Resolve title and context from workloadStore item or fallback body data (Telos items)
+  const title = item?.title ?? body.data.title;
+  if (!title) {
```
dimakis, line 1922: 🔵 **style**: When the item doesn't exist and no `body.data.title` is provided, the error response is `404` with `'Item not found and no title provided'`. A `404` is debatable when the real issue is a missing required field — `400` would be more accurate for the 'no title provided' case. Consider splitting: `404` when `!item && !body.data.title`, vs keeping `404` for a true 'not found'. Minor since the frontend always sends a title for Telos items. `[fixable]`

dimakis (commented):
## Centaur Review

_Compaction indicator is well-implemented with proper state lifecycle (including defensive reset on SDK crash). Promote fallback logic is sound, but the 'no broadcast' test is broken — it monkey-patches a non-existent app property instead of the module-scoped callback, passing vacuously._

- 🟡 **bugs** (L550): Test 'fallback does not broadcast workload update' monkey-patches `(app as any)._workloadBroadcast`, but the actual broadcast callback is a module-scoped closure variable `onWorkloadBroadcast` set via `setWorkloadBroadcast()`. The test creates a property on the `app` object that nothing reads, so the assertion passes vacuously — the broadcast is never intercepted. Furthermore, `setWorkloadBroadcast` is never called in the test setup, so `onWorkloadBroadcast` is `null` and no broadcast fires at all. This test proves nothing about the fallback path's broadcast behavior. `[fixable]`
- 🔵 **style** (L512): The test description was renamed to 'returns 404 for missing item with no body title' but the test body only sends `{ description: '' }` — consider also adding a test that sends neither `title` nor `description` (empty body) to cover the pure 404 case. `[fixable]`


### `server/app.ts`
_Compaction indicator is well-implemented with proper state lifecycle (including defensive reset on SDK crash). Promote fallback logic is sound, but the 'no broadcast' test is broken — it monkey-patches a non-existent app property instead of the module-scoped callback, passing vacuously._

- 🔵 **style** (L1922): When the item doesn't exist and no `body.data.title` is provided, the error response is `404` with `'Item not found and no title provided'`. A `404` is debatable when the real issue is a missing required field — `400` would be more accurate for the 'no title provided' case. Consider splitting: `404` when `!item && !body.data.title`, vs keeping `404` for a true 'not found'. Minor since the frontend always sends a title for Telos items. `[fixable]`

PR #430, opened 2026-07-02
docs(design): transport SSOT architecture

dimakis:
## Summary

- Design doc for server-authoritative transport architecture to permanently fix the recurring "send twice to wake agent" bug
- Based on industry analysis of Anthropic SDK, OpenAI, LibreChat, LobeChat, and Vercel AI SDK transport patterns
- 5-phase migration: P0 foundation (state events, crash recovery, SSE default), P1 server-authoritative state, P2 always-send (remove client routing), P3 clean reconnect (cursor-based, multi-client fan-out), P4 cleanup (remove WS)
- Identifies 6 distinct root causes across 45+ fix attempts and 3 transport architectures

## Resolved Decisions

1. **SSE-first:** Flip SSE to default in P0, remove WS in P4
2. **Multi-client fan-out:** Designed in from the start — any client can send, all clients receive events, no primary/secondary
3. **EventStore migration:** Backwards-compatible — add `state` alongside `is_active`, migrate readers, drop `is_active` in P4
4. **Heartbeat:** Default 20s, configurable via `MITZO_HEARTBEAT_INTERVAL_MS`

## Test plan

- [ ] Review design doc for completeness
- [ ] Validate assumptions against codebase (done — code audit confirms all assumptions)
- [ ] Resolve open questions (done — all 4 resolved)
- [ ] Centaur review (2 rounds completed)
- [ ] Begin P0 implementation

### DIFF
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -1,6 +1,6 @@
 # Transport Single Source of Truth
 
-**Status:** Proposed
+**Status:** Approved
 **Author:** Claude Opus 4.6 + Dimitri Saridakis
 **Created:** 2026-07-02
 **Telos:** 97361f814c5f54df
@@ -77,7 +77,7 @@ This is a chat application transport layer. Thousands of production systems have
 CLIENT                          SERVER
 ------                          ------
 
-  |--- POST /send ----------------->|
+  |--- POST /api/chat/send -------->|
   |<-- 202 { accepted, jobId } -----|    (write acknowledged)
   |                                 |
   |<-- SSE: state=running ----------|    (authoritative state)
@@ -86,7 +86,7 @@ CLIENT                          SERVER
   |<-- SSE: block_delta ------------|
   |<-- SSE: state=idle -------------|    (turn complete)
   |                                 |
-  |--- POST /send ----------------->|    (next turn, always immediate)
+  |--- POST /api/chat/send -------->|    (next turn, always immediate)
   |<-- 202 { accepted } ------------|
   |                                 |
   |  [SSE drops]                    |
@@ -137,11 +137,16 @@ type SessionStateEvent = {
   type: 'session_state_changed';
   sessionId: string;
   state: 'idle' | 'running' | 'requires_action';
-  // Maps from the 7-state internal machine:
-  // CREATED/ENDED           -> 'idle'
-  // STARTING/ACTIVE         -> 'running'
-  // DETACHED/SUSPENDED      -> preserve last emitted (server buffers)
-  // permission_request      -> 'requires_action'
+  // Two emission paths:
+  //
+  // 1. From setSessionState() (lifecycle transitions):
+  //    CREATED/ENDED           -> 'idle'
+  //    STARTING/ACTIVE         -> 'running'
+  //    DETACHED/SUSPENDED      -> preserve last emitted (server buffers)
+  //
+  // 2. From permission_request v2 events (cross-cutting, not a SessionState):
+  //    permission_request      -> 'requires_action'
+  //    permission resolved     -> restore previous state ('running')
 };
 ```
 
@@ -171,9 +176,12 @@ Callers must remember to update all three. They diverge.
 
 **Changes:**
 
-- Remove `is_active` boolean column from EventStore (redundant with `state`)
+- Add `state` column alongside `is_active` in EventStore (backwards-compatible; Phase 0)
+- Write both `state` and `is_active` in `setSessionState()` during transition period
+- Migrate all `is_active` readers to use `state` (Phase 1)
+- Drop `is_active` column once all consumers migrated (Phase 4)
 - Replace `handleReconnect`'s `storeMeta.isActive` check with state-based logic
-- Add `recoverStaleSessions()` on startup: any session in ACTIVE/STARTING/DETACHED → ENDED
+- Add `recoverStaleSessions()` on startup: any session in ACTIVE/STARTING/DETACHED/SUSPENDED → ENDED
 - Expose `state` in `/api/sessions/:id/meta` response
 - Emit `session_state_changed` on every `setSessionState()` call
 
@@ -214,7 +222,53 @@ Combined with the state event, reconnect is trivial:
 
 **Current:** Various heartbeat mechanisms across transports, inconsistent.
 
-**Target:** Server sends SSE `ping` every 20 seconds. Client tracks `lastHeartbeatAt`. If no heartbeat for 40 seconds, client tears down SSE and reconnects (with cursor). This is proactive failure detection -- the client knows the connection is dead before the user tries to send.
+**Target:** Server sends SSE `ping` every `MITZO_HEARTBEAT_INTERVAL_MS` (default 20s, matching Anthropic SDK). Client tracks `lastHeartbeatAt`. If no heartbeat for 2x the interval, client tears down SSE and reconnects (with cursor). This is proactive failure detection -- the client knows the connection is dead before the user tries to send.
+
+Configurable via environment variable for development and testing (e.g. `MITZO_HEARTBEAT_INTERVAL_MS=5000` for faster detection during dev).
+
+### Multi-Client Fan-Out
+
+#### Design Decision
+
+Multiple clients can connect to the same session simultaneously (e.g. phone + laptop). This is designed in from the start, not bolted on later.
+
+#### Model
+
+- **Any client can send.** POST `/api/chat/send` accepts from any authenticated connection subscribed to the session. Server queues via AsyncQueue (already works). No primary/secondary client distinction.
+- **All clients receive all events.** `session_state_changed`, `message_start`, `block_delta`, etc. are fanned out to every SSE connection subscribed to that session.
+- **Concurrent sends are serialized.** If two clients POST simultaneously, AsyncQueue serializes them. The SDK processes one turn at a time. Both clients see both messages and both responses in sequence order.
+
+#### ConnectionRegistry Changes
+
+**Current:** `ConnectionRegistry` maps one connection per session (owner model).
+
+**Target:** `ConnectionRegistry` tracks a **set** of connections per session:
+
+```typescript
+// Current
+sessions: Map<sessionId, { connectionId: string; cursor: number }>;
+
+// Target
+sessions: Map<sessionId, Set<{ connectionId: string; cursor: number }>>;
+```
+
+Every `session_state_changed` and protocol event fans out to all connections in the set. Connection add/remove is independent -- one client disconnecting doesn't affect others.
+
+#### Event Fan-Out
+
+When `setSessionState()` or `appendEvent()` fires:
+
+1. Look up all connections subscribed to that session
+2. Write the SSE event to each connection's stream
+3. Each connection tracks its own cursor independently
+
+This is the same pattern used by every multi-device chat system. The EventStore already stores events durably with sequence numbers, so a reconnecting client catches up from its own cursor regardless of what other clients have received.
+
+#### What We Don't Need
+
+- **Presence awareness.** Clients don't need to know about each other. If we want "also viewing on phone" indicators later, that's a separate feature.
+- **Send coordination.** No locks, no "typing" exclusion. If two humans both send, they both get responses. This is unlikely in practice and harmless if it happens.
+- **Session ownership transfer.** With no primary/secondary model, there's nothing to transfer. Any client can send, any client can interrupt.
 
 ## Migration Plan
 
@@ -225,8 +279,11 @@ Combined with the state event, reconnect is trivial:
 3. Add `state` field to `/api/sessions/:id/meta` response
 4. Client receives and logs `session_state_changed` but doesn't act on it yet
 5. Add crash recovery: `recoverStaleSessions()` on server startup
+6. Flip SSE to default transport (currently WS default, SSE opt-in via localStorage)
+7. Add `state` column to EventStore sessions table alongside `is_active` (backwards-compatible)
+8. Migrate `setSessionState()` to write both `state` and `is_active`
 
-**Tests:** Verify state events are emitted on every transition. Verify crash recovery marks stale sessions as ENDED.
+**Tests:** Verify state events are emitted on every transition. Verify crash recovery marks stale sessions as ENDED. Verify SSE transport works as default. Verify `state` column is populated and consistent with `is_active`.
 
 **Dead code removal:** None yet. Additive only.
 
@@ -236,17 +293,21 @@ Combined with the state event, reconnect is trivial:
 2. `running` in messages reducer derived from `session_state_changed.state === 'running'`
 3. Remove `syncRunningState()` REST polling (state events make it unnecessary)
 4. Remove `running` field from `reconnected` and `session_switched` messages (replaced by replayed state events)
+5. Migrate all server-side `is_active` readers to use `state` instead
+6. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
 
-**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch).
+**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch). Verify reconnect works with state-based check.
 
 **Dead code removal:** `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
 
 ### Phase 2: Always-send (remove client routing)
 
+**Prerequisite:** Verify `clientMsgId` idempotency guard from PR #386 is still present and keyed correctly in `chat-rest-handler.ts`. If refactored or removed, re-implement before enabling retries.
+
 1. Remove `wasRunning` branch from `sendMessage()`
 2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
 3. Client awaits POST response before marking message as sent
-4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key -- server already has idempotency checking (#386), so retried POSTs are deduplicated server-side
+4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key so retried POSTs are deduplicated server-side
 5. Add `sendError` UI state for permanent failures
 6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
 
@@ -258,39 +319,40 @@ Combined with the state event, reconnect is trivial:
 
 1. SSE reconnect uses `?from=<lastSeq>` parameter
 2. Server replays events from cursor (already supported by EventStore)
-3. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
-4. Remove `handleReconnect` ownership dance
-5. Remove `doReconnectPost` and associated state management
-6. Remove periodic sync polling (cursor-based replay replaces it)
-7. Remove `is_active` boolean column from EventStore (safe now that all consumers use `state`)
-8. Remove `markSessionInactive()` (replaced by `setSessionState(ENDED)`)
+3. Remove `handleReconnect` ownership dance
+4. Remove `doReconnectPost` and associated state management
+5. Remove periodic sync polling (cursor-based replay replaces it)
+6. `ConnectionRegistry` fan-out: track set of connections per session, not single owner
 
-**Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect.
+**Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect. Verify multi-client fan-out (two SSE connections receive same events).
 
-**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (packages/client/src/sse-connection.ts), periodic sync timer and `shouldSync` logic (packages/harness/src/connection-registry.ts), `reconnect` message type, `markSessionInactive()`, `is_active` column.
+**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (packages/client/src/sse-connection.ts), periodic sync timer and `startPeriodicSync()`/`stopPeriodicSync()` logic (packages/harness/src/connection-registry.ts), `reconnect` message type.
 
 ### Phase 4: Cleanup
 
-1. Remove WS transport code (if SSE migration is complete by this point)
+1. Remove WS transport code (SSE is default since Phase 0)
 2. Remove `SessionSseRegistry` if consolidated into `ConnectionRegistry`
 3. Remove stale connectionId/ownership management code from PR #401 revert residue
-4. Update all design docs to reflect new architecture
-5. Archive `session-state-machine.md` Phase 4 (crash recovery is now Phase 0)
+4. Remove `is_active` column from EventStore (all consumers migrated to `state` in Phase 1)
+5. Remove `markSessionInactive()` (replaced by `setSessionState(ENDED)`)
+6. Update all design docs to reflect new architecture
+7. Archive `session-state-machine.md` Phase 4 (crash recovery is now Phase 0)
 
 ## Files Involved
 
 ### Server
 
-| File                                          | Changes                                                                    |
-| --------------------------------------------- | -------------------------------------------------------------------------- |
-| `server/query-loop.ts`                        | Emit `session_state_changed` on state transitions                          |
-| `server/chat.ts`                              | Emit state events from `startChat`/`sendToChat`; crash recovery on startup |
-| `server/ws-handler-v2.ts`                     | Remove `handleReconnect`; replace `is_active` checks with state-based      |
-| `server/index.ts`                             | Call `recoverStaleSessions()` on startup                                   |
-| `server/app.ts`                               | Add `state` to meta endpoint response                                      |
-| `packages/protocol/src/event-store.ts`        | Remove `is_active` column; emit events from `setSessionState()`            |
-| `packages/protocol/src/types.ts`              | Add `SessionStateEvent` type                                               |
-| `packages/harness/src/connection-registry.ts` | Remove periodic sync; simplify to cursor-based replay                      |
+| File                                          | Changes                                                                                                    |
+| --------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `server/query-loop.ts`                        | Emit `session_state_changed` on state transitions                                                          |
+| `server/chat.ts`                              | Emit state events from `startChat`/`sendToChat`; crash recovery on startup                                 |
+| `server/ws-handler-v2.ts`                     | Replace `is_active` checks with state-based (P1); remove `handleReconnect` (P3); remove file (P4)          |
+| `server/index.ts`                             | Call `recoverStaleSessions()` on startup                                                                   |
+| `server/app.ts`                               | Add `state` to meta endpoint response                                                                      |
+| `packages/protocol/src/event-store.ts`        | Add `state` column alongside `is_active` (P0); emit events from `setSessionState()`; drop `is_active` (P4) |
+| `packages/protocol/src/types.ts`              | Add `SessionStateEvent` type                                                                               |
+| `packages/harness/src/connection-registry.ts` | Multi-client fan-out: Set of connections per session (P3); remove periodic sync                            |
+| `frontend/src/client-store.ts`                | Flip SSE to default transport (P0)                                                                         |
 
 ### Client
 
@@ -299,39 +361,44 @@ Combined with the state event, reconnect is trivial:
 | `packages/client/src/store.ts`           | Remove `wasRunning` queuing; await POST ack; remove `syncRunningState`           |
 | `packages/client/src/protocol-parser.ts` | Handle `session_state_changed`; remove `pendingSend` drain; remove `SET_RUNNING` |
 | `packages/client/src/sse-connection.ts`  | Reconnect with `?from=lastSeq`; remove `doReconnectPost`; add heartbeat timeout  |
-| `packages/client/src/connection.ts`      | Same cursor-based reconnect if WS still active                                   |
+| `packages/client/src/connection.ts`      | Remove in Phase 4 (WS transport removal)                                         |
 | `frontend/src/components/ChatInput.tsx`  | No changes (useQueuedMessages is a separate user-facing feature)                 |
 
-### Dead Code Candidates (cumulative removal)
+### Dead Code Candidates (by phase)
 
 ```
-// Client
+// Phase 1 — Server-authoritative state
+- SET_RUNNING action type (replaced by session_state_changed)
+- syncRunningState() REST polling
+- running field in reconnected/session_switched messages
+
+// Phase 2 — Always-send
 - PENDING_SEND_TIMEOUT_MS constant
 - pendingSend[] array and all drain logic
 - wasRunning queuing branch in sendMessage()
 - clearPendingSendTimer()
 - onSendQueued callback
-- SET_RUNNING action type (replaced by session_state_changed)
-- syncRunningState() REST polling
+
+// Phase 3 — Clean reconnect
+- handleReconnect() ownership dance
 - doReconnectPost() and associated state
-- periodic sync timer
+- periodic sync timer / startPeriodicSync() / stopPeriodicSync() logic
+- reconnect message type
 
-// Server (Phase 1)
-- running field in reconnected/session_switched messages
-
-// Server (Phase 3)
+// Phase 4 — Cleanup
+- WS transport code (MitzoConnection, ws-handler-v2.ts)
 - markSessionInactive() (replaced by setSessionState(ENDED))
 - is_active column in EventStore sessions table
-- handleReconnect() ownership dance
-- shouldSync() / periodic sync replay logic
+- SessionSseRegistry (if consolidated)
+- connectionId ownership management (PR #401 residue)
 ```
 
-## Open Questions
+## Resolved Decisions
 
-1. **WS removal timeline:** Phase 3 assumes SSE is the primary transport. If WS is still active, cursor-based reconnect needs to work for both. Should we complete the SSE migration first?
+1. **WS removal timeline:** Complete SSE migration. Flip SSE to default transport in Phase 0. Remove WS transport code in Phase 4 cleanup. Maintaining cursor-based reconnect for both transports doubles the surface area for exactly the kind of bug we're eliminating.
 
-2. **Multi-device fan-out:** The current architecture assumes one client per session. If we want multi-device (phone + laptop viewing same session), the job-based model enables it naturally. Worth designing for now or deferring?
+2. **Multi-device fan-out:** Design for multiple clients per session from the start. See [Multi-Client Fan-Out](#multi-client-fan-out) section. Any connected client can send, server queues via AsyncQueue, all clients see the result. No primary/secondary distinction.
 
-3. **EventStore migration:** Removing `is_active` requires a SQLite migration. Should we add a backwards-compatible `state`-only path first, or migrate in one step?
+3. **EventStore migration:** Backwards-compatible. Add `state` column alongside `is_active` in Phase 0. Migrate all readers to use `state` through Phases 1-3. Drop `is_active` in Phase 4 cleanup.
 
-4. **Heartbeat interval:** 20s matches Anthropic's SDK. Should we make it configurable for development (faster detection during testing)?
+4. **Heartbeat interval:** Default 20s (matches Anthropic SDK). Configurable via `MITZO_HEARTBEAT_INTERVAL_MS` environment variable for development and testing.
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -313,7 +313,9 @@ This is the same pattern used by every multi-device chat system. The EventStore 
 5. Add `sendError` UI state for permanent failures
 6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
 
-**Tests:** Verify send-while-running works (server queues via AsyncQueue). Verify POST failure shows error UI. Verify heartbeat timeout triggers reconnect. Verify no message loss on reconnect during generation.
+**UX note:** The `USER_SEND` reducer action (which renders the user bubble immediately) stays as-is -- the UI remains optimistic. The POST acknowledgment controls whether the message is considered _delivered_, not _displayed_. On POST failure, the existing bubble gets a "failed to send" indicator with retry. This preserves the instant-feel UX while adding delivery confidence.
+
+**Tests:** Verify send-while-running works (server queues via AsyncQueue). Verify POST failure shows error UI with retry. Verify heartbeat timeout triggers reconnect. Verify no message loss on reconnect during generation.
 
 **Dead code removal:** `pendingSend`, `onSendQueued`, `clearPendingSendTimer`, `PENDING_SEND_TIMEOUT_MS`, `drainOne`, safety-net timer in `sendMessage`, `pendingSend` drain in `session_end` handler, `pendingSend = []` in switch/new/error handlers.
 
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -3,7 +3,7 @@
 **Status:** Proposed
 **Author:** Claude Opus 4.6 + Dimitri Saridakis
 **Created:** 2026-07-02
-**Telos:** TBD
+**Telos:** 97361f814c5f54df
 **Supersedes:** Parts of `session-state-machine.md`, `streaming-input-session-control.md`
 **Related PRs:** #396 (never merged), #401 (merged, reverted via #403), #407
 
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -3,7 +3,7 @@
 **Status:** Approved
 **Author:** Claude Opus 4.6 + Dimitri Saridakis
 **Created:** 2026-07-02
-**Telos:** 97361f814c5f54df
+**Telos:** b138b20e472bd324
 **Incorporates and replaces:** `session-state-machine.md` (never implemented) and supersedes `streaming-input-session-control.md` (implemented in PR #33, but scope expanded here). Crash recovery from `session-state-machine.md` Phase 4 is pulled forward to Phase 0 here.
 **Related PRs:** #396 (never merged), #401 (merged, reverted via #403), #407
 
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -4,7 +4,7 @@
 **Author:** Claude Opus 4.6 + Dimitri Saridakis
 **Created:** 2026-07-02
 **Telos:** 97361f814c5f54df
-**Supersedes:** Parts of `session-state-machine.md`, `streaming-input-session-control.md`
+**Incorporates and replaces:** `session-state-machine.md` (Status: Proposed, never implemented), `streaming-input-session-control.md`. Crash recovery from `session-state-machine.md` Phase 4 is pulled forward to Phase 0 here.
 **Related PRs:** #396 (never merged), #401 (merged, reverted via #403), #407
 
 ## Problem Statement
@@ -236,19 +236,17 @@ Combined with the state event, reconnect is trivial:
 2. `running` in messages reducer derived from `session_state_changed.state === 'running'`
 3. Remove `syncRunningState()` REST polling (state events make it unnecessary)
 4. Remove `running` field from `reconnected` and `session_switched` messages (replaced by replayed state events)
-5. Remove `is_active` boolean column from EventStore
-6. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
 
-**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch). Verify reconnect works with state-based check.
+**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch).
 
-**Dead code removal:** `markSessionInactive()`, `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
+**Dead code removal:** `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
 
 ### Phase 2: Always-send (remove client routing)
 
 1. Remove `wasRunning` branch from `sendMessage()`
 2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
 3. Client awaits POST response before marking message as sent
-4. Add retry with exponential backoff on POST failure (max 3 attempts)
+4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key -- server already has idempotency checking (#386), so retried POSTs are deduplicated server-side
 5. Add `sendError` UI state for permanent failures
 6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
 
@@ -260,13 +258,16 @@ Combined with the state event, reconnect is trivial:
 
 1. SSE reconnect uses `?from=<lastSeq>` parameter
 2. Server replays events from cursor (already supported by EventStore)
-3. Remove `handleReconnect` ownership dance
-4. Remove `doReconnectPost` and associated state management
-5. Remove periodic sync polling (cursor-based replay replaces it)
+3. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
+4. Remove `handleReconnect` ownership dance
+5. Remove `doReconnectPost` and associated state management
+6. Remove periodic sync polling (cursor-based replay replaces it)
+7. Remove `is_active` boolean column from EventStore (safe now that all consumers use `state`)
+8. Remove `markSessionInactive()` (replaced by `setSessionState(ENDED)`)
 
 **Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect.
 
-**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (sse-connection.ts), periodic sync timer and `shouldSync` logic (connection-registry.ts), `reconnect` message type.
+**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (packages/client/src/sse-connection.ts), periodic sync timer and `shouldSync` logic (packages/harness/src/connection-registry.ts), `reconnect` message type, `markSessionInactive()`, `is_active` column.
 
 ### Phase 4: Cleanup
 
@@ -280,16 +281,16 @@ Combined with the state event, reconnect is trivial:
 
 ### Server
 
-| File                                   | Changes                                                                    |
-| -------------------------------------- | -------------------------------------------------------------------------- |
-| `server/query-loop.ts`                 | Emit `session_state_changed` on state transitions                          |
-| `server/chat.ts`                       | Emit state events from `startChat`/`sendToChat`; crash recovery on startup |
-| `server/ws-handler-v2.ts`              | Remove `handleReconnect`; replace `is_active` checks with state-based      |
-| `server/index.ts`                      | Call `recoverStaleSessions()` on startup                                   |
-| `server/app.ts`                        | Add `state` to meta endpoint response                                      |
-| `packages/protocol/src/event-store.ts` | Remove `is_active` column; emit events from `setSessionState()`            |
-| `packages/protocol/src/types.ts`       | Add `SessionStateEvent` type                                               |
-| `server/connection-registry.ts`        | Remove periodic sync; simplify to cursor-based replay                      |
+| File                                          | Changes                                                                    |
+| --------------------------------------------- | -------------------------------------------------------------------------- |
+| `server/query-loop.ts`                        | Emit `session_state_changed` on state transitions                          |
+| `server/chat.ts`                              | Emit state events from `startChat`/`sendToChat`; crash recovery on startup |
+| `server/ws-handler-v2.ts`                     | Remove `handleReconnect`; replace `is_active` checks with state-based      |
+| `server/index.ts`                             | Call `recoverStaleSessions()` on startup                                   |
+| `server/app.ts`                               | Add `state` to meta endpoint response                                      |
+| `packages/protocol/src/event-store.ts`        | Remove `is_active` column; emit events from `setSessionState()`            |
+| `packages/protocol/src/types.ts`              | Add `SessionStateEvent` type                                               |
+| `packages/harness/src/connection-registry.ts` | Remove periodic sync; simplify to cursor-based replay                      |
 
 ### Client
 
@@ -315,11 +316,13 @@ Combined with the state event, reconnect is trivial:
 - doReconnectPost() and associated state
 - periodic sync timer
 
-// Server
+// Server (Phase 1)
+- running field in reconnected/session_switched messages
+
+// Server (Phase 3)
 - markSessionInactive() (replaced by setSessionState(ENDED))
 - is_active column in EventStore sessions table
 - handleReconnect() ownership dance
-- running field in reconnected/session_switched messages
 - shouldSync() / periodic sync replay logic
 ```
 
--- a/docs/design/transport-ssot.md
+++ b/docs/design/transport-ssot.md
@@ -1,0 +1,334 @@
+# Transport Single Source of Truth
+
+**Status:** Proposed
+**Author:** Claude Opus 4.6 + Dimitri Saridakis
+**Created:** 2026-07-02
+**Telos:** TBD
+**Supersedes:** Parts of `session-state-machine.md`, `streaming-input-session-control.md`
+**Related PRs:** #396 (never merged), #401 (merged, reverted via #403), #407
+
+## Problem Statement
+
+### The Symptom
+
+User sends a prompt. Agent doesn't respond. User sends a second prompt. Agent wakes up and responds to both. This has been the single most persistent bug in Mitzo's history.
+
+### The History
+
+**45+ commits** across 3 transport architectures (raw WS, v2 multiplexed WS, SSE) have attempted to fix variants of this bug. 3 fixes were reverted after causing regressions. Every transport migration re-introduced the problem. The most recent attempt (PR #401, stable client connectionId) was merged and reverted within 16 hours because the connectionId was scoped per-tab instead of per-session, causing cross-session event routing.
+
+### The Root Causes (Taxonomy)
+
+| #   | Root Cause                                           | Fix Attempts                       | Current Status                 |
+| --- | ---------------------------------------------------- | ---------------------------------- | ------------------------------ |
+| 1   | Dead transport, messages silently dropped            | #53, #288 (proactive suspend)      | Mostly fixed                   |
+| 2   | Dual source of truth (SessionRegistry vs EventStore) | #257, #362 (state machine routing) | Fixed for send/interrupt paths |
+| 3   | iOS kills WS without readyState update               | heartbeat, #368 (force reconnect)  | Fixed for iOS                  |
+| 4   | Ownership race on reconnect (connectionId changes)   | #401 (stable cid) -- **reverted**  | **Open**                       |
+| 5   | Running state not synced on foreground               | #353, #407 (query meta)            | Fixed                          |
+| 6   | Stale `running` in `newSession()` (never reset)      | Never attempted                    | **Open**                       |
+
+### The Architectural Deficiency
+
+There is no single, authoritative model of "is this session alive and accepting input?" Instead there are four overlapping, sometimes-stale signals:
+
+| Location                        | Signal              | Updated By                         |
+| ------------------------------- | ------------------- | ---------------------------------- |
+| `EventStore.state` (SQLite)     | `SessionState` enum | `setSessionState()` callers        |
+| `EventStore.is_active` (SQLite) | boolean             | `markSessionInactive()`            |
+| `SessionRegistry` (in-memory)   | Map presence        | `register()`/`remove()`            |
+| Client `messages.running`       | boolean             | Protocol events (sometimes missed) |
+
+Each fix patches one signal without reconciling the others. Each transport migration creates new divergence paths between them. This is why the bug keeps coming back.
+
+### Why This Is a Solved Problem
+
+This is a chat application transport layer. Thousands of production systems have solved it. The industry has converged on a clear architecture:
+
+| Platform               | Transport               | Generation Model                                 | State Authority                                             |
+| ---------------------- | ----------------------- | ------------------------------------------------ | ----------------------------------------------------------- |
+| Anthropic (Claude.ai)  | SSE + HTTP POST         | Server-side with worker epochs                   | Explicit state machine (`idle`/`running`/`requires_action`) |
+| OpenAI (Responses API) | SSE or WebSocket        | Server-side, chained via `previous_response_id`  | Connection-scoped, explicit events                          |
+| LibreChat              | SSE + HTTP POST         | `GenerationJobManager` with persisted job status | Server job store (Redis/memory)                             |
+| LobeChat               | SSE via tRPC            | Operation-based tracking                         | Server pushes canonical message snapshots                   |
+| Vercel AI SDK          | Pluggable (SSE default) | Server-side stream                               | `activeStreamId` with resume                                |
+
+**Common pattern across all:**
+
+1. SSE for reads, HTTP POST for writes (with acknowledgment)
+2. Generation is a server-side job that survives connection drops
+3. Client "running" state is derived from server events, never used for routing
+4. Sequence-based resumption on reconnect
+5. Heartbeat-based liveness detection
+
+## Solution: Server-Authoritative Transport
+
+### Design Principles
+
+1. **Server is the single source of truth.** The server owns session state. The client is a cache with explicit invalidation. Period.
+2. **Client never queues based on local state.** `sendMessage()` always sends. The server decides what to do.
+3. **Generation survives connection.** A generation is a server-side job. If the client disconnects, the generation continues. When the client reconnects, it re-subscribes.
+4. **Explicit state transitions, not derived state.** The server emits authoritative state events. The client renders them. No guessing from event absence.
+5. **Write acknowledgment, not fire-and-forget.** HTTP POST returns success or error. The message is not "sent" until acknowledged.
+
+### Architecture
+
+```
+CLIENT                          SERVER
+------                          ------
+
+  |--- POST /send ----------------->|
+  |<-- 202 { accepted, jobId } -----|    (write acknowledged)
+  |                                 |
+  |<-- SSE: state=running ----------|    (authoritative state)
+  |<-- SSE: message_start ----------|
+  |<-- SSE: block_delta ------------|
+  |<-- SSE: block_delta ------------|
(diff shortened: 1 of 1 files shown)


docs/design/transport-ssot.md
```diff
@@ -0,0 +1,334 @@
...
+| -------------------------------------- | -------------------------------------------------------------------------- |
+| `server/query-loop.ts`                 | Emit `session_state_changed` on state transitions                          |
+| `server/chat.ts`                       | Emit state events from `startChat`/`sendToChat`; crash recovery on startup |
+| `server/ws-handler-v2.ts`              | Remove `handleReconnect`; replace `is_active` checks with state-based      |
+| `server/index.ts`                      | Call `recoverStaleSessions()` on startup                                   |
+| `server/app.ts`                        | Add `state` to meta endpoint response                                      |
+| `packages/protocol/src/event-store.ts` | Remove `is_active` column; emit events from `setSessionState()`            |
+| `packages/protocol/src/types.ts`       | Add `SessionStateEvent` type                                               |
+| `server/connection-registry.ts`        | Remove periodic sync; simplify to cursor-based replay                      |
```
dimakis: 🟡 **bugs**: Files Involved table lists `server/connection-registry.ts`, but the actual file is `packages/harness/src/connection-registry.ts`. The harness package owns this module. Using the wrong path will mislead implementers. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,334 @@
+# Transport Single Source of Truth
+
+**Status:** Proposed
+**Author:** Claude Opus 4.6 + Dimitri Saridakis
+**Created:** 2026-07-02
+**Telos:** 97361f814c5f54df
+**Supersedes:** Parts of `session-state-machine.md`, `streaming-input-session-control.md`
```
dimakis: 🔵 **style**: "Supersedes: Parts of `session-state-machine.md`" — that doc is still Status: Proposed (never implemented). Consider saying "Incorporates and replaces" rather than "Supersedes" since there's nothing live to supersede. The crash recovery concept (`recoverStaleSessions`) originates from that doc's Phase 4 and is pulled forward here to Phase 0, which is worth calling out explicitly. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,334 @@
...
+**Dead code removal:** `markSessionInactive()`, `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
+
+### Phase 2: Always-send (remove client routing)
+
+1. Remove `wasRunning` branch from `sendMessage()`
+2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
+3. Client awaits POST response before marking message as sent
+4. Add retry with exponential backoff on POST failure (max 3 attempts)
+5. Add `sendError` UI state for permanent failures
+6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
```
dimakis, line 314: 🟡 **unsafe_assumptions**: Phase 1 step 5 says "Remove `is_active` boolean column from EventStore" — but `is_active` is used in 4+ SQL queries in event-store.ts (session listing, meta endpoint, reconnect checks) and exposed via the `/api/sessions/:id/meta` response (`isActive` field). Removing it in Phase 1 while `handleReconnect` still exists (removed in Phase 3) creates a gap. Either the migration plan needs a compatibility shim in Phase 1, or `is_active` removal should move to Phase 3 alongside `handleReconnect` removal. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,334 @@
...
+3. Remove `handleReconnect` ownership dance
+4. Remove `doReconnectPost` and associated state management
+5. Remove periodic sync polling (cursor-based replay replaces it)
+
+**Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect.
+
+**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (sse-connection.ts), periodic sync timer and `shouldSync` logic (connection-registry.ts), `reconnect` message type.
+
+### Phase 4: Cleanup
+
```
dimakis, line 336: 🔵 **unsafe_assumptions**: Phase 2 step 4 proposes "retry with exponential backoff on POST failure (max 3 attempts)" but doesn't address idempotency. If the POST succeeds server-side but the response is lost (network error), the client would retry and send the message again. The `seq` number in the 202 response could serve as an idempotency key, but the design doesn't specify this. Worth noting in the design or in Open Questions. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,334 @@
...
+## Files Involved
+
+### Server
+
+| File                                   | Changes                                                                    |
+| -------------------------------------- | -------------------------------------------------------------------------- |
+| `server/query-loop.ts`                 | Emit `session_state_changed` on state transitions                          |
+| `server/chat.ts`                       | Emit state events from `startChat`/`sendToChat`; crash recovery on startup |
+| `server/ws-handler-v2.ts`              | Remove `handleReconnect`; replace `is_active` checks with state-based      |
+| `server/index.ts`                      | Call `recoverStaleSessions()` on startup                                   |
```
dimakis: 🔵 **style**: The Files Involved tables list server-side files as `server/` paths but `connection-registry.ts` and `session-registry.ts` actually live in `packages/harness/src/`. The `event-store.ts` path is correctly listed as `packages/protocol/src/event-store.ts`. For consistency, harness files should use their actual package paths (`packages/harness/src/connection-registry.ts`). `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-researched design doc with accurate diagnosis of the transport bug taxonomy. Two file path inaccuracies in the Files Involved table (`connection-registry.ts` is in `packages/harness/`, not `server/`), and the `is_active` column removal in Phase 1 conflicts with its continued use in `handleReconnect` which isn't removed until Phase 3._

- 🟡 **bugs** (L292): Files Involved table lists `server/connection-registry.ts`, but the actual file is `packages/harness/src/connection-registry.ts`. The harness package owns this module. Using the wrong path will mislead implementers. `[fixable]`
- 🔵 **style** (L7): "Supersedes: Parts of `session-state-machine.md`" — that doc is still Status: Proposed (never implemented). Consider saying "Incorporates and replaces" rather than "Supersedes" since there's nothing live to supersede. The crash recovery concept (`recoverStaleSessions`) originates from that doc's Phase 4 and is pulled forward here to Phase 0, which is worth calling out explicitly. `[fixable]`
- 🟡 **unsafe_assumptions** (L253): Phase 1 step 5 says "Remove `is_active` boolean column from EventStore" — but `is_active` is used in 4+ SQL queries in event-store.ts (session listing, meta endpoint, reconnect checks) and exposed via the `/api/sessions/:id/meta` response (`isActive` field). Removing it in Phase 1 while `handleReconnect` still exists (removed in Phase 3) creates a gap. Either the migration plan needs a compatibility shim in Phase 1, or `is_active` removal should move to Phase 3 alongside `handleReconnect` removal. `[fixable]`
- 🔵 **unsafe_assumptions** (L272): Phase 2 step 4 proposes "retry with exponential backoff on POST failure (max 3 attempts)" but doesn't address idempotency. If the POST succeeds server-side but the response is lost (network error), the client would retry and send the message again. The `seq` number in the 202 response could serve as an idempotency key, but the design doesn't specify this. Worth noting in the design or in Open Questions. `[fixable]`
- 🔵 **style** (L288): The Files Involved tables list server-side files as `server/` paths but `connection-registry.ts` and `session-registry.ts` actually live in `packages/harness/src/`. The `event-store.ts` path is correctly listed as `packages/protocol/src/event-store.ts`. For consistency, harness files should use their actual package paths (`packages/harness/src/connection-registry.ts`). `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,337 @@
...
+3. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
+4. Remove `handleReconnect` ownership dance
+5. Remove `doReconnectPost` and associated state management
+6. Remove periodic sync polling (cursor-based replay replaces it)
+7. Remove `is_active` boolean column from EventStore (safe now that all consumers use `state`)
+8. Remove `markSessionInactive()` (replaced by `setSessionState(ENDED)`)
+
+**Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect.
+
+**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (packages/client/src/sse-connection.ts), periodic sync timer and `shouldSync` logic (packages/harness/src/connection-registry.ts), `reconnect` message type, `markSessionInactive()`, `is_active` column.
```
dimakis: 🟡 **bugs**: `shouldSync()` is referenced as dead code to remove from `connection-registry.ts`, but no function by that name exists. The actual methods are `startPeriodicSync()` and `stopPeriodicSync()`. Same issue on line 326. Update to reference the real function names to avoid confusion during implementation. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,337 @@
...
+- `reconnected` message (carries `running: boolean`)
+- `syncRunningState()` REST polling on foreground
+
+**Target:** Server emits `session_state_changed` events with the authoritative state machine value:
+
+```typescript
+type SessionStateEvent = {
+  type: 'session_state_changed';
+  sessionId: string;
+  state: 'idle' | 'running' | 'requires_action';
```
dimakis, line 139: 🟡 **unsafe_assumptions**: The `SessionStateEvent` maps `permission_request -> 'requires_action'`, but `permission_request` is a v2 protocol event type, not a `SessionState` enum value. The 7-state `SessionState` has no `requires_action` equivalent — it's a cross-cutting concern (permissions can arrive in ACTIVE state). The doc should clarify that `requires_action` is derived from permission events, not from `setSessionState()`, which contradicts the "Emit `session_state_changed` on every `setSessionState()` call" instruction in Section 3. This needs a second emission path (from `permission_request` events) that isn't mentioned in the implementation plan. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,337 @@
...
+5. **Write acknowledgment, not fire-and-forget.** HTTP POST returns success or error. The message is not "sent" until acknowledged.
+
+### Architecture
+
+```
+CLIENT                          SERVER
+------                          ------
+
+  |--- POST /send ----------------->|
+  |<-- 202 { accepted, jobId } -----|    (write acknowledged)
```
dimakis, line 81: 🔵 **style**: The architecture diagram shows `POST /send` but Section 4 (Write Acknowledgment) references `POST /api/chat/send`. The existing codebase uses the `/api/chat/send` path (via `chat-rest-handler.ts`). The diagram should use the full path for consistency and to avoid ambiguity during implementation. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,337 @@
...
+**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch).
+
+**Dead code removal:** `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
+
+### Phase 2: Always-send (remove client routing)
+
+1. Remove `wasRunning` branch from `sendMessage()`
+2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
+3. Client awaits POST response before marking message as sent
+4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key -- server already has idempotency checking (#386), so retried POSTs are deduplicated server-side
```
dimakis: 🔵 **unsafe_assumptions**: Phase 2 references `clientMsgId` as an idempotency key and claims "server already has idempotency checking (#386)". This should be verified during implementation — if the idempotency check is keyed differently or has been removed/refactored since #386, retried POSTs could create duplicate messages. Consider adding a Phase 2 prerequisite to verify the idempotency mechanism exists and is keyed on `clientMsgId`.

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,337 @@
...
+| --------------- | -------------------------------------- | ---------------------------------- |
+| Lifecycle phase | `EventStore.state` (the state machine) | Routing, reconnect, crash recovery |
+| Process alive?  | `SessionRegistry.has()`                | Query loop management only         |
+| Client display  | `session_state_changed` events         | UI rendering only                  |
+
+**Changes:**
+
+- Remove `is_active` boolean column from EventStore (redundant with `state`)
+- Replace `handleReconnect`'s `storeMeta.isActive` check with state-based logic
+- Add `recoverStaleSessions()` on startup: any session in ACTIVE/STARTING/DETACHED → ENDED
```
dimakis: 🔵 **style**: Phase 0 proposes `recoverStaleSessions()` to transition ACTIVE/STARTING/DETACHED sessions to ENDED on startup, but doesn't mention SUSPENDED. The `SessionState` type includes SUSPENDED as a state where the session is not cleanly ended. If the server crashes while a session is SUSPENDED, it should also be recovered. Consider adding SUSPENDED to the recovery set, or document why it's excluded. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-researched design doc with accurate codebase references. All major claims verified against real code. A few naming inaccuracies (`shouldSync` doesn't exist) and the `requires_action` state needs a clearer emission path since it's orthogonal to the `SessionState` machine._

- 🟡 **bugs** (L270): `shouldSync()` is referenced as dead code to remove from `connection-registry.ts`, but no function by that name exists. The actual methods are `startPeriodicSync()` and `stopPeriodicSync()`. Same issue on line 326. Update to reference the real function names to avoid confusion during implementation. `[fixable]`
- 🟡 **unsafe_assumptions** (L139): The `SessionStateEvent` maps `permission_request -> 'requires_action'`, but `permission_request` is a v2 protocol event type, not a `SessionState` enum value. The 7-state `SessionState` has no `requires_action` equivalent — it's a cross-cutting concern (permissions can arrive in ACTIVE state). The doc should clarify that `requires_action` is derived from permission events, not from `setSessionState()`, which contradicts the "Emit `session_state_changed` on every `setSessionState()` call" instruction in Section 3. This needs a second emission path (from `permission_request` events) that isn't mentioned in the implementation plan. `[fixable]`
- 🔵 **style** (L81): The architecture diagram shows `POST /send` but Section 4 (Write Acknowledgment) references `POST /api/chat/send`. The existing codebase uses the `/api/chat/send` path (via `chat-rest-handler.ts`). The diagram should use the full path for consistency and to avoid ambiguity during implementation. `[fixable]`
- 🔵 **unsafe_assumptions** (L249): Phase 2 references `clientMsgId` as an idempotency key and claims "server already has idempotency checking (#386)". This should be verified during implementation — if the idempotency check is keyed differently or has been removed/refactored since #386, retried POSTs could create duplicate messages. Consider adding a Phase 2 prerequisite to verify the idempotency mechanism exists and is keyed on `clientMsgId`.
- 🔵 **style** (L176): Phase 0 proposes `recoverStaleSessions()` to transition ACTIVE/STARTING/DETACHED sessions to ENDED on startup, but doesn't mention SUSPENDED. The `SessionState` type includes SUSPENDED as a state where the session is not cleanly ended. If the server crashes while a session is SUSPENDED, it should also be recovered. Consider adding SUSPENDED to the recovery set, or document why it's excluded. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,404 @@
...
+5. Migrate all server-side `is_active` readers to use `state` instead
+6. Replace `handleReconnect`'s `storeMeta.isActive` check with `storeState`-based logic
+
+**Tests:** Verify `running` state transitions match server state in all scenarios (new session, reconnect, iOS foreground, session switch). Verify reconnect works with state-based check.
+
+**Dead code removal:** `syncRunningState()`, `SET_RUNNING` action type, `running` field in reconnect/switch messages.
+
+### Phase 2: Always-send (remove client routing)
+
+**Prerequisite:** Verify `clientMsgId` idempotency guard from PR #386 is still present and keyed correctly in `chat-rest-handler.ts`. If refactored or removed, re-implement before enabling retries.
```
dimakis: 🟡 **unsafe_assumptions**: Phase 2 prerequisite says to verify `clientMsgId` idempotency guard is "present and keyed correctly in `chat-rest-handler.ts`". The guard actually lives in `server/chat.ts` (`storeAndEchoIfNew()` at line ~1162, which calls `eventStore.hasUserMessage()`). `chat-rest-handler.ts` passes `clientMsgId` through to `handleSendV2()` which calls `sendToChat()` in `chat.ts`, but the dedup logic itself is not in `chat-rest-handler.ts`. The prerequisite should reference `chat.ts` (or both files) so an implementer doesn't look in the wrong place and conclude the guard is missing. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,404 @@
+# Transport Single Source of Truth
+
+**Status:** Approved
+**Author:** Claude Opus 4.6 + Dimitri Saridakis
+**Created:** 2026-07-02
+**Telos:** 97361f814c5f54df
+**Incorporates and replaces:** `session-state-machine.md` (Status: Proposed, never implemented), `streaming-input-session-control.md`. Crash recovery from `session-state-machine.md` Phase 4 is pulled forward to Phase 0 here.
+**Related PRs:** #396 (never merged), #401 (merged, reverted via #403), #407
```
dimakis, line 8: 🔵 **style**: `streaming-input-session-control.md` status is listed implicitly as never implemented (via the "Incorporates and replaces" line grouping it with `session-state-machine.md`). However, `streaming-input-session-control.md` has `Status: Implemented (v0.1.0)` and was shipped with protocol v2 (PR #33). The relationship is different — this doc replaces a *shipped* design, not a proposed one. Consider clarifying the distinction (e.g., "replaces session-state-machine.md (never implemented) and supersedes streaming-input-session-control.md (implemented, but scope expanded here)"). `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,404 @@
...
+
+## Files Involved
+
+### Server
+
+| File                                          | Changes                                                                                                    |
+| --------------------------------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `server/query-loop.ts`                        | Emit `session_state_changed` on state transitions                                                          |
+| `server/chat.ts`                              | Emit state events from `startChat`/`sendToChat`; crash recovery on startup                                 |
+| `server/ws-handler-v2.ts`                     | Replace `is_active` checks with state-based (P1); remove `handleReconnect` (P3); remove file (P4)          |
```
dimakis, line 353: 🔵 **unsafe_assumptions**: The Files Involved table lists `frontend/src/client-store.ts` for the SSE default transport flip. This file does exist and contains the transport selector logic (`localStorage.getItem('mitzo:transport') === 'sse'`). However, if the SSE flip also involves changes to `packages/client/src/store.ts` (where the `MitzoConnection` is instantiated), that file should be listed in the table as well. Currently the Server table lists the store file for removing `wasRunning` but the Phase 0 SSE flip may span both locations. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,404 @@
...
+| --------------- | -------------------------------------- | ---------------------------------- |
+| Lifecycle phase | `EventStore.state` (the state machine) | Routing, reconnect, crash recovery |
+| Process alive?  | `SessionRegistry.has()`                | Query loop management only         |
+| Client display  | `session_state_changed` events         | UI rendering only                  |
+
+**Changes:**
+
+- Add `state` column alongside `is_active` in EventStore (backwards-compatible; Phase 0)
+- Write both `state` and `is_active` in `setSessionState()` during transition period
+- Migrate all `is_active` readers to use `state` (Phase 1)
```
dimakis, line 183: 🔵 **style**: The `SessionStateEvent` type comment says `setSessionState()` maps `DETACHED/SUSPENDED -> 'preserve last emitted (server buffers)'`. This is an important behavioral detail but it's buried in a code comment. The buffering mechanism (how the server remembers and replays the last emitted client-facing state after a transient internal state) should be called out explicitly in the design text, as implementers need to build this state tracking. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-grounded design doc — all referenced code artifacts verified as existing. One inaccurate file reference for the clientMsgId idempotency guard (it's in chat.ts, not chat-rest-handler.ts) and minor clarification opportunities._

- 🟡 **unsafe_assumptions** (L305): Phase 2 prerequisite says to verify `clientMsgId` idempotency guard is "present and keyed correctly in `chat-rest-handler.ts`". The guard actually lives in `server/chat.ts` (`storeAndEchoIfNew()` at line ~1162, which calls `eventStore.hasUserMessage()`). `chat-rest-handler.ts` passes `clientMsgId` through to `handleSendV2()` which calls `sendToChat()` in `chat.ts`, but the dedup logic itself is not in `chat-rest-handler.ts`. The prerequisite should reference `chat.ts` (or both files) so an implementer doesn't look in the wrong place and conclude the guard is missing. `[fixable]`
- 🔵 **style** (L8): `streaming-input-session-control.md` status is listed implicitly as never implemented (via the "Incorporates and replaces" line grouping it with `session-state-machine.md`). However, `streaming-input-session-control.md` has `Status: Implemented (v0.1.0)` and was shipped with protocol v2 (PR #33). The relationship is different — this doc replaces a *shipped* design, not a proposed one. Consider clarifying the distinction (e.g., "replaces session-state-machine.md (never implemented) and supersedes streaming-input-session-control.md (implemented, but scope expanded here)"). `[fixable]`
- 🔵 **unsafe_assumptions** (L349): The Files Involved table lists `frontend/src/client-store.ts` for the SSE default transport flip. This file does exist and contains the transport selector logic (`localStorage.getItem('mitzo:transport') === 'sse'`). However, if the SSE flip also involves changes to `packages/client/src/store.ts` (where the `MitzoConnection` is instantiated), that file should be listed in the table as well. Currently the Server table lists the store file for removing `wasRunning` but the Phase 0 SSE flip may span both locations. `[fixable]`
- 🔵 **style** (L181): The `SessionStateEvent` type comment says `setSessionState()` maps `DETACHED/SUSPENDED -> 'preserve last emitted (server buffers)'`. This is an important behavioral detail but it's buried in a code comment. The buffering mechanism (how the server remembers and replays the last emitted client-facing state after a transient internal state) should be called out explicitly in the design text, as implementers need to build this state tracking. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+## Migration Plan
+
+### Phase 0: Foundation (no behavior change)
+
+1. Add `session_state_changed` event to the protocol
+2. Emit it from `setSessionState()` alongside existing events
+3. Add `state` field to `/api/sessions/:id/meta` response
+4. Client receives and logs `session_state_changed` but doesn't act on it yet
+5. Add crash recovery: `recoverStaleSessions()` on server startup
+6. Flip SSE to default transport (currently WS default, SSE opt-in via localStorage)
```
dimakis, line 284: 🟡 **style**: Phase 0 item 7 ('Add `state` column to EventStore sessions table alongside `is_active`') is already implemented — the migration exists in `packages/protocol/src/event-store.ts:282-293` and `setSessionState()` already writes to it. This should be marked as 'Done' or noted as pre-existing to avoid confusion during implementation. Similarly, the `setSessionState` method and `VALID_TRANSITIONS` table already exist. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+
+#### 1. Remove Client-Side Routing on `running`
+
+**Current:** `sendMessage()` checks `wasRunning`. If true, queues locally in `pendingSend[]`. Drains on `session_end` or after 5s timeout.
+
+**Target:** `sendMessage()` always sends via HTTP POST. Server returns acknowledgment. No local queuing. No `pendingSend`. No safety-net timer.
+
+The server already handles send-while-running correctly: `handleSendV2()` routes to `sendToChat()` which pushes onto the `AsyncQueue`. The SDK processes it on the next turn. This path is well-tested (used for takeover/reattach scenarios). The client-side queue is redundant and is the primary source of the current bug.
+
+**Remove:**
```
dimakis, line 109: 🔵 **style**: The doc says the `wasRunning` branch is '~50 lines' to remove. The actual `wasRunning` branch in `packages/client/src/store.ts:377-398` is 22 lines. The ~50 figure may include related code scattered elsewhere (`pendingSend` declarations, timer helpers, drain logic), but as written it overstates the size of the single branch. Consider saying '~50 lines across store and parser' for accuracy. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+- **Presence awareness.** Clients don't need to know about each other. If we want "also viewing on phone" indicators later, that's a separate feature.
+- **Send coordination.** No locks, no "typing" exclusion. If two humans both send, they both get responses. This is unlikely in practice and harmless if it happens.
+- **Session ownership transfer.** With no primary/secondary model, there's nothing to transfer. Any client can send, any client can interrupt.
+
+## Migration Plan
+
+### Phase 0: Foundation (no behavior change)
+
+1. Add `session_state_changed` event to the protocol
+2. Emit it from `setSessionState()` alongside existing events
```
dimakis, line 280: 🟡 **unsafe_assumptions**: Phase 0 item 8 says 'Migrate `setSessionState()` to write both `state` and `is_active`'. Currently `setSessionState()` (event-store.ts:190-196) only writes `state` and `last_state_change` — it does NOT write `is_active`. Meanwhile `markSessionInactive()` only writes `is_active`. This means during the transition period, a session ended via `setSessionState(ENDED)` won't flip `is_active = 0`, causing stale `is_active = 1` for ended sessions. The doc correctly identifies this gap but should explicitly call out the risk: until this migration step is done, any code path that uses `setSessionState()` to end a session without also calling `markSessionInactive()` will leave a stale `is_active`. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+
+- **Any client can send.** POST `/api/chat/send` accepts from any authenticated connection subscribed to the session. Server queues via AsyncQueue (already works). No primary/secondary client distinction.
+- **All clients receive all events.** `session_state_changed`, `message_start`, `block_delta`, etc. are fanned out to every SSE connection subscribed to that session.
+- **Concurrent sends are serialized.** If two clients POST simultaneously, AsyncQueue serializes them. The SDK processes one turn at a time. Both clients see both messages and both responses in sequence order.
+
+#### ConnectionRegistry Changes
+
+**Current:** `ConnectionRegistry` maps one connection per session (owner model).
+
+**Target:** `ConnectionRegistry` tracks a **set** of connections per session:
```
dimakis, line 247: 🟡 **unsafe_assumptions**: The Multi-Client Fan-Out section says 'Concurrent sends are serialized. If two clients POST simultaneously, AsyncQueue serializes them. The SDK processes one turn at a time.' This assumes the SDK handles interleaved user messages gracefully. If the SDK's `query()` doesn't support multiple sequential user messages pushed mid-generation (i.e., a second message arriving while the first is still generating), this could produce unexpected behavior. The doc should note whether this is a tested SDK capability or an assumption that needs verification. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-researched design doc with accurate problem analysis. Main issues: Phase 0 describes already-implemented work (state column migration) without noting it, the is_active/state synchronization gap during transition needs explicit risk callout, and the UX impact of moving from optimistic to acknowledged sends should be addressed._

- 🟡 **style** (L284): Phase 0 item 7 ('Add `state` column to EventStore sessions table alongside `is_active`') is already implemented — the migration exists in `packages/protocol/src/event-store.ts:282-293` and `setSessionState()` already writes to it. This should be marked as 'Done' or noted as pre-existing to avoid confusion during implementation. Similarly, the `setSessionState` method and `VALID_TRANSITIONS` table already exist. `[fixable]`
- 🔵 **style** (L109): The doc says the `wasRunning` branch is '~50 lines' to remove. The actual `wasRunning` branch in `packages/client/src/store.ts:377-398` is 22 lines. The ~50 figure may include related code scattered elsewhere (`pendingSend` declarations, timer helpers, drain logic), but as written it overstates the size of the single branch. Consider saying '~50 lines across store and parser' for accuracy. `[fixable]`
- 🟡 **unsafe_assumptions** (L280): Phase 0 item 8 says 'Migrate `setSessionState()` to write both `state` and `is_active`'. Currently `setSessionState()` (event-store.ts:190-196) only writes `state` and `last_state_change` — it does NOT write `is_active`. Meanwhile `markSessionInactive()` only writes `is_active`. This means during the transition period, a session ended via `setSessionState(ENDED)` won't flip `is_active = 0`, causing stale `is_active = 1` for ended sessions. The doc correctly identifies this gap but should explicitly call out the risk: until this migration step is done, any code path that uses `setSessionState()` to end a session without also calling `markSessionInactive()` will leave a stale `is_active`. `[fixable]`
- 🟡 **unsafe_assumptions** (L247): The Multi-Client Fan-Out section says 'Concurrent sends are serialized. If two clients POST simultaneously, AsyncQueue serializes them. The SDK processes one turn at a time.' This assumes the SDK handles interleaved user messages gracefully. If the SDK's `query()` doesn't support multiple sequential user messages pushed mid-generation (i.e., a second message arriving while the first is still generating), this could produce unexpected behavior. The doc should note whether this is a tested SDK capability or an assumption that needs verification. `[fixable]`
- 🔵 **style** (L215): The `SessionStateEvent` type comment says 'CREATED/ENDED -> idle', 'STARTING/ACTIVE -> running'. The DETACHED/SUSPENDED mapping says 'preserve last emitted (server buffers)'. However, CLOSING state is not mentioned in the mapping. Given the 7-state server model includes CLOSING (referenced by the 'Why 3 states, not 7' paragraph), the mapping should specify CLOSING's client-facing state — likely 'running' (still generating) or 'idle' (winding down). `[fixable]`
- 🟡 **regressions** (L316): Phase 2 removes `sendMessage()`'s `wasRunning` guard and makes all sends go via HTTP POST with acknowledgment. But the current `sendMessage()` fallback path (line 400-403 in store.ts) handles `connection.send()` returning false with a `sendError` message. Phase 2 step 3 says 'Client awaits POST response before marking message as sent' — this changes the UX from immediate-optimistic (message appears instantly, error shown later) to acknowledged (message appears only after 202). The doc should explicitly note this UX change and whether the `USER_SEND` reducer action (which adds the user bubble immediately at line 364) should be deferred until acknowledgment or kept as-is with a 'pending' state. `[fixable]`
- 🔵 **style** (L336): Phase 3 item 6 says 'ConnectionRegistry fan-out: track set of connections per session, not single owner'. But the current ConnectionRegistry (packages/harness/src/connection-registry.ts) already tracks per-connection per-session cursors via a two-level Map, and connections have `watchedSessions: Set<string>`. The inverse mapping (session -> connections) may need addition, but the doc should clarify what specifically needs to change vs what's already there. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+
+The client marks the message as "sent" only after receiving 202. On failure, it shows an error with retry. No ambiguity about whether the server received the message.
+
+**Note:** This is already partially implemented -- the SSE transport uses HTTP POST for sends. The gap is that the client doesn't wait for or act on the response.
+
+#### 5. Sequence-Based Resumption
+
+**Current:** EventStore already assigns monotonic sequence numbers to events. The periodic sync mechanism uses cursors. But reconnect still attempts full state reconstruction via `handleReconnect`.
+
+**Target:** SSE reconnect sends `?from=<lastSeq>`. Server replays events from that sequence number. No full state reconstruction, no `handleReconnect` ownership dance.
```
dimakis, line 215: 🔵 **style**: The `SessionStateEvent` type comment says 'CREATED/ENDED -> idle', 'STARTING/ACTIVE -> running'. The DETACHED/SUSPENDED mapping says 'preserve last emitted (server buffers)'. However, CLOSING state is not mentioned in the mapping. Given the 7-state server model includes CLOSING (referenced by the 'Why 3 states, not 7' paragraph), the mapping should specify CLOSING's client-facing state — likely 'running' (still generating) or 'idle' (winding down). `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+**Prerequisite:** Verify `clientMsgId` idempotency guard from PR #386 is still present and keyed correctly. The dedup logic lives in `server/chat.ts` (`storeAndEchoIfNew()` which calls `eventStore.hasUserMessage()`), not in `chat-rest-handler.ts` (which only passes `clientMsgId` through). If refactored or removed, re-implement before enabling retries.
+
+1. Remove `wasRunning` branch from `sendMessage()`
+2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
+3. Client awaits POST response before marking message as sent
+4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key so retried POSTs are deduplicated server-side
+5. Add `sendError` UI state for permanent failures
+6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
+
+**Tests:** Verify send-while-running works (server queues via AsyncQueue). Verify POST failure shows error UI. Verify heartbeat timeout triggers reconnect. Verify no message loss on reconnect during generation.
```
dimakis: 🟡 **regressions**: Phase 2 removes `sendMessage()`'s `wasRunning` guard and makes all sends go via HTTP POST with acknowledgment. But the current `sendMessage()` fallback path (line 400-403 in store.ts) handles `connection.send()` returning false with a `sendError` message. Phase 2 step 3 says 'Client awaits POST response before marking message as sent' — this changes the UX from immediate-optimistic (message appears instantly, error shown later) to acknowledged (message appears only after 202). The doc should explicitly note this UX change and whether the `USER_SEND` reducer action (which adds the user bubble immediately at line 364) should be deferred until acknowledgment or kept as-is with a 'pending' state. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,406 @@
...
+6. `ConnectionRegistry` fan-out: track set of connections per session, not single owner
+
+**Tests:** Verify reconnect replays missed events correctly. Verify no duplicate events. Verify generation continues during disconnect and events are available on reconnect. Verify multi-client fan-out (two SSE connections receive same events).
+
+**Dead code removal:** `handleReconnect` (ws-handler-v2.ts), `doReconnectPost` (packages/client/src/sse-connection.ts), periodic sync timer and `startPeriodicSync()`/`stopPeriodicSync()` logic (packages/harness/src/connection-registry.ts), `reconnect` message type.
+
+### Phase 4: Cleanup
+
+1. Remove WS transport code (SSE is default since Phase 0)
+2. Remove `SessionSseRegistry` if consolidated into `ConnectionRegistry`
```
dimakis, line 338: 🔵 **style**: Phase 3 item 6 says 'ConnectionRegistry fan-out: track set of connections per session, not single owner'. But the current ConnectionRegistry (packages/harness/src/connection-registry.ts) already tracks per-connection per-session cursors via a two-level Map, and connections have `watchedSessions: Set<string>`. The inverse mapping (session -> connections) may need addition, but the doc should clarify what specifically needs to change vs what's already there. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+
+#### ConnectionRegistry Changes
+
+**Current:** `ConnectionRegistry` maps one connection per session (owner model).
+
+**Target:** `ConnectionRegistry` tracks a **set** of connections per session:
+
+```typescript
+// Current
+sessions: Map<sessionId, { connectionId: string; cursor: number }>;
```
dimakis, line 251: 🟡 **bugs**: The "Current" pseudo-code `sessions: Map<sessionId, { connectionId, cursor }>` misrepresents ConnectionRegistry's actual structure. In reality, the primary map is `connections: Map<connectionId, Connection>` where each Connection has `watchedSessions: Set<string>`, and cursors are tracked separately in `cursors: Map<connectionId, Map<sessionId, number>>`. The relationship is inverted (connection→sessions, not session→connection). The current code already supports multiple connections watching the same session. This means the "Target" change may be smaller than described, and implementers working from this doc will start from the wrong mental model. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+
+// Target
+sessions: Map<sessionId, Set<{ connectionId: string; cursor: number }>>;
```
dimakis, line 254: 🟡 **bugs**: The target type `Set<{ connectionId: string; cursor: number }>` won't work correctly in TypeScript/JavaScript — `Set` uses reference equality for objects, making lookup and deletion by value impossible. A `Map<string, number>` (connectionId → cursor) would be the correct data structure here. Given that the existing `ConnectionRegistry` already uses per-connection cursor maps, the target should be described in terms of the actual existing structure rather than an invented one. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+
+**Current:** `sendMessage()` checks `wasRunning`. If true, queues locally in `pendingSend[]`. Drains on `session_end` or after 5s timeout.
+
+**Target:** `sendMessage()` always sends via HTTP POST. Server returns acknowledgment. No local queuing. No `pendingSend`. No safety-net timer.
+
+The server already handles send-while-running correctly: `handleSendV2()` routes to `sendToChat()` which pushes onto the `AsyncQueue`. The SDK processes it on the next turn. This path is well-tested (used for takeover/reattach scenarios). The client-side queue is redundant and is the primary source of the current bug.
+
+**Remove:**
+
+- `wasRunning` branch in `store.sendMessage()` (~50 lines)
```
dimakis, line 111: 🟡 **unsafe_assumptions**: The doc says the `wasRunning` branch is "~50 lines" but the actual implementation spans lines 339-404 in `packages/client/src/store.ts` (66 lines). Minor inaccuracy, but an implementer using this doc to scope removal work would underestimate. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+1. Remove `wasRunning` branch from `sendMessage()`
+2. Remove `pendingSend[]`, `PENDING_SEND_TIMEOUT_MS`, drain logic, timer, callbacks
+3. Client awaits POST response before marking message as sent
+4. Add retry with exponential backoff on POST failure (max 3 attempts). Use `clientMsgId` as idempotency key so retried POSTs are deduplicated server-side
+5. Add `sendError` UI state for permanent failures
+6. Heartbeat liveness: 20s ping, 40s timeout, auto-reconnect
+
+**UX note:** The `USER_SEND` reducer action (which renders the user bubble immediately) stays as-is -- the UI remains optimistic. The POST acknowledgment controls whether the message is considered _delivered_, not _displayed_. On POST failure, the existing bubble gets a "failed to send" indicator with retry. This preserves the instant-feel UX while adding delivery confidence.
+
+**Tests:** Verify send-while-running works (server queues via AsyncQueue). Verify POST failure shows error UI with retry. Verify heartbeat timeout triggers reconnect. Verify no message loss on reconnect during generation.
```
dimakis, line 318: 🟡 **unsafe_assumptions**: Phase 2 prerequisite says idempotency guard `storeAndEchoIfNew()` lives in `server/chat.ts` and calls `eventStore.hasUserMessage()`. This is accurate today, but the doc warns it may have been "refactored or removed" since PR #386 — this safety net is good. However, the Phase 2 plan adds retry with exponential backoff (up to 3 attempts) relying on this dedup, but doesn't specify whether the HTTP POST handler (`chat-rest-handler.ts`) also needs dedup at the HTTP layer (e.g., returning 202 for already-accepted messages instead of processing them again up to the dedup check). If the dedup is only in `storeAndEchoIfNew()`, the agent will still receive the message twice via AsyncQueue before dedup fires.

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+  state: 'idle' | 'running' | 'requires_action';
+  // Two emission paths:
+  //
+  // 1. From setSessionState() (lifecycle transitions):
+  //    CREATED/ENDED           -> 'idle'
+  //    STARTING/ACTIVE         -> 'running'
+  //    DETACHED/SUSPENDED      -> preserve last emitted (server buffers)
+  //
+  // 2. From permission_request v2 events (cross-cutting, not a SessionState):
+  //    permission_request      -> 'requires_action'
```
dimakis, line 148: 🔵 **style**: The `SessionStateEvent` type comment says DETACHED/SUSPENDED → 'preserve last emitted (server buffers)', but the 3-state type `'idle' | 'running' | 'requires_action'` doesn't include a 'no change' sentinel. The prose below explains `lastEmittedClientState` buffering, but the inline type comment could mislead — someone reading just the type definition might expect a fourth state value. Consider clarifying inline that these server states suppress emission entirely (no event sent) rather than emit a special value. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+| `server/ws-handler-v2.ts`                     | Replace `is_active` checks with state-based (P1); remove `handleReconnect` (P3); remove file (P4)          |
+| `server/index.ts`                             | Call `recoverStaleSessions()` on startup                                                                   |
+| `server/app.ts`                               | Add `state` to meta endpoint response                                                                      |
+| `packages/protocol/src/event-store.ts`        | Add `state` column alongside `is_active` (P0); emit events from `setSessionState()`; drop `is_active` (P4) |
+| `packages/protocol/src/types.ts`              | Add `SessionStateEvent` type                                                                               |
+| `packages/harness/src/connection-registry.ts` | Multi-client fan-out: Set of connections per session (P3); remove periodic sync                            |
+| `frontend/src/client-store.ts`                | Flip SSE to default transport (P0)                                                                         |
+
```
dimakis, line 360: 🔵 **style**: The Files Involved table lists `frontend/src/client-store.ts` for "Flip SSE to default transport (P0)" but the actual SSE/WS transport selection logic lives in `packages/client/src/store.ts` (where `MitzoConnection` is configured). `frontend/src/client-store.ts` re-exports from the package. If the flip is just a config default change, the doc should reference the actual file where the transport choice is made. `[fixable]`

dimakis (commented):
## Centaur Review

Found **6** issue(s) (4 warning).


### `docs/design/transport-ssot.md`
_Well-structured design doc with strong problem analysis and phased migration plan. Key concern: the ConnectionRegistry "Current" vs "Target" comparison misrepresents the existing data model (it's already connection→sessions, not session→connection), which could mislead implementers about the scope of multi-client fan-out changes._

- 🟡 **bugs** (L251): The "Current" pseudo-code `sessions: Map<sessionId, { connectionId, cursor }>` misrepresents ConnectionRegistry's actual structure. In reality, the primary map is `connections: Map<connectionId, Connection>` where each Connection has `watchedSessions: Set<string>`, and cursors are tracked separately in `cursors: Map<connectionId, Map<sessionId, number>>`. The relationship is inverted (connection→sessions, not session→connection). The current code already supports multiple connections watching the same session. This means the "Target" change may be smaller than described, and implementers working from this doc will start from the wrong mental model. `[fixable]`
- 🟡 **bugs** (L254): The target type `Set<{ connectionId: string; cursor: number }>` won't work correctly in TypeScript/JavaScript — `Set` uses reference equality for objects, making lookup and deletion by value impossible. A `Map<string, number>` (connectionId → cursor) would be the correct data structure here. Given that the existing `ConnectionRegistry` already uses per-connection cursor maps, the target should be described in terms of the actual existing structure rather than an invented one. `[fixable]`
- 🟡 **unsafe_assumptions** (L111): The doc says the `wasRunning` branch is "~50 lines" but the actual implementation spans lines 339-404 in `packages/client/src/store.ts` (66 lines). Minor inaccuracy, but an implementer using this doc to scope removal work would underestimate. `[fixable]`
- 🟡 **unsafe_assumptions** (L318): Phase 2 prerequisite says idempotency guard `storeAndEchoIfNew()` lives in `server/chat.ts` and calls `eventStore.hasUserMessage()`. This is accurate today, but the doc warns it may have been "refactored or removed" since PR #386 — this safety net is good. However, the Phase 2 plan adds retry with exponential backoff (up to 3 attempts) relying on this dedup, but doesn't specify whether the HTTP POST handler (`chat-rest-handler.ts`) also needs dedup at the HTTP layer (e.g., returning 202 for already-accepted messages instead of processing them again up to the dedup check). If the dedup is only in `storeAndEchoIfNew()`, the agent will still receive the message twice via AsyncQueue before dedup fires.
- 🔵 **style** (L148): The `SessionStateEvent` type comment says DETACHED/SUSPENDED → 'preserve last emitted (server buffers)', but the 3-state type `'idle' | 'running' | 'requires_action'` doesn't include a 'no change' sentinel. The prose below explains `lastEmittedClientState` buffering, but the inline type comment could mislead — someone reading just the type definition might expect a fourth state value. Consider clarifying inline that these server states suppress emission entirely (no event sent) rather than emit a special value. `[fixable]`
- 🔵 **style** (L360): The Files Involved table lists `frontend/src/client-store.ts` for "Flip SSE to default transport (P0)" but the actual SSE/WS transport selection logic lives in `packages/client/src/store.ts` (where `MitzoConnection` is configured). `frontend/src/client-store.ts` re-exports from the package. If the flip is just a config default change, the doc should reference the actual file where the transport choice is made. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+**Target:** Server emits `session_state_changed` events with the authoritative state machine value:
+
+```typescript
+type SessionStateEvent = {
+  type: 'session_state_changed';
+  sessionId: string;
+  state: 'idle' | 'running' | 'requires_action';
+  // Two emission paths:
+  //
```
dimakis, line 141: 🟡 **unsafe_assumptions**: The design specifies DETACHED/SUSPENDED should 'preserve last emitted' client state via a `lastEmittedClientState` field per session and suppress emission. However, the already-implemented `toClientState()` in `packages/protocol/src/event-store.ts:35-37` returns `'idle'` as a safe default for DETACHED/SUSPENDED with a comment noting it can't track last-emitted state at the EventStore level, and emission is not suppressed. The design should either acknowledge this gap as a Phase 1+ follow-up or be updated to match the current behavior. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+## Migration Plan
+
+### Phase 0: Foundation (no behavior change)
+
+1. Add `session_state_changed` event to the protocol
+2. Emit it from `setSessionState()` alongside existing events
+3. Add `state` field to `/api/sessions/:id/meta` response
+4. Client receives and logs `session_state_changed` but doesn't act on it yet
+5. Add crash recovery: `recoverStaleSessions()` on server startup
+6. Flip SSE to default transport (currently WS default, SSE opt-in via localStorage)
```
dimakis, line 284: 🔵 **style**: Phase 0 items 1-2 (session_state_changed event, emission from setSessionState), item 5 (recoverStaleSessions), items 7-8 (state column, setSessionState writing both columns) are already implemented in the codebase. The migration plan reads as future work but these are completed. Consider marking them as done (e.g., with checkboxes or a 'Status: Complete' annotation) to avoid confusion about what remains. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+Configurable via environment variable for development and testing (e.g. `MITZO_HEARTBEAT_INTERVAL_MS=5000` for faster detection during dev).
+
+### Multi-Client Fan-Out
+
+#### Design Decision
+
+Multiple clients can connect to the same session simultaneously (e.g. phone + laptop). This is designed in from the start, not bolted on later.
+
+#### Model
+
```
dimakis, line 238: 🟡 **unsafe_assumptions**: The ConnectionRegistry multi-client fan-out section shows the target structure as `Map<sessionId, Set<{ connectionId, cursor }>>`, but object literals in a Set won't deduplicate by value — Set uses reference equality. This would need to be a Map<sessionId, Map<connectionId, { cursor }>> or similar structure in practice. Since this is a design doc (not code), this is minor, but implementers may copy the pattern verbatim. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+| Process alive?  | `SessionRegistry.has()`                | Query loop management only         |
+| Client display  | `session_state_changed` events         | UI rendering only                  |
+
+**Changes:**
+
+- Add `state` column alongside `is_active` in EventStore (backwards-compatible; Phase 0)
+- Write both `state` and `is_active` in `setSessionState()` during transition period
+- Migrate all `is_active` readers to use `state` (Phase 1)
+- Drop `is_active` column once all consumers migrated (Phase 4)
+- Replace `handleReconnect`'s `storeMeta.isActive` check with state-based logic
```
dimakis, line 185: 🔵 **unsafe_assumptions**: The HTTP POST send endpoint returns 202 with `{ accepted, sessionId, seq }`. The design doesn't specify behavior when the client POSTs to a session that exists in EventStore but has no live process in SessionRegistry (e.g., session in ENDED or DETACHED state without a running agent). Should this return 404, 409, or attempt to restart? This edge case is worth documenting since it's the exact failure mode the design is trying to eliminate. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+| `packages/client/src/protocol-parser.ts` | Handle `session_state_changed`; remove `pendingSend` drain; remove `SET_RUNNING` |
+| `packages/client/src/sse-connection.ts`  | Reconnect with `?from=lastSeq`; remove `doReconnectPost`; add heartbeat timeout  |
+| `packages/client/src/connection.ts`      | Remove in Phase 4 (WS transport removal)                                         |
+| `frontend/src/components/ChatInput.tsx`  | No changes (useQueuedMessages is a separate user-facing feature)                 |
+
+### Dead Code Candidates (by phase)
+
+```
+// Phase 1 — Server-authoritative state
+- SET_RUNNING action type (replaced by session_state_changed)
```
dimakis, line 375: 🔵 **style**: The Files Involved table lists `frontend/src/client-store.ts` for 'Flip SSE to default transport (P0)', but the table header column says 'File' with paths like `server/query-loop.ts` (no leading slash). The client-store entry is the only frontend file in the Server table — consider moving it to the Client table or adding a note that this is a cross-cutting change. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-researched design doc with accurate codebase references. Main concern: Phase 0 items are already implemented but presented as future work, and the DETACHED/SUSPENDED buffering design doesn't match the current implementation — either the doc or the code needs updating to close the gap._

- 🟡 **unsafe_assumptions** (L141): The design specifies DETACHED/SUSPENDED should 'preserve last emitted' client state via a `lastEmittedClientState` field per session and suppress emission. However, the already-implemented `toClientState()` in `packages/protocol/src/event-store.ts:35-37` returns `'idle'` as a safe default for DETACHED/SUSPENDED with a comment noting it can't track last-emitted state at the EventStore level, and emission is not suppressed. The design should either acknowledge this gap as a Phase 1+ follow-up or be updated to match the current behavior. `[fixable]`
- 🔵 **style** (L284): Phase 0 items 1-2 (session_state_changed event, emission from setSessionState), item 5 (recoverStaleSessions), items 7-8 (state column, setSessionState writing both columns) are already implemented in the codebase. The migration plan reads as future work but these are completed. Consider marking them as done (e.g., with checkboxes or a 'Status: Complete' annotation) to avoid confusion about what remains. `[fixable]`
- 🟡 **unsafe_assumptions** (L238): The ConnectionRegistry multi-client fan-out section shows the target structure as `Map<sessionId, Set<{ connectionId, cursor }>>`, but object literals in a Set won't deduplicate by value — Set uses reference equality. This would need to be a Map<sessionId, Map<connectionId, { cursor }>> or similar structure in practice. Since this is a design doc (not code), this is minor, but implementers may copy the pattern verbatim. `[fixable]`
- 🔵 **unsafe_assumptions** (L185): The HTTP POST send endpoint returns 202 with `{ accepted, sessionId, seq }`. The design doesn't specify behavior when the client POSTs to a session that exists in EventStore but has no live process in SessionRegistry (e.g., session in ENDED or DETACHED state without a running agent). Should this return 404, 409, or attempt to restart? This edge case is worth documenting since it's the exact failure mode the design is trying to eliminate. `[fixable]`
- 🔵 **style** (L375): The Files Involved table lists `frontend/src/client-store.ts` for 'Flip SSE to default transport (P0)', but the table header column says 'File' with paths like `server/query-loop.ts` (no leading slash). The client-store entry is the only frontend file in the Server table — consider moving it to the Client table or adding a note that this is a cross-cutting change. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+
+- `useQueuedMessages` in ChatInput (user-facing explicit queue feature, separate concern)
+- `interruptMessage()` guard on `!running` (interrupt only makes sense when running)
+
+#### 2. Explicit Server State Events
+
+**Current:** Client derives `running` from:
+
+- `SET_RUNNING` actions dispatched in various handlers
+- `session_end` events (set `running: false`)
```
dimakis, line 129: 🟡 **unsafe_assumptions**: DETACHED/SUSPENDED buffering described here (`preserve last emitted (server buffers)` with `lastEmittedClientState`) is not implemented. The current `toClientState()` in `packages/protocol/src/event-store.ts:35-36` maps both DETACHED and SUSPENDED to `'idle'` and always emits — no suppression, no buffering. During multi-client fan-out (Phase 3), other clients watching the same session would incorrectly see `idle` when one client's transport drops while the agent is still running. The doc should note this as an open implementation gap, or the mapping should be fixed before Phase 3. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+- `clearPendingSendTimer()` helper
+- `pendingSend = []` in `switchSession`/`newSession`/error handler
+
+**Keep:**
+
+- `useQueuedMessages` in ChatInput (user-facing explicit queue feature, separate concern)
+- `interruptMessage()` guard on `!running` (interrupt only makes sense when running)
+
+#### 2. Explicit Server State Events
+
```
dimakis, line 125: 🔵 **style**: The `SessionStateEvent` comment block maps `CREATED/ENDED -> 'idle'` and `STARTING/ACTIVE -> 'running'` but omits `CLOSING`. The actual `toClientState()` implementation maps CLOSING to `'idle'`. Add it to the CREATED/ENDED line for completeness: `CREATED/CLOSING/ENDED -> 'idle'`. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+Multiple clients can connect to the same session simultaneously (e.g. phone + laptop). This is designed in from the start, not bolted on later.
+
+#### Model
+
+- **Any client can send.** POST `/api/chat/send` accepts from any authenticated connection subscribed to the session. Server queues via AsyncQueue (already works). No primary/secondary client distinction.
+- **All clients receive all events.** `session_state_changed`, `message_start`, `block_delta`, etc. are fanned out to every SSE connection subscribed to that session.
+- **Concurrent sends are serialized.** If two clients POST simultaneously, AsyncQueue serializes them. The SDK processes one turn at a time. Both clients see both messages and both responses in sequence order.
+
+#### ConnectionRegistry Changes
+
```
dimakis, line 244: 🔵 **style**: The ConnectionRegistry pseudocode shows `sessions: Map<sessionId, { connectionId, cursor }>` as the current structure, but the actual implementation is connection-centric: `Map<connectionId, Connection>` where each `Connection` has `watchedSessions: Set<string>`, with cursors tracked in a separate `Map<connectionId, Map<sessionId, number>>`. The 'Target' pseudocode should be revised to match the actual data model being extended, not an idealized version that doesn't correspond to the code. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+- **Send coordination.** No locks, no "typing" exclusion. If two humans both send, they both get responses. This is unlikely in practice and harmless if it happens.
+- **Session ownership transfer.** With no primary/secondary model, there's nothing to transfer. Any client can send, any client can interrupt.
+
+## Migration Plan
+
+### Phase 0: Foundation (no behavior change)
+
+1. Add `session_state_changed` event to the protocol
+2. Emit it from `setSessionState()` alongside existing events
+3. Add `state` field to `/api/sessions/:id/meta` response
```
dimakis, line 281: 🔵 **style**: Phase 0 items (state column, recoverStaleSessions, SSE default, session_state_changed emission) and Phase 1 items (syncRunningState removal, SET_RUNNING removal) are already implemented in the codebase but the doc reads as if they are future work. Consider adding completion markers (e.g. checkboxes or a status column) to each phase so readers know the current progress. `[fixable]`

dimakis (commented):
## Centaur Review

### `docs/design/transport-ssot.md`
_Well-structured design doc with clear problem taxonomy and phased migration. Main concern: the DETACHED/SUSPENDED buffering spec is not implemented — current `toClientState()` contradicts the design by mapping those states to 'idle' instead of preserving the last emitted state. This will matter at Phase 3 (multi-client fan-out). Three minor suggestions on pseudocode accuracy and phase tracking._

- 🟡 **unsafe_assumptions** (L129): DETACHED/SUSPENDED buffering described here (`preserve last emitted (server buffers)` with `lastEmittedClientState`) is not implemented. The current `toClientState()` in `packages/protocol/src/event-store.ts:35-36` maps both DETACHED and SUSPENDED to `'idle'` and always emits — no suppression, no buffering. During multi-client fan-out (Phase 3), other clients watching the same session would incorrectly see `idle` when one client's transport drops while the agent is still running. The doc should note this as an open implementation gap, or the mapping should be fixed before Phase 3. `[fixable]`
- 🔵 **style** (L125): The `SessionStateEvent` comment block maps `CREATED/ENDED -> 'idle'` and `STARTING/ACTIVE -> 'running'` but omits `CLOSING`. The actual `toClientState()` implementation maps CLOSING to `'idle'`. Add it to the CREATED/ENDED line for completeness: `CREATED/CLOSING/ENDED -> 'idle'`. `[fixable]`
- 🔵 **style** (L244): The ConnectionRegistry pseudocode shows `sessions: Map<sessionId, { connectionId, cursor }>` as the current structure, but the actual implementation is connection-centric: `Map<connectionId, Connection>` where each `Connection` has `watchedSessions: Set<string>`, with cursors tracked in a separate `Map<connectionId, Map<sessionId, number>>`. The 'Target' pseudocode should be revised to match the actual data model being extended, not an idealized version that doesn't correspond to the code. `[fixable]`
- 🔵 **style** (L281): Phase 0 items (state column, recoverStaleSessions, SSE default, session_state_changed emission) and Phase 1 items (syncRunningState removal, SET_RUNNING removal) are already implemented in the codebase but the doc reads as if they are future work. Consider adding completion markers (e.g. checkboxes or a status column) to each phase so readers know the current progress. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+- `interruptMessage()` guard on `!running` (interrupt only makes sense when running)
+
+#### 2. Explicit Server State Events
+
+**Current:** Client derives `running` from:
+
+- `SET_RUNNING` actions dispatched in various handlers
+- `session_end` events (set `running: false`)
+- `reconnected` message (carries `running: boolean`)
+- `syncRunningState()` REST polling on foreground
```
dimakis, line 131: 🟡 **bugs**: Section 2 claims the client uses `SET_RUNNING` actions and lists removing `SET_RUNNING` as Phase 1 dead code. However, the codebase already uses `SESSION_STATE_CHANGED` action type (in `packages/client/src/slices/messages.ts:160`) with the exact 3-state model (`idle | running | requires_action`) this doc proposes. The protocol-parser already dispatches `{ type: 'SESSION_STATE_CHANGED', state }` in multiple places and the reducer derives `running` from it. Phase 1 steps 1-2 appear to be already implemented — the doc describes the current state as the target state, which will confuse implementors. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+```typescript
+// Current
+sessions: Map<sessionId, { connectionId: string; cursor: number }>;
+
+// Target
+sessions: Map<sessionId, Set<{ connectionId: string; cursor: number }>>;
+```
+
+Every `session_state_changed` and protocol event fans out to all connections in the set. Connection add/remove is independent -- one client disconnecting doesn't affect others.
+
```
dimakis, line 258: 🟡 **bugs**: Phase 0 step 6 ('Flip SSE to default transport') is already done. `frontend/src/client-store.ts` lines 19-26 explicitly document SSE + HTTP POST as the default transport with a comment: 'This is an intentional flip from WS-default per the transport-ssot design doc.' The migration plan should note which steps are already complete vs. remaining. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+Combined with the state event, reconnect is trivial:
+
+1. Client opens SSE with `?from=lastSeq`
+2. Server replays missed events (including any `session_state_changed`)
+3. Client is caught up. Done.
+
+#### 6. Heartbeat Liveness
+
+**Current:** Various heartbeat mechanisms across transports, inconsistent.
```
dimakis, line 225: 🟡 **bugs**: The ConnectionRegistry current-state description ('maps one connection per session (owner model)') is inaccurate. The actual model in `packages/harness/src/connection-registry.ts` is one connection per browser tab/client, where each connection watches multiple sessions and has one active session. The `ConnectionRegistry` uses `connections: Map<string, Connection>` (keyed by connectionId) with `watchedSessions: Set<string>` per connection — not a session→connection map. The proposed 'Target' change to `Map<sessionId, Set<...>>` is actually inverting the current key structure, which is a larger change than described. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+// Current
+sessions: Map<sessionId, { connectionId: string; cursor: number }>;
+
+// Target
+sessions: Map<sessionId, Set<{ connectionId: string; cursor: number }>>;
+```
+
+Every `session_state_changed` and protocol event fans out to all connections in the set. Connection add/remove is independent -- one client disconnecting doesn't affect others.
+
+#### Event Fan-Out
```
dimakis, line 259: 🟡 **unsafe_assumptions**: Phase 0 step 7 ('Add state column to EventStore sessions table alongside is_active') may already be partially done. The EventStore in `packages/protocol/src/event-store.ts` already has a `state` property in SessionMeta (line 80) and a `setSessionState()` method (line 609). The doc should clarify what's already present vs. what needs to be added to avoid redundant migration work. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+| `server/app.ts`                               | Add `state` to meta endpoint response                                                                      |
+| `packages/protocol/src/event-store.ts`        | Add `state` column alongside `is_active` (P0); emit events from `setSessionState()`; drop `is_active` (P4) |
+| `packages/protocol/src/types.ts`              | Add `SessionStateEvent` type                                                                               |
+| `packages/harness/src/connection-registry.ts` | Multi-client fan-out: Set of connections per session (P3); remove periodic sync                            |
+| `frontend/src/client-store.ts`                | Flip SSE to default transport (P0)                                                                         |
+
+### Client
+
+| File                                     | Changes                                                                          |
+| ---------------------------------------- | -------------------------------------------------------------------------------- |
```
dimakis, line 364: 🔵 **style**: The 'Files Involved' table lists `frontend/src/client-store.ts` for Phase 0 ('Flip SSE to default transport'), but this is already implemented per that file's current contents. Consider adding a 'Status' column to the files table or marking completed items, so implementors can distinguish completed from remaining work. `[fixable]`

docs/design/transport-ssot.md
```diff
@@ -0,0 +1,408 @@
...
+| `packages/client/src/store.ts`           | Remove `wasRunning` queuing; await POST ack; remove `syncRunningState`           |
+| `packages/client/src/protocol-parser.ts` | Handle `session_state_changed`; remove `pendingSend` drain; remove `SET_RUNNING` |
+| `packages/client/src/sse-connection.ts`  | Reconnect with `?from=lastSeq`; remove `doReconnectPost`; add heartbeat timeout  |
+| `packages/client/src/connection.ts`      | Remove in Phase 4 (WS transport removal)                                         |
+| `frontend/src/components/ChatInput.tsx`  | No changes (useQueuedMessages is a separate user-facing feature)                 |
+
+### Dead Code Candidates (by phase)
+
+```
```
dimakis, line 373: 🔵 **style**: The 'Dead Code Candidates' section lists `SET_RUNNING action type` for Phase 1 removal, but this action type doesn't exist in the codebase (only `SESSION_STATE_CHANGED` does). This entry should be removed or corrected to reflect the actual dead code that needs cleaning up. `[fixable]`

dimakis (commented):
## Centaur Review

Found **6** issue(s) (4 warning).


### `docs/design/transport-ssot.md`
_Design doc is well-structured and the architecture is sound, but several claims about the current codebase are stale — key Phase 0/1 items (SSE default, SESSION_STATE_CHANGED action type, state column) appear already implemented, and the ConnectionRegistry data model description is inaccurate. These should be corrected before implementation begins to avoid confusion._

- 🟡 **bugs** (L131): Section 2 claims the client uses `SET_RUNNING` actions and lists removing `SET_RUNNING` as Phase 1 dead code. However, the codebase already uses `SESSION_STATE_CHANGED` action type (in `packages/client/src/slices/messages.ts:160`) with the exact 3-state model (`idle | running | requires_action`) this doc proposes. The protocol-parser already dispatches `{ type: 'SESSION_STATE_CHANGED', state }` in multiple places and the reducer derives `running` from it. Phase 1 steps 1-2 appear to be already implemented — the doc describes the current state as the target state, which will confuse implementors. `[fixable]`
- 🟡 **bugs** (L258): Phase 0 step 6 ('Flip SSE to default transport') is already done. `frontend/src/client-store.ts` lines 19-26 explicitly document SSE + HTTP POST as the default transport with a comment: 'This is an intentional flip from WS-default per the transport-ssot design doc.' The migration plan should note which steps are already complete vs. remaining. `[fixable]`
- 🟡 **bugs** (L225): The ConnectionRegistry current-state description ('maps one connection per session (owner model)') is inaccurate. The actual model in `packages/harness/src/connection-registry.ts` is one connection per browser tab/client, where each connection watches multiple sessions and has one active session. The `ConnectionRegistry` uses `connections: Map<string, Connection>` (keyed by connectionId) with `watchedSessions: Set<string>` per connection — not a session→connection map. The proposed 'Target' change to `Map<sessionId, Set<...>>` is actually inverting the current key structure, which is a larger change than described. `[fixable]`
- 🟡 **unsafe_assumptions** (L259): Phase 0 step 7 ('Add state column to EventStore sessions table alongside is_active') may already be partially done. The EventStore in `packages/protocol/src/event-store.ts` already has a `state` property in SessionMeta (line 80) and a `setSessionState()` method (line 609). The doc should clarify what's already present vs. what needs to be added to avoid redundant migration work. `[fixable]`
- 🔵 **style** (L364): The 'Files Involved' table lists `frontend/src/client-store.ts` for Phase 0 ('Flip SSE to default transport'), but this is already implemented per that file's current contents. Consider adding a 'Status' column to the files table or marking completed items, so implementors can distinguish completed from remaining work. `[fixable]`
- 🔵 **style** (L373): The 'Dead Code Candidates' section lists `SET_RUNNING action type` for Phase 1 removal, but this action type doesn't exist in the codebase (only `SESSION_STATE_CHANGED` does). This entry should be removed or corrected to reflect the actual dead code that needs cleaning up. `[fixable]`

PR #431, opened 2026-07-02
feat(transport): transport SSOT phase 0 foundation

dimakis:
## Summary

- Transport SSOT Phase 0: additive foundation with no behavior change
- Emits `session_state_changed` events from `setSessionState()` with 7-state → 3-state mapping (idle/running/requires_action)
- Syncs `is_active` from `state` on every transition (backwards-compatible for P4 removal)
- Adds `recoverStaleSessions()` — marks ACTIVE/STARTING/DETACHED/SUSPENDED → ENDED on server startup
- Flips SSE to default transport (WS opt-in via `localStorage.setItem('mitzo:transport', 'ws')`)
- Adds `state` to `/api/sessions/:id/meta` response
- Protocol parser handles `session_state_changed` (no-op, ready for P1)

## Context

Design doc: PR #430 (`design/transport-ssot` branch)
Telos: 97361f814c5f54df

## Test plan

- [x] TypeScript build passes (`tsc -b`)
- [x] No new test failures (39 pre-existing on main, unchanged)
- [ ] Centaur review
- [ ] Verify SSE works as default on phone + laptop
- [ ] Verify crash recovery on server restart (check logs for "recovered stale sessions")

### DIFF
--- a/frontend/src/client-store.ts
+++ b/frontend/src/client-store.ts
@@ -17,13 +17,14 @@ import { eventBus } from './lib/event-bus-singleton';
 import { getPreferredModel } from './lib/model-preference';
 
 /**
- * Transport selector — set localStorage 'mitzo:transport' to 'sse' to use
- * SSE + HTTP POST instead of WebSocket. Default is 'ws'.
+ * Transport selector — SSE + HTTP POST is the default transport (Transport SSOT P0).
+ * This is an intentional flip from WS-default per the transport-ssot design doc.
+ * Set localStorage 'mitzo:transport' to 'ws' to fall back to WebSocket.
  *
- * Toggle from console: localStorage.setItem('mitzo:transport', 'sse'); location.reload();
- * Revert:              localStorage.removeItem('mitzo:transport'); location.reload();
+ * Force WS:   localStorage.setItem('mitzo:transport', 'ws'); location.reload();
+ * Revert SSE: localStorage.removeItem('mitzo:transport'); location.reload();
  */
-const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') === 'sse';
+const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') !== 'ws';
 
 const sseConfig: SseConnectionConfig | undefined = useSSE
   ? {
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -292,6 +292,15 @@ export function parseServerMessage(
       callbacks.onSessionRenamed?.(msg.name as string);
       break;
 
+    case 'session_state_changed':
+      // P0: log for observability, no UI action yet (Phase 1 will bind to running state)
+      console.debug('[mitzo] session_state_changed', {
+        sessionId: msg.sessionId,
+        state: msg.state,
+        internalState: msg.internalState,
+      });
+      break;
+
     case 'message_start':
       result.messagesActions.push({
         type: 'MESSAGE_START',
--- a/packages/protocol/src/event-store.ts
+++ b/packages/protocol/src/event-store.ts
@@ -5,11 +5,42 @@ import type {
   SessionMeta,
   SessionSearchResult,
   SessionState,
+  ClientSessionState,
   EventStoreLogger,
 } from './types.js';
 
 // Re-export types for consumer convenience
-export type { StoredEvent, SessionMeta, SessionSearchResult, SessionState, EventStoreLogger };
+export type {
+  StoredEvent,
+  SessionMeta,
+  SessionSearchResult,
+  SessionState,
+  ClientSessionState,
+  EventStoreLogger,
+};
+
+/**
+ * Map internal 7-state lifecycle to client-facing 3-state.
+ * Note: 'requires_action' is never returned here — it is emitted separately
+ * by the permission_request handler (Phase 1), not from lifecycle transitions.
+ */
+function toClientState(state: SessionState): ClientSessionState {
+  switch (state) {
+    case 'STARTING':
+    case 'ACTIVE':
+      return 'running';
+    case 'CREATED':
+    case 'CLOSING':
+    case 'ENDED':
+    case 'DETACHED':
+    case 'SUSPENDED':
+      return 'idle';
+    default: {
+      const _exhaustive: never = state;
+      return _exhaustive;
+    }
+  }
+}
 
 const noopLogger: EventStoreLogger = { info() {} };
 
@@ -191,6 +222,7 @@ export class EventStore {
         `UPDATE sessions SET
           state = ?,
           last_state_change = ?,
+          is_active = ?,
           updated_at = unixepoch('now', 'subsec') * 1000
         WHERE session_id = ?`,
       ),
@@ -596,12 +628,25 @@ export class EventStore {
       }
     }
 
-    this.stmts.setSessionState.run(newState, now, sessionId);
+    // Sync is_active from state (backwards-compatible, P0).
+    // Only ENDED/CLOSING are inactive. CREATED is transient — see recoverStaleSessions().
+    const isActive = newState !== 'ENDED' && newState !== 'CLOSING' ? 1 : 0;
+    this.stmts.setSessionState.run(newState, now, isActive, sessionId);
+
+    // Emit session_state_changed event for client consumption (P0)
+    const clientState = toClientState(newState);
+    this.append(sessionId, 'session_state_changed', {
+      sessionId,
+      state: clientState,
+      internalState: newState,
+      timestamp: now,
+    });
 
     this.log.info('session state transition', {
       sessionId,
       fromState,
       toState: newState,
+      clientState,
       clientId: opts?.clientId,
       reason: opts?.reason,
     });
@@ -610,6 +655,39 @@ export class EventStore {
   getSessionState(sessionId: string): SessionState | null {
     const row = this.stmts.getSessionState.get(sessionId) as { state: string | null } | undefined;
     return (row?.state as SessionState) ?? null;
+  }
+
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, SUSPENDED, or CLOSING is transitioned to ENDED.
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    // CREATED excluded: transient state, moves to STARTING synchronously in startChat().
+    // The crash window between CREATED and STARTING is negligible.
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED', 'CLOSING'];
+    const placeholders = staleStates.map(() => '?').join(', ');
+    // Inline prepare is intentional — this runs once at startup, not worth caching.
+    const rows = this.db!.prepare(
+      `SELECT session_id FROM sessions WHERE state IN (${placeholders})`,
+    ).all(...staleStates) as Array<{ session_id: string }>;
+
+    for (const row of rows) {
+      this.setSessionState(row.session_id, 'ENDED', {
+        reason: 'server_restart',
+        force: true,
+      });
+    }
+
+    if (rows.length > 0) {
+      this.log.info('recovered stale sessions on startup', {
+        count: rows.length,
+        sessionIds: rows.map((r) => r.session_id),
+      });
+    }
+
+    return rows.length;
   }
 
   recordUsage(
--- a/packages/protocol/src/types.ts
+++ b/packages/protocol/src/types.ts
@@ -208,6 +208,22 @@ export type SessionState =
   | 'CLOSING'
   | 'ENDED';
 
+/**
+ * Client-facing session state derived from the internal 7-state machine.
+ * Mirrors Anthropic SDK convention: idle | running | requires_action.
+ */
+export type ClientSessionState = 'idle' | 'running' | 'requires_action';
+
+/** Server-authoritative state event emitted on every lifecycle transition. */
+export interface SessionStateEvent {
+  type: 'session_state_changed';
+  sessionId: string;
+  state: ClientSessionState;
+  /** Internal lifecycle state for debugging (not used for UI). */
+  internalState: SessionState;
+  timestamp: number;
+}
+
 export interface Session {
   id: string;
   summary: string;
--- a/server/index.ts
+++ b/server/index.ts
@@ -1024,6 +1024,11 @@ checkPort(PORT).then((inUse) => {
     // Start periodic sync for connection-level delivery guarantee
     connRegistry.startPeriodicSync();
 
+    // Recover sessions left in incomplete states after crash/restart (Transport SSOT P0).
+    // Must run before reconcileSessionsBackground() so reconciliation sees ENDED states.
+    // recoverStaleSessions() logs internally — no need to log here.
+    eventStore.recoverStaleSessions();
+
     // Eagerly reconcile sessions so the first /api/sessions request is fast and accurate.
     reconcileSessionsBackground();
     // Clean up stale worktrees across all repos.
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -965,6 +965,48 @@ describe('boot_context', () => {
   });
 });
 
+// ─── Session state (Transport SSOT P0) ──────────────────────────────────────
+
+describe('session_state_changed', () => {
+  it('produces no message actions (P0: observability only)', () => {
+    const r = parseServerMessage(
+      {
+        type: 'session_state_changed',
+        sessionId: 'sid-1',
+        state: 'running',
+        internalState: 'ACTIVE',
+        timestamp: 1234567890,
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(r.messagesActions).toHaveLength(0);
+  });
+
+  it('logs via console.debug', () => {
+    const spy = vi.spyOn(console, 'debug').mockImplementation(() => {});
+    parseServerMessage(
+      {
+        type: 'session_state_changed',
+        sessionId: 'sid-1',
+        state: 'idle',
+        internalState: 'ENDED',
+        timestamp: 1234567890,
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(spy).toHaveBeenCalledWith('[mitzo] session_state_changed', {
+      sessionId: 'sid-1',
+      state: 'idle',
+      internalState: 'ENDED',
+    });
+    spy.mockRestore();
+  });
+});
+
 // ─── Subagent cancellation ───────────────────────────────────────────────────
 
 describe('subagent_cancelled', () => {
--- a/packages/protocol/__tests__/event-store.test.ts
+++ b/packages/protocol/__tests__/event-store.test.ts
@@ -642,6 +642,215 @@ describe('EventStore', () => {
     });
   });
 
+  describe('toClientState mapping (via session_state_changed events)', () => {
+    const sid = 'client-state-test';
+
+    beforeEach(() => {
+      store.upsertSession({ sessionId: sid });
+    });
+
+    it('maps CREATED to idle', () => {
+      store.setSessionState(sid, 'CREATED', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('idle');
+      expect(stateEvent?.payload.internalState).toBe('CREATED');
+    });
+
+    it('maps STARTING to running', () => {
+      store.setSessionState(sid, 'STARTING', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('running');
+    });
+
+    it('maps ACTIVE to running', () => {
+      store.setSessionState(sid, 'ACTIVE', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('running');
+    });
+
+    it('maps DETACHED to idle', () => {
+      store.setSessionState(sid, 'DETACHED', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('idle');
+    });
+
+    it('maps SUSPENDED to idle', () => {
+      store.setSessionState(sid, 'SUSPENDED', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('idle');
+    });
+
+    it('maps CLOSING to idle', () => {
+      store.setSessionState(sid, 'CLOSING', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('idle');
+    });
+
+    it('maps ENDED to idle', () => {
+      store.setSessionState(sid, 'ENDED', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.state).toBe('idle');
+    });
+
+    it('includes timestamp in event payload', () => {
+      const before = Date.now();
+      store.setSessionState(sid, 'ACTIVE', { force: true });
+      const events = store.getSessionEvents(sid);
+      const stateEvent = events.find((e) => e.type === 'session_state_changed');
+      expect(stateEvent?.payload.timestamp).toBeGreaterThanOrEqual(before);
+    });
+  });
+
+  describe('setSessionState syncs is_active', () => {
+    const sid = 'is-active-sync-test';
+
+    beforeEach(() => {
+      store.upsertSession({ sessionId: sid });
+    });
+
+    it('sets is_active=true for ACTIVE', () => {
+      store.setSessionState(sid, 'ACTIVE', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(true);
+    });
+
+    it('sets is_active=true for DETACHED', () => {
+      store.setSessionState(sid, 'DETACHED', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(true);
+    });
+
+    it('sets is_active=true for SUSPENDED', () => {
+      store.setSessionState(sid, 'SUSPENDED', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(true);
+    });
+
+    it('sets is_active=false for ENDED', () => {
+      store.setSessionState(sid, 'ENDED', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(false);
+    });
+
+    it('sets is_active=false for CLOSING', () => {
+      store.setSessionState(sid, 'CLOSING', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(false);
+    });
+
+    it('sets is_active=true for CREATED', () => {
+      store.setSessionState(sid, 'CREATED', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(true);
+    });
+
+    it('sets is_active=true for STARTING', () => {
+      store.setSessionState(sid, 'STARTING', { force: true });
+      expect(store.getSession(sid)!.isActive).toBe(true);
+    });
+  });
+
+  describe('recoverStaleSessions', () => {
+    it('transitions ACTIVE sessions to ENDED', () => {
+      store.upsertSession({ sessionId: 'active-1' });
+      store.setSessionState('active-1', 'ACTIVE', { force: true });
+
+      const count = store.recoverStaleSessions();
+
+      expect(count).toBe(1);
+      expect(store.getSessionState('active-1')).toBe('ENDED');
+    });
+
+    it('transitions STARTING sessions to ENDED', () => {
+      store.upsertSession({ sessionId: 'starting-1' });
+      store.setSessionState('starting-1', 'STARTING', { force: true });
+
+      store.recoverStaleSessions();
+
+      expect(store.getSessionState('starting-1')).toBe('ENDED');
+    });
+
+    it('transitions DETACHED sessions to ENDED', () => {
+      store.upsertSession({ sessionId: 'detached-1' });
+      store.setSessionState('detached-1', 'DETACHED', { force: true });
+
+      store.recoverStaleSessions();
+
+      expect(store.getSessionState('detached-1')).toBe('ENDED');
+    });
+
+    it('transitions SUSPENDED sessions to ENDED', () => {
+      store.upsertSession({ sessionId: 'suspended-1' });
+      store.setSessionState('suspended-1', 'SUSPENDED', { force: true });
+
+      store.recoverStaleSessions();
+
+      expect(store.getSessionState('suspended-1')).toBe('ENDED');
+    });
+
+    it('transitions CLOSING sessions to ENDED', () => {
+      store.upsertSession({ sessionId: 'closing-1' });
+      store.setSessionState('closing-1', 'CLOSING', { force: true });
+
+      store.recoverStaleSessions();
+
+      expect(store.getSessionState('closing-1')).toBe('ENDED');
+    });
+
+    it('does not touch ENDED sessions', () => {
+      store.upsertSession({ sessionId: 'ended-1' });
+      store.setSessionState('ended-1', 'ENDED', { force: true });
+
+      const count = store.recoverStaleSessions();
+
+      expect(count).toBe(0);
+      expect(store.getSessionState('ended-1')).toBe('ENDED');
+    });
+
+    it('does not touch CREATED sessions', () => {
+      store.upsertSession({ sessionId: 'created-1' });
+      store.setSessionState('created-1', 'CREATED', { force: true });
+
+      const count = store.recoverStaleSessions();
+
+      expect(count).toBe(0);
+      expect(store.getSessionState('created-1')).toBe('CREATED');
+    });
+
+    it('returns correct count for multiple stale sessions', () => {
+      store.upsertSession({ sessionId: 'stale-1' });
+      store.upsertSession({ sessionId: 'stale-2' });
+      store.upsertSession({ sessionId: 'ok-1' });
+      store.setSessionState('stale-1', 'ACTIVE', { force: true });
+      store.setSessionState('stale-2', 'DETACHED', { force: true });
+      store.setSessionState('ok-1', 'ENDED', { force: true });
+
+      const count = store.recoverStaleSessions();
+
+      expect(count).toBe(2);
+    });
+
+    it('emits session_state_changed events for recovered sessions', () => {
+      store.upsertSession({ sessionId: 'recover-1' });
+      store.setSessionState('recover-1', 'ACTIVE', { force: true });
+
+      // Clear events from setup
+      const beforeCount = store.getSessionEvents('recover-1').length;
+
+      store.recoverStaleSessions();
+
+      const events = store.getSessionEvents('recover-1');
+      // Should have new session_state_changed event from recovery
+      const recoveryEvent = events
+        .slice(beforeCount)
+        .find((e) => e.type === 'session_state_changed');
+      expect(recoveryEvent).toBeDefined();
+      expect(recoveryEvent?.payload.state).toBe('idle');
+      expect(recoveryEvent?.payload.internalState).toBe('ENDED');
+    });
+  });
+
   describe('close', () => {
     it('is safe to call multiple times', () => {
       store.close();
--- a/packages/protocol/src/index.ts
+++ b/packages/protocol/src/index.ts
@@ -19,6 +19,8 @@ export type {
   Session,
   SessionClosedBy,
   SessionState,
+  ClientSessionState,
+  SessionStateEvent,
   StoredEvent,
   SessionMeta,
   SessionSearchResult,
--- a/server/app.ts
+++ b/server/app.ts
@@ -1218,6 +1218,7 @@ app.get('/api/sessions/:id/meta', (req, res) => {
     cwd: meta.cwd,
     mode: meta.mode,
     isActive: meta.isActive,
+    state: meta.state,
     totalTokens,
     totalCostUsd: meta.totalCostUsd,
     numTurns: meta.numTurns,


packages/protocol/src/event-store.ts
```diff
@@ -612,6 +655,35 @@ export class EventStore {
     return (row?.state as SessionState) ?? null;
   }
 
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, or SUSPENDED is transitioned to ENDED.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED'];
```
dimakis: 🟡 **bugs**: recoverStaleSessions() omits CLOSING state. After a crash, sessions left in CLOSING will never reach ENDED — the process performing the graceful shutdown is gone. These sessions become zombies with is_active=0 but state≠ENDED, which may confuse downstream code that checks state vs. is_active. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -598,10 +625,26 @@ export class EventStore {
 
     this.stmts.setSessionState.run(newState, now, sessionId);
 
+    // Backwards-compatible: sync is_active from state (P0)
+    const isActive = newState !== 'ENDED' && newState !== 'CLOSING' ? 1 : 0;
+    this.db!.prepare(
```
dimakis: 🟡 **unsafe_assumptions**: The is_active sync in setSessionState() creates a new prepared statement via this.db!.prepare() on every call instead of using a pre-compiled statement (like the existing stmts.markInactive). On hot paths with frequent state transitions this re-parses the SQL each time. Should be added to the stmts object in init(). `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -5,11 +5,38 @@ import type {
...
+  StoredEvent,
+  SessionMeta,
+  SessionSearchResult,
+  SessionState,
+  ClientSessionState,
+  EventStoreLogger,
+};
+
+/** Map internal 7-state lifecycle to client-facing 3-state. */
+function toClientState(state: SessionState): ClientSessionState {
```
dimakis, line 27: 🔵 **style**: toClientState() never returns 'requires_action', making it a dead value in the ClientSessionState union. The function is module-private so callers can't extend it. Consider either removing 'requires_action' from the type until Phase 1 adds it, or adding a comment to the function noting the intentional gap — otherwise readers will wonder if it's a bug. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +655,35 @@ export class EventStore {
     return (row?.state as SessionState) ?? null;
   }
 
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, or SUSPENDED is transitioned to ENDED.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
```
dimakis, line 666: 🟡 **missing_tests**: recoverStaleSessions() has no test coverage. The existing event-store.test.ts has comprehensive state machine tests — this method should be tested there: verify it transitions ACTIVE/STARTING/DETACHED/SUSPENDED to ENDED, returns the correct count, emits session_state_changed events, and is idempotent when no stale sessions exist. `[fixable]`

packages/protocol/src/event-store.ts
dimakis, line 27: 🔵 **missing_tests**: toClientState() has no unit tests. It maps all 7 internal states to 3 client states — this mapping is a key contract for the protocol and should be verified exhaustively (one assertion per SessionState value). `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -598,10 +625,26 @@ export class EventStore {
 
     this.stmts.setSessionState.run(newState, now, sessionId);
 
+    // Backwards-compatible: sync is_active from state (P0)
+    const isActive = newState !== 'ENDED' && newState !== 'CLOSING' ? 1 : 0;
+    this.db!.prepare(
+      "UPDATE sessions SET is_active = ?, updated_at = unixepoch('now', 'subsec') * 1000 WHERE session_id = ?",
+    ).run(isActive, sessionId);
+
+    // Emit session_state_changed event for client consumption (P0)
```
dimakis, line 636: 🔵 **missing_tests**: The session_state_changed event emission inside setSessionState() is untested. Existing state machine tests don't verify that append() is called with the correct event payload. At minimum, verify the event is stored and contains the expected clientState mapping. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -598,10 +625,26 @@ export class EventStore {
 
     this.stmts.setSessionState.run(newState, now, sessionId);
 
+    // Backwards-compatible: sync is_active from state (P0)
```
dimakis: 🔵 **regressions**: The is_active sync sets DETACHED and SUSPENDED sessions to is_active=1, but the previous code path (markSessionInactive called from query-loop.ts) may have set is_active=0 for detached sessions. If any callers still use markSessionInactive() independently of setSessionState(), the two can get out of sync — is_active becomes a dual-writer field. Verify that all state changes go through setSessionState() now.

dimakis (commented):
## Centaur Review

_Solid P0 foundation — the SSE default flip, crash recovery, and state event plumbing are correct in the happy path. Main concerns: CLOSING state omitted from crash recovery (potential zombie sessions), inline db.prepare() on every state transition (perf), and no tests for the three new behaviors (recoverStaleSessions, toClientState, event emission)._

- 🟡 **bugs** (L664): recoverStaleSessions() omits CLOSING state. After a crash, sessions left in CLOSING will never reach ENDED — the process performing the graceful shutdown is gone. These sessions become zombies with is_active=0 but state≠ENDED, which may confuse downstream code that checks state vs. is_active. `[fixable]`
- 🟡 **unsafe_assumptions** (L630): The is_active sync in setSessionState() creates a new prepared statement via this.db!.prepare() on every call instead of using a pre-compiled statement (like the existing stmts.markInactive). On hot paths with frequent state transitions this re-parses the SQL each time. Should be added to the stmts object in init(). `[fixable]`
- 🔵 **style** (L23): toClientState() never returns 'requires_action', making it a dead value in the ClientSessionState union. The function is module-private so callers can't extend it. Consider either removing 'requires_action' from the type until Phase 1 adds it, or adding a comment to the function noting the intentional gap — otherwise readers will wonder if it's a bug. `[fixable]`
- 🟡 **missing_tests** (L663): recoverStaleSessions() has no test coverage. The existing event-store.test.ts has comprehensive state machine tests — this method should be tested there: verify it transitions ACTIVE/STARTING/DETACHED/SUSPENDED to ENDED, returns the correct count, emits session_state_changed events, and is idempotent when no stale sessions exist. `[fixable]`
- 🔵 **missing_tests** (L23): toClientState() has no unit tests. It maps all 7 internal states to 3 client states — this mapping is a key contract for the protocol and should be verified exhaustively (one assertion per SessionState value). `[fixable]`
- 🔵 **missing_tests** (L634): The session_state_changed event emission inside setSessionState() is untested. Existing state machine tests don't verify that append() is called with the correct event payload. At minimum, verify the event is stored and contains the expected clientState mapping. `[fixable]`
- 🔵 **regressions** (L628): The is_active sync sets DETACHED and SUSPENDED sessions to is_active=1, but the previous code path (markSessionInactive called from query-loop.ts) may have set is_active=0 for detached sessions. If any callers still use markSessionInactive() independently of setSessionState(), the two can get out of sync — is_active becomes a dual-writer field. Verify that all state changes go through setSessionState() now.


_Solid P0 foundation — the SSE default flip, crash recovery, and state event plumbing are correct in the happy path. Main concerns: CLOSING state omitted from crash recovery (potential zombie sessions), inline db.prepare() on every state transition (perf), and no tests for the three new behaviors (recoverStaleSessions, toClientState, event emission)._

- 🔵 **style** (L295): The session_state_changed case is a silent no-op with a comment referencing 'Phase 1'. Consider logging via console.debug or a callbacks hook so the event is observable during development, and to validate the message is actually received. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -292,6 +292,10 @@ export function parseServerMessage(
       callbacks.onSessionRenamed?.(msg.name as string);
       break;
 
+    case 'session_state_changed':
```
dimakis, line 295: 🔵 **style**: The session_state_changed case is a silent no-op with a comment referencing 'Phase 1'. Consider logging via console.debug or a callbacks hook so the event is observable during development, and to validate the message is actually received. `[fixable]`

server/index.ts
```diff
@@ -1024,6 +1024,12 @@ checkPort(PORT).then((inUse) => {
     // Start periodic sync for connection-level delivery guarantee
     connRegistry.startPeriodicSync();
 
+    // Recover sessions left in incomplete states after crash/restart (Transport SSOT P0)
+    const recovered = eventStore.recoverStaleSessions();
+    if (recovered > 0) {
+      log.info(`recovered ${recovered} stale session(s) on startup`);
```
dimakis: 🔵 **style**: Duplicate log: `recoverStaleSessions()` already logs 'recovered stale sessions on startup' with session IDs (event-store.ts:682), and each individual transition is also logged by `setSessionState` (event-store.ts:646). The `log.info` at this call site adds a third, redundant log line. Consider removing it or at minimum removing the EventStore-internal one, since callers have better context. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -292,6 +292,17 @@ export function parseServerMessage(
...
```
dimakis, line 295: 🟡 **missing_tests**: The new `session_state_changed` case in `parseServerMessage` has no test coverage in the protocol-parser test suite. Even though it's currently console-only, a test verifying it doesn't throw (and doesn't produce spurious actions) would prevent regressions when Phase 1 adds real behavior. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -292,6 +292,17 @@ export function parseServerMessage(
...
+      // P0: log for observability, no UI action yet (Phase 1 will bind to running state)
+      if (typeof console !== 'undefined') {
```
dimakis: 🔵 **style**: The `if (typeof console !== 'undefined')` guard is unnecessary — `console` is defined in every environment where this code runs (browsers, Node, React Native). All other `console.warn` calls in this file (e.g., lines 168, 338, 468) are unguarded. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +658,36 @@ export class EventStore {
...
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, SUSPENDED, or CLOSING is transitioned to ENDED.
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED', 'CLOSING'];
+    const placeholders = staleStates.map(() => '?').join(', ');
+    const rows = this.db!.prepare(
```
dimakis, line 672: 🔵 **unsafe_assumptions**: Minor: `recoverStaleSessions` creates a new prepared statement on every call via `this.db!.prepare(...)`. Since this only runs once at startup it's not a performance concern, but it's inconsistent with the rest of the file which pre-prepares all statements in the constructor. If this method is ever called more frequently (e.g., periodic health checks), cache the statement. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -5,11 +5,42 @@ import type {
...
+  ClientSessionState,
+  EventStoreLogger,
+};
+
+/**
+ * Map internal 7-state lifecycle to client-facing 3-state.
+ * Note: 'requires_action' is never returned here — it is emitted separately
+ * by the permission_request handler (Phase 1), not from lifecycle transitions.
+ */
+function toClientState(state: SessionState): ClientSessionState {
```
dimakis, line 27: 🔵 **style**: `toClientState` is a module-private function with no direct unit tests. It's covered indirectly through `setSessionState` integration tests, which is acceptable for P0, but consider exporting it (or a test-only alias) so the mapping table can be tested in isolation — especially once Phase 1 adds `requires_action`. `[fixable]`

frontend/src/client-store.ts
```diff
@@ -17,13 +17,13 @@ import { eventBus } from './lib/event-bus-singleton';
...
+ * Transport selector — SSE + HTTP POST is the default transport (Transport SSOT P0).
+ * Set localStorage 'mitzo:transport' to 'ws' to fall back to WebSocket.
  *
- * Toggle from console: localStorage.setItem('mitzo:transport', 'sse'); location.reload();
- * Revert:              localStorage.removeItem('mitzo:transport'); location.reload();
+ * Force WS:   localStorage.setItem('mitzo:transport', 'ws'); location.reload();
+ * Revert SSE: localStorage.removeItem('mitzo:transport'); location.reload();
  */
-const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') === 'sse';
+const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') !== 'ws';
```
dimakis, line 27: 🟡 **regressions**: Changing the default transport from WS to SSE is a user-visible behavioral change. Any client with no `mitzo:transport` localStorage key will silently switch from WebSocket to SSE on upgrade. If SSE has gaps (e.g., the parser only logs `session_state_changed` rather than acting on it), this could regress existing users. Ensure the SSE path has been validated end-to-end before merging, or gate behind a more explicit opt-in.

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid P0 foundation: state machine, crash recovery, and SSE default are well-structured with good test coverage on the EventStore side. Main concerns are the SSE default flip being a silent behavioral change for existing clients, missing protocol-parser test for the new event type, and minor style nits (duplicate logging, unnecessary console guard)._

- 🔵 **style** (L1030): Duplicate log: `recoverStaleSessions()` already logs 'recovered stale sessions on startup' with session IDs (event-store.ts:682), and each individual transition is also logged by `setSessionState` (event-store.ts:646). The `log.info` at this call site adds a third, redundant log line. Consider removing it or at minimum removing the EventStore-internal one, since callers have better context. `[fixable]`


_Solid P0 foundation: state machine, crash recovery, and SSE default are well-structured with good test coverage on the EventStore side. Main concerns are the SSE default flip being a silent behavioral change for existing clients, missing protocol-parser test for the new event type, and minor style nits (duplicate logging, unnecessary console guard)._

- 🟡 **missing_tests** (L295): The new `session_state_changed` case in `parseServerMessage` has no test coverage in the protocol-parser test suite. Even though it's currently console-only, a test verifying it doesn't throw (and doesn't produce spurious actions) would prevent regressions when Phase 1 adds real behavior. `[fixable]`
- 🔵 **style** (L297): The `if (typeof console !== 'undefined')` guard is unnecessary — `console` is defined in every environment where this code runs (browsers, Node, React Native). All other `console.warn` calls in this file (e.g., lines 168, 338, 468) are unguarded. `[fixable]`


_Solid P0 foundation: state machine, crash recovery, and SSE default are well-structured with good test coverage on the EventStore side. Main concerns are the SSE default flip being a silent behavioral change for existing clients, missing protocol-parser test for the new event type, and minor style nits (duplicate logging, unnecessary console guard)._

- 🔵 **unsafe_assumptions** (L670): Minor: `recoverStaleSessions` creates a new prepared statement on every call via `this.db!.prepare(...)`. Since this only runs once at startup it's not a performance concern, but it's inconsistent with the rest of the file which pre-prepares all statements in the constructor. If this method is ever called more frequently (e.g., periodic health checks), cache the statement. `[fixable]`
- 🔵 **style** (L27): `toClientState` is a module-private function with no direct unit tests. It's covered indirectly through `setSessionState` integration tests, which is acceptable for P0, but consider exporting it (or a test-only alias) so the mapping table can be tested in isolation — especially once Phase 1 adds `requires_action`. `[fixable]`


_Solid P0 foundation: state machine, crash recovery, and SSE default are well-structured with good test coverage on the EventStore side. Main concerns are the SSE default flip being a silent behavioral change for existing clients, missing protocol-parser test for the new event type, and minor style nits (duplicate logging, unnecessary console guard)._

- 🟡 **regressions** (L26): Changing the default transport from WS to SSE is a user-visible behavioral change. Any client with no `mitzo:transport` localStorage key will silently switch from WebSocket to SSE on upgrade. If SSE has gaps (e.g., the parser only logs `session_state_changed` rather than acting on it), this could regress existing users. Ensure the SSE path has been validated end-to-end before merging, or gate behind a more explicit opt-in.

dimakis:
## Centaur Review

### `server/app.ts`
_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🔴 **regressions** (L1221): Adding `state: meta.state` to the `/api/sessions/:id/meta` response will break the existing route test in `server/__tests__/routes.test.ts` (line ~450). The test uses `toEqual` (strict equality), so the unexpected `state` field will cause an assertion failure. The test fixture mock also lacks a `state` property, so `meta.state` will be `undefined` in tests. `[fixable]`


_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🟡 **regressions**: `SessionMetaResponse` interface (around line 49) is missing the new `state` field that the server now returns. Frontend consumers of this type won't have type-safe access to session state. Add `state: string | null;` to the interface. `[fixable]`


_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🟡 **missing_tests**: The default transport flip from WS to SSE (`!== 'ws'` instead of `=== 'sse'`) is a significant behavioral change but has no test coverage. If there's existing code or documentation that assumes WS is the default, this could cause breakage. Consider at minimum a test that verifies `useSSE` is true when localStorage has no `mitzo:transport` key. `[fixable]`


### `server/query-loop.ts`
_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🔵 **style** (L1401): Redundant write: `markSessionInactive()` (line 1401) sets `is_active=0`, then `setSessionState('ENDED')` (line 1402) also sets `is_active=0` via the new sync logic. The comment on event-store.ts:633 notes this migration should happen in P1, but it's worth noting this is now a double-write to the same column in the same transaction. `[fixable]`


_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🔵 **unsafe_assumptions** (L639): Every `setSessionState` call now appends a `session_state_changed` event via `store.append()`, but events appended outside the query loop (CREATED, STARTING, DETACHED, SUSPENDED from chat.ts/app.ts) are not pushed in real-time — they rely on the 5-second periodic sync for delivery. This is acceptable for P0 observability, but worth documenting as a known latency gap for Phase 1 when UI will consume these events.


### `packages/protocol/src/types.ts`
_Solid P0 foundation. One critical regression (route test will break from new `state` field with strict `toEqual`), one missing client-side type update, and a few documentation/cleanup suggestions for the P0→P1 transition._

- 🔵 **missing_tests**: The `SessionStateEvent` interface is defined but not included in the frontend `ServerMessage` union type (`frontend/src/types/ws-messages.ts`). While `@mitzo/client` uses a loose `{ type: string }` interface that handles it, the frontend types are out of sync with the actual protocol. Consider adding it to `ServerMessage` for type parity. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. Main concern is DETACHED mapping to 'idle' while is_active=true, creating a semantic mismatch that should be documented or resolved in P1._

- 🟡 **regressions**: SessionStateEvent (or a local SessionStateChangedMsg interface) is not added to the ServerMessage union. The protocol parser handles it, so it won't crash, and this matches the existing pattern (e.g., boot_context is also missing from the union). However, if any consumer ever narrows on ServerMessage['type'], 'session_state_changed' won't be a valid discriminant. Worth adding to the union for Phase 1 when UI binds to this event. `[fixable]`


_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. Main concern is DETACHED mapping to 'idle' while is_active=true, creating a semantic mismatch that should be documented or resolved in P1._

- 🟡 **bugs**: toClientState maps DETACHED and SUSPENDED to 'idle', but the comment says 'preserve last emitted state' and acknowledges the EventStore can't do this. In practice this means a session that is DETACHED (agent still alive, just disconnected from WS) will emit state='idle' to the client — potentially misleading since the agent is still running. The is_active sync correctly keeps is_active=1 for DETACHED, creating a mismatch: is_active=true but client state='idle'. Consider returning 'running' for DETACHED to match the semantic that the generation is still alive, or document this as a known P0 limitation that P1 resolves. `[fixable]`
- 🔵 **unsafe_assumptions**: recoverStaleSessions() uses an inline db.prepare() with string interpolation of placeholders. This is safe because the placeholders are '?' joined from a fixed-length array, but a future maintainer adding a state with special chars to staleStates could be surprised. The comment 'runs once at startup, not worth caching' is a good callout — no action needed, just noting for awareness.
- 🔵 **style**: toClientState is a module-level function but SessionStateEvent (defined in types.ts) is not imported or referenced by the function signature — it returns ClientSessionState. The function's return type is inferred. Adding an explicit return type annotation `function toClientState(state: SessionState): ClientSessionState` would make the contract clearer. (It already has this — just confirming, no issue.)


_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. Main concern is DETACHED mapping to 'idle' while is_active=true, creating a semantic mismatch that should be documented or resolved in P1._

- 🔵 **missing_tests**: recoverStaleSessions tests don't verify that is_active is synced to 0 for recovered sessions. The setSessionState path does set is_active=0 for ENDED, but an explicit assertion like `expect(store.getSession('active-1')!.isActive).toBe(false)` after recovery would document this guarantee. `[fixable]`


_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. Main concern is DETACHED mapping to 'idle' while is_active=true, creating a semantic mismatch that should be documented or resolved in P1._

- 🔵 **missing_tests**: The SSE-default flip (localStorage !== 'ws' instead of === 'sse') is a behavioral change affecting all clients. There are no client-side tests for the transport selector logic. If a user had previously set localStorage to 'sse' explicitly, that value is now ignored (they'd get SSE anyway, which is fine) — but there's no test documenting the new default or the fallback behavior. `[fixable]`

dimakis:
## Centaur Review

_Solid P0 foundation — the SSE default flip, state event emission, crash recovery, and is_active sync are well-tested and internally consistent. Main concern is the DETACHED/SUSPENDED→idle mapping sending a misleading client state while is_active=1, and the missing ServerMessage union entry for the new event type._

- 🟡 **regressions** (L221): The `ServerMessage` union type does not include a `session_state_changed` variant. While the protocol parser handles it via the untyped switch on `msg.type`, any code that exhaustively matches on `ServerMessage` won't know about this event. Add a `SessionStateChangedMsg` interface and include it in the union for type completeness. `[fixable]`


_Solid P0 foundation — the SSE default flip, state event emission, crash recovery, and is_active sync are well-tested and internally consistent. Main concern is the DETACHED/SUSPENDED→idle mapping sending a misleading client state while is_active=1, and the missing ServerMessage union entry for the new event type._

- 🟡 **bugs** (L39): `toClientState` maps DETACHED and SUSPENDED to 'idle', but the comment says "preserve last emitted state" — which it doesn't actually do. Meanwhile `is_active` is set to 1 for these states (line 634), signaling the session is alive. Emitting 'idle' to the client for DETACHED/SUSPENDED is misleading: from the client's perspective the agent may still be running server-side. The comment acknowledges this limitation but Phase 1 consumers could rely on this incorrect signal. Consider returning 'running' for DETACHED (since the query loop is still alive) or documenting in the `ClientSessionState` type that 'idle' may not mean the agent is truly idle. `[fixable]`
- 🔵 **style** (L27): `toClientState` is a pure mapping function that could live in `types.ts` alongside the types it maps between, making it available to both the event-store and future server-layer overrides. Currently it's module-private in event-store.ts, so the server layer that the comment on line 36-38 mentions would need to duplicate the logic. `[fixable]`
- 🔵 **unsafe_assumptions** (L671): `recoverStaleSessions` uses `this.db!.prepare(...)` with a non-null assertion. If `close()` was called before `recoverStaleSessions()` (e.g., during a fast shutdown-restart cycle), this throws a generic error. The other methods use cached prepared statements which would also fail, but this is the only place with an inline `prepare` on a potentially-null `db`. `[fixable]`


_Solid P0 foundation — the SSE default flip, state event emission, crash recovery, and is_active sync are well-tested and internally consistent. Main concern is the DETACHED/SUSPENDED→idle mapping sending a misleading client state while is_active=1, and the missing ServerMessage union entry for the new event type._

- 🔵 **missing_tests**: No test covers `recoverStaleSessions` being called when the session does not exist (i.e., the SELECT returns a session_id but `setSessionState` is called on it — the `getSession` call inside `setSessionState` returns null, so `fromState` is null). This path works because `force: true` is passed, but an explicit test would document that contract. `[fixable]`
- 🔵 **missing_tests**: The `setSessionState` is_active sync tests cover ACTIVE, DETACHED, SUSPENDED, ENDED, and CLOSING — but not CREATED and STARTING. These would confirm the full mapping table and prevent regressions if the is_active logic changes. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -5,11 +5,42 @@ import type {
...
+  switch (state) {
+    case 'STARTING':
+    case 'ACTIVE':
+      return 'running';
+    case 'CREATED':
+    case 'CLOSING':
+    case 'ENDED':
+      return 'idle';
+    // DETACHED/SUSPENDED: preserve last emitted state.
```
dimakis: 🔵 **style**: Comment says 'DETACHED/SUSPENDED: preserve last emitted state' but the code actually always returns 'idle' — it does not preserve anything. The comment describes aspirational P1 behavior, not what the code does. Suggest rewording to 'DETACHED/SUSPENDED: return idle as a safe default. Phase 1 will preserve last emitted state.' `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -596,12 +628,26 @@ export class EventStore {
       }
     }
 
-    this.stmts.setSessionState.run(newState, now, sessionId);
+    // Sync is_active from state in the same UPDATE (backwards-compatible, P0).
+    // DETACHED/SUSPENDED keep is_active=1 because the generation is still alive server-side.
+    // markSessionInactive() callers should migrate to setSessionState(ENDED) in P1.
+    const isActive = newState !== 'ENDED' && newState !== 'CLOSING' ? 1 : 0;
```
dimakis, line 633: 🟡 **regressions**: is_active is now set to 1 for CREATED state (via the `newState !== 'ENDED' && newState !== 'CLOSING'` check). This is consistent with the schema default (DEFAULT 1) so it's not a regression today. However, CREATED means the session hasn't started the query loop — if any consumer relies on is_active=true meaning 'the agent is running', this could be misleading. The existing markSessionInactive() in query-loop.ts (line 1401) is called right before setSessionState(ENDED) at line 1402, causing a redundant double-write. The PR comment acknowledges this P1 migration, which is good.

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +658,37 @@ export class EventStore {
...
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, SUSPENDED, or CLOSING is transitioned to ENDED.
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED', 'CLOSING'];
+    const placeholders = staleStates.map(() => '?').join(', ');
+    // Inline prepare is intentional — this runs once at startup, not worth caching.
+    const rows = this.db!.prepare(
```
dimakis, line 672: 🔵 **unsafe_assumptions**: recoverStaleSessions() calls this.db!.prepare() with a non-null assertion. If recoverStaleSessions() were ever called after close(), db would be null and this would throw. Currently safe because it's only called at startup in server/index.ts, but a guard (or asserting this.db in the method) would be more defensive. `[fixable]`

packages/protocol/src/types.ts
```diff
@@ -208,6 +208,22 @@ export type SessionState =
   | 'CLOSING'
   | 'ENDED';
 
+/**
+ * Client-facing session state derived from the internal 7-state machine.
+ * Mirrors Anthropic SDK convention: idle | running | requires_action.
+ */
+export type ClientSessionState = 'idle' | 'running' | 'requires_action';
+
+/** Server-authoritative state event emitted on every lifecycle transition. */
```
dimakis, line 217: 🔵 **style**: SessionStateEvent interface is defined but never imported or used anywhere in the codebase. If it's scaffolding for Phase 1, consider adding a brief comment (e.g., '// Used by Phase 1 UI binding') or deferring the definition to the PR that consumes it, to avoid dead code. `[fixable]`

dimakis (commented):
## Centaur Review

_Well-structured P0 foundation. No bugs or critical issues — the state machine, crash recovery, SSE default flip, and event emission are all correctly implemented with thorough test coverage. Findings are minor style/documentation improvements._

- 🔵 **style** (L36): Comment says 'DETACHED/SUSPENDED: preserve last emitted state' but the code actually always returns 'idle' — it does not preserve anything. The comment describes aspirational P1 behavior, not what the code does. Suggest rewording to 'DETACHED/SUSPENDED: return idle as a safe default. Phase 1 will preserve last emitted state.' `[fixable]`
- 🟡 **regressions** (L634): is_active is now set to 1 for CREATED state (via the `newState !== 'ENDED' && newState !== 'CLOSING'` check). This is consistent with the schema default (DEFAULT 1) so it's not a regression today. However, CREATED means the session hasn't started the query loop — if any consumer relies on is_active=true meaning 'the agent is running', this could be misleading. The existing markSessionInactive() in query-loop.ts (line 1401) is called right before setSessionState(ENDED) at line 1402, causing a redundant double-write. The PR comment acknowledges this P1 migration, which is good.
- 🔵 **unsafe_assumptions** (L671): recoverStaleSessions() calls this.db!.prepare() with a non-null assertion. If recoverStaleSessions() were ever called after close(), db would be null and this would throw. Currently safe because it's only called at startup in server/index.ts, but a guard (or asserting this.db in the method) would be more defensive. `[fixable]`


_Well-structured P0 foundation. No bugs or critical issues — the state machine, crash recovery, SSE default flip, and event emission are all correctly implemented with thorough test coverage. Findings are minor style/documentation improvements._

- 🔵 **missing_tests**: No test verifies that setSessionState on a non-existent session handles gracefully (getSession returns null, fromState is null). The code handles it via the `if (fromState && !opts?.force)` guard, but an explicit test would document the contract for crash-recovery scenarios where the session row might not exist. `[fixable]`


### `packages/protocol/src/types.ts`
_Well-structured P0 foundation. No bugs or critical issues — the state machine, crash recovery, SSE default flip, and event emission are all correctly implemented with thorough test coverage. Findings are minor style/documentation improvements._

- 🔵 **style** (L217): SessionStateEvent interface is defined but never imported or used anywhere in the codebase. If it's scaffolding for Phase 1, consider adding a brief comment (e.g., '// Used by Phase 1 UI binding') or deferring the definition to the PR that consumes it, to avoid dead code. `[fixable]`

frontend/src/client-store.ts
```diff
@@ -17,13 +17,14 @@ import { eventBus } from './lib/event-bus-singleton';
...
+ * This is an intentional flip from WS-default per the transport-ssot design doc.
+ * Set localStorage 'mitzo:transport' to 'ws' to fall back to WebSocket.
  *
- * Toggle from console: localStorage.setItem('mitzo:transport', 'sse'); location.reload();
- * Revert:              localStorage.removeItem('mitzo:transport'); location.reload();
+ * Force WS:   localStorage.setItem('mitzo:transport', 'ws'); location.reload();
+ * Revert SSE: localStorage.removeItem('mitzo:transport'); location.reload();
  */
-const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') === 'sse';
+const useSSE = typeof window !== 'undefined' && localStorage.getItem('mitzo:transport') !== 'ws';
```
dimakis, line 27: 🟡 **regressions**: Flipping the default transport from WS to SSE is a behavioral change that affects all clients on upgrade. Users with `mitzo:transport` unset in localStorage (the common case) will silently switch from WebSocket to SSE. If the SSE transport has any feature gaps or bugs, this becomes a production regression with no server-side rollback — each client must manually set `localStorage('mitzo:transport', 'ws')`. Consider whether the SSE path has been validated at parity with WS before making it the default, or gate this behind a more visible opt-in (e.g., server-sent feature flag).

packages/protocol/src/event-store.ts
dimakis, line 633: 🟡 **bugs**: The `is_active` derivation treats `CREATED` as active (`is_active=1`), but `recoverStaleSessions` does NOT recover `CREATED` sessions. This is internally consistent (CREATED is not stale), but creates a subtle invariant: a session in CREATED state will appear active forever if `startChat` never runs. If this is intentional (CREATED is a transient state that doesn't persist long), it's fine — but worth documenting why CREATED is excluded from recovery yet kept `is_active=1`. `[fixable]`

packages/protocol/src/event-store.ts
dimakis, line 672: 🔵 **style**: `this.db!` non-null assertion in `recoverStaleSessions()` — if `close()` has been called before startup completes (unlikely but possible in tests or error paths), this will throw an unhandled error. Other methods in this class use the cached `this.stmts` which would also fail, so this is consistent with the existing pattern, but a guard-and-return-0 would be more defensive for a startup-only method. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +658,37 @@ export class EventStore {
...
   }
 
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, SUSPENDED, or CLOSING is transitioned to ENDED.
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED', 'CLOSING'];
```
dimakis, line 669: 🔵 **style**: The `staleStates` array duplicates the set of states that map to `is_active=1` minus `CREATED`. Consider deriving it from the same source as the `isActive` logic (line 634) to keep them in sync, or add a comment cross-referencing the two. If a new state is added to `SessionState` in the future, both locations need updating independently. `[fixable]`

server/index.ts
```diff
@@ -1024,6 +1024,10 @@ checkPort(PORT).then((inUse) => {
     // Start periodic sync for connection-level delivery guarantee
     connRegistry.startPeriodicSync();
 
+    // Recover sessions left in incomplete states after crash/restart (Transport SSOT P0).
+    // recoverStaleSessions() logs internally — no need to log here.
```
dimakis, line 1029: 🔵 **unsafe_assumptions**: `recoverStaleSessions()` runs synchronously in the `server.listen` callback, which is fine for SQLite. But it runs before `reconcileSessionsBackground()` — if reconciliation also touches session state, the ordering dependency should be documented (recovery must happen first so reconciliation sees ENDED states). `[fixable]`

dimakis (commented):
## Centaur Review

_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. The main concern is the default transport flip from WS to SSE, which is a silent behavioral change for all existing clients._

- 🟡 **regressions** (L27): Flipping the default transport from WS to SSE is a behavioral change that affects all clients on upgrade. Users with `mitzo:transport` unset in localStorage (the common case) will silently switch from WebSocket to SSE. If the SSE transport has any feature gaps or bugs, this becomes a production regression with no server-side rollback — each client must manually set `localStorage('mitzo:transport', 'ws')`. Consider whether the SSE path has been validated at parity with WS before making it the default, or gate this behind a more visible opt-in (e.g., server-sent feature flag).


_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. The main concern is the default transport flip from WS to SSE, which is a silent behavioral change for all existing clients._

- 🟡 **bugs** (L634): The `is_active` derivation treats `CREATED` as active (`is_active=1`), but `recoverStaleSessions` does NOT recover `CREATED` sessions. This is internally consistent (CREATED is not stale), but creates a subtle invariant: a session in CREATED state will appear active forever if `startChat` never runs. If this is intentional (CREATED is a transient state that doesn't persist long), it's fine — but worth documenting why CREATED is excluded from recovery yet kept `is_active=1`. `[fixable]`
- 🔵 **style** (L671): `this.db!` non-null assertion in `recoverStaleSessions()` — if `close()` has been called before startup completes (unlikely but possible in tests or error paths), this will throw an unhandled error. Other methods in this class use the cached `this.stmts` which would also fail, so this is consistent with the existing pattern, but a guard-and-return-0 would be more defensive for a startup-only method. `[fixable]`
- 🔵 **style** (L668): The `staleStates` array duplicates the set of states that map to `is_active=1` minus `CREATED`. Consider deriving it from the same source as the `isActive` logic (line 634) to keep them in sync, or add a comment cross-referencing the two. If a new state is added to `SessionState` in the future, both locations need updating independently. `[fixable]`


_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. The main concern is the default transport flip from WS to SSE, which is a silent behavioral change for all existing clients._

- 🔵 **missing_tests**: The `setSessionState syncs is_active` tests cover ACTIVE, DETACHED, SUSPENDED, ENDED, and CLOSING — but not CREATED and STARTING. Adding assertions for these two states would complete the matrix and guard against regressions in the `isActive` derivation logic. `[fixable]`


### `server/index.ts`
_Solid P0 foundation — state machine, crash recovery, and event emission are well-tested and correctly implemented. The main concern is the default transport flip from WS to SSE, which is a silent behavioral change for all existing clients._

- 🔵 **unsafe_assumptions** (L1028): `recoverStaleSessions()` runs synchronously in the `server.listen` callback, which is fine for SQLite. But it runs before `reconcileSessionsBackground()` — if reconciliation also touches session state, the ordering dependency should be documented (recovery must happen first so reconciliation sees ENDED states). `[fixable]`

packages/protocol/src/event-store.ts
dimakis: 🔵 **style**: Comment says "preserve last emitted state" but the code returns a hardcoded 'idle'. The comment describes aspirational P1 behavior, not what the code does. Consider rewriting to: "DETACHED/SUSPENDED: return 'idle' as a safe default. The server layer can override if needed." (drop the misleading "preserve last emitted" line). `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -5,11 +5,42 @@ import type {
...
 } from './types.js';
 
 // Re-export types for consumer convenience
-export type { StoredEvent, SessionMeta, SessionSearchResult, SessionState, EventStoreLogger };
+export type {
+  StoredEvent,
+  SessionMeta,
+  SessionSearchResult,
+  SessionState,
+  ClientSessionState,
```
dimakis, line 18: 🟡 **regressions**: ClientSessionState is re-exported from event-store.ts (the server-side entry point '@mitzo/protocol/event-store') but not from the main package index (packages/protocol/src/index.ts). Frontend consumers importing from '@mitzo/protocol' won't have access to this type. This will become a problem when Phase 1 needs the type in frontend code. `[fixable]`

packages/protocol/src/event-store.ts
dimakis, line 27: 🟡 **bugs**: toClientState lacks a default case, so TypeScript won't catch a missing branch if a new SessionState variant is added. The return type is ClientSessionState but a new unmatched variant would return undefined at runtime. Add `default: { const _exhaustive: never = state; return _exhaustive; }` for compile-time exhaustiveness checking. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -596,12 +628,29 @@ export class EventStore {
       }
     }
 
-    this.stmts.setSessionState.run(newState, now, sessionId);
+    // Sync is_active from state in the same UPDATE (backwards-compatible, P0).
```
dimakis: 🔵 **style**: The 6-line comment block above `const isActive = ...` is heavy for what is a straightforward boolean derivation. The migration note about markSessionInactive() is useful but could be a single line; the CREATED explanation is better placed near recoverStaleSessions where it actually matters. `[fixable]`

packages/protocol/src/types.ts
```diff
@@ -208,6 +208,22 @@ export type SessionState =
...
+export interface SessionStateEvent {
```
dimakis, line 218: 🔵 **style**: SessionStateEvent interface is defined and exported from types.ts but never imported or used anywhere in the codebase. If it's meant to be a documented contract for Phase 1 consumers, consider exporting it from the package index. Otherwise it's dead code. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid P0 foundation — correct state machine mapping, good test coverage, clean crash recovery. Main concern is the missing exhaustiveness guard in toClientState which could cause a silent undefined return if a new state is added._

- 🔵 **style** (L36): Comment says "preserve last emitted state" but the code returns a hardcoded 'idle'. The comment describes aspirational P1 behavior, not what the code does. Consider rewriting to: "DETACHED/SUSPENDED: return 'idle' as a safe default. The server layer can override if needed." (drop the misleading "preserve last emitted" line). `[fixable]`
- 🟡 **regressions** (L18): ClientSessionState is re-exported from event-store.ts (the server-side entry point '@mitzo/protocol/event-store') but not from the main package index (packages/protocol/src/index.ts). Frontend consumers importing from '@mitzo/protocol' won't have access to this type. This will become a problem when Phase 1 needs the type in frontend code. `[fixable]`
- 🟡 **bugs** (L27): toClientState lacks a default case, so TypeScript won't catch a missing branch if a new SessionState variant is added. The return type is ClientSessionState but a new unmatched variant would return undefined at runtime. Add `default: { const _exhaustive: never = state; return _exhaustive; }` for compile-time exhaustiveness checking. `[fixable]`
- 🔵 **style** (L631): The 6-line comment block above `const isActive = ...` is heavy for what is a straightforward boolean derivation. The migration note about markSessionInactive() is useful but could be a single line; the CREATED explanation is better placed near recoverStaleSessions where it actually matters. `[fixable]`


_Solid P0 foundation — correct state machine mapping, good test coverage, clean crash recovery. Main concern is the missing exhaustiveness guard in toClientState which could cause a silent undefined return if a new state is added._

- 🔵 **missing_tests**: No test for the toClientState exhaustiveness guarantee. If a new SessionState variant is added to the union, the switch in toClientState will silently return undefined (no default case). Consider adding a test that asserts toClientState covers all SessionState values, or adding an exhaustive check (e.g., `default: state satisfies never`). `[fixable]`


### `packages/protocol/src/types.ts`
_Solid P0 foundation — correct state machine mapping, good test coverage, clean crash recovery. Main concern is the missing exhaustiveness guard in toClientState which could cause a silent undefined return if a new state is added._

- 🔵 **style** (L218): SessionStateEvent interface is defined and exported from types.ts but never imported or used anywhere in the codebase. If it's meant to be a documented contract for Phase 1 consumers, consider exporting it from the package index. Otherwise it's dead code. `[fixable]`

dimakis:
## Centaur Review

_Solid P0 foundation — state machine mapping, crash recovery, and SSE default flip are well-tested and correct. Findings are minor: a misleading comment on DETACHED/SUSPENDED mapping, missing type in the frontend union, no test for the SSE default flip, and a narrow edge case with stuck CREATED sessions._

- 🔵 **style** (L36): Comment says "preserve last emitted state" but the code returns a hardcoded 'idle'. The implementation is correct for P0 (EventStore doesn't track last-emitted), but the comment implies an intent that contradicts what the code does. Suggest: "DETACHED/SUSPENDED: return 'idle' as safe default — EventStore doesn't track last-emitted state. Server layer can override if needed." `[fixable]`
- 🔵 **unsafe_assumptions** (L670): CREATED is excluded from `recoverStaleSessions()` by design, but a crash between `upsertSession()` (sets CREATED) and `startChat()` (sets STARTING) leaves the session stuck in CREATED with `is_active=1` permanently. The window is narrow, but over many restarts these could accumulate as phantom active sessions. Consider either including CREATED in stale recovery or adding an age-based filter (e.g., CREATED sessions older than 5 minutes). `[fixable]`
- 🔵 **unsafe_assumptions** (L676): `this.db!` non-null assertion is safe at startup but fragile — if `recoverStaleSessions()` is ever called after `close()`, it throws a confusing TypeError instead of a descriptive error. A guard like `if (!this.db) return 0;` would be defensive and self-documenting. `[fixable]`


_Solid P0 foundation — state machine mapping, crash recovery, and SSE default flip are well-tested and correct. Findings are minor: a misleading comment on DETACHED/SUSPENDED mapping, missing type in the frontend union, no test for the SSE default flip, and a narrow edge case with stuck CREATED sessions._

- 🔵 **missing_tests** (L27): The default transport flipped from WS to SSE (the behavioral change gated by `!== 'ws'` instead of `=== 'sse'`), but there is no test asserting the new default. A unit test for useSSE resolution (no localStorage → true, 'ws' → false, 'sse' → true) would guard against accidental reversion of this intentional flip. `[fixable]`


_Solid P0 foundation — state machine mapping, crash recovery, and SSE default flip are well-tested and correct. Findings are minor: a misleading comment on DETACHED/SUSPENDED mapping, missing type in the frontend union, no test for the SSE default flip, and a narrow edge case with stuck CREATED sessions._

- 🟡 **regressions** (L221): The `ServerMessage` discriminated union in ws-messages.ts does not include `session_state_changed`. While the runtime is safe (the `@mitzo/client` parser uses a loose `type: string` interface), this file is the canonical type documentation for server→client messages. Adding a `SessionStateChangedMsg` entry keeps it in sync with the protocol and makes the new event discoverable for frontend consumers building on the typed union. `[fixable]`

dimakis:
## Centaur Review

_Solid P0 foundation. The one actionable gap is the routes test (server/__tests__/routes.test.ts) which doesn't assert or mock the new `state` field added to the /api/sessions/:id/meta response, meaning the field has no integration-level coverage._

- 🟡 **regressions** (L450): The `GET /api/sessions/:id/meta` route now returns `state: meta.state` (app.ts:1221), but the test assertion at line 450-460 uses `toEqual` which is exact-match. The mock at line 85 doesn't include `state`, so `meta.state` will be `undefined`. The response will contain `state: undefined` (which JSON.stringify omits), so the test likely still passes by accident. However, this means the new `state` field has zero test coverage in the routes test. Add `state` to the mock and the assertion. `[fixable]`


_Solid P0 foundation. The one actionable gap is the routes test (server/__tests__/routes.test.ts) which doesn't assert or mock the new `state` field added to the /api/sessions/:id/meta response, meaning the field has no integration-level coverage._

- 🔵 **missing_tests** (L642): The `toClientState` mapping tests only verify the `state` field in the emitted event. Consider a direct test for `toClientState('STARTING')` etc. if it were exported — but since it's a private function, the current approach of testing through `setSessionState` is fine. However, there's no test verifying that the `session_state_changed` event shape matches the `SessionStateEvent` interface (e.g., that `type` field is `'session_state_changed'`). The tests check `payload.state` and `payload.internalState` but never assert `e.type === 'session_state_changed'` on the event itself — they only use `.find(e => e.type === 'session_state_changed')` as a filter, not as an assertion. `[fixable]`


_Solid P0 foundation. The one actionable gap is the routes test (server/__tests__/routes.test.ts) which doesn't assert or mock the new `state` field added to the /api/sessions/:id/meta response, meaning the field has no integration-level coverage._

- 🔵 **style** (L32): The two `return 'idle'` branches (lines 32-35 for CREATED/CLOSING/ENDED, and lines 38-40 for DETACHED/SUSPENDED) could be merged into a single fallthrough block. The separate block with a comment explaining DETACHED/SUSPENDED is reasonable documentation, but the comment says 'the server layer can override if needed' — this escape hatch doesn't exist in the current code and could mislead future readers into thinking there's a mechanism for it. `[fixable]`
- 🔵 **unsafe_assumptions** (L676): `recoverStaleSessions` uses `this.db!` (non-null assertion). If called after `close()`, this will throw a cryptic runtime error. The method is only called at startup (server/index.ts:1031) so this is safe in practice, but it's inconsistent with other methods that presumably also rely on `this.db` being non-null.
- 🔵 **bugs** (L636): The `is_active` derivation (`newState !== 'ENDED' && newState !== 'CLOSING'`) maps DETACHED/SUSPENDED to `is_active=true`, while `toClientState` maps them to `'idle'`. This is semantically correct (detached sessions are still alive, just disconnected), but the divergence between `isActive=true` and `clientState='idle'` could confuse consumers. The comment at line 634 explains the rationale, which helps. Not a bug, but worth being aware of when Phase 1 adds UI bindings.

packages/protocol/src/event-store.ts
```diff
@@ -596,12 +631,25 @@ export class EventStore {
       }
     }
 
-    this.stmts.setSessionState.run(newState, now, sessionId);
+    // Sync is_active from state (backwards-compatible, P0).
+    // Only ENDED/CLOSING are inactive. CREATED is transient — see recoverStaleSessions().
+    const isActive = newState !== 'ENDED' && newState !== 'CLOSING' ? 1 : 0;
```
dimakis, line 633: 🟡 **regressions**: `markSessionInactive()` (line 574) still directly sets `is_active=0` without updating `state` or emitting `session_state_changed`. Now that `setSessionState()` derives `is_active` from state, there are two divergent mutation paths. In `query-loop.ts:1401` both are called sequentially (redundant but safe), but any future caller of `markSessionInactive` alone will leave `state` and `is_active` inconsistent — e.g., `state=ACTIVE` with `is_active=0`. Consider deprecating `markSessionInactive` in favor of `setSessionState('ENDED')` to make state the single source of truth. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +660,39 @@ export class EventStore {
...
   }
 
+  /**
+   * Recover sessions left in incomplete states after a server crash/restart.
+   * Any session in ACTIVE, STARTING, DETACHED, SUSPENDED, or CLOSING is transitioned to ENDED.
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    // CREATED is excluded: it's transient and moves to STARTING immediately (see isActive comment
```
dimakis: 🟡 **bugs**: `recoverStaleSessions()` excludes CREATED, but `setSessionState` sets `is_active=1` for CREATED (line 636). If the server crashes between `setSessionState('CREATED')` (chat.ts:870) and `setSessionState('STARTING')`, the session is stuck as CREATED+active forever — no recovery path cleans it up. The window is small but real. Consider including CREATED in recovery, or at least adding a comment explaining why an orphaned CREATED is acceptable. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -5,11 +5,45 @@ import type {
...
+  switch (state) {
+    case 'STARTING':
+    case 'ACTIVE':
+      return 'running';
+    case 'CREATED':
+    case 'CLOSING':
+    case 'ENDED':
+      return 'idle';
+    // DETACHED/SUSPENDED: return 'idle' as safe default. The server layer
+    // can override if needed (e.g. to reflect last-known running state).
```
dimakis: 🔵 **style**: The DETACHED/SUSPENDED cases (lines 36-40) are separated from the CREATED/CLOSING/ENDED cases (lines 32-35) with a comment, but they all return the same value ('idle'). Combining them into a single fall-through block would be clearer: `case 'CREATED': case 'CLOSING': case 'ENDED': case 'DETACHED': case 'SUSPENDED': return 'idle';` — the comment about server override can go above the combined block. `[fixable]`

packages/protocol/src/event-store.ts
```diff
@@ -612,6 +660,39 @@ export class EventStore {
...
+   * CLOSING is included because the process performing graceful shutdown is gone after a crash.
+   * Returns the number of sessions recovered.
+   */
+  recoverStaleSessions(): number {
+    // CREATED is excluded: it's transient and moves to STARTING immediately (see isActive comment
+    // in setSessionState). All other non-terminal states are recovered.
+    const staleStates = ['ACTIVE', 'STARTING', 'DETACHED', 'SUSPENDED', 'CLOSING'];
+    const placeholders = staleStates.map(() => '?').join(', ');
+    // Inline prepare is intentional — this runs once at startup, not worth caching.
+    const rows = this.db!.prepare(
```
dimakis, line 672: 🔵 **unsafe_assumptions**: `this.db!` non-null assertion in `recoverStaleSessions()` will throw if called after `close()`. All other methods use `this.stmts` which shares the same assumption, so this is consistent — but since `recoverStaleSessions` is only called at startup, the risk is negligible.

packages/client/src/protocol-parser.ts
```diff
@@ -292,6 +292,15 @@ export function parseServerMessage(
       callbacks.onSessionRenamed?.(msg.name as string);
       break;
 
+    case 'session_state_changed':
```
dimakis, line 295: 🔵 **style**: The `session_state_changed` handler uses `console.debug` for production observability. If the project already has a structured logging pattern for the client package (e.g., via a logger utility), consider using that instead for consistency. If not, `console.debug` is fine for P0. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid P0 foundation — the core state machine, crash recovery, and event emission are well-tested. Main concern is the dual `is_active` mutation path (`markSessionInactive` vs `setSessionState`) which should be consolidated before Phase 1 to make state the true SSOT._

- 🟡 **regressions** (L636): `markSessionInactive()` (line 574) still directly sets `is_active=0` without updating `state` or emitting `session_state_changed`. Now that `setSessionState()` derives `is_active` from state, there are two divergent mutation paths. In `query-loop.ts:1401` both are called sequentially (redundant but safe), but any future caller of `markSessionInactive` alone will leave `state` and `is_active` inconsistent — e.g., `state=ACTIVE` with `is_active=0`. Consider deprecating `markSessionInactive` in favor of `setSessionState('ENDED')` to make state the single source of truth. `[fixable]`
- 🟡 **bugs** (L670): `recoverStaleSessions()` excludes CREATED, but `setSessionState` sets `is_active=1` for CREATED (line 636). If the server crashes between `setSessionState('CREATED')` (chat.ts:870) and `setSessionState('STARTING')`, the session is stuck as CREATED+active forever — no recovery path cleans it up. The window is small but real. Consider including CREATED in recovery, or at least adding a comment explaining why an orphaned CREATED is acceptable. `[fixable]`
- 🔵 **style** (L37): The DETACHED/SUSPENDED cases (lines 36-40) are separated from the CREATED/CLOSING/ENDED cases (lines 32-35) with a comment, but they all return the same value ('idle'). Combining them into a single fall-through block would be clearer: `case 'CREATED': case 'CLOSING': case 'ENDED': case 'DETACHED': case 'SUSPENDED': return 'idle';` — the comment about server override can go above the combined block. `[fixable]`
- 🔵 **unsafe_assumptions** (L675): `this.db!` non-null assertion in `recoverStaleSessions()` will throw if called after `close()`. All other methods use `this.stmts` which shares the same assumption, so this is consistent — but since `recoverStaleSessions` is only called at startup, the risk is negligible.


_Solid P0 foundation — the core state machine, crash recovery, and event emission are well-tested. Main concern is the dual `is_active` mutation path (`markSessionInactive` vs `setSessionState`) which should be consolidated before Phase 1 to make state the true SSOT._

- 🔵 **missing_tests**: No test verifies the interaction between `markSessionInactive()` and `setSessionState()`. Given that both now modify `is_active`, a test confirming that calling `markSessionInactive` followed by `setSessionState('ACTIVE', { force: true })` re-enables `is_active` (or vice versa) would catch future inconsistencies. `[fixable]`
- 🔵 **missing_tests**: No test covers `recoverStaleSessions()` emitting the correct event count when multiple sessions are recovered simultaneously. The existing test ('emits session_state_changed events for recovered sessions') only checks a single session. A test recovering 3+ sessions and verifying each got exactly one new `session_state_changed` event would guard against event duplication. `[fixable]`


_Solid P0 foundation — the core state machine, crash recovery, and event emission are well-tested. Main concern is the dual `is_active` mutation path (`markSessionInactive` vs `setSessionState`) which should be consolidated before Phase 1 to make state the true SSOT._

- 🔵 **style** (L295): The `session_state_changed` handler uses `console.debug` for production observability. If the project already has a structured logging pattern for the client package (e.g., via a logger utility), consider using that instead for consistency. If not, `console.debug` is fine for P0. `[fixable]`

PR #433, opened 2026-07-02
feat(transport): transport SSOT phase 1 — server-authoritative state

dimakis:
## Summary

- Replace client-side `SET_RUNNING` with server-authoritative `session_state_changed` events
- Remove `syncRunningState()` polling — state is now fully event-driven
- Migrate server-side `is_active` readers to use `state` column
- Remove `running` field from reconnected/session_switched messages

Builds on P0 (PR #431, merged). Net -371 lines — more dead code removed than new code added.

## Test plan

- [x] 247 tests pass across protocol-parser, store, and ws-handler-v2
- [x] P0 no-op tests updated to verify P1 action dispatch
- [ ] Manual test: start session, verify running state reflects from server events
- [ ] Manual test: reconnect, verify no `syncRunningState` poll

Supersedes #432 (closed when P0 base merged).

### DIFF
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -185,6 +185,8 @@ export function ChatView() {
 
   const handleStop = useCallback(() => {
     storeStopGeneration();
+    // Optimistic update — server confirms via session_state_changed event,
+    // but we set idle immediately for responsive UI on the stop button.
     storeDispatchMessages({ type: 'SESSION_STATE_CHANGED', state: 'idle' });
   }, [storeStopGeneration, storeDispatchMessages]);
 
--- a/frontend/src/pages/ChatView.tsx
+++ b/frontend/src/pages/ChatView.tsx
@@ -185,7 +185,7 @@ export function ChatView() {
 
   const handleStop = useCallback(() => {
     storeStopGeneration();
-    storeDispatchMessages({ type: 'SET_RUNNING', running: false });
+    storeDispatchMessages({ type: 'SESSION_STATE_CHANGED', state: 'idle' });
   }, [storeStopGeneration, storeDispatchMessages]);
 
   function handlePermission(
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -946,6 +946,41 @@ describe('session_state_changed', () => {
     );
     expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'idle' });
   });
+
+  it('dispatches requires_action state', () => {
+    const r = parseServerMessage(
+      {
+        type: 'session_state_changed',
+        sessionId: 'sid-1',
+        state: 'requires_action',
+        internalState: 'ACTIVE',
+        timestamp: 1234567890,
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'requires_action' });
+  });
+
+  it('warns on unknown state', () => {
+    const spy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+    const r = parseServerMessage(
+      {
+        type: 'session_state_changed',
+        sessionId: 'sid-1',
+        state: 'bogus',
+        internalState: 'ACTIVE',
+        timestamp: 1234567890,
+      },
+      makeState(),
+      makeCallbacks(),
+      POOL_KEY,
+    );
+    expect(r.messagesActions).toHaveLength(0);
+    expect(spy).toHaveBeenCalledWith('[mitzo] unknown session state:', 'bogus');
+    spy.mockRestore();
+  });
 });
 
 // ─── Subagent cancellation ───────────────────────────────────────────────────
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -50,7 +50,7 @@ describe('pool lifecycle events', () => {
 // ─── Reattach ─────────────────────────────────────────────────────────────────
 
 describe('reattach', () => {
-  it('reattached sets running and calls onSessionAssigned', () => {
+  it('reattached calls onSessionAssigned and setWsRunning (P1: no SET_RUNNING)', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'reattached', clientId: 'c1', sessionId: 'sid-1', running: true },
@@ -58,13 +58,14 @@ describe('reattach', () => {
       cb,
       POOL_KEY,
     );
-    expect(r.messagesActions).toEqual([{ type: 'SET_RUNNING', running: true }]);
+    // P1: running state comes from session_state_changed events, not reattached
+    expect(r.messagesActions).toHaveLength(0);
     expect(cb.onSessionAssigned).toHaveBeenCalledWith('sid-1');
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, true);
     expect(r.connectionUpdate).toEqual({ status: 'connected' });
   });
 
-  it('reattach_failed sets running=false and marks connection connected', () => {
+  it('reattach_failed marks connection connected (P1: no SET_RUNNING)', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'reattach_failed', clientId: 'old' },
@@ -72,7 +73,8 @@ describe('reattach', () => {
       cb,
       POOL_KEY,
     );
-    expect(r.messagesActions).toEqual([{ type: 'SET_RUNNING', running: false }]);
+    // P1: running state comes from session_state_changed events
+    expect(r.messagesActions).toHaveLength(0);
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, false);
     expect(r.connectionUpdate).toEqual({ status: 'connected' });
   });
@@ -165,7 +167,8 @@ describe('session lifecycle', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage({ type: 'session_end', sessionId: 'sid' }, state, cb, POOL_KEY);
     expect(r.messagesActions).toContainEqual({ type: 'SESSION_END', sessionId: 'sid' });
-    expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: true });
+    // P1: no SET_RUNNING — running state from server's session_state_changed
+    expect(r.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
     expect(cb.sendQueued).toHaveBeenCalledWith(POOL_KEY, { type: 'send', prompt: 'follow-up' });
     // Second message stays queued
     expect(state.pendingSend).toEqual([{ type: 'send', prompt: 'second' }]);
@@ -385,7 +388,7 @@ describe('error handling', () => {
 // ─── Subscribed ──────────────────────────────────────────────────────────────
 
 describe('subscribed', () => {
-  it('subscribed with running=true sets running', () => {
+  it('subscribed with running=true calls setWsRunning (P1: no SET_RUNNING)', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'subscribed', sessionId: 'sid', running: true },
@@ -393,7 +396,8 @@ describe('subscribed', () => {
       cb,
       POOL_KEY,
     );
-    expect(r.messagesActions).toEqual([{ type: 'SET_RUNNING', running: true }]);
+    // P1: running state from session_state_changed events, not subscribed
+    expect(r.messagesActions).toHaveLength(0);
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, true);
   });
 
@@ -591,7 +595,7 @@ describe('misc', () => {
 // ─── session_takeover ────────────────────────────────────────────────────────
 
 describe('session_takeover', () => {
-  it('produces SET_RUNNING false and ERROR message', () => {
+  it('produces ERROR message (P1: no SET_RUNNING, state from server event)', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'session_takeover', sessionId: 'sess-1' },
@@ -599,7 +603,8 @@ describe('session_takeover', () => {
       cb,
       POOL_KEY,
     );
-    expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: false });
+    // P1: running=false comes from server's session_state_changed event
+    expect(r.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
     expect(r.messagesActions).toContainEqual(
       expect.objectContaining({ type: 'ERROR', error: expect.stringContaining('another device') }),
     );
@@ -616,75 +621,17 @@ describe('reconnected', () => {
     expect(onReconnected).toHaveBeenCalled();
   });
 
-  it('dispatches SET_RUNNING false when active session reports not running', () => {
+  it('P1: reconnected does not dispatch SET_RUNNING — state from replayed events', () => {
     const state = makeState({ currentSessionId: 'sid-1' });
     const r = parseServerMessage(
-      { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 0, running: false }] },
+      { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 3 }] },
       state,
       makeCallbacks(),
       POOL_KEY,
     );
-    expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: false });
-  });
-
-  it('dispatches SET_RUNNING true when active session is still running', () => {
-    const setWsRunning = vi.fn();
-    const state = makeState({ currentSessionId: 'sid-1' });
-    const r = parseServerMessage(
-      { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 0, running: true }] },
-      state,
-      makeCallbacks({ setWsRunning }),
-      POOL_KEY,
-    );
-    expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: true });
-    expect(setWsRunning).toHaveBeenCalledWith(POOL_KEY, true);
-  });
-
-  it('no-ops when no currentSessionId', () => {
-    const state = makeState({ currentSessionId: undefined });
-    const r = parseServerMessage(
-      { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 0, running: false }] },
-      state,
-      makeCallbacks(),
-      POOL_KEY,
-    );
-    expect(r.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
-  });
-
-  it('no-ops when sessions field is undefined (backward compat)', () => {
-    const state = makeState({ currentSessionId: 'sid-1' });
-    const r = parseServerMessage({ type: 'reconnected' }, state, makeCallbacks(), POOL_KEY);
-    expect(r.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
-  });
-
-  it('no-ops when sessions has invalid shape (runtime validation)', () => {
-    const state = makeState({ currentSessionId: 'sid-1' });
-    // Invalid: sessions is not an array
-    const r1 = parseServerMessage(
-      { type: 'reconnected', sessions: { invalid: 'shape' } },
-      state,
-      makeCallbacks(),
-      POOL_KEY,
-    );
-    expect(r1.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
-
-    // Invalid: array contains entries missing required fields
-    const r2 = parseServerMessage(
-      { type: 'reconnected', sessions: [{ sessionId: 'sid-1' }] }, // missing running field
-      state,
-      makeCallbacks(),
-      POOL_KEY,
-    );
-    expect(r2.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
-
-    // Invalid: running field has wrong type
-    const r3 = parseServerMessage(
-      { type: 'reconnected', sessions: [{ sessionId: 'sid-1', running: 'yes' }] },
-      state,
-      makeCallbacks(),
-      POOL_KEY,
-    );
-    expect(r3.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
+    // Running state comes from replayed session_state_changed events, not reconnected payload
+    expect(r.messagesActions).toHaveLength(0);
+    expect(r.connectionUpdate).toEqual({ status: 'connected' });
   });
 });
 
@@ -965,10 +912,10 @@ describe('boot_context', () => {
   });
 });
 
-// ─── Session state (Transport SSOT P0) ──────────────────────────────────────
+// ─── Session state (Transport SSOT P1) ──────────────────────────────────────
 
 describe('session_state_changed', () => {
-  it('produces no message actions (P0: observability only)', () => {
+  it('dispatches SESSION_STATE_CHANGED action with state', () => {
     const r = parseServerMessage(
       {
         type: 'session_state_changed',
@@ -981,12 +928,11 @@ describe('session_state_changed', () => {
       makeCallbacks(),
       POOL_KEY,
     );
-    expect(r.messagesActions).toHaveLength(0);
+    expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'running' });
   });
 
-  it('logs via console.debug', () => {
-    const spy = vi.spyOn(console, 'debug').mockImplementation(() => {});
-    parseServerMessage(
+  it('dispatches idle state on ENDED', () => {
+    const r = parseServerMessage(
       {
         type: 'session_state_changed',
         sessionId: 'sid-1',
@@ -998,12 +944,7 @@ describe('session_state_changed', () => {
       makeCallbacks(),
       POOL_KEY,
     );
-    expect(spy).toHaveBeenCalledWith('[mitzo] session_state_changed', {
-      sessionId: 'sid-1',
-      state: 'idle',
-      internalState: 'ENDED',
-    });
-    spy.mockRestore();
+    expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'idle' });
   });
 });
 
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -167,8 +167,8 @@ describe('session lifecycle', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage({ type: 'session_end', sessionId: 'sid' }, state, cb, POOL_KEY);
     expect(r.messagesActions).toContainEqual({ type: 'SESSION_END', sessionId: 'sid' });
-    // P1: no SET_RUNNING — running state from server's session_state_changed
-    expect(r.messagesActions).not.toContainEqual(expect.objectContaining({ type: 'SET_RUNNING' }));
+    // P1: optimistic running=true when draining pending send
+    expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'running' });
     expect(cb.sendQueued).toHaveBeenCalledWith(POOL_KEY, { type: 'send', prompt: 'follow-up' });
     // Second message stays queued
     expect(state.pendingSend).toEqual([{ type: 'send', prompt: 'second' }]);
@@ -960,7 +960,10 @@ describe('session_state_changed', () => {
       makeCallbacks(),
       POOL_KEY,
     );
-    expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'requires_action' });
+    expect(r.messagesActions).toContainEqual({
+      type: 'SESSION_STATE_CHANGED',
+      state: 'requires_action',
+    });
   });
 
   it('warns on unknown state', () => {
--- a/packages/client/__tests__/protocol-parser.test.ts
+++ b/packages/client/__tests__/protocol-parser.test.ts
@@ -50,7 +50,7 @@ describe('pool lifecycle events', () => {
 // ─── Reattach ─────────────────────────────────────────────────────────────────
 
 describe('reattach', () => {
-  it('reattached calls onSessionAssigned and setWsRunning (P1: no SET_RUNNING)', () => {
+  it('reattached calls onSessionAssigned and setWsRunning', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'reattached', clientId: 'c1', sessionId: 'sid-1', running: true },
@@ -58,14 +58,14 @@ describe('reattach', () => {
       cb,
       POOL_KEY,
     );
-    // P1: running state comes from session_state_changed events, not reattached
+    // Running state comes from session_state_changed events, not reattached
     expect(r.messagesActions).toHaveLength(0);
     expect(cb.onSessionAssigned).toHaveBeenCalledWith('sid-1');
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, true);
     expect(r.connectionUpdate).toEqual({ status: 'connected' });
   });
 
-  it('reattach_failed marks connection connected (P1: no SET_RUNNING)', () => {
+  it('reattach_failed marks connection connected', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'reattach_failed', clientId: 'old' },
@@ -73,7 +73,7 @@ describe('reattach', () => {
       cb,
       POOL_KEY,
     );
-    // P1: running state comes from session_state_changed events
+    // Running state comes from session_state_changed events
     expect(r.messagesActions).toHaveLength(0);
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, false);
     expect(r.connectionUpdate).toEqual({ status: 'connected' });
@@ -167,7 +167,7 @@ describe('session lifecycle', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage({ type: 'session_end', sessionId: 'sid' }, state, cb, POOL_KEY);
     expect(r.messagesActions).toContainEqual({ type: 'SESSION_END', sessionId: 'sid' });
-    // P1: optimistic running=true when draining pending send
+    // Optimistic running=true when draining pending send
     expect(r.messagesActions).toContainEqual({ type: 'SESSION_STATE_CHANGED', state: 'running' });
     expect(cb.sendQueued).toHaveBeenCalledWith(POOL_KEY, { type: 'send', prompt: 'follow-up' });
     // Second message stays queued
@@ -388,7 +388,7 @@ describe('error handling', () => {
 // ─── Subscribed ──────────────────────────────────────────────────────────────
 
 describe('subscribed', () => {
-  it('subscribed with running=true calls setWsRunning (P1: no SET_RUNNING)', () => {
+  it('subscribed with running=true calls setWsRunning', () => {
     const cb = makeCallbacks();
     const r = parseServerMessage(
       { type: 'subscribed', sessionId: 'sid', running: true },
@@ -396,7 +396,7 @@ describe('subscribed', () => {
       cb,
       POOL_KEY,
     );
-    // P1: running state from session_state_changed events, not subscribed
+    // Running state from session_state_changed events, not subscribed
     expect(r.messagesActions).toHaveLength(0);
     expect(cb.setWsRunning).toHaveBeenCalledWith(POOL_KEY, true);
   });
@@ -620,7 +620,7 @@ describe('reconnected', () => {
     expect(onReconnected).toHaveBeenCalled();
   });
 
-  it('P1: reconnected does not dispatch SET_RUNNING — state from replayed events', () => {
+  it('reconnected does not dispatch SET_RUNNING — state from replayed events', () => {
     const state = makeState({ currentSessionId: 'sid-1' });
     const r = parseServerMessage(
       { type: 'reconnected', sessions: [{ sessionId: 'sid-1', replayed: 3 }] },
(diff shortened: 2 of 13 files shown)


dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Solid P1 migration from client-derived to server-authoritative session state. The main concern is a running-state gap on session switch: the old immediate `running` field was removed but no immediate replacement exists — the client relies on periodic sync (5s) or the next live event, which creates a noticeable regression when switching to sessions paused during tool execution._

- 🟡 **regressions**: handleSwitchSession no longer sends `running` in session_switched, and the client comment says 'running state restored via replayed session_state_changed events' — but handleSwitchSession does NOT replay events. It only calls connRegistry.watch(), so the client must wait for either the next live query-loop event (immediate if streaming, seconds if in a tool-execution pause) or the ConnectionRegistry's periodic sync (up to 5 seconds). Previously, running was communicated immediately via the session_switched payload. This creates a visible regression: switching to a running-but-paused session will show the send button instead of the stop button for up to 5 seconds. `[fixable]`


_Solid P1 migration from client-derived to server-authoritative session state. The main concern is a running-state gap on session switch: the old immediate `running` field was removed but no immediate replacement exists — the client relies on periodic sync (5s) or the next live event, which creates a noticeable regression when switching to sessions paused during tool execution._

- 🟡 **regressions**: Removing SET_RUNNING from the `subscribed` handler means the client no longer derives running state from subscription acknowledgment. Combined with the session_switched removal, there is no immediate-delivery path for running state on session switch — only the deferred periodic sync (5s interval) or a new live event from the query loop. Consider having handleSwitchSession replay the latest session_state_changed event (single event, not full history) to close this gap. `[fixable]`
- 🔵 **unsafe_assumptions**: session_state_changed handler checks `typeof msg.state === 'string'` but does not validate against the ClientSessionState union ('idle' | 'running' | 'requires_action'). An unexpected string value (e.g., from a server version mismatch) would silently map to running=false in the reducer. Consider validating against the known values, or at minimum logging unknown states. This is consistent with the rest of the parser's loose typing but is more consequential now that this event drives primary UI state. `[fixable]`


_Solid P1 migration from client-derived to server-authoritative session state. The main concern is a running-state gap on session switch: the old immediate `running` field was removed but no immediate replacement exists — the client relies on periodic sync (5s) or the next live event, which creates a noticeable regression when switching to sessions paused during tool execution._

- 🔵 **missing_tests**: session_state_changed tests cover 'running' and 'idle' states but omit 'requires_action', the third valid ClientSessionState. Since the reducer maps requires_action to running=false (same as idle), this behavior should be explicitly tested to prevent future regressions if the mapping changes. `[fixable]`


_Solid P1 migration from client-derived to server-authoritative session state. The main concern is a running-state gap on session switch: the old immediate `running` field was removed but no immediate replacement exists — the client relies on periodic sync (5s) or the next live event, which creates a noticeable regression when switching to sessions paused during tool execution._

- 🔵 **style**: handleStop dispatches SESSION_STATE_CHANGED('idle') as a client-side optimistic update, which contradicts the PR's stated goal of 'server-authoritative state'. This is likely intentional for UX responsiveness, but worth a brief inline comment noting this is an intentional departure — otherwise a future reader may assume it's a leftover from the old pattern and try to remove it. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -410,22 +407,13 @@ export async function handleSwitchSession(
...
       // Cross-reference registry with durable state to avoid reporting
-      // a zombie query loop as running.
-      const found = ctx.sessionRegistry.findBySessionId(msg.sessionId);
-      const storeState = ctx.eventStore.getSessionState(msg.sessionId);
-      const running =
-        found && storeState !== 'ENDED' && storeState !== 'CLOSING' && storeState !== null
-          ? ctx.sessionRegistry.isActive(found.clientId)
-          : false;
-
       ctx.connRegistry.get(connectionId)?.transport.send({
```
dimakis, line 411: 🟡 **regressions**: handleSwitchSession no longer sends `running` in the `session_switched` response, and does not replay past events. When a user switches to an already-running session, the client resets to `INITIAL_MESSAGES_STATE` (running=false) via switchSession() in store.ts:267. The client only learns the session is running when the next live `session_state_changed` event arrives — but if the session is steadily in ACTIVE state with no transition, no such event fires. The UI will show a send button instead of the stop button until the agent emits a new message_start/block_delta event. Consider either replaying the latest session_state_changed event during switch, or sending a one-shot state snapshot in the session_switched response. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -381,14 +351,13 @@ export function parseServerMessage(
       if (msg.sessionId && !state.currentSessionId) {
         callbacks.onSessionAssigned(msg.sessionId as string);
       }
+      // P1: pending send still drains queued messages, but running state
+      // comes from server's session_state_changed event (P2 removes pendingSend entirely)
       const pending = state.pendingSend.shift();
       if (pending) {
```
dimakis, line 370: 🟡 **regressions**: The `session_end` handler previously dispatched `SET_RUNNING: true` immediately when draining a queued message via `onSendQueued`. Now, `running` goes false (from the SESSION_END reducer case) and stays false until the server round-trips a `session_state_changed: running` event. This creates a brief UI flicker where the stop button disappears and the send button appears between the queued send and the server's state event. The old behavior was optimistic — setting running=true at the same time as sending the queued message. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -293,12 +265,10 @@ export function parseServerMessage(
       break;
 
     case 'session_state_changed':
-      // P0: log for observability, no UI action yet (Phase 1 will bind to running state)
-      console.debug('[mitzo] session_state_changed', {
-        sessionId: msg.sessionId,
-        state: msg.state,
-        internalState: msg.internalState,
-      });
+      // P1: server-authoritative state — derive running from this event only
```
dimakis: 🔵 **style**: The `typeof msg.state === 'string'` check accepts any string, not just valid `ClientSessionState` values. Since this is a system boundary (server → client), consider validating against the known values ('idle' | 'running' | 'requires_action') to avoid silently accepting malformed server messages that would set `running: false` for unexpected states. `[fixable]`

packages/client/src/store.ts
```diff
@@ -385,9 +361,7 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
             parserState.pendingSendTimer = undefined;
             return;
           }
-          set((s) => ({
-            messages: messagesReducer(s.messages, { type: 'SET_RUNNING', running: true }),
-          }));
+          // P1: running state from server's session_state_changed event
```
dimakis: 🟡 **regressions**: Same issue in the pendingSendTimer callback: the timer fires `connection.send(pending)` but no longer sets `running: true`. The UI stays in idle state until the server responds with `session_state_changed`. Previously `SET_RUNNING: true` was dispatched synchronously. For a safety-net timer that fires after a 5s delay, the additional round-trip latency compounds the gap. `[fixable]`

packages/client/src/slices/messages.ts
```diff
@@ -573,8 +574,8 @@ export function messagesReducer(state: MessagesState, action: MessagesAction): M
         running: true,
       };
 
-    case 'SET_RUNNING':
-      return { ...state, running: action.running };
+    case 'SESSION_STATE_CHANGED':
+      return { ...state, running: action.state === 'running' };
```
dimakis: 🔵 **unsafe_assumptions**: `requires_action` maps to `running: false` since `action.state === 'running'` is false. While `requires_action` is currently never emitted from `toClientState()` (the comment at event-store.ts:24 confirms this), the type contract `ClientSessionState = 'idle' | 'running' | 'requires_action'` allows it. If a future phase emits it, the stop button would disappear during permission prompts. Consider `action.state !== 'idle'` to be safe, or document the assumption. `[fixable]`

packages/client/__tests__/protocol-parser.test.ts
```diff
@@ -981,12 +928,11 @@ describe('session_state_changed', () => {
       makeCallbacks(),
```
dimakis, line 927: 🔵 **missing_tests**: Tests cover `state: 'running'` and `state: 'idle'` (mapped from ENDED) but there is no test for `state: 'requires_action'` to verify the expected behavior (currently maps to running=false). Adding a test would document the intended semantics and catch regressions if the mapping changes. `[fixable]`

packages/client/__tests__/store.test.ts
```diff
@@ -1299,139 +1299,6 @@ describe('foreground recovery', () => {
     expect(state.messages[0].messageId).toBe('user-1');
   });
```
dimakis, line 1301: 🔵 **style**: The removed `syncRunningState` tests are replaced with a two-line comment. The foreground recovery section (`describe('foreground recovery')`) still exists but now only tests message re-fetch — not running state. Consider adding a test that verifies `session_state_changed` events received after foreground restore correctly update `running`, to prove the new mechanism works end-to-end. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/ws-handler-v2.ts`
_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🟡 **regressions** (L410): handleSwitchSession no longer sends `running` in the `session_switched` response, and does not replay past events. When a user switches to an already-running session, the client resets to `INITIAL_MESSAGES_STATE` (running=false) via switchSession() in store.ts:267. The client only learns the session is running when the next live `session_state_changed` event arrives — but if the session is steadily in ACTIVE state with no transition, no such event fires. The UI will show a send button instead of the stop button until the agent emits a new message_start/block_delta event. Consider either replaying the latest session_state_changed event during switch, or sending a one-shot state snapshot in the session_switched response. `[fixable]`


_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🟡 **regressions** (L357): The `session_end` handler previously dispatched `SET_RUNNING: true` immediately when draining a queued message via `onSendQueued`. Now, `running` goes false (from the SESSION_END reducer case) and stays false until the server round-trips a `session_state_changed: running` event. This creates a brief UI flicker where the stop button disappears and the send button appears between the queued send and the server's state event. The old behavior was optimistic — setting running=true at the same time as sending the queued message. `[fixable]`
- 🔵 **style** (L268): The `typeof msg.state === 'string'` check accepts any string, not just valid `ClientSessionState` values. Since this is a system boundary (server → client), consider validating against the known values ('idle' | 'running' | 'requires_action') to avoid silently accepting malformed server messages that would set `running: false` for unexpected states. `[fixable]`


_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🟡 **regressions** (L364): Same issue in the pendingSendTimer callback: the timer fires `connection.send(pending)` but no longer sets `running: true`. The UI stays in idle state until the server responds with `session_state_changed`. Previously `SET_RUNNING: true` was dispatched synchronously. For a safety-net timer that fires after a 5s delay, the additional round-trip latency compounds the gap. `[fixable]`


_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🔵 **unsafe_assumptions** (L578): `requires_action` maps to `running: false` since `action.state === 'running'` is false. While `requires_action` is currently never emitted from `toClientState()` (the comment at event-store.ts:24 confirms this), the type contract `ClientSessionState = 'idle' | 'running' | 'requires_action'` allows it. If a future phase emits it, the stop button would disappear during permission prompts. Consider `action.state !== 'idle'` to be safe, or document the assumption. `[fixable]`


_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🔵 **missing_tests** (L928): Tests cover `state: 'running'` and `state: 'idle'` (mapped from ENDED) but there is no test for `state: 'requires_action'` to verify the expected behavior (currently maps to running=false). Adding a test would document the intended semantics and catch regressions if the mapping changes. `[fixable]`


_Clean migration from client-derived to server-authoritative running state. Main concern: switchSession to an already-running session will show running=false until the next live state event, since session_switched no longer carries running and no events are replayed — this is a user-visible regression where the stop button disappears._

- 🔵 **style** (L1301): The removed `syncRunningState` tests are replaced with a two-line comment. The foreground recovery section (`describe('foreground recovery')`) still exists but now only tests message re-fetch — not running state. Consider adding a test that verifies `session_state_changed` events received after foreground restore correctly update `running`, to prove the new mechanism works end-to-end. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean migration from client-driven SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. One UX regression worth addressing: pending-send drain no longer sets running=true optimistically, causing a brief stop→send button flicker. Missing test for the critical session_state_changed follow-up after session_switched._

- 🟡 **regressions** (L356): Removing the optimistic `SET_RUNNING true` when draining a pending send on `session_end` creates a brief UI flicker. Flow: SESSION_END sets running=false → pending message sent to server → server eventually emits session_state_changed with running. During that round-trip, the send button replaces the stop button even though a message is in flight. Same issue exists in the store's `drainOne` timer. The old code bridged this gap with an immediate SET_RUNNING true. Consider dispatching `{ type: 'SESSION_STATE_CHANGED', state: 'running' }` optimistically when draining a queued message, mirroring the optimistic update pattern already used in `handleStop`. `[fixable]`


_Clean migration from client-driven SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. One UX regression worth addressing: pending-send drain no longer sets running=true optimistically, causing a brief stop→send button flicker. Missing test for the critical session_state_changed follow-up after session_switched._

- 🟡 **missing_tests** (L648): The test 'P1: session_switched does not include running field' only checks `transport.sent[0]` for the absence of `running`. It does not verify that a `session_state_changed` event follows as `transport.sent[1]`, which is the critical new behavior added in `handleSwitchSession`. Without this assertion, a regression that removes the follow-up event would go undetected. `[fixable]`


_Clean migration from client-driven SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. One UX regression worth addressing: pending-send drain no longer sets running=true optimistically, causing a brief stop→send button flicker. Missing test for the critical session_state_changed follow-up after session_switched._

- 🔵 **missing_tests** (L967): The `dispatchMessages` test only covers `SESSION_STATE_CHANGED` with `state: 'running'` → `running: true`. There's no corresponding test for `state: 'idle'` → `running: false` or `state: 'requires_action'` → `running: false`. Since `requires_action` mapping to `running=false` is a design decision (permission requests hide the stop button), a test would document this intent and catch accidental regressions if the mapping changes. `[fixable]`
- 🔵 **style** (L1299): The comment `// P1: syncRunningState tests removed — running state is now server-authoritative` is left inside the `foreground recovery` describe block, which now has only one test. Consider removing the orphaned comment or collapsing the describe block if the remaining test is better grouped elsewhere. `[fixable]`


### `server/query-loop.ts`
_Clean migration from client-driven SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. One UX regression worth addressing: pending-send drain no longer sets running=true optimistically, causing a brief stop→send button flicker. Missing test for the critical session_state_changed follow-up after session_switched._

- 🔵 **style** (L1396): `markSessionInactive()` is now dead code — `setSessionState('ENDED')` atomically syncs `is_active` via line 633 of event-store.ts. The method definition, its prepared statement, and its tests in `event-store.test.ts` could be removed to avoid confusion about which path is authoritative. Fine to defer to a cleanup PR. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -381,14 +356,13 @@ export function parseServerMessage(
       if (msg.sessionId && !state.currentSessionId) {
```
dimakis, line 364: 🟡 **regressions**: Removing the optimistic `SET_RUNNING true` when draining a pending send on `session_end` creates a brief UI flicker. Flow: SESSION_END sets running=false → pending message sent to server → server eventually emits session_state_changed with running. During that round-trip, the send button replaces the stop button even though a message is in flight. Same issue exists in the store's `drainOne` timer. The old code bridged this gap with an immediate SET_RUNNING true. Consider dispatching `{ type: 'SESSION_STATE_CHANGED', state: 'running' }` optimistically when draining a queued message, mirroring the optimistic update pattern already used in `handleStop`. `[fixable]`

server/__tests__/ws-handler-v2.test.ts
```diff
@@ -648,40 +648,9 @@ describe('handleSwitchSession', () => {
     await handleSwitchSession('c1', { type: 'switch_session', sessionId: 'sess-1' }, ctx);
```
dimakis, line 648: 🟡 **missing_tests**: The test 'P1: session_switched does not include running field' only checks `transport.sent[0]` for the absence of `running`. It does not verify that a `session_state_changed` event follows as `transport.sent[1]`, which is the critical new behavior added in `handleSwitchSession`. Without this assertion, a regression that removes the follow-up event would go undetected. `[fixable]`

packages/client/__tests__/store.test.ts
```diff
@@ -967,7 +967,7 @@ describe('setModel', () => {
 describe('dispatchMessages', () => {
```
dimakis, line 967: 🔵 **missing_tests**: The `dispatchMessages` test only covers `SESSION_STATE_CHANGED` with `state: 'running'` → `running: true`. There's no corresponding test for `state: 'idle'` → `running: false` or `state: 'requires_action'` → `running: false`. Since `requires_action` mapping to `running=false` is a design decision (permission requests hide the stop button), a test would document this intent and catch accidental regressions if the mapping changes. `[fixable]`

packages/client/__tests__/store.test.ts
```diff
@@ -1299,139 +1299,6 @@ describe('foreground recovery', () => {
     expect(state.messages[0].messageId).toBe('user-1');
```
dimakis, line 1299: 🔵 **style**: The comment `// P1: syncRunningState tests removed — running state is now server-authoritative` is left inside the `foreground recovery` describe block, which now has only one test. Consider removing the orphaned comment or collapsing the describe block if the remaining test is better grouped elsewhere. `[fixable]`

server/query-loop.ts
```diff
@@ -1396,9 +1396,8 @@ async function _runQueryLoopInner(
           durationApiMs: 0, // only available from SDK result
```
dimakis, line 1396: 🔵 **style**: `markSessionInactive()` is now dead code — `setSessionState('ENDED')` atomically syncs `is_active` via line 633 of event-store.ts. The method definition, its prepared statement, and its tests in `event-store.test.ts` could be removed to avoid confusion about which path is authoritative. Fine to defer to a cleanup PR. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -121,36 +122,15 @@ export function parseServerMessage(
...
-        if (active) {
-          result.messagesActions.push({ type: 'SET_RUNNING', running: active.running });
-          if (active.running) callbacks.setWsRunning?.(poolKey, true);
-        }
-      }
       callbacks.onReconnected?.();
       break;
-    }
 
     case 'session_takeover':
```
dimakis, line 136: 🟡 **regressions**: session_takeover no longer sets running=false. The comment says 'running=false comes from server's session_state_changed event', but the server sends session_takeover to the old transport and then unwatches it (ws-handler-v2.ts:553), so the old client will never receive a subsequent session_state_changed event. The UI will remain in 'running' state with a stop button after takeover, instead of showing the error in an idle state. Consider keeping an inline SESSION_STATE_CHANGED dispatch here, or emitting session_state_changed from the server before unwatching. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -381,14 +359,17 @@ export function parseServerMessage(
       if (msg.sessionId && !state.currentSessionId) {
         callbacks.onSessionAssigned(msg.sessionId as string);
       }
+      // Drain first queued message. Optimistic running=true avoids UI flicker
+      // between the send and the server's session_state_changed confirmation.
       const pending = state.pendingSend.shift();
       if (pending) {
```
dimakis, line 370: 🔵 **style**: Unnecessary `as ClientSessionState` cast on a string literal: `'running' as ClientSessionState`. The literal 'running' already satisfies the union type. The cast is needed on line 274 (`msg.state as ClientSessionState`) because it comes from runtime validation, but this one is a compile-time constant. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -439,16 +420,16 @@ export function parseServerMessage(
           command: 'close',
           content: 'Session closing... The agent will commit work and write a summary.',
         });
-        result.messagesActions.push({ type: 'SET_RUNNING', running: false });
+        // P1: running=false from server's session_state_changed (CLOSING → idle)
```
dimakis: 🔵 **regressions**: session_close_ack with accepted=true previously dispatched SET_RUNNING false. Now the comment says 'running=false from server's session_state_changed (CLOSING → idle)'. This is correct only if the server transitions to CLOSING before or concurrently with the ack. Verified: closeSessionByUser calls the agent's closeout flow which should eventually trigger state transitions. However, the 'no active agent' path in closeSessionByUser (chat.ts:1470-1493) calls upsertSession({ isActive: false }) but does NOT call setSessionState('ENDED'), so no session_state_changed event is emitted. This could leave the client in a running state if the agent wasn't active when close was requested. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -292,14 +265,19 @@ export function parseServerMessage(
...
-    case 'session_state_changed':
-      // P0: log for observability, no UI action yet (Phase 1 will bind to running state)
-      console.debug('[mitzo] session_state_changed', {
-        sessionId: msg.sessionId,
-        state: msg.state,
-        internalState: msg.internalState,
-      });
+    case 'session_state_changed': {
+      // P1: server-authoritative state — derive running from this event only
+      const validStates: ReadonlySet<string> = new Set(['idle', 'running', 'requires_action']);
```
dimakis: 🔵 **style**: The Set of valid states is allocated on every call to parseServerMessage. Since it's a ReadonlySet of constants, hoist it to module scope to avoid repeated allocations in a hot path. `[fixable]`

server/__tests__/ws-handler-v2.test.ts
```diff
@@ -618,7 +618,7 @@ describe('handleSwitchSession', () => {
     expect(ctx.connRegistry.hasOpenWatchers('sess-1')).toBe(true);
   });
 
-  it('returns running: true when session has active query loop', async () => {
+  it('P1: session_switched does not include running field', async () => {
```
dimakis: 🟡 **missing_tests**: The test 'P1: session_switched does not include running field' mocks getSessionState('ACTIVE') but only asserts that session_switched lacks a running field. It does not verify that a session_state_changed message is sent as the second transport message — the core new behavior of handleSwitchSession. Add an assertion like: expect(transport.sent[1]).toMatchObject({ type: 'session_state_changed', state: 'running', internalState: 'ACTIVE' }). `[fixable]`

dimakis: 🔵 **missing_tests**: No test covers handleSwitchSession when getSessionState returns null (session exists in EventStore but has no state yet). The production code guards with `if (currentState)`, but no test verifies that only session_switched is sent (no session_state_changed) in that case. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean architectural shift from client-inferred to server-authoritative running state, but session_takeover has a regression where the old client never receives the session_state_changed event to clear running state, and the new session_state_changed emission in handleSwitchSession lacks test coverage._

- 🟡 **regressions** (L132): session_takeover no longer sets running=false. The comment says 'running=false comes from server's session_state_changed event', but the server sends session_takeover to the old transport and then unwatches it (ws-handler-v2.ts:553), so the old client will never receive a subsequent session_state_changed event. The UI will remain in 'running' state with a stop button after takeover, instead of showing the error in an idle state. Consider keeping an inline SESSION_STATE_CHANGED dispatch here, or emitting session_state_changed from the server before unwatching. `[fixable]`
- 🔵 **style** (L365): Unnecessary `as ClientSessionState` cast on a string literal: `'running' as ClientSessionState`. The literal 'running' already satisfies the union type. The cast is needed on line 274 (`msg.state as ClientSessionState`) because it comes from runtime validation, but this one is a compile-time constant. `[fixable]`
- 🔵 **regressions** (L423): session_close_ack with accepted=true previously dispatched SET_RUNNING false. Now the comment says 'running=false from server's session_state_changed (CLOSING → idle)'. This is correct only if the server transitions to CLOSING before or concurrently with the ack. Verified: closeSessionByUser calls the agent's closeout flow which should eventually trigger state transitions. However, the 'no active agent' path in closeSessionByUser (chat.ts:1470-1493) calls upsertSession({ isActive: false }) but does NOT call setSessionState('ENDED'), so no session_state_changed event is emitted. This could leave the client in a running state if the agent wasn't active when close was requested. `[fixable]`
- 🔵 **style** (L270): The Set of valid states is allocated on every call to parseServerMessage. Since it's a ReadonlySet of constants, hoist it to module scope to avoid repeated allocations in a hot path. `[fixable]`


_Clean architectural shift from client-inferred to server-authoritative running state, but session_takeover has a regression where the old client never receives the session_state_changed event to clear running state, and the new session_state_changed emission in handleSwitchSession lacks test coverage._

- 🟡 **missing_tests** (L621): The test 'P1: session_switched does not include running field' mocks getSessionState('ACTIVE') but only asserts that session_switched lacks a running field. It does not verify that a session_state_changed message is sent as the second transport message — the core new behavior of handleSwitchSession. Add an assertion like: expect(transport.sent[1]).toMatchObject({ type: 'session_state_changed', state: 'running', internalState: 'ACTIVE' }). `[fixable]`
- 🔵 **missing_tests** (L621): No test covers handleSwitchSession when getSessionState returns null (session exists in EventStore but has no state yet). The production code guards with `if (currentState)`, but no test verifies that only session_switched is sent (no session_state_changed) in that case. `[fixable]`

server/query-loop.ts
```diff
@@ -1396,9 +1396,8 @@ async function _runQueryLoopInner(
...
         });
       }
-      // Mark session as inactive in durable store
+      // Mark session as ended in durable store (P1: setSessionState syncs is_active)
```
dimakis, line 1399: 🔴 **bugs**: Comment says 'P1: setSessionState syncs is_active' but setSessionState does NOT update the is_active column — it only updates state, last_state_change, and updated_at (see event-store.ts:190-195). Removing markSessionInactive without either (a) adding is_active sync to setSessionState or (b) migrating ALL is_active readers means the is_active column becomes permanently stale for normal query completions. Two unmigrated readers remain: app.ts:1220 (GET /api/sessions/:id/meta returns isActive: meta.isActive) and chat.ts:1806 (session list filter uses !m.isActive). The meta endpoint is no longer critical since syncRunningState is removed, but the comment is factually wrong and will mislead future maintainers. Either add is_active sync to setSessionState (making the comment true), or fix the comment to say 'is_active is no longer the source of truth; callers use getSessionState() instead'. `[fixable]`

dimakis, line 1399: 🟡 **regressions**: GET /api/sessions/:id/meta (app.ts:1220) still returns isActive from the EventStore's is_active column. After this PR, normally-completed sessions will have is_active=1 forever (markSessionInactive removed) while state='ENDED'. Any consumer of this API (external tools, future features) will see stale isActive=true for completed sessions. Consider either keeping markSessionInactive alongside setSessionState, or having setSessionState also sync is_active when transitioning to ENDED/CLOSING. `[fixable]`

packages/client/__tests__/protocol-parser.test.ts
```diff
@@ -616,75 +620,17 @@ describe('reconnected', () => {
...
-      state,
-      makeCallbacks(),
-      POOL_KEY,
-    );
-    expect(r.messagesActions).toContainEqual({ type: 'SET_RUNNING', running: false });
-  });
-
-  it('dispatches SET_RUNNING true when active session is still running', () => {
-    const setWsRunning = vi.fn();
+  it('P1: reconnected does not dispatch SET_RUNNING — state from replayed events', () => {
```
dimakis: 🟡 **regressions**: Five defensive tests for the 'reconnected' handler were deleted: no-ops for undefined currentSessionId, undefined sessions field, non-array sessions, missing required fields, and wrong field types. While the handler no longer processes running state from reconnected, these tests documented runtime validation behavior that guards against malformed server messages. The handler still reads msg.sessions (even if only for replaying events elsewhere), so defensive tests remain valuable. Consider keeping at least the 'no currentSessionId' and 'invalid sessions shape' tests adapted to the new behavior. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -121,36 +126,17 @@ export function parseServerMessage(
 
     // ── v2 handshake events ────────────────────────────────────────────────
```
dimakis, line 128: 🔵 **style**: Temporal 'P1:' prefix comments throughout the code and test names (e.g., 'P1: running state from session_state_changed events', test names like 'P1: no SET_RUNNING') will become meaningless after Phase 1 ships. These should describe the design rationale ('running state derived from session_state_changed events') without referencing the migration phase, since future readers won't know what P1 means. `[fixable]`

packages/client/src/store.ts
```diff
@@ -217,31 +216,8 @@ export function createMitzoStore(options: MitzoStoreOptions): StoreApi<MitzoStor
...
-        }
-      })
-      .catch(() => {
-        // Non-fatal — running state will correct on next event or user action
-      })
-      .finally(() => {
-        runningSyncInFlight = false;
-      });
-  }
+  // P1: syncRunningState removed — running state is server-authoritative
```
dimakis: 🔵 **unsafe_assumptions**: The comment block 'P1: syncRunningState removed — running state is server-authoritative via session_state_changed events. No polling needed.' is reasonable, but there's an implicit assumption that session_state_changed events will always be delivered during foreground recovery. If the WebSocket didn't disconnect during iOS background (just suspended), no reconnect happens, and periodic sync is the only delivery path. Verify that the ConnectionRegistry periodic sync interval is short enough to avoid a stale running indicator on foreground return.

dimakis (commented):
## Centaur Review

### `server/query-loop.ts`
_Sound architectural migration from client-inferred to server-authoritative running state, but the is_active column becomes stale for normal query completions due to markSessionInactive removal without syncing is_active in setSessionState — a misleading comment claims otherwise. Two API consumers still read the now-stale field._

- 🔴 **bugs** (L1399): Comment says 'P1: setSessionState syncs is_active' but setSessionState does NOT update the is_active column — it only updates state, last_state_change, and updated_at (see event-store.ts:190-195). Removing markSessionInactive without either (a) adding is_active sync to setSessionState or (b) migrating ALL is_active readers means the is_active column becomes permanently stale for normal query completions. Two unmigrated readers remain: app.ts:1220 (GET /api/sessions/:id/meta returns isActive: meta.isActive) and chat.ts:1806 (session list filter uses !m.isActive). The meta endpoint is no longer critical since syncRunningState is removed, but the comment is factually wrong and will mislead future maintainers. Either add is_active sync to setSessionState (making the comment true), or fix the comment to say 'is_active is no longer the source of truth; callers use getSessionState() instead'. `[fixable]`
- 🟡 **regressions** (L1399): GET /api/sessions/:id/meta (app.ts:1220) still returns isActive from the EventStore's is_active column. After this PR, normally-completed sessions will have is_active=1 forever (markSessionInactive removed) while state='ENDED'. Any consumer of this API (external tools, future features) will see stale isActive=true for completed sessions. Consider either keeping markSessionInactive alongside setSessionState, or having setSessionState also sync is_active when transitioning to ENDED/CLOSING. `[fixable]`


_Sound architectural migration from client-inferred to server-authoritative running state, but the is_active column becomes stale for normal query completions due to markSessionInactive removal without syncing is_active in setSessionState — a misleading comment claims otherwise. Two API consumers still read the now-stale field._

- 🟡 **missing_tests**: The session_close_ack handler (protocol-parser.ts:426-435) was changed from SET_RUNNING to SESSION_STATE_CHANGED but has zero test coverage — no test for session_close_ack exists in the protocol parser test suite. This is a safety-net path ('server's no active agent path may not emit session_state_changed') that should be tested. `[fixable]`
- 🟡 **regressions** (L623): Five defensive tests for the 'reconnected' handler were deleted: no-ops for undefined currentSessionId, undefined sessions field, non-array sessions, missing required fields, and wrong field types. While the handler no longer processes running state from reconnected, these tests documented runtime validation behavior that guards against malformed server messages. The handler still reads msg.sessions (even if only for replaying events elsewhere), so defensive tests remain valuable. Consider keeping at least the 'no currentSessionId' and 'invalid sessions shape' tests adapted to the new behavior. `[fixable]`


_Sound architectural migration from client-inferred to server-authoritative running state, but the is_active column becomes stale for normal query completions due to markSessionInactive removal without syncing is_active in setSessionState — a misleading comment claims otherwise. Two API consumers still read the now-stale field._

- 🔵 **style** (L128): Temporal 'P1:' prefix comments throughout the code and test names (e.g., 'P1: running state from session_state_changed events', test names like 'P1: no SET_RUNNING') will become meaningless after Phase 1 ships. These should describe the design rationale ('running state derived from session_state_changed events') without referencing the migration phase, since future readers won't know what P1 means. `[fixable]`


_Sound architectural migration from client-inferred to server-authoritative running state, but the is_active column becomes stale for normal query completions due to markSessionInactive removal without syncing is_active in setSessionState — a misleading comment claims otherwise. Two API consumers still read the now-stale field._

- 🔵 **unsafe_assumptions** (L219): The comment block 'P1: syncRunningState removed — running state is server-authoritative via session_state_changed events. No polling needed.' is reasonable, but there's an implicit assumption that session_state_changed events will always be delivered during foreground recovery. If the WebSocket didn't disconnect during iOS background (just suspended), no reconnect happens, and periodic sync is the only delivery path. Verify that the ConnectionRegistry periodic sync interval is short enough to avoid a stale running indicator on foreground return.

packages/client/src/protocol-parser.ts
```diff
@@ -292,14 +271,18 @@ export function parseServerMessage(
...
-      console.debug('[mitzo] session_state_changed', {
-        sessionId: msg.sessionId,
-        state: msg.state,
-        internalState: msg.internalState,
-      });
+    case 'session_state_changed': {
+      // Server-authoritative state — derive running from this event only
+      if (typeof msg.state === 'string' && VALID_CLIENT_STATES.has(msg.state)) {
+        result.messagesActions.push({
+          type: 'SESSION_STATE_CHANGED',
```
dimakis, line 278: 🔵 **unsafe_assumptions**: Non-string `msg.state` (undefined, null, number) is silently dropped without a warning. The `else if` only catches unknown *string* values. Adding an `else { console.warn(...) }` for non-string types would help diagnose serialization bugs or protocol mismatches that could otherwise be invisible. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -179,14 +158,14 @@ export function parseServerMessage(
     // ── v1 handshake events (kept for backward compat) ─────────────────────
 
     case 'reattached':
-      result.messagesActions.push({ type: 'SET_RUNNING', running: true });
+      // Running state from server's session_state_changed events
       callbacks.setWsRunning?.(poolKey, true);
       result.connectionUpdate = { status: 'connected' };
       if (msg.sessionId) callbacks.onSessionAssigned(msg.sessionId as string);
       break;
```
dimakis, line 165: 🔵 **style**: Comment says 'Running state restored via replayed session_state_changed events' but `handleSwitchSession` emits a *fresh* `session_state_changed` event immediately after `session_switched` — it is not a replay from the event store. Consider 'Running state delivered via a follow-up session_state_changed event' for accuracy. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -101,6 +102,10 @@ export interface ParseResult {
     | { type: 'workload_batch_updated'; items: WorkloadItem[]; created: number };
 }
 
+// ─── Constants ──────────────────────────────────────────────────────────────
```
dimakis, line 105: 🔵 **unsafe_assumptions**: `VALID_CLIENT_STATES` is defined as a manual `Set(['idle', 'running', 'requires_action'])` that must stay in sync with the `ClientSessionState` type. If a new state is added to the type, this set must also be updated manually. Consider deriving the set from a const array that also drives the type, e.g., `const CLIENT_STATES = ['idle', 'running', 'requires_action'] as const; type ClientSessionState = typeof CLIENT_STATES[number];` `[fixable]`

frontend/src/pages/ChatView.tsx
```diff
@@ -185,7 +185,9 @@ export function ChatView() {
 
   const handleStop = useCallback(() => {
     storeStopGeneration();
-    storeDispatchMessages({ type: 'SET_RUNNING', running: false });
+    // Optimistic update — server confirms via session_state_changed event,
```
dimakis, line 188: 🔵 **style**: The `handleStop` callback (with its optimistic update comment) is duplicated identically in both `ChatView.tsx` and `DesktopChatView.tsx`. This is pre-existing duplication, but since both files are modified in this PR, it may be worth extracting to a shared hook or utility. `[fixable]`

dimakis (commented):
## Centaur Review

Found **5** issue(s).


_Clean, well-structured migration from client-managed SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. All SET_RUNNING references properly replaced, test coverage is comprehensive, and the server-side state delivery (handleSwitchSession emitting session_state_changed, handleReconnect using getSessionState for stale cleanup, periodic sync for iOS gaps) is correct. Only minor robustness and style improvements suggested._

- 🔵 **unsafe_assumptions** (L278): Non-string `msg.state` (undefined, null, number) is silently dropped without a warning. The `else if` only catches unknown *string* values. Adding an `else { console.warn(...) }` for non-string types would help diagnose serialization bugs or protocol mismatches that could otherwise be invisible. `[fixable]`
- 🔵 **style** (L165): Comment says 'Running state restored via replayed session_state_changed events' but `handleSwitchSession` emits a *fresh* `session_state_changed` event immediately after `session_switched` — it is not a replay from the event store. Consider 'Running state delivered via a follow-up session_state_changed event' for accuracy. `[fixable]`
- 🔵 **unsafe_assumptions** (L105): `VALID_CLIENT_STATES` is defined as a manual `Set(['idle', 'running', 'requires_action'])` that must stay in sync with the `ClientSessionState` type. If a new state is added to the type, this set must also be updated manually. Consider deriving the set from a const array that also drives the type, e.g., `const CLIENT_STATES = ['idle', 'running', 'requires_action'] as const; type ClientSessionState = typeof CLIENT_STATES[number];` `[fixable]`


_Clean, well-structured migration from client-managed SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. All SET_RUNNING references properly replaced, test coverage is comprehensive, and the server-side state delivery (handleSwitchSession emitting session_state_changed, handleReconnect using getSessionState for stale cleanup, periodic sync for iOS gaps) is correct. Only minor robustness and style improvements suggested._

- 🔵 **missing_tests**: The new `session_close_ack` tests only cover `accepted: true` and `accepted: false`. Missing: test for when `session_close_ack` arrives with `accepted: false` and a `reason` field — the current handler silently ignores `reason`, so the user gets no feedback on why their close request was rejected. `[fixable]`


_Clean, well-structured migration from client-managed SET_RUNNING to server-authoritative SESSION_STATE_CHANGED. All SET_RUNNING references properly replaced, test coverage is comprehensive, and the server-side state delivery (handleSwitchSession emitting session_state_changed, handleReconnect using getSessionState for stale cleanup, periodic sync for iOS gaps) is correct. Only minor robustness and style improvements suggested._

- 🔵 **style** (L188): The `handleStop` callback (with its optimistic update comment) is duplicated identically in both `ChatView.tsx` and `DesktopChatView.tsx`. This is pre-existing duplication, but since both files are modified in this PR, it may be worth extracting to a shared hook or utility. `[fixable]`

PR #436, opened 2026-07-04
fix(task-board): fix session-task linkage and stuck workflow advancement

dimakis:
## Summary

- **Orphan detection was broken**: `getActiveSessionIds()` returned SDK `sessionId` UUIDs but `task.session_id` stores `clientId` — they never matched, so spawned tasks were never reclaimed when their session died
- **No bidirectional linkage**: spawned task sessions never got `taskContext` set, so the session overview couldn't show which task a session was working on
- **Stuck workflows**: when a spawned session ended (success or crash), nothing called `orchestrator.tick()` — the workflow stayed frozen on the completed task forever

## Test plan

- [x] Existing 47 orchestrator tests pass
- [x] New test: `reclaims spawned task sessions using clientId matching` — verifies `task:` prefixed clientIds are correctly matched in orphan detection
- [ ] Centaur review
- [ ] Manual: start a PR shepherd workflow, verify spawned session appears in session list with task linkage
- [ ] Manual: verify workflow advances after spawned session completes

### DIFF
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -912,6 +912,37 @@ describe('TaskOrchestrator', () => {
       });
     });
 
+    it('tick after spawn session ends reclaims orphaned task and advances', async () => {
+      const spawnSession = vi.fn().mockResolvedValue('task:spawned-1');
+      const deps = createTestDeps(store);
+      deps.spawnSession = spawnSession;
+      let activeClients = new Set(['task:spawned-1']);
+      deps.getActiveSessionIds = () => activeClients;
+      const orch = new TaskOrchestrator(deps);
+
+      const goal = store.create({ title: 'Goal' });
+      const t1 = store.create({ title: 'Spawn task', parentId: goal.id });
+      const t2 = store.create({ title: 'Next task', parentId: goal.id, sessionPolicy: 'reuse' });
+
+      orch.start(goal.id);
+
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalled();
+      });
+
+      // Simulate session completing the task
+      store.update(t1.id, { status: 'done' });
+      store.cascadeStatus(t1.id);
+
+      // Simulate .finally() wiring: session removed from registry, tick called
+      activeClients = new Set();
+      orch.onTaskCompleted(t1.id);
+
+      // t2 should now be active (workflow advanced)
+      expect(store.get(t2.id)!.status).toBe('active');
+      expect(orch.getStatus().activeTaskId).toBe(t2.id);
+    });
+
     it('reuse policy tasks use pinned session as before', () => {
       const spawnSession = vi.fn().mockResolvedValue('spawned-client-1');
       const deps = createTestDeps(store);
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -1015,6 +1015,45 @@ describe('TaskOrchestrator', () => {
       expect(spawnSession).toHaveBeenCalledTimes(1); // no re-dispatch
     });
 
+    it('does not advance old goal when orchestrator moved to a new goal', async () => {
+      const spawnSession = vi.fn().mockResolvedValue('task:spawned-1');
+      const deps = createTestDeps(store);
+      deps.spawnSession = spawnSession;
+      let activeClients = new Set(['task:spawned-1']);
+      deps.getActiveSessionIds = () => activeClients;
+      const orch = new TaskOrchestrator(deps);
+
+      const goal1 = store.create({ title: 'Goal 1' });
+      store.create({ title: 'Spawn task', parentId: goal1.id });
+
+      const goal2 = store.create({ title: 'Goal 2' });
+      const g2task = store.create({
+        title: 'Reuse task',
+        parentId: goal2.id,
+        sessionPolicy: 'reuse',
+      });
+
+      orch.start(goal1.id);
+
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(1);
+      });
+
+      // User switches to a new goal while spawned session is still running
+      orch.stop();
+      orch.start(goal2.id);
+      expect(orch.getStatus().goalId).toBe(goal2.id);
+      expect(orch.getStatus().activeTaskId).toBe(g2task.id);
+
+      // Old spawned session ends — tick should NOT interfere with goal2
+      activeClients = new Set();
+      orch.tick();
+
+      // Goal2 state should be unchanged
+      expect(orch.getStatus().goalId).toBe(goal2.id);
+      expect(orch.getStatus().activeTaskId).toBe(g2task.id);
+    });
+
     it('reuse policy tasks use pinned session as before', () => {
       const spawnSession = vi.fn().mockResolvedValue('spawned-client-1');
       const deps = createTestDeps(store);
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -977,6 +977,40 @@ describe('TaskOrchestrator', () => {
       });
     });
 
+    it('tick is a no-op when orchestrator is stopped before session ends', async () => {
+      // Edge case: orchestrator stopped (idle) while a spawned session is
+      // still in-flight. .finally() fires → tick() → returns early (guarded
+      // by state !== 'running'). No crash, no side effects.
+      const spawnSession = vi.fn().mockResolvedValue('task:spawned-1');
+      const deps = createTestDeps(store);
+      deps.spawnSession = spawnSession;
+      // Return spawned clientId while session is alive
+      let activeClients = new Set(['task:spawned-1']);
+      deps.getActiveSessionIds = () => activeClients;
+      const orch = new TaskOrchestrator(deps);
+
+      const goal = store.create({ title: 'Goal' });
+      const t1 = store.create({ title: 'Spawn task', parentId: goal.id });
+
+      orch.start(goal.id);
+
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(1);
+      });
+
+      // User stops the orchestrator while session is still running
+      orch.stop();
+      expect(orch.getStatus().state).toBe('idle');
+
+      // .finally() fires after session ends — tick is a no-op in idle state
+      activeClients = new Set();
+      orch.tick();
+
+      expect(orch.getStatus().state).toBe('idle');
+      expect(store.get(t1.id)?.status).toBe('active'); // unchanged
+      expect(spawnSession).toHaveBeenCalledTimes(1); // no re-dispatch
+    });
+
     it('reuse policy tasks use pinned session as before', () => {
       const spawnSession = vi.fn().mockResolvedValue('spawned-client-1');
       const deps = createTestDeps(store);
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -967,6 +967,10 @@ describe('TaskOrchestrator', () => {
       // Orchestrator should be paused (all tasks active, none pending)
       expect(orch.getStatus().state).toBe('paused');
 
+      // Verify the task is still active (pre-condition for orphan reclaim)
+      const task = store.getChildren(goal.id)[0];
+      expect(store.get(task.id)!.status).toBe('active');
+
       // .finally() fires: session removed from registry, resume() called
       activeClients = new Set();
       orch.resume();
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -912,7 +912,7 @@ describe('TaskOrchestrator', () => {
       });
     });
 
-    it('tick after spawn session ends reclaims orphaned task and advances', async () => {
+    it('tick after spawn session completes advances workflow', async () => {
       const spawnSession = vi.fn().mockResolvedValue('task:spawned-1');
       const deps = createTestDeps(store);
       deps.spawnSession = spawnSession;
@@ -930,17 +930,51 @@ describe('TaskOrchestrator', () => {
         expect(spawnSession).toHaveBeenCalled();
       });
 
-      // Simulate session completing the task
+      // Agent completed the task during the session
       store.update(t1.id, { status: 'done' });
       store.cascadeStatus(t1.id);
 
-      // Simulate .finally() wiring: session removed from registry, tick called
+      // .finally() fires: session removed from registry, tick() called directly
       activeClients = new Set();
-      orch.onTaskCompleted(t1.id);
+      orch.tick();
 
       // t2 should now be active (workflow advanced)
       expect(store.get(t2.id)!.status).toBe('active');
       expect(orch.getStatus().activeTaskId).toBe(t2.id);
+    });
+
+    it('resume after spawn session dies reclaims unfinished task', async () => {
+      // Primary scenario for .finally(): session crashes without calling
+      // TaskComplete. The task stays active but the session is gone.
+      // After all tasks are spawned, the orchestrator pauses (no pending left).
+      // .finally() calls resume() → tick() → orphan reclaim → re-dispatch.
+      const spawnSession = vi.fn().mockResolvedValue('task:spawned-1');
+      const deps = createTestDeps(store);
+      deps.spawnSession = spawnSession;
+      let activeClients = new Set(['task:spawned-1']);
+      deps.getActiveSessionIds = () => activeClients;
+      const orch = new TaskOrchestrator(deps);
+
+      const goal = store.create({ title: 'Goal' });
+      store.create({ title: 'Spawn task', parentId: goal.id });
+
+      orch.start(goal.id);
+
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(1);
+      });
+
+      // Orchestrator should be paused (all tasks active, none pending)
+      expect(orch.getStatus().state).toBe('paused');
+
+      // .finally() fires: session removed from registry, resume() called
+      activeClients = new Set();
+      orch.resume();
+
+      // Orphan reclaim detected the dead session and re-dispatched the task
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalledTimes(2);
+      });
     });
 
     it('reuse policy tasks use pinned session as before', () => {
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -967,9 +967,10 @@ describe('TaskOrchestrator', () => {
       // Orchestrator should be paused (all tasks active, none pending)
       expect(orch.getStatus().state).toBe('paused');
 
-      // Verify the task is still active (pre-condition for orphan reclaim)
+      // Verify the task is active and has sessionId set (pre-condition for orphan reclaim)
       const task = store.getChildren(goal.id)[0];
       expect(store.get(task.id)!.status).toBe('active');
+      expect(store.get(task.id)!.sessionId).toBe('task:spawned-1');
 
       // .finally() fires: session removed from registry, resume() called
       activeClients = new Set();
--- a/server/__tests__/task-orchestrator.test.ts
+++ b/server/__tests__/task-orchestrator.test.ts
@@ -312,18 +312,19 @@ describe('TaskOrchestrator', () => {
 
   describe('orphan detection', () => {
     it('reclaims orphaned tasks during tick', () => {
-      // Create deps with getActiveSessionIds
+      // getActiveSessionIds returns clientIds (not SDK sessionIds)
+      // to match what setSessionId stores on tasks
       const depsWithOrphan = createTestDeps(store);
-      depsWithOrphan.getActiveSessionIds = () => new Set(['alive-session']);
+      depsWithOrphan.getActiveSessionIds = () => new Set(['alive-client']);
       const orch = new TaskOrchestrator(depsWithOrphan);
 
       const goal = store.create({ title: 'Goal' });
       const c1 = store.create({ title: 'Orphan', parentId: goal.id });
       store.create({ title: 'Next', parentId: goal.id });
 
-      // Simulate c1 assigned to dead session
+      // Simulate c1 assigned to dead session (clientId not in active set)
       store.update(c1.id, { status: 'active' });
-      store.setSessionId(c1.id, 'dead-session');
+      store.setSessionId(c1.id, 'dead-client');
 
       orch.start(goal.id);
 
@@ -335,7 +336,7 @@ describe('TaskOrchestrator', () => {
 
     it('does not reclaim tasks with alive sessions', () => {
       const depsWithOrphan = createTestDeps(store);
-      depsWithOrphan.getActiveSessionIds = () => new Set(['alive-session']);
+      depsWithOrphan.getActiveSessionIds = () => new Set(['alive-client']);
       const orch = new TaskOrchestrator(depsWithOrphan);
 
       const goal = store.create({ title: 'Goal' });
@@ -343,12 +344,41 @@ describe('TaskOrchestrator', () => {
       const c2 = store.create({ title: 'Next', parentId: goal.id });
 
       store.update(c1.id, { status: 'active' });
-      store.setSessionId(c1.id, 'alive-session');
+      store.setSessionId(c1.id, 'alive-client');
 
       orch.start(goal.id);
 
       // c1 is alive, so tick should skip it and pick c2
       expect(orch.getStatus().activeTaskId).toBe(c2.id);
+    });
+
+    it('reclaims spawned task sessions using clientId matching', () => {
+      // Simulates the real scenario: task.session_id stores clientId
+      // (e.g. 'task:abc123'), active set contains clientIds from registry
+      const depsWithOrphan = createTestDeps(store);
+      depsWithOrphan.getActiveSessionIds = () => new Set(['task:alive-wt']);
+      const orch = new TaskOrchestrator(depsWithOrphan);
+
+      const goal = store.create({ title: 'Goal' });
+      const orphan = store.create({ title: 'Dead spawn', parentId: goal.id });
+      const alive = store.create({ title: 'Alive spawn', parentId: goal.id });
+      store.create({ title: 'Pending', parentId: goal.id });
+
+      // Orphan: session ended, clientId no longer in registry
+      store.update(orphan.id, { status: 'active' });
+      store.setSessionId(orphan.id, 'task:dead-wt');
+
+      // Alive: session still running
+      store.update(alive.id, { status: 'active' });
+      store.setSessionId(alive.id, 'task:alive-wt');
+
+      orch.start(goal.id);
+
+      // Orphan should be reclaimed and re-dispatched
+      expect(store.get(orphan.id)!.sessionId).toBeNull();
+      // Alive should NOT be reclaimed
+      expect(store.get(alive.id)!.sessionId).toBe('task:alive-wt');
+      expect(store.get(alive.id)!.status).toBe('active');
     });
   });
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -235,15 +235,16 @@ const orchestrator = new TaskOrchestrator({
     sseRegistry.broadcast('task_state', data);
   },
   getActiveSessionIds: () => {
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
     const ids = new Set<string>();
     for (const [clientId] of registry.entries()) {
-      const session = registry.get(clientId);
-      if (session?.sessionId) ids.add(session.sessionId);
+      ids.add(clientId);
     }
     return ids;
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
-    const clientId = `headless:${generateWtId()}`;
+    const clientId = `task:${generateWtId()}`;
 
     try {
       const transport = new NullTransport();
@@ -257,16 +258,33 @@ const orchestrator = new TaskOrchestrator({
         telosTaskId: goalId,
         taskContext: { currentTaskId: taskId, goalId },
         onSessionResolved: (sessionId) => {
-          log.info('spawned headless session resolved', { taskId, sessionId, clientId });
+          log.info('task session resolved', { taskId, sessionId, clientId });
           sseRegistry.broadcast('sessions_changed', {});
         },
-      }).catch((err) => {
-        log.error('spawned session failed', {
-          taskId,
-          clientId,
-          error: (err as Error).message,
+      })
+        .catch((err) => {
+          log.error('task session failed', {
+            taskId,
+            clientId,
+            error: (err as Error).message,
+          });
+        })
+        .finally(() => {
+          // Session ended (success or failure) — clean up registry entry
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          if (registry.get(clientId)) {
+            registry.remove(clientId);
+          }
+          orchestratorRef?.tick();
         });
-      });
+
+      // Link session to task for session overview visibility.
+      // register() is synchronous inside startChat, so the session exists here.
+      const session = registry.get(clientId);
+      if (session) {
+        session.taskContext = { currentTaskId: taskId, goalId };
+      }
 
       return clientId;
     } catch (err) {
--- a/server/index.ts
+++ b/server/index.ts
@@ -273,10 +273,10 @@ const orchestrator = new TaskOrchestrator({
           // Only advance if the orchestrator is still on the same goal.
           // A user may have paused manually or started a new goal while
           // this session was in-flight — don't override that.
-          const status = orchestratorRef?.getStatus();
-          if (status && status.goalId === goalId) {
-            const loopState = status.state;
-            if (loopState === 'paused') {
+          if (!orchestratorRef) return;
+          const status = orchestratorRef.getStatus();
+          if (status.goalId === goalId) {
+            if (status.state === 'paused') {
               orchestratorRef.resume();
             } else {
               orchestratorRef.tick();
--- a/server/index.ts
+++ b/server/index.ts
@@ -270,7 +270,16 @@ const orchestrator = new TaskOrchestrator({
           // (no WS close to trigger normal removal) and advance the
           // orchestrator so orphan reclaim picks up unfinished tasks.
           registry.remove(clientId);
-          orchestratorRef?.tick();
+          // After all tasks are spawned, the orchestrator pauses (no pending
+          // tasks left). tick() is a no-op when paused, so resume() is needed
+          // to re-enter the running state and trigger orphan reclaim.
+          if (orchestratorRef) {
+            if (orchestratorRef.getStatus().state === 'paused') {
+              orchestratorRef.resume();
+            } else {
+              orchestratorRef.tick();
+            }
+          }
         });
 
       return clientId;
--- a/server/index.ts
+++ b/server/index.ts
@@ -274,18 +274,26 @@ const orchestrator = new TaskOrchestrator({
           // A user may have paused manually or started a new goal while
           // this session was in-flight — don't override that.
           if (!orchestratorRef) return;
-          const status = orchestratorRef.getStatus();
-          if (status.goalId === goalId) {
-            if (status.state === 'paused') {
-              orchestratorRef.resume();
+          try {
+            const status = orchestratorRef.getStatus();
+            if (status.goalId === goalId) {
+              if (status.state === 'paused') {
+                orchestratorRef.resume();
+              } else {
+                orchestratorRef.tick();
+              }
             } else {
-              orchestratorRef.tick();
+              log.info('task session ended after loop stopped or goal changed', {
+                taskId,
+                clientId,
+                goalId,
+              });
             }
-          } else {
-            log.info('task session ended after loop stopped or goal changed', {
+          } catch (err: unknown) {
+            log.error('orchestrator advance failed after task session end', {
               taskId,
               clientId,
-              goalId,
+              error: err instanceof Error ? err.message : String(err),
             });
           }
         });
--- a/server/index.ts
+++ b/server/index.ts
@@ -237,9 +237,7 @@ const orchestrator = new TaskOrchestrator({
   getActiveSessionIds: () => {
     // Return clientIds — task.session_id stores clientId, not SDK sessionId.
     // Using sessionId here caused orphan detection to never match spawned tasks.
-    const ids = new Set<string>();
-    for (const [clientId] of registry.entries()) ids.add(clientId);
-    return ids;
+    return new Set(registry.keys());
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
     const clientId = `task:${generateWtId()}`;
@@ -276,11 +274,18 @@ const orchestrator = new TaskOrchestrator({
           // A user may have paused manually or started a new goal while
           // this session was in-flight — don't override that.
           if (orchestratorRef && orchestratorRef.getStatus().goalId === goalId) {
-            if (orchestratorRef.getStatus().state === 'paused') {
+            const loopState = orchestratorRef.getStatus().state;
+            if (loopState === 'paused') {
               orchestratorRef.resume();
             } else {
               orchestratorRef.tick();
             }
+          } else {
+            log.info('task session ended after loop stopped or goal changed', {
+              taskId,
+              clientId,
+              goalId,
+            });
           }
         });
 
--- a/server/index.ts
+++ b/server/index.ts
@@ -273,8 +273,9 @@ const orchestrator = new TaskOrchestrator({
           // Only advance if the orchestrator is still on the same goal.
           // A user may have paused manually or started a new goal while
           // this session was in-flight — don't override that.
-          if (orchestratorRef && orchestratorRef.getStatus().goalId === goalId) {
-            const loopState = orchestratorRef.getStatus().state;
+          const status = orchestratorRef?.getStatus();
+          if (status && status.goalId === goalId) {
+            const loopState = status.state;
             if (loopState === 'paused') {
               orchestratorRef.resume();
             } else {
--- a/server/index.ts
+++ b/server/index.ts
@@ -237,7 +237,9 @@ const orchestrator = new TaskOrchestrator({
   getActiveSessionIds: () => {
     // Return clientIds — task.session_id stores clientId, not SDK sessionId.
     // Using sessionId here caused orphan detection to never match spawned tasks.
-    return new Set(Array.from(registry.entries(), ([clientId]) => clientId));
+    const ids = new Set<string>();
+    for (const [clientId] of registry.entries()) ids.add(clientId);
+    return ids;
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
     const clientId = `task:${generateWtId()}`;
@@ -270,10 +272,10 @@ const orchestrator = new TaskOrchestrator({
           // (no WS close to trigger normal removal) and advance the
           // orchestrator so orphan reclaim picks up unfinished tasks.
           registry.remove(clientId);
-          // After all tasks are spawned, the orchestrator pauses (no pending
-          // tasks left). tick() is a no-op when paused, so resume() is needed
-          // to re-enter the running state and trigger orphan reclaim.
-          if (orchestratorRef) {
+          // Only advance if the orchestrator is still on the same goal.
+          // A user may have paused manually or started a new goal while
+          // this session was in-flight — don't override that.
+          if (orchestratorRef && orchestratorRef.getStatus().goalId === goalId) {
             if (orchestratorRef.getStatus().state === 'paused') {
               orchestratorRef.resume();
             } else {
--- a/server/index.ts
+++ b/server/index.ts
@@ -237,11 +237,7 @@ const orchestrator = new TaskOrchestrator({
   getActiveSessionIds: () => {
     // Return clientIds — task.session_id stores clientId, not SDK sessionId.
     // Using sessionId here caused orphan detection to never match spawned tasks.
-    const ids = new Set<string>();
-    for (const [clientId] of registry.entries()) {
-      ids.add(clientId);
-    }
-    return ids;
+    return new Set(Array.from(registry.entries(), ([clientId]) => clientId));
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
     const clientId = `task:${generateWtId()}`;
@@ -273,18 +269,9 @@ const orchestrator = new TaskOrchestrator({
           // Session ended (success or failure) — clean up registry entry
           // (no WS close to trigger normal removal) and advance the
           // orchestrator so orphan reclaim picks up unfinished tasks.
-          if (registry.get(clientId)) {
-            registry.remove(clientId);
-          }
+          registry.remove(clientId);
           orchestratorRef?.tick();
         });
-
-      // Link session to task for session overview visibility.
-      // register() is synchronous inside startChat, so the session exists here.
-      const session = registry.get(clientId);
-      if (session) {
-        session.taskContext = { currentTaskId: taskId, goalId };
-      }
 
       return clientId;
     } catch (err) {
--- a/packages/harness/src/session-registry.ts
+++ b/packages/harness/src/session-registry.ts
@@ -155,6 +155,10 @@ export class SessionRegistry {
     return this.sessions.get(clientId);
   }
 
+  keys(): IterableIterator<string> {
+    return this.sessions.keys();
+  }
+
   entries(): IterableIterator<[string, ManagedSession]> {
     return this.sessions.entries();
   }


server/index.ts
```diff
@@ -256,16 +257,33 @@ const orchestrator = new TaskOrchestrator({
...
+            taskId,
+            clientId,
+            error: (err as Error).message,
+          });
+        })
+        .finally(() => {
+          // Session ended (success or failure) — clean up registry entry
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          if (registry.get(clientId)) {
```
dimakis: 🔵 **style**: The `if (registry.get(clientId))` guard before `registry.remove(clientId)` is redundant — `remove()` already handles missing entries gracefully (checks `if (session)` internally). The guard adds visual noise without preventing anything. Consider just calling `registry.remove(clientId)` directly. `[fixable]`

server/index.ts
```diff
@@ -235,15 +235,16 @@ const orchestrator = new TaskOrchestrator({
     sseRegistry.broadcast('task_state', data);
   },
   getActiveSessionIds: () => {
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
```
dimakis, line 239: 🔵 **style**: The loop `for (const [clientId] of registry.entries()) { ids.add(clientId); }` destructures tuples just to get the key. If SessionRegistry exposes a `keys()` iterator or a `size`/`clientIds` accessor, that would be cleaner. Minor — current form is clear enough. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


### `server/index.ts`
_Correct fix for a real bug — tasks stored clientIds but orphan detection compared against SDK sessionIds, so spawned tasks were never matched. The .finally() cleanup for NullTransport sessions is the right approach. Two minor style nits, one suggestion for wiring-level test coverage._

- 🔵 **style** (L275): The `if (registry.get(clientId))` guard before `registry.remove(clientId)` is redundant — `remove()` already handles missing entries gracefully (checks `if (session)` internally). The guard adds visual noise without preventing anything. Consider just calling `registry.remove(clientId)` directly. `[fixable]`
- 🔵 **style** (L239): The loop `for (const [clientId] of registry.entries()) { ids.add(clientId); }` destructures tuples just to get the key. If SessionRegistry exposes a `keys()` iterator or a `size`/`clientIds` accessor, that would be cleaner. Minor — current form is clear enough. `[fixable]`


_Correct fix for a real bug — tasks stored clientIds but orphan detection compared against SDK sessionIds, so spawned tasks were never matched. The .finally() cleanup for NullTransport sessions is the right approach. Two minor style nits, one suggestion for wiring-level test coverage._

- 🔵 **missing_tests**: The `.finally()` cleanup in spawnSession (registry removal + tick trigger) is critical to fixing stuck workflows but is only covered by integration-level wiring in index.ts, not by a unit test. The orchestrator tests validate orphan detection logic itself, which is the important part, but a test confirming that a completed spawn removes its registry entry and triggers tick would guard against regressions in the wiring.

dimakis:
## Centaur Review

_Correct fix — the clientId/sessionId mismatch in orphan detection was a real bug that caused spawned tasks to never be reclaimed, and the missing .finally() cleanup left NullTransport sessions stranded in the registry. The fix is well-scoped with good test additions; minor suggestions for test coverage of the cleanup path and prefix consistency._

- 🔵 **missing_tests**: The `.finally()` cleanup in `index.ts` (registry removal + re-tick) is untested. This is the primary mechanism that unblocks stuck workflows — when a spawned task session ends without the agent calling TaskComplete, registry cleanup + tick triggers orphan reclaim. A test exercising the spawnSession→finally→tick→reclaim flow would guard the core fix. Admittedly this is integration-level wiring in index.ts that's harder to unit test, but a test with a mock spawnSession that resolves immediately and verifies orphan reclaim would add confidence. `[fixable]`


### `server/index.ts`
_Correct fix — the clientId/sessionId mismatch in orphan detection was a real bug that caused spawned tasks to never be reclaimed, and the missing .finally() cleanup left NullTransport sessions stranded in the registry. The fix is well-scoped with good test additions; minor suggestions for test coverage of the cleanup path and prefix consistency._

- 🔵 **style** (L241): The loop `for (const [clientId] of registry.entries()) { ids.add(clientId); }` can be simplified to `return new Set(Array.from(registry.entries(), ([id]) => id))` or even `return new Set(registry.keys())` if `ConnectionRegistry` exposes a `keys()` method. Minor — the current form is clear enough. `[fixable]`


### `server/app.ts`
_Correct fix — the clientId/sessionId mismatch in orphan detection was a real bug that caused spawned tasks to never be reclaimed, and the missing .finally() cleanup left NullTransport sessions stranded in the registry. The fix is well-scoped with good test additions; minor suggestions for test coverage of the cleanup path and prefix consistency._

- 🟡 **regressions** (L501): The REST API session-creation path in `app.ts` still uses the `headless:` prefix for its clientIds. These sessions are now included in `getActiveSessionIds()` (which returns all clientIds from the registry). This is correct behavior — API-created headless sessions should not be orphan-reclaimed. However, the inconsistent prefix naming (`headless:` vs `task:`) makes it less clear which sessions are orchestrator-spawned vs API-spawned. Consider renaming the `app.ts` prefix too (or documenting the distinction), since both are NullTransport-based headless sessions that differ only in origin. `[fixable]`

server/__tests__/task-orchestrator.test.ts
```diff
@@ -839,6 +869,41 @@ describe('TaskOrchestrator', () => {
       expect(deps.setTaskContext).not.toHaveBeenCalled();
```
dimakis, line 869: 🔵 **missing_tests**: The new test 'tick after spawn session ends reclaims orphaned task and advances' covers the happy path (task completes, then session ends). There's no test for the failure path: a spawned session dies without completing the task. In production, `.finally()` fires → `registry.remove()` → `tick()` → orphan reclaim resets the still-active task to pending. This is the primary scenario the `.finally()` wiring exists to handle and should be tested explicitly. `[fixable]`

server/__tests__/task-orchestrator.test.ts
```diff
@@ -839,6 +869,41 @@ describe('TaskOrchestrator', () => {
...
+        expect(spawnSession).toHaveBeenCalled();
+      });
+
+      // Simulate session completing the task
+      store.update(t1.id, { status: 'done' });
+      store.cascadeStatus(t1.id);
+
+      // Simulate .finally() wiring: session removed from registry, tick called
+      activeClients = new Set();
+      orch.onTaskCompleted(t1.id);
```
dimakis: 🔵 **style**: The test comment says 'Simulate .finally() wiring' but calls `orch.onTaskCompleted(t1.id)`, while the actual `.finally()` block calls `orchestratorRef?.tick()`. These are different entry points: `onTaskCompleted` has a `state !== 'running'` guard and resets `spawnDepth` before calling `tick()`. The test works because tick is idempotent, but it doesn't precisely model the production code path. Using `orch.tick()` directly (after manually clearing the active set) would be a more faithful simulation. `[fixable]`

server/index.ts
```diff
@@ -256,16 +253,31 @@ const orchestrator = new TaskOrchestrator({
...
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          registry.remove(clientId);
+          orchestratorRef?.tick();
         });
-      });
+
+      // Link session to task for session overview visibility.
+      // register() is synchronous inside startChat, so the session exists here.
+      const session = registry.get(clientId);
```
dimakis: 🟡 **unsafe_assumptions**: The comment says 'register() is synchronous inside startChat, so the session exists here', but `startChat` is fire-and-forget (not awaited). This depends on `startChat` synchronously calling `registry.register()` before its first `await`. If `startChat` is ever refactored to defer registration (e.g. behind an async worktree check), `registry.get(clientId)` will silently return undefined, the `if (session)` guard will skip silently, and `taskContext` won't be set — breaking session overview visibility with no error logged. Consider logging when `session` is null, or moving this assignment into `startChat` itself via the options bag. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid bug fix — the clientId/sessionId mismatch in orphan detection was a clear correctness issue, and the `.finally()` wiring properly handles spawned session lifecycle. Minor gaps in test coverage for the failure path and a fragile synchrony assumption on taskContext assignment._

- 🔵 **missing_tests** (L869): The new test 'tick after spawn session ends reclaims orphaned task and advances' covers the happy path (task completes, then session ends). There's no test for the failure path: a spawned session dies without completing the task. In production, `.finally()` fires → `registry.remove()` → `tick()` → orphan reclaim resets the still-active task to pending. This is the primary scenario the `.finally()` wiring exists to handle and should be tested explicitly. `[fixable]`
- 🔵 **style** (L900): The test comment says 'Simulate .finally() wiring' but calls `orch.onTaskCompleted(t1.id)`, while the actual `.finally()` block calls `orchestratorRef?.tick()`. These are different entry points: `onTaskCompleted` has a `state !== 'running'` guard and resets `spawnDepth` before calling `tick()`. The test works because tick is idempotent, but it doesn't precisely model the production code path. Using `orch.tick()` directly (after manually clearing the active set) would be a more faithful simulation. `[fixable]`


### `server/index.ts`
_Solid bug fix — the clientId/sessionId mismatch in orphan detection was a clear correctness issue, and the `.finally()` wiring properly handles spawned session lifecycle. Minor gaps in test coverage for the failure path and a fragile synchrony assumption on taskContext assignment._

- 🟡 **unsafe_assumptions** (L277): The comment says 'register() is synchronous inside startChat, so the session exists here', but `startChat` is fire-and-forget (not awaited). This depends on `startChat` synchronously calling `registry.register()` before its first `await`. If `startChat` is ever refactored to defer registration (e.g. behind an async worktree check), `registry.get(clientId)` will silently return undefined, the `if (session)` guard will skip silently, and `taskContext` won't be set — breaking session overview visibility with no error logged. Consider logging when `session` is null, or moving this assignment into `startChat` itself via the options bag. `[fixable]`

server/__tests__/task-orchestrator.test.ts
dimakis, line 869: 🟡 **missing_tests**: The new 'tick after spawn session ends' test simulates the .finally() wiring using orch.onTaskCompleted(t1.id), but the actual .finally() in index.ts calls orchestratorRef?.tick() directly. onTaskCompleted resets spawnDepth to 0 (line 197 of task-orchestrator.ts) while tick() does not, making the test more permissive than reality. The test should call orch.tick() directly (after manually updating store state) to match the real code path. `[fixable]`

server/index.ts
```diff
@@ -256,16 +253,31 @@ const orchestrator = new TaskOrchestrator({
...
+        })
+        .finally(() => {
+          // Session ended (success or failure) — clean up registry entry
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          registry.remove(clientId);
+          orchestratorRef?.tick();
         });
-      });
+
```
dimakis, line 300: 🔵 **unsafe_assumptions**: The comment 'register() is synchronous inside startChat, so the session exists here' is correct today (registry.register at chat.ts:848 runs before any await in _startChatInner), but this is a fragile implicit contract. If a future refactor adds an await before register() (e.g. async validation), the registry.get(clientId) here would silently return undefined and skip the taskContext assignment. Consider moving the taskContext assignment into the spawnSession callback after setSessionId in task-orchestrator.ts, or documenting the contract at the register() call site. `[fixable]`

server/index.ts
```diff
@@ -235,15 +235,12 @@ const orchestrator = new TaskOrchestrator({
...
-    const ids = new Set<string>();
-    for (const [clientId] of registry.entries()) {
-      const session = registry.get(clientId);
-      if (session?.sessionId) ids.add(session.sessionId);
-    }
-    return ids;
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
+    return new Set(Array.from(registry.entries(), ([clientId]) => clientId));
   },
```
dimakis, line 241: 🔵 **style**: Array.from(registry.entries(), ([clientId]) => clientId) allocates an intermediate array just to feed it into a Set. A clearer alternative: new Set(registry.keys()) if SessionRegistry exposes a keys() method, or the existing loop pattern (for...of with ids.add) which avoids the allocation. `[fixable]`

dimakis (commented):
## Centaur Review

_Correct fix for a real bug (clientId/sessionId mismatch broke orphan detection for spawned tasks), with good test coverage of the happy path, but missing test coverage for the failure-path orphan reclaim scenario that motivated the fix._

- 🟡 **missing_tests** (L869): The new 'tick after spawn session ends' test simulates the .finally() wiring using orch.onTaskCompleted(t1.id), but the actual .finally() in index.ts calls orchestratorRef?.tick() directly. onTaskCompleted resets spawnDepth to 0 (line 197 of task-orchestrator.ts) while tick() does not, making the test more permissive than reality. The test should call orch.tick() directly (after manually updating store state) to match the real code path. `[fixable]`
- 🟡 **missing_tests**: No test covers the failure path — the primary scenario this fix addresses. When a spawned session fails (.catch branch) without calling TaskComplete, the .finally() fires, registry.remove() makes the task orphaned, and tick() should reclaim it back to 'pending'. This scenario (task still 'active' when session ends) is distinct from the tested happy path (task already 'done' when session ends) and exercises the orphan detection that was previously broken. `[fixable]`


### `server/index.ts`
_Correct fix for a real bug (clientId/sessionId mismatch broke orphan detection for spawned tasks), with good test coverage of the happy path, but missing test coverage for the failure-path orphan reclaim scenario that motivated the fix._

- 🔵 **unsafe_assumptions** (L274): The comment 'register() is synchronous inside startChat, so the session exists here' is correct today (registry.register at chat.ts:848 runs before any await in _startChatInner), but this is a fragile implicit contract. If a future refactor adds an await before register() (e.g. async validation), the registry.get(clientId) here would silently return undefined and skip the taskContext assignment. Consider moving the taskContext assignment into the spawnSession callback after setSessionId in task-orchestrator.ts, or documenting the contract at the register() call site. `[fixable]`
- 🔵 **style** (L241): Array.from(registry.entries(), ([clientId]) => clientId) allocates an intermediate array just to feed it into a Set. A clearer alternative: new Set(registry.keys()) if SessionRegistry exposes a keys() method, or the existing loop pattern (for...of with ids.add) which avoids the allocation. `[fixable]`

server/__tests__/task-orchestrator.test.ts
```diff
@@ -839,6 +869,74 @@ describe('TaskOrchestrator', () => {
...
+
+      orch.start(goal.id);
+
+      await vi.waitFor(() => {
+        expect(spawnSession).toHaveBeenCalled();
+      });
+
+      // Agent completed the task during the session
+      store.update(t1.id, { status: 'done' });
+      store.cascadeStatus(t1.id);
```
dimakis, line 935: 🔵 **missing_tests**: The 'tick after spawn session completes advances workflow' test doesn't assert that `spawnSession` was called exactly once. Adding `expect(spawnSession).toHaveBeenCalledTimes(1)` after verifying t2 is active would prove t2 used the reuse path (not spawn), which is the key workflow-advancement contract being tested. `[fixable]`

server/index.ts
```diff
@@ -235,15 +235,12 @@ const orchestrator = new TaskOrchestrator({
...
   getActiveSessionIds: () => {
-    const ids = new Set<string>();
-    for (const [clientId] of registry.entries()) {
-      const session = registry.get(clientId);
-      if (session?.sessionId) ids.add(session.sessionId);
-    }
-    return ids;
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
+    return new Set(Array.from(registry.entries(), ([clientId]) => clientId));
```
dimakis: 🔵 **style**: Minor: `Array.from(registry.entries(), ([clientId]) => clientId)` iterates key-value pairs but only uses keys. Consider adding a `keys()` method to `SessionRegistry` (it wraps a Map) for cleaner intent, or at minimum a comment explaining why `entries()` is used (because no `keys()` exists). `[fixable]`

server/index.ts
```diff
@@ -256,16 +253,46 @@ const orchestrator = new TaskOrchestrator({
...
+            }
+          }
+        });
+
+      // Link session to task for session overview visibility.
+      // register() runs synchronously before startChat's first await,
+      // so the session exists here. Log if that assumption ever breaks.
+      const session = registry.get(clientId);
+      if (session) {
+        session.taskContext = { currentTaskId: taskId, goalId };
```
dimakis: 🔵 **unsafe_assumptions**: The `taskContext` linkage assumes `registry.get(clientId)` succeeds because `register()` runs synchronously before `startChat`'s first `await`. The assumption is currently valid (verified: no `await` before `register()` in `_startChatInner`), and the `log.warn` fallback is good defensive coding. However, this is a fragile coupling — if someone adds an `await` before `register()` in `chat.ts`, this breaks silently. Consider adding a brief code comment in `chat.ts` near `register()` noting that callers depend on its synchronous execution relative to the function's return. `[fixable]`

server/index.ts
```diff
@@ -235,15 +235,12 @@ const orchestrator = new TaskOrchestrator({
...
-      if (session?.sessionId) ids.add(session.sessionId);
-    }
-    return ids;
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
+    return new Set(Array.from(registry.entries(), ([clientId]) => clientId));
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
-    const clientId = `headless:${generateWtId()}`;
+    const clientId = `task:${generateWtId()}`;
```
dimakis, line 243: 🔵 **style**: The `task:` prefix for spawned sessions is a good improvement over `headless:`, but `app.ts:501` still uses `headless:` for externally-created headless sessions (POST /api/sessions). The inconsistency is intentional (different use cases), but worth noting that both code paths create NullTransport sessions — only the task path gets `.finally()` cleanup. If external headless sessions also need orphan-aware cleanup in the future, the same pattern would be needed there.

dimakis (commented):
## Centaur Review

Found **4** issue(s).


_Correct fix for a real bug — `getActiveSessionIds` was returning SDK sessionIds but `setSessionId` stores clientIds, causing orphan detection to silently skip all spawned tasks. The `.finally()` wiring and `taskContext` linkage are well-structured. All findings are suggestions; no blocking issues._

- 🔵 **missing_tests** (L895): The 'tick after spawn session completes advances workflow' test doesn't assert that `spawnSession` was called exactly once. Adding `expect(spawnSession).toHaveBeenCalledTimes(1)` after verifying t2 is active would prove t2 used the reuse path (not spawn), which is the key workflow-advancement contract being tested. `[fixable]`


### `server/index.ts`
_Correct fix for a real bug — `getActiveSessionIds` was returning SDK sessionIds but `setSessionId` stores clientIds, causing orphan detection to silently skip all spawned tasks. The `.finally()` wiring and `taskContext` linkage are well-structured. All findings are suggestions; no blocking issues._

- 🔵 **style** (L240): Minor: `Array.from(registry.entries(), ([clientId]) => clientId)` iterates key-value pairs but only uses keys. Consider adding a `keys()` method to `SessionRegistry` (it wraps a Map) for cleaner intent, or at minimum a comment explaining why `entries()` is used (because no `keys()` exists). `[fixable]`
- 🔵 **unsafe_assumptions** (L289): The `taskContext` linkage assumes `registry.get(clientId)` succeeds because `register()` runs synchronously before `startChat`'s first `await`. The assumption is currently valid (verified: no `await` before `register()` in `_startChatInner`), and the `log.warn` fallback is good defensive coding. However, this is a fragile coupling — if someone adds an `await` before `register()` in `chat.ts`, this breaks silently. Consider adding a brief code comment in `chat.ts` near `register()` noting that callers depend on its synchronous execution relative to the function's return. `[fixable]`
- 🔵 **style** (L243): The `task:` prefix for spawned sessions is a good improvement over `headless:`, but `app.ts:501` still uses `headless:` for externally-created headless sessions (POST /api/sessions). The inconsistency is intentional (different use cases), but worth noting that both code paths create NullTransport sessions — only the task path gets `.finally()` cleanup. If external headless sessions also need orphan-aware cleanup in the future, the same pattern would be needed there.

server/index.ts
```diff
@@ -256,16 +253,46 @@ const orchestrator = new TaskOrchestrator({
...
+        })
+        .finally(() => {
+          // Session ended (success or failure) — clean up registry entry
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          registry.remove(clientId);
+          // After all tasks are spawned, the orchestrator pauses (no pending
+          // tasks left). tick() is a no-op when paused, so resume() is needed
+          // to re-enter the running state and trigger orphan reclaim.
+          if (orchestratorRef) {
```
dimakis: 🟡 **unsafe_assumptions**: The `.finally()` handler resumes a paused orchestrator without checking if it's still on the same goal. If a user manually pauses the orchestrator (`/api/loop/pause`) while a spawned session is still in-flight, the `.finally()` will override the manual pause by calling `resume()`. The `spawnSession` closure already has `goalId` in scope — add a guard like `orchestratorRef.getStatus().goalId === goalId` to only advance the same goal's workflow, matching the existing guard pattern in the orchestrator's spawn `.then()` callback (task-orchestrator.ts:364). `[fixable]`

server/index.ts
dimakis: 🔵 **style**: `new Set(Array.from(registry.entries(), ([clientId]) => clientId))` allocates an intermediate array. The registry is a Map wrapper, so this could be simplified to `new Set(registry.sessions.keys())` — though that would require exposing `keys()` on SessionRegistry. Consider adding a `keys()` method to `SessionRegistry` for a cleaner API. Minor — current approach is correct, just slightly more allocation than needed. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid fix for the clientId-vs-sessionId mismatch in orphan detection and the missing spawn lifecycle cleanup. The `.finally()` handler should guard against stale goals to avoid overriding manual pauses._

- 🟡 **unsafe_assumptions** (L275): The `.finally()` handler resumes a paused orchestrator without checking if it's still on the same goal. If a user manually pauses the orchestrator (`/api/loop/pause`) while a spawned session is still in-flight, the `.finally()` will override the manual pause by calling `resume()`. The `spawnSession` closure already has `goalId` in scope — add a guard like `orchestratorRef.getStatus().goalId === goalId` to only advance the same goal's workflow, matching the existing guard pattern in the orchestrator's spawn `.then()` callback (task-orchestrator.ts:364). `[fixable]`
- 🔵 **style** (L240): `new Set(Array.from(registry.entries(), ([clientId]) => clientId))` allocates an intermediate array. The registry is a Map wrapper, so this could be simplified to `new Set(registry.sessions.keys())` — though that would require exposing `keys()` on SessionRegistry. Consider adding a `keys()` method to `SessionRegistry` for a cleaner API. Minor — current approach is correct, just slightly more allocation than needed. `[fixable]`


_Solid fix for the clientId-vs-sessionId mismatch in orphan detection and the missing spawn lifecycle cleanup. The `.finally()` handler should guard against stale goals to avoid overriding manual pauses._

- 🔵 **missing_tests**: No test covers the `.finally()` path when the orchestrator has been stopped (state='idle') before the spawned session finishes. In production, `tick()` returns immediately (guarded by `state !== 'running'`), so it's safe — but a test would document this behavior and prevent regressions if the guard changes. `[fixable]`

dimakis:
## Centaur Review

Found **3** issue(s).


_Correct fix for a real bug — getActiveSessionIds was returning SDK sessionIds while tasks store clientIds, causing orphan detection to never match. The .finally() handler for lifecycle cleanup is well-structured. Main suggestion: rename `getActiveSessionIds` → `getActiveClientIds` to prevent re-introduction of the semantic mismatch._

- 🔵 **style** (L37): The interface field `getActiveSessionIds` and its JSDoc ("Get active session IDs for orphan detection") now returns clientIds, not SDK sessionIds. The comment in `index.ts:238-239` explains this, but the contract-level name is misleading — a future contributor wiring a new implementation could reasonably return SDK sessionIds and re-introduce the bug this PR fixes. Consider renaming to `getActiveClientIds` (with a matching update to the orphan detection call site) to make the contract self-documenting. `[fixable]`


_Correct fix for a real bug — getActiveSessionIds was returning SDK sessionIds while tasks store clientIds, causing orphan detection to never match. The .finally() handler for lifecycle cleanup is well-structured. Main suggestion: rename `getActiveSessionIds` → `getActiveClientIds` to prevent re-introduction of the semantic mismatch._

- 🔵 **missing_tests**: No test covers the `.finally()` handler firing after the orchestrator has been `stop()`-ed (e.g., user manually stops the loop while spawned sessions are still running). The code is safe — `tick()` guards on `state !== 'running'` and `resume()` guards on `state !== 'paused'` — but a test would document that invariant and prevent regressions if the guards change. `[fixable]`


### `server/index.ts`
_Correct fix for a real bug — getActiveSessionIds was returning SDK sessionIds while tasks store clientIds, causing orphan detection to never match. The .finally() handler for lifecycle cleanup is well-structured. Main suggestion: rename `getActiveSessionIds` → `getActiveClientIds` to prevent re-introduction of the semantic mismatch._

- 🔵 **unsafe_assumptions** (L287): The comment says `register() runs synchronously before startChat's first await` — this is true today but is a coupling to startChat's internal implementation order. The fallback `log.warn` mitigates this well. Just flagging that if startChat is ever refactored to defer registration (e.g., after worktree creation), this would silently degrade. The logging is the right defense; no action needed.

server/index.ts
dimakis: 🔵 **unsafe_assumptions**: The `.finally()` handler reads `orchestratorRef.getStatus().state` and branches on `paused` vs other. If the orchestrator was `stop()`ped between spawn and session completion (user stopped the loop), calling `tick()` is harmless (returns early due to `state !== 'running'`), and `resume()` won't fire since state is `idle` not `paused`. This is safe — but a brief log line in the `else` branch of the `orchestratorRef` check (or when state is `idle`) would help debug "session ended after loop stopped" situations in production. `[fixable]`

server/index.ts
dimakis: 🔵 **style**: `new Set(Array.from(registry.entries(), ([clientId]) => clientId))` can be simplified to `new Set(registry.keys())` if `SessionRegistry` exposes a `keys()` method, or `new Set(Array.from(registry.entries()).map(([k]) => k))` isn't more readable. Current form works but `Array.from(iterable, mapFn)` with destructuring is slightly unusual. Consider adding a `keys()` method to `SessionRegistry` — it would make this intent obvious. `[fixable]`

server/__tests__/task-orchestrator.test.ts
```diff
@@ -839,6 +869,74 @@ describe('TaskOrchestrator', () => {
...
+      // .finally() fires: session removed from registry, tick() called directly
+      activeClients = new Set();
+      orch.tick();
+
+      // t2 should now be active (workflow advanced)
+      expect(store.get(t2.id)!.status).toBe('active');
+      expect(orch.getStatus().activeTaskId).toBe(t2.id);
+    });
+
+    it('resume after spawn session dies reclaims unfinished task', async () => {
```
dimakis, line 946: 🔵 **missing_tests**: The "resume after spawn session dies" test covers the crash → reclaim → re-dispatch path well, but doesn't verify the reclaimed task's status transitions (active → pending → active). Adding `expect(store.get(task.id)!.status).toBe('pending')` before the `resume()` call would document the pre-condition more clearly and guard against store-level regressions. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


### `server/index.ts`
_Solid bug fix — the root cause (clientId vs sessionId mismatch in orphan detection) is correctly identified and fixed, with good test coverage for the new `.finally()` lifecycle wiring. No correctness issues found._

- 🔵 **unsafe_assumptions** (L275): The `.finally()` handler reads `orchestratorRef.getStatus().state` and branches on `paused` vs other. If the orchestrator was `stop()`ped between spawn and session completion (user stopped the loop), calling `tick()` is harmless (returns early due to `state !== 'running'`), and `resume()` won't fire since state is `idle` not `paused`. This is safe — but a brief log line in the `else` branch of the `orchestratorRef` check (or when state is `idle`) would help debug "session ended after loop stopped" situations in production. `[fixable]`
- 🔵 **style** (L240): `new Set(Array.from(registry.entries(), ([clientId]) => clientId))` can be simplified to `new Set(registry.keys())` if `SessionRegistry` exposes a `keys()` method, or `new Set(Array.from(registry.entries()).map(([k]) => k))` isn't more readable. Current form works but `Array.from(iterable, mapFn)` with destructuring is slightly unusual. Consider adding a `keys()` method to `SessionRegistry` — it would make this intent obvious. `[fixable]`


_Solid bug fix — the root cause (clientId vs sessionId mismatch in orphan detection) is correctly identified and fixed, with good test coverage for the new `.finally()` lifecycle wiring. No correctness issues found._

- 🔵 **missing_tests** (L906): The "resume after spawn session dies" test covers the crash → reclaim → re-dispatch path well, but doesn't verify the reclaimed task's status transitions (active → pending → active). Adding `expect(store.get(task.id)!.status).toBe('pending')` before the `resume()` call would document the pre-condition more clearly and guard against store-level regressions. `[fixable]`

dimakis:
## Centaur Review

_Correct fix for a real bug (clientId vs sessionId mismatch in orphan detection). The `.finally()` handler for spawn lifecycle is well-guarded. A few stale comments and minor test gaps, plus a subtle change in `getActiveSessionIds` scope that's benign today but could surprise future callers._

- 🔵 **style** (L40): JSDoc still says "Spawn a new headless session" and line 349 says "Spawn a dedicated headless session" despite the PR renaming the clientId prefix from `headless:` to `task:`. These comments should be updated for consistency. `[fixable]`


_Correct fix for a real bug (clientId vs sessionId mismatch in orphan detection). The `.finally()` handler for spawn lifecycle is well-guarded. A few stale comments and minor test gaps, plus a subtle change in `getActiveSessionIds` scope that's benign today but could surprise future callers._

- 🔵 **missing_tests**: No test covers the `.finally()` handler's goal-change guard: if the user starts a new goal while an old spawned session is still in-flight, the `.finally()` handler should not resume/tick the orchestrator (because `getStatus().goalId !== oldGoalId`). The 'stopped' test covers the idle-state guard but not the goal-mismatch case. `[fixable]`
- 🔵 **missing_tests**: No test covers the `taskContext` assignment path (index.ts lines 289-297) — specifically the case where `registry.get(clientId)` returns undefined (the warn-log branch). This is a defensive guard worth exercising even if it can't happen today, since it documents an assumption about `startChat`'s synchronous register behavior. `[fixable]`


### `server/index.ts`
_Correct fix for a real bug (clientId vs sessionId mismatch in orphan detection). The `.finally()` handler for spawn lifecycle is well-guarded. A few stale comments and minor test gaps, plus a subtle change in `getActiveSessionIds` scope that's benign today but could surprise future callers._

- 🟡 **regressions** (L241): `getActiveSessionIds()` now returns ALL clientIds from the registry (including `headless:*` sessions created via `POST /api/sessions` in app.ts). Previously it returned SDK sessionIds which effectively excluded headless sessions. This is benign today — `getOrphaned()` only matches tasks with a non-null `session_id`, and only the orchestrator's `setSessionId` sets it (using `task:*` prefix) — but any future code that calls `store.setSessionId()` with a `headless:*` clientId would create a phantom liveness match. Consider filtering to `task:*` prefixed entries if the active set should only cover orchestrator-managed sessions. `[fixable]`

server/index.ts
```diff
@@ -256,16 +253,53 @@ const orchestrator = new TaskOrchestrator({
...
+        })
+        .finally(() => {
+          // Session ended (success or failure) — clean up registry entry
+          // (no WS close to trigger normal removal) and advance the
+          // orchestrator so orphan reclaim picks up unfinished tasks.
+          registry.remove(clientId);
+          // Only advance if the orchestrator is still on the same goal.
+          // A user may have paused manually or started a new goal while
+          // this session was in-flight — don't override that.
+          if (orchestratorRef && orchestratorRef.getStatus().goalId === goalId) {
```
dimakis: 🟡 **unsafe_assumptions**: The `.finally()` callback checks `orchestratorRef.getStatus().goalId === goalId` to decide whether to advance. However, when the orchestrator is in `running` state with the *same* goalId, calling `tick()` is safe but could double-dispatch if another source (e.g., `SignalProcessor` at line 183, or `onTaskCompleted`) triggers `tick()` in the same microtask queue. Since Node.js is single-threaded and `tick()` is synchronous this is safe today, but the lack of a re-entrancy guard makes it fragile if `tick()` ever becomes async. Low risk given current design, noting for awareness.

server/index.ts
```diff
@@ -256,16 +253,53 @@ const orchestrator = new TaskOrchestrator({
...
+              clientId,
+              goalId,
+            });
+          }
+        });
+
+      // Link session to task for session overview visibility.
+      // register() runs synchronously before startChat's first await,
+      // so the session exists here. Log if that assumption ever breaks.
+      const session = registry.get(clientId);
```
dimakis: 🟡 **unsafe_assumptions**: The comment says `register() runs synchronously before startChat's first await` — this assumes `startChat` calls `registry.register(clientId, ...)` synchronously before its first `await`. If `startChat` ever changes to do an async operation before registering (e.g., async worktree creation), `registry.get(clientId)` would return undefined, and `taskContext` would silently not be set. The `log.warn` fallback correctly detects this, but the session would run without taskContext, which could cause the session overview to miss it. The defensive logging is good; just flagging the coupling.

server/index.ts
```diff
@@ -256,16 +253,53 @@ const orchestrator = new TaskOrchestrator({
...
-        log.error('spawned session failed', {
+      })
+        .catch((err) => {
+          log.error('task session failed', {
+            taskId,
+            clientId,
+            error: (err as Error).message,
+          });
+        })
+        .finally(() => {
```
dimakis, line 268: 🔵 **style**: The `.finally()` block is 22 lines with branching logic (paused vs running, goal match vs mismatch). Consider extracting this into a named function (e.g., `onTaskSessionEnded(clientId, taskId, goalId)`) to improve readability and make the intent self-documenting. The current inline approach works but is dense for a fire-and-forget chain. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid fix — correctly aligns orphan detection to use clientIds instead of SDK sessionIds, adds proper cleanup via .finally(), and includes good test coverage for the main scenarios. The few findings are low-severity: a missing edge-case test for goal-change, and minor style/coupling observations._

- 🟡 **unsafe_assumptions** (L275): The `.finally()` callback checks `orchestratorRef.getStatus().goalId === goalId` to decide whether to advance. However, when the orchestrator is in `running` state with the *same* goalId, calling `tick()` is safe but could double-dispatch if another source (e.g., `SignalProcessor` at line 183, or `onTaskCompleted`) triggers `tick()` in the same microtask queue. Since Node.js is single-threaded and `tick()` is synchronous this is safe today, but the lack of a re-entrancy guard makes it fragile if `tick()` ever becomes async. Low risk given current design, noting for awareness.
- 🟡 **unsafe_assumptions** (L294): The comment says `register() runs synchronously before startChat's first await` — this assumes `startChat` calls `registry.register(clientId, ...)` synchronously before its first `await`. If `startChat` ever changes to do an async operation before registering (e.g., async worktree creation), `registry.get(clientId)` would return undefined, and `taskContext` would silently not be set. The `log.warn` fallback correctly detects this, but the session would run without taskContext, which could cause the session overview to miss it. The defensive logging is good; just flagging the coupling.
- 🔵 **style** (L267): The `.finally()` block is 22 lines with branching logic (paused vs running, goal match vs mismatch). Consider extracting this into a named function (e.g., `onTaskSessionEnded(clientId, taskId, goalId)`) to improve readability and make the intent self-documenting. The current inline approach works but is dense for a fire-and-forget chain. `[fixable]`
- 🔵 **style** (L243): The prefix was renamed from `headless:` to `task:` which is clearer. However, `log.info` on line 256 still says `'task session resolved'` and line 261 says `'task session failed'` — these are fine. But the catch block at line 306 still says `'failed to spawn session for task'`, which could be confused with line 261's `'task session failed'` during log searching. Consider differentiating: one is the async session failure (agent error) and the other is the synchronous spawn failure (e.g., transport creation). Minor log clarity issue. `[fixable]`


_Solid fix — correctly aligns orphan detection to use clientIds instead of SDK sessionIds, adds proper cleanup via .finally(), and includes good test coverage for the main scenarios. The few findings are low-severity: a missing edge-case test for goal-change, and minor style/coupling observations._

- 🔵 **missing_tests**: The new tests cover the happy paths (tick advances, resume reclaims, idle no-op), but there's no test for the goal-changed guard: what happens when `.finally()` fires but the orchestrator has moved to a *different* goalId (via `stop()+start(newGoal)`)? The production code handles it (line 275 checks `goalId === goalId`), but there's no test asserting that the old goal's tasks are left untouched. `[fixable]`

server/index.ts
```diff
@@ -235,15 +235,12 @@ const orchestrator = new TaskOrchestrator({
...
-      if (session?.sessionId) ids.add(session.sessionId);
-    }
-    return ids;
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
+    return new Set(registry.keys());
   },
   spawnSession: async (taskId: string, prompt: string, goalId: string) => {
-    const clientId = `headless:${generateWtId()}`;
+    const clientId = `task:${generateWtId()}`;
```
dimakis, line 243: 🔵 **style**: The prefix was renamed from `headless:` to `task:` which is clearer. However, `log.info` on line 256 still says `'task session resolved'` and line 261 says `'task session failed'` — these are fine. But the catch block at line 306 still says `'failed to spawn session for task'`, which could be confused with line 261's `'task session failed'` during log searching. Consider differentiating: one is the async session failure (agent error) and the other is the synchronous spawn failure (e.g., transport creation). Minor log clarity issue. `[fixable]`

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Solid fix for a real bug — orphan detection was comparing clientIds against SDK sessionIds, so it never matched spawned tasks. The `.finally() handler for registry cleanup and orchestrator advancement is well-guarded against all edge cases (stopped, goal changed, paused). Tests cover the critical paths. Only minor suggestions._

- 🔵 **missing_tests**: The `.finally()` handler in `server/index.ts` has an else branch for when the goal has changed or the orchestrator was stopped (logs 'task session ended after loop stopped or goal changed'). The 'stopped' case is tested ('tick is a no-op when orchestrator is stopped'), but the 'goal changed while session in-flight' scenario is not — e.g., user starts a new goal while a spawned session from the old goal is still running. Consider adding a test where `goalId` no longer matches. `[fixable]`


_Solid fix for a real bug — orphan detection was comparing clientIds against SDK sessionIds, so it never matched spawned tasks. The `.finally() handler for registry cleanup and orchestrator advancement is well-guarded against all edge cases (stopped, goal changed, paused). Tests cover the critical paths. Only minor suggestions._

- 🔵 **missing_tests**: No unit test file exists for `SessionRegistry`. The new `keys()` method is a trivial one-liner, but there are no tests for any of the registry's methods (entries, get, remove, etc.). Low priority since these are thin wrappers over Map, but worth noting for coverage. `[fixable]`


### `server/index.ts`
_Solid fix for a real bug — orphan detection was comparing clientIds against SDK sessionIds, so it never matched spawned tasks. The `.finally() handler for registry cleanup and orchestrator advancement is well-guarded against all edge cases (stopped, goal changed, paused). Tests cover the critical paths. Only minor suggestions._

- 🔵 **style**: Minor: `orchestratorRef.getStatus()` is called twice in the `.finally()` handler (once to check `goalId`, once to check `state`). Since this is synchronous single-threaded code there's no correctness issue, but a single `const status = orchestratorRef.getStatus()` would be marginally cleaner. `[fixable]`

server/index.ts
```diff
@@ -257,16 +254,41 @@ const orchestrator = new TaskOrchestrator({
...
-          error: (err as Error).message,
+      })
+        .catch((err) => {
+          log.error('task session failed', {
+            taskId,
+            clientId,
+            error: (err as Error).message,
+          });
+        })
+        .finally(() => {
```
dimakis, line 268: 🟡 **bugs**: The .finally() handler calls orchestratorRef.tick() or .resume() without a try-catch. These methods call into TaskStore (SQLite queries) which can throw on database errors. Since .finally() is on a promise chain, an exception here becomes an unhandled promise rejection — the session cleanup (registry.remove) has already happened, but the error is silently swallowed in production or crashes the process depending on Node.js --unhandled-rejections mode. Wrap the orchestrator calls in try-catch with a log.error fallback. `[fixable]`

server/index.ts
dimakis, line 243: 🔵 **style**: The prefix changed from 'headless:' to 'task:' which is a clearer name. However, no migration handling exists for tasks in SQLite that may still reference 'headless:' session_ids from before this change. On server restart, such tasks would be orphaned and reclaimed (correct behavior), but worth noting in the commit message that this is intentional — the prefix change applies only to new sessions.

server/__tests__/task-orchestrator.test.ts
```diff
@@ -882,6 +912,148 @@ describe('TaskOrchestrator', () => {
...
```
dimakis, line 946: 🔵 **missing_tests**: The 'resume after spawn session dies' test relies on the .then() callback from the mock spawnSession setting sessionId on the task (via queueMicrotask flushing during vi.waitFor). This implicit dependency makes the test fragile — if vi.waitFor timing changes, the sessionId might not be set when resume() runs, causing orphan detection to miss the task (no session_id = no orphan). Consider adding an explicit assertion that task.sessionId is set before simulating the crash, e.g. `expect(store.get(task.id)!.sessionId).toBe('task:spawned-1')` before clearing activeClients. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_Solid fix for a real bug where orphan detection never matched spawned tasks due to clientId vs sessionId mismatch. The .finally() handler for session lifecycle advancement is well-guarded with goal comparison. Tests are comprehensive. Minor concern: wrap orchestrator calls in .finally() with try-catch to avoid unhandled rejections on store errors._

- 🟡 **bugs** (L268): The .finally() handler calls orchestratorRef.tick() or .resume() without a try-catch. These methods call into TaskStore (SQLite queries) which can throw on database errors. Since .finally() is on a promise chain, an exception here becomes an unhandled promise rejection — the session cleanup (registry.remove) has already happened, but the error is silently swallowed in production or crashes the process depending on Node.js --unhandled-rejections mode. Wrap the orchestrator calls in try-catch with a log.error fallback. `[fixable]`
- 🔵 **style** (L243): The prefix changed from 'headless:' to 'task:' which is a clearer name. However, no migration handling exists for tasks in SQLite that may still reference 'headless:' session_ids from before this change. On server restart, such tasks would be orphaned and reclaimed (correct behavior), but worth noting in the commit message that this is intentional — the prefix change applies only to new sessions.


_Solid fix for a real bug where orphan detection never matched spawned tasks due to clientId vs sessionId mismatch. The .finally() handler for session lifecycle advancement is well-guarded with goal comparison. Tests are comprehensive. Minor concern: wrap orchestrator calls in .finally() with try-catch to avoid unhandled rejections on store errors._

- 🔵 **missing_tests** (L946): The 'resume after spawn session dies' test relies on the .then() callback from the mock spawnSession setting sessionId on the task (via queueMicrotask flushing during vi.waitFor). This implicit dependency makes the test fragile — if vi.waitFor timing changes, the sessionId might not be set when resume() runs, causing orphan detection to miss the task (no session_id = no orphan). Consider adding an explicit assertion that task.sessionId is set before simulating the crash, e.g. `expect(store.get(task.id)!.sessionId).toBe('task:spawned-1')` before clearing activeClients. `[fixable]`

dimakis:
## Centaur Review

_Core bug fix is correct — orphan detection was comparing clientIds against SDK sessionIds, causing spawned tasks to never match. The `.finally()` lifecycle hook and goal-guard logic are well-designed. Tests cover the main scenarios (orphan reclaim, idle guard, goal switch) but one test title is misleading and the startChat-rejection retry loop is untested._

- 🔵 **missing_tests** (L158): New `keys()` method has no unit test in `packages/harness/__tests__/session-registry.test.ts`. It's a trivial delegation to `Map.keys()`, but every other accessor (`get`, `entries`) is exercised through tests. A one-liner test would keep coverage consistent. `[fixable]`


_Core bug fix is correct — orphan detection was comparing clientIds against SDK sessionIds, causing spawned tasks to never match. The `.finally()` lifecycle hook and goal-guard logic are well-designed. Tests cover the main scenarios (orphan reclaim, idle guard, goal switch) but one test title is misleading and the startChat-rejection retry loop is untested._

- 🔵 **style** (L36): The JSDoc for `getActiveSessionIds` says "session IDs" but the PR's whole point is that these are **clientIds**, not SDK sessionIds. Updating the doc to `/** Get active client IDs (registry keys) for orphan detection */` would prevent the same confusion from recurring. `[fixable]`


_Core bug fix is correct — orphan detection was comparing clientIds against SDK sessionIds, causing spawned tasks to never match. The `.finally()` lifecycle hook and goal-guard logic are well-designed. Tests cover the main scenarios (orphan reclaim, idle guard, goal switch) but one test title is misleading and the startChat-rejection retry loop is untested._

- 🔵 **missing_tests** (L915): Test "tick after spawn session completes advances workflow" is misleading: t2 (reuse policy) is already dispatched to active during the initial `start()` tick chain (spawn t1 → microtask tick → dispatch t2 as reuse). The subsequent `orch.tick()` is actually a no-op that pauses the orchestrator. The test passes but doesn't verify that `.finally()` tick *advances* anything — it only confirms it doesn't break already-advanced state. Consider either (a) making both tasks spawn-policy so t2 stays pending until t1 completes, or (b) renaming the test to reflect what it actually asserts. `[fixable]`


### `server/index.ts`
_Core bug fix is correct — orphan detection was comparing clientIds against SDK sessionIds, causing spawned tasks to never match. The `.finally()` lifecycle hook and goal-guard logic are well-designed. Tests cover the main scenarios (orphan reclaim, idle guard, goal switch) but one test title is misleading and the startChat-rejection retry loop is untested._

- 🟡 **missing_tests** (L261): The `.catch()` handler (startChat rejection) logs the error but doesn't mark the task as failed/blocked. When startChat throws, `.finally()` fires → `registry.remove()` → `tick()` → orphan reclaim resets the task to pending → re-dispatch → potentially infinite retry loop. This differs from the orchestrator's own spawn-failure path (line 397 in task-orchestrator.ts) which marks the task `blocked` with an annotation. No test covers the startChat-rejection → `.finally()` → orphan-reclaim cycle to verify it terminates. `[fixable]`
- 🔵 **unsafe_assumptions** (L282): When `.finally()` calls `orchestratorRef.tick()` directly (state is 'running'), `spawnDepth` is not reset — only `broadcastAndTick()` and explicit event handlers (`onTaskCompleted`, etc.) reset it. In practice this is safe because reaching state='running' implies a prior reset, but calling `tick()` with a potentially stale `spawnDepth` is fragile if new call sites are added later. The `resume()` path correctly resets via `broadcastAndTick()`. `[fixable]`

server/index.ts
dimakis, line 243: 🔵 **style**: Prefix changed from `headless:` to `task:` — this is a good semantic improvement, but any external logging/monitoring/alerting that greps for `headless:` will silently stop matching. If there are log queries or dashboards filtering on the old prefix, they should be updated. Not a code issue, but worth noting in the PR description.

server/index.ts
```diff
@@ -235,15 +235,12 @@ const orchestrator = new TaskOrchestrator({
...
   getActiveSessionIds: () => {
-    const ids = new Set<string>();
-    for (const [clientId] of registry.entries()) {
-      const session = registry.get(clientId);
-      if (session?.sessionId) ids.add(session.sessionId);
-    }
-    return ids;
+    // Return clientIds — task.session_id stores clientId, not SDK sessionId.
+    // Using sessionId here caused orphan detection to never match spawned tasks.
+    return new Set(registry.keys());
```
dimakis, line 240: 🔵 **unsafe_assumptions**: The new `getActiveSessionIds` returns ALL clientIds from the registry via `registry.keys()`, including regular browser-connected sessions. Previously it only returned SDK sessionIds from active sessions. This means orphan detection now compares task `session_id` (which stores clientIds like `task:abc123`) against ALL registry keys (which also include browser client IDs). This works correctly because browser clientIds will never match a `task:*` sessionId stored on a task, but it's a broader set than strictly necessary. Not a bug — just a noted semantic change.

dimakis (commented):
## Centaur Review

Found **3** issue(s).


_Solid bug fix — the core issue (sessionId vs clientId mismatch in orphan detection) is correctly identified and fixed, the `.finally()` handler has proper guards against goal drift and state transitions, and the new tests cover the key lifecycle scenarios (success advancement, crash recovery, idle no-op, goal change)._

- 🔵 **missing_tests**: The `.finally()` handler in index.ts has a `resume()` path (when `status.state === 'paused'`) and a `tick()` path (otherwise). The 'resume after spawn session dies' test at line 946 exercises the resume path by calling `orch.resume()` directly, but doesn't test the goal-change guard on that path. Consider adding a test where the orchestrator is paused AND the goal has changed, verifying that `.resume()` on the new goal doesn't cause unexpected behavior from the old session's finally handler. This is low-risk since the `.finally()` guard in index.ts checks `status.goalId === goalId` before either branch. `[fixable]`


### `server/index.ts`
_Solid bug fix — the core issue (sessionId vs clientId mismatch in orphan detection) is correctly identified and fixed, the `.finally()` handler has proper guards against goal drift and state transitions, and the new tests cover the key lifecycle scenarios (success advancement, crash recovery, idle no-op, goal change)._

- 🔵 **style** (L243): Prefix changed from `headless:` to `task:` — this is a good semantic improvement, but any external logging/monitoring/alerting that greps for `headless:` will silently stop matching. If there are log queries or dashboards filtering on the old prefix, they should be updated. Not a code issue, but worth noting in the PR description.
- 🔵 **unsafe_assumptions** (L240): The new `getActiveSessionIds` returns ALL clientIds from the registry via `registry.keys()`, including regular browser-connected sessions. Previously it only returned SDK sessionIds from active sessions. This means orphan detection now compares task `session_id` (which stores clientIds like `task:abc123`) against ALL registry keys (which also include browser client IDs). This works correctly because browser clientIds will never match a `task:*` sessionId stored on a task, but it's a broader set than strictly necessary. Not a bug — just a noted semantic change.

server/index.ts
```diff
@@ -257,16 +254,49 @@ const orchestrator = new TaskOrchestrator({
...
+          try {
+            const status = orchestratorRef.getStatus();
+            if (status.goalId === goalId) {
+              if (status.state === 'paused') {
+                orchestratorRef.resume();
+              } else {
+                orchestratorRef.tick();
+              }
+            } else {
+              log.info('task session ended after loop stopped or goal changed', {
```
dimakis, line 286: 🟡 **bugs**: The `.finally()` handler calls `resume()` whenever `state === 'paused'`, but cannot distinguish between auto-pause (all tasks dispatched — correct to resume) and user-initiated `pause()` (should stay paused). If a user manually pauses the orchestrator while spawned sessions are in-flight, the next session to end will call `resume()` and override the user's intent. Consider tracking a `manuallyPaused` flag on the orchestrator, or changing the `.finally()` to always call `tick()` — and having `tick()` internally handle orphan reclaim even in paused state. `[fixable]`

dimakis (commented):
## Centaur Review

### `server/index.ts`
_The core fix (clientId-based orphan detection replacing broken sessionId matching) is correct and well-tested with comprehensive edge-case coverage; the one gap is that `.finally()` cannot distinguish user-initiated pause from auto-pause, which could override a manual pause._

- 🟡 **bugs** (L286): The `.finally()` handler calls `resume()` whenever `state === 'paused'`, but cannot distinguish between auto-pause (all tasks dispatched — correct to resume) and user-initiated `pause()` (should stay paused). If a user manually pauses the orchestrator while spawned sessions are in-flight, the next session to end will call `resume()` and override the user's intent. Consider tracking a `manuallyPaused` flag on the orchestrator, or changing the `.finally()` to always call `tick()` — and having `tick()` internally handle orphan reclaim even in paused state. `[fixable]`


_The core fix (clientId-based orphan detection replacing broken sessionId matching) is correct and well-tested with comprehensive edge-case coverage; the one gap is that `.finally()` cannot distinguish user-initiated pause from auto-pause, which could override a manual pause._

- 🟡 **missing_tests**: No test for 'user manually pauses while a spawned session is in-flight; session ends; orchestrator should remain paused'. The existing 'does not resume old goal when paused orchestrator moved to a new goal' test covers the goal-change guard but not the same-goal manual-pause scenario. This is the complement of the auto-pause/resume test at line 939 and would expose the warning above. `[fixable]`

PR #396, opened 2026-06-25
fix(transport): prevent SSE replay storms and silent message loss

dimakis:
## Summary

Fixes the SSE transport regression causing 8+ minute message delivery delays and silent message drops after the SSE migration (#381).

**Root cause chain:**
1. `ProcessTransport.interrupt()` throws unhandled → server crashes → loses all in-memory state
2. Clients reconnect with stale cursors → periodic sync replays thousands of events at 50/5s
3. `isSessionActive` only checked a boolean, not session STATE → SUSPENDED/DETACHED sessions replayed indefinitely
4. Failed `send` POSTs (404 from stale connectionId) were silently dropped → user messages lost

**Three fixes:**

- **Periodic sync replay storm prevention** (`connection-registry.ts`):
  - Replace `isSessionActive` with `shouldSync` — checks session STATE (only ACTIVE/STARTING/CREATED sync)
  - Add `MAX_REPLAY_GAP` (200 events) — skip ahead when cursor is far behind head instead of replaying 6K+ events
  - Add `getHeadSeq` to EventStore for gap detection

- **POST retry for failed message sends** (`sse-connection.ts`):
  - Re-queue `send` POSTs that get HTTP errors or network errors
  - Messages flush on next successful reconnect instead of being silently dropped

- **Guard ProcessTransport.interrupt() crash** (`chat.ts`):
  - Wrap `queryInstance.interrupt()` in try/catch to prevent server process death

## Test plan
- [x] 76 tests passing (connection-registry + sse-connection)
- [x] Type check clean
- [ ] Centaur review (2 rounds)
- [ ] Manual test: send messages during agent tool execution, verify delivery
- [ ] Manual test: restart server, verify no replay storm in logs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #440, opened 2026-07-04
refactor(transport): SSOT phase 3 — clean reconnect

dimakis:
## Summary

- **Remove ownership dance from `handleReconnect`** — reattach/rekey/zombie cleanup removed (~60 lines). `handleSendV2` already handles all of this on the first user message, making it redundant on reconnect.
- **Remove periodic sync** — 5s timer in `ConnectionRegistry` that retried missed events is eliminated (~130 lines). Cursor-based replay on welcome covers the gap.
- **Fire-and-forget reconnect POST** — client no longer defers `_connected` until the reconnect POST succeeds (~100 lines). Connection is marked immediately on welcome; the POST is advisory.
- **Remove `ReconnectMessage` from WS union** — reconnect is REST-only, not dispatched via WebSocket.

Net: **-1,043 lines** across 9 files. All 177 affected tests pass.

## Context

Phase 3 of the Transport SSOT series (P0 #431, P1 #433). The reconnect "ownership dance" duplicated logic already handled by `handleSendV2` on the first user message. Periodic sync was a safety net made redundant by cursor replay on welcome.

## What's kept

- `/reconnect` REST endpoint + `handleReconnect` handler (watch + cursor reset + EventStore replay + boot context + suspend resume)
- Client-side `doPost('reconnect', ...)` on welcome (fire-and-forget)
- `ConnectionRegistry.resetCursor()` and cursor tracking in `broadcast()`

## Test plan

- [x] `npx vitest run` — all 177 tests pass (ws-handler-v2, connection-registry, sse-connection)
- [ ] Manual: kill server mid-chat, restart → verify auto-reconnect + event replay
- [ ] Manual: background iOS app, foreground → verify suspend/resume
- [ ] Manual: send message after reconnect → verify session resumes via handleSendV2

### DIFF
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -225,33 +225,35 @@ export class SseConnection implements ChatConnection {
               lastSeq,
             }))
           : null);
-      this._connected = true;
       if (sessions) {
         this._pendingReconnectSessions = sessions;
         // Capture connectionId to detect stale callbacks — if a new welcome
         // arrives while this POST is in-flight, the callback should no-op
         // to avoid double-flushing pending sends.
         const postConnectionId = this._connectionId;
+        // Don't mark connected until POST succeeds — prevents send() from
+        // bypassing the pending queue and arriving before cursor setup.
         this.doPost('reconnect', { type: 'reconnect', sessions }).then(
           () => {
             if (this._connectionId !== postConnectionId) return; // stale callback
             this._pendingReconnectSessions = null;
-            // Flush pending sends AFTER reconnect so the server processes
-            // handleReconnect (cursor reset, replay) before user messages.
+            this._connected = true;
             this.flushPendingSends();
+            this.listener?.({ type: '_open' });
           },
           () => {
             if (this._connectionId !== postConnectionId) return; // stale callback
             // doPost already logs the warning. Keep _pendingReconnectSessions
             // so the next EventSource reconnect retries automatically.
-            // Still flush — handleSendV2 handles ownership independently.
-            this.flushPendingSends();
+            // Don't flush — server hasn't set up cursor/replay. Don't mark
+            // connected — sends stay queued until next successful reconnect.
           },
         );
       } else {
+        this._connected = true;
         this.flushPendingSends();
+        this.listener?.({ type: '_open' });
       }
-      this.listener?.({ type: '_open' });
       this._isReconnect = true;
     });
 
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -214,17 +214,7 @@ export class SseConnection implements ChatConnection {
       // have been cleared (clearSession) since the retry was queued.
       // Refresh lastSeq from current map — SSE events may have advanced it
       // since the original failure, avoiding unnecessary replay.
-      const pending = this._pendingReconnectSessions
-        ?.filter((s) => this.seqBySession.has(s.sessionId))
-        .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
-      const sessions =
-        (pending && pending.length > 0 ? pending : null) ??
-        (this._isReconnect && this.seqBySession.size > 0
-          ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
-              sessionId,
-              lastSeq,
-            }))
-          : null);
+      const sessions = this.getReconnectSessions();
       if (sessions) {
         this._pendingReconnectSessions = sessions;
         // Capture connectionId to detect stale callbacks — if a new welcome
@@ -288,7 +278,9 @@ export class SseConnection implements ChatConnection {
   }
 
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
-    if (!this._connectionId) return;
+    if (!this._connectionId) {
+      throw new Error('doPost called without connectionId');
+    }
     try {
       const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
@@ -352,6 +344,30 @@ export class SseConnection implements ChatConnection {
       default:
         return null;
     }
+  }
+
+  /**
+   * Build the session list for a reconnect POST.
+   *
+   * Priority: (1) pending retry sessions (filtered against current seqBySession,
+   * with refreshed lastSeq), (2) all tracked sessions if this is a reconnect.
+   * Returns null when there's nothing to reconnect.
+   */
+  private getReconnectSessions(): Array<{ sessionId: string; lastSeq: number }> | null {
+    const pending = this._pendingReconnectSessions
+      ?.filter((s) => this.seqBySession.has(s.sessionId))
+      .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
+
+    if (pending && pending.length > 0) return pending;
+
+    if (this._isReconnect && this.seqBySession.size > 0) {
+      return Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+        sessionId,
+        lastSeq,
+      }));
+    }
+
+    return null;
   }
 
   // ─── Browser lifecycle ─────────────────────────────────────────────────────
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -64,6 +64,7 @@ export class SseConnection implements ChatConnection {
       this.es = null;
     }
     this._connected = false;
+    this._pendingReconnectSessions = null;
   }
 
   /**
@@ -209,8 +210,13 @@ export class SseConnection implements ChatConnection {
       // Fire reconnect POST if reconnecting with sessions, or retry a
       // previously failed reconnect. handleSendV2 handles ownership on first
       // message, and replayed events arrive via SSE regardless.
+      // Filter pending retries against current seqBySession — sessions may
+      // have been cleared (clearSession) since the retry was queued.
+      const pending = this._pendingReconnectSessions?.filter((s) =>
+        this.seqBySession.has(s.sessionId),
+      );
       const sessions =
-        this._pendingReconnectSessions ??
+        (pending && pending.length > 0 ? pending : null) ??
         (this._isReconnect && this.seqBySession.size > 0
           ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
               sessionId,
@@ -274,7 +280,7 @@ export class SseConnection implements ChatConnection {
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
     if (!this._connectionId) return;
     try {
-      await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
+      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
@@ -282,6 +288,9 @@ export class SseConnection implements ChatConnection {
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok) {
+        throw new Error(`HTTP ${res.status}`);
+      }
     } catch (err) {
       // POST failures are non-fatal — the server may be temporarily
       // unreachable. SSE EventSource auto-reconnects and replays missed events
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -23,7 +23,6 @@ export interface SseConnectionConfig {
   fetch: (url: string, init?: RequestInit) => Promise<Response>;
   /** Factory for EventSource — allows injection for testing. */
   createEventSource?: (url: string) => EventSource;
-  reconnectDelayMs?: number;
   /** URL for the sendBeacon suspend fallback. */
   suspendUrl?: string;
 }
@@ -38,7 +37,6 @@ export class SseConnection implements ChatConnection {
   private listener: ConnectionListener | null = null;
   private seqBySession = new Map<string, number>();
   private pendingSends: Array<{ endpoint: string; body: Record<string, unknown> }> = [];
-  private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
   private boundOnVisibility: (() => void) | null = null;
   private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
   private boundOnPageHide: (() => void) | null = null;
@@ -47,7 +45,6 @@ export class SseConnection implements ChatConnection {
   constructor(config: SseConnectionConfig) {
     this.config = {
       createEventSource: (url: string) => new EventSource(url),
-      reconnectDelayMs: 500,
       suspendUrl: '',
       ...config,
     };
@@ -60,10 +57,6 @@ export class SseConnection implements ChatConnection {
 
   disconnect(): void {
     this.removeBrowserListeners();
-    if (this.reconnectTimer) {
-      clearTimeout(this.reconnectTimer);
-      this.reconnectTimer = null;
-    }
     if (this.es) {
       this.es.close();
       this.es = null;
@@ -92,8 +85,8 @@ export class SseConnection implements ChatConnection {
       return true;
     }
 
-    // Queue if reconnecting
-    if (this.reconnectTimer || this.es) {
+    // Queue if EventSource exists (reconnecting)
+    if (this.es) {
       if (this.pendingSends.length >= MAX_PENDING_SENDS) {
         this.pendingSends.shift();
       }
@@ -175,7 +168,6 @@ export class SseConnection implements ChatConnection {
    */
   checkAndReconnect(force = false): void {
     if (!force && this._connected) return;
-    if (this.reconnectTimer) return;
     if (this.es) {
       this.es.close();
       this.es = null;
@@ -192,11 +184,6 @@ export class SseConnection implements ChatConnection {
 
   private doConnect(): void {
     if (this.es) return;
-
-    if (this.reconnectTimer) {
-      clearTimeout(this.reconnectTimer);
-      this.reconnectTimer = null;
-    }
 
     // Always use the base URL — reconnect sessions are sent via POST in the
     // welcome handler. This avoids the bug where EventSource auto-reconnect
@@ -217,18 +204,21 @@ export class SseConnection implements ChatConnection {
       }
       this._connectionId = msg.connectionId as string;
 
-      // _connected deferred until doReconnectPost succeeds — prevents
-      // external send() from bypassing the pending queue mid-reconnect.
-      // Capture both connectionId and ES instance for the staleness guard.
-      const welcomeConnectionId = this._connectionId;
-      const welcomeEs = this.es;
+      // Fire reconnect POST (fire-and-forget) if reconnecting with sessions.
+      // No need to defer _connected — handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
       if (this._isReconnect && this.seqBySession.size > 0) {
-        this.doReconnectPost(welcomeConnectionId, welcomeEs);
-      } else {
-        this._connected = true;
-        this.flushPendingSends();
-        this.listener?.({ type: '_open' });
+        this.doPost('reconnect', {
+          type: 'reconnect',
+          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          })),
+        });
       }
+      this._connected = true;
+      this.flushPendingSends();
+      this.listener?.({ type: '_open' });
       this._isReconnect = true;
     });
 
@@ -260,66 +250,6 @@ export class SseConnection implements ChatConnection {
 
     // EventSource fires 'open' when the connection is established,
     // but we wait for the 'welcome' event before marking as connected.
-  }
-
-  /**
-   * Send the reconnect POST and only mark connected on success.
-   *
-   * On failure the client stays disconnected — the next EventSource
-   * auto-reconnect will trigger a fresh welcome + retry. This prevents
-   * flushing pending sends into the void when the server never ran
-   * handleReconnect (no watch, no reattach, no replay).
-   */
-  private async doReconnectPost(
-    welcomeConnectionId: string,
-    welcomeEs: EventSource | null,
-  ): Promise<void> {
-    try {
-      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/reconnect`, {
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          'X-Connection-ID': welcomeConnectionId,
-        },
-        body: JSON.stringify({
-          type: 'reconnect',
-          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
-            sessionId,
-            lastSeq,
-          })),
-        }),
-      });
-
-      // Guard: bail if disconnect() was called, a newer welcome arrived,
-      // or checkAndReconnect replaced the EventSource while in-flight.
-      if (!this.es || this.es !== welcomeEs || this._connectionId !== welcomeConnectionId) return;
-
-      if (res.ok) {
-        this._connected = true;
-        this.flushPendingSends();
-        this.listener?.({ type: '_open' });
-      } else {
-        console.warn('[SseConnection] reconnect POST returned', res.status);
-        this.scheduleReconnect();
-      }
-    } catch (err) {
-      if (!this.es || this.es !== welcomeEs || this._connectionId !== welcomeConnectionId) return;
-      console.warn('[SseConnection] reconnect POST failed', err);
-      this.scheduleReconnect();
-    }
-  }
-
-  /** Tear down and reconnect after a delay to avoid tight retry loops. */
-  private scheduleReconnect(): void {
-    if (this.reconnectTimer) return;
-    if (this.es) {
-      this.es.close();
-      this.es = null;
-    }
-    this.reconnectTimer = setTimeout(() => {
-      this.reconnectTimer = null;
-      this.doConnect();
-    }, this.config.reconnectDelayMs);
   }
 
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -37,6 +37,8 @@ export class SseConnection implements ChatConnection {
   private listener: ConnectionListener | null = null;
   private seqBySession = new Map<string, number>();
   private pendingSends: Array<{ endpoint: string; body: Record<string, unknown> }> = [];
+  /** Sessions that need a reconnect POST — set when reconnect fails, retried on next welcome. */
+  private _pendingReconnectSessions: Array<{ sessionId: string; lastSeq: number }> | null = null;
   private boundOnVisibility: (() => void) | null = null;
   private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
   private boundOnPageHide: (() => void) | null = null;
@@ -81,7 +83,7 @@ export class SseConnection implements ChatConnection {
     if (!endpoint) return false;
 
     if (this._connected && this._connectionId) {
-      this.doPost(endpoint, msg);
+      this.doPost(endpoint, msg).catch(() => {});
       return true;
     }
 
@@ -143,7 +145,7 @@ export class SseConnection implements ChatConnection {
 
     // Try POST first
     if (this._connected && this._connectionId) {
-      this.doPost('suspend', { type: 'session_suspend', sessions });
+      this.doPost('suspend', { type: 'session_suspend', sessions }).catch(() => {});
       return;
     }
 
@@ -204,20 +206,37 @@ export class SseConnection implements ChatConnection {
       }
       this._connectionId = msg.connectionId as string;
 
-      // Fire reconnect POST (fire-and-forget) if reconnecting with sessions.
-      // No need to defer _connected — handleSendV2 handles ownership on first
+      // Fire reconnect POST if reconnecting with sessions, or retry a
+      // previously failed reconnect. handleSendV2 handles ownership on first
       // message, and replayed events arrive via SSE regardless.
-      if (this._isReconnect && this.seqBySession.size > 0) {
-        this.doPost('reconnect', {
-          type: 'reconnect',
-          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
-            sessionId,
-            lastSeq,
-          })),
-        });
-      }
+      const sessions =
+        this._pendingReconnectSessions ??
+        (this._isReconnect && this.seqBySession.size > 0
+          ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+              sessionId,
+              lastSeq,
+            }))
+          : null);
       this._connected = true;
-      this.flushPendingSends();
+      if (sessions) {
+        this._pendingReconnectSessions = sessions;
+        this.doPost('reconnect', { type: 'reconnect', sessions }).then(
+          () => {
+            this._pendingReconnectSessions = null;
+            // Flush pending sends AFTER reconnect so the server processes
+            // handleReconnect (cursor reset, replay) before user messages.
+            this.flushPendingSends();
+          },
+          () => {
+            // doPost already logs the warning. Keep _pendingReconnectSessions
+            // so the next EventSource reconnect retries automatically.
+            // Still flush — handleSendV2 handles ownership independently.
+            this.flushPendingSends();
+          },
+        );
+      } else {
+        this.flushPendingSends();
+      }
       this.listener?.({ type: '_open' });
       this._isReconnect = true;
     });
@@ -263,9 +282,14 @@ export class SseConnection implements ChatConnection {
         },
         body: JSON.stringify(body),
       });
-    } catch {
+    } catch (err) {
       // POST failures are non-fatal — the server may be temporarily
-      // unreachable. The SSE stream will reconnect and replay missed events.
+      // unreachable. SSE EventSource auto-reconnects and replays missed events
+      // from the EventStore. However, a failed reconnect POST means the server
+      // won't reset the cursor or re-send boot context until the next
+      // reconnect cycle. Client-side seq dedup prevents duplicate delivery.
+      console.warn(`[mitzo] ${endpoint} POST failed:`, err instanceof Error ? err.message : err);
+      throw err;
     }
   }
 
@@ -274,7 +298,7 @@ export class SseConnection implements ChatConnection {
     const toFlush = this.pendingSends;
     this.pendingSends = [];
     for (const { endpoint, body } of toFlush) {
-      this.doPost(endpoint, body);
+      this.doPost(endpoint, body).catch(() => {});
     }
   }
 
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -212,9 +212,11 @@ export class SseConnection implements ChatConnection {
       // message, and replayed events arrive via SSE regardless.
       // Filter pending retries against current seqBySession — sessions may
       // have been cleared (clearSession) since the retry was queued.
-      const pending = this._pendingReconnectSessions?.filter((s) =>
-        this.seqBySession.has(s.sessionId),
-      );
+      // Refresh lastSeq from current map — SSE events may have advanced it
+      // since the original failure, avoiding unnecessary replay.
+      const pending = this._pendingReconnectSessions
+        ?.filter((s) => this.seqBySession.has(s.sessionId))
+        .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
       const sessions =
         (pending && pending.length > 0 ? pending : null) ??
         (this._isReconnect && this.seqBySession.size > 0
@@ -226,14 +228,20 @@ export class SseConnection implements ChatConnection {
       this._connected = true;
       if (sessions) {
         this._pendingReconnectSessions = sessions;
+        // Capture connectionId to detect stale callbacks — if a new welcome
+        // arrives while this POST is in-flight, the callback should no-op
+        // to avoid double-flushing pending sends.
+        const postConnectionId = this._connectionId;
         this.doPost('reconnect', { type: 'reconnect', sessions }).then(
           () => {
+            if (this._connectionId !== postConnectionId) return; // stale callback
             this._pendingReconnectSessions = null;
             // Flush pending sends AFTER reconnect so the server processes
             // handleReconnect (cursor reset, replay) before user messages.
             this.flushPendingSends();
           },
           () => {
+            if (this._connectionId !== postConnectionId) return; // stale callback
             // doPost already logs the warning. Keep _pendingReconnectSessions
             // so the next EventSource reconnect retries automatically.
             // Still flush — handleSendV2 handles ownership independently.
--- a/packages/protocol/src/ws-schemas-v2.ts
+++ b/packages/protocol/src/ws-schemas-v2.ts
@@ -124,9 +124,10 @@ export const V2SetModeMessage = z.object({
 
 // ─── Union ──────────────────────────────────────────────────────────────────
 
+// ReconnectMessage is handled via REST POST (not WS) — exported for
+// the REST handler but excluded from the WS union.
 export const IncomingWsMessageV2 = z.discriminatedUnion('type', [
   HelloMessage,
-  ReconnectMessage,
   WatchMessage,
   UnwatchMessage,
   SwitchSessionMessage,
--- a/packages/protocol/src/ws-schemas-v2.ts
+++ b/packages/protocol/src/ws-schemas-v2.ts
@@ -124,10 +124,11 @@ export const V2SetModeMessage = z.object({
 
 // ─── Union ──────────────────────────────────────────────────────────────────
 
-// ReconnectMessage is handled via REST POST (not WS) — exported for
-// the REST handler but excluded from the WS union.
+// ReconnectMessage is primarily handled via REST POST (SSE transport), but WS
+// clients still send it over WS until P4 removes the WS chat transport.
 export const IncomingWsMessageV2 = z.discriminatedUnion('type', [
   HelloMessage,
+  ReconnectMessage,
   WatchMessage,
   UnwatchMessage,
   SwitchSessionMessage,
--- a/server/ws-handler-v2.ts
+++ b/server/ws-handler-v2.ts
@@ -264,24 +264,38 @@ export function handleReconnect(
         // transport (refreshed), and ownership checks on send/interrupt
         // handle the rekey atomically.
         const found = ctx.sessionRegistry.findBySessionId(entry.sessionId);
+        const storeState = ctx.eventStore.getSessionState(entry.sessionId);
+
+        // Skip reattach for zombie sessions — if EventStore says the session
+        // is ENDED or CLOSING, don't reattach. The session's query loop is
+        // finished and reattaching would resurrect a dead transport binding.
         if (found && ctx.sessionRegistry.isActive(found.clientId)) {
-          const ownerConnection = getOwnerConnection(found.clientId);
-          // Reattach if: (a) same connection owns it, OR (b) old owner
-          // connection is gone (device restart gave us a new connectionId).
-          const ownerGone =
-            ownerConnection !== connectionId && !ctx.connRegistry.get(ownerConnection);
-          if (
-            (ownerConnection === connectionId || ownerGone) &&
-            !ctx.sessionRegistry.isAttached(found.clientId)
-          ) {
-            const transport = ctx.connRegistry.get(connectionId)?.transport;
-            if (transport) {
-              reattachChat(found.clientId, transport);
-              log.info('reattached detached session on reconnect', {
-                connectionId,
-                sessionId: entry.sessionId,
-                ownerGone,
-              });
+          if (storeState === 'ENDED' || storeState === 'CLOSING') {
+            log.info('skipping reattach for zombie session on reconnect', {
+              connectionId,
+              sessionId: entry.sessionId,
+              clientId: found.clientId,
+              storeState,
+            });
+          } else {
+            const ownerConnection = getOwnerConnection(found.clientId);
+            // Reattach if: (a) same connection owns it, OR (b) old owner
+            // connection is gone (device restart gave us a new connectionId).
+            const ownerGone =
+              ownerConnection !== connectionId && !ctx.connRegistry.get(ownerConnection);
+            if (
+              (ownerConnection === connectionId || ownerGone) &&
+              !ctx.sessionRegistry.isAttached(found.clientId)
+            ) {
+              const transport = ctx.connRegistry.get(connectionId)?.transport;
+              if (transport) {
+                reattachChat(found.clientId, transport);
+                log.info('reattached detached session on reconnect', {
+                  connectionId,
+                  sessionId: entry.sessionId,
+                  ownerGone,
+                });
+              }
             }
           }
         }
@@ -576,6 +590,9 @@ export function handleSendV2(
             // with seq <= lastProcessedSeq. The two HTTP requests (reconnect
             // POST and send POST) can arrive as separate event loop ticks in any
             // order, but duplicate delivery is always harmless thanks to seq dedup.
+            // TODO(P4): Unify cursor reset so watch() initializes the cursor from
+            // the client's lastSeq, removing the dependency on client-side seq dedup
+            // for the window between watch() and the reconnect POST arriving.
             ctx.connRegistry.setActive(connectionId, sessionId);
             sendToChat(activeClientId, prompt, msg.images, msg.contextBlocks, msg.clientMsgId);
             span.setAttribute('routing.decision', isOwner ? 'active' : 'takeover');
@@ -950,7 +967,7 @@ export async function dispatchV2Message(
       // Already handled at routing layer, ignore duplicate
       break;
     case 'reconnect':
-      // WS clients still send reconnect over WS (removed in P4).
+      // WS clients may also send reconnect over WS.
       handleReconnect(connectionId, msg, ctx);
       break;
     case 'watch':
--- a/server/ws-handler-v2.ts
+++ b/server/ws-handler-v2.ts
@@ -247,58 +247,21 @@ export function handleReconnect(
           } as Record<string, unknown>);
         }
 
-        // Reset cursor to last replayed seq — prevents duplicate delivery from
-        // periodic sync. If no events replayed, cursor stays at client's lastSeq.
+        // Reset cursor to last replayed seq so broadcast() doesn't re-deliver.
+        // If no events replayed, cursor stays at client's lastSeq.
         const newCursor = events.length > 0 ? events[events.length - 1].seq : entry.lastSeq;
         ctx.connRegistry.resetCursor(connectionId, entry.sessionId, newCursor);
 
-        // Cross-reference with the durable EventStore: state=ENDED in the store
-        // is ground truth that the query loop has finished (P1: use state, not is_active).
-        const found = ctx.sessionRegistry.findBySessionId(entry.sessionId);
-        const storeState = ctx.eventStore.getSessionState(entry.sessionId);
-        let running = found ? ctx.sessionRegistry.isActive(found.clientId) : false;
-        if (running && (storeState === 'ENDED' || storeState === 'CLOSING')) {
-          running = false;
-          log.info('removing stale session from registry (state-based)', {
-            connectionId,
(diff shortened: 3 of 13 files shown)


dimakis (commented):
## Centaur Review

### `packages/protocol/src/ws-schemas-v2.ts`
_Sound simplification overall, but removing ReconnectMessage from the WS union breaks the WS transport fallback (MitzoConnection still sends reconnect over WS). The fire-and-forget design also introduces a subtle cursor race between send and reconnect POST arrival order._

- 🔴 **regressions** (L129): ReconnectMessage removed from WS union breaks the WS transport fallback. `MitzoConnection` (packages/client/src/connection.ts:228) still sends `{ type: 'reconnect' }` over WS on reconnect. Since it's no longer in the discriminated union, the server's `dispatchV2Message` Zod parse rejects it and the message is silently dropped. This means WS clients get no event replay, no cursor reset, and no boot_context re-send on reconnect. Either update `MitzoConnection` to send reconnect via REST POST (like `SseConnection` does), or keep `ReconnectMessage` in the WS union with a no-op handler. `[fixable]`


### `server/ws-handler-v2.ts`
_Sound simplification overall, but removing ReconnectMessage from the WS union breaks the WS transport fallback (MitzoConnection still sends reconnect over WS). The fire-and-forget design also introduces a subtle cursor race between send and reconnect POST arrival order._

- 🔵 **style** (L915): Dead code: `case 'reconnect':` can never be reached because `reconnect` was removed from the `IncomingWsMessageV2` union. The Zod parse on line 900 rejects reconnect messages before this switch. Remove this case to avoid confusion. `[fixable]`


_Sound simplification overall, but removing ReconnectMessage from the WS union breaks the WS transport fallback (MitzoConnection still sends reconnect over WS). The fire-and-forget design also introduces a subtle cursor race between send and reconnect POST arrival order._

- 🟡 **unsafe_assumptions** (L211): Fire-and-forget reconnect POST means sends can arrive at the server before handleReconnect runs. handleSendV2 calls `watch()` (line 541 of ws-handler-v2.ts) but does NOT call `resetCursor()`, so the cursor starts at 0 for that connection+session. If a broadcast fires during the race window (between send and reconnect POST processing), the cursor advances. When handleReconnect then calls `resetCursor(lastSeq)`, it could roll the cursor backward, potentially causing duplicate event delivery on the next reconnect. Consider adding a `resetCursor` call in the handleSendV2 takeover/reattach path, or documenting why this is acceptable. `[fixable]`


_Sound simplification overall, but removing ReconnectMessage from the WS union breaks the WS transport fallback (MitzoConnection still sends reconnect over WS). The fire-and-forget design also introduces a subtle cursor race between send and reconnect POST arrival order._

- 🟡 **missing_tests**: The test for 'sends reconnect POST fire-and-forget on reconnect welcome' verifies the POST is sent, but there's no test for the case where a send is queued AND the reconnect POST races with it. Specifically: queue a send during reconnect, receive welcome, verify both reconnect and send are dispatched without ordering guarantees (the new 'flushes pending sends immediately' test covers this but doesn't test that the server can handle either arriving first).


_Sound simplification overall, but removing ReconnectMessage from the WS union breaks the WS transport fallback (MitzoConnection still sends reconnect over WS). The fire-and-forget design also introduces a subtle cursor race between send and reconnect POST arrival order._

- 🟡 **regressions**: The `connection.test.ts` tests at line 197 ('reconnects after close and sends hello + reconnect with tracked sessions') still assert that reconnect is sent over WS and expect it to work. These tests pass because they only check the client-side behavior (message is sent), but they're now testing a path that's broken server-side (reconnect over WS is silently dropped). These tests give false confidence. `[fixable]`


---
_Reviewed at fe4025746dc0d5048face40e238a7848f521a0ea_

packages/protocol/src/ws-schemas-v2.ts
```diff
@@ -124,9 +124,10 @@ export const V2SetModeMessage = z.object({
 
 // ─── Union ──────────────────────────────────────────────────────────────────
 
+// ReconnectMessage is handled via REST POST (not WS) — exported for
+// the REST handler but excluded from the WS union.
 export const IncomingWsMessageV2 = z.discriminatedUnion('type', [
```
dimakis, line 129: 🔴 **regressions**: ReconnectMessage removed from WS union breaks the WS transport fallback. `MitzoConnection` (packages/client/src/connection.ts:228) still sends `{ type: 'reconnect' }` over WS on reconnect. Since it's no longer in the discriminated union, the server's `dispatchV2Message` Zod parse rejects it and the message is silently dropped. This means WS clients get no event replay, no cursor reset, and no boot_context re-send on reconnect. Either update `MitzoConnection` to send reconnect via REST POST (like `SseConnection` does), or keep `ReconnectMessage` in the WS union with a no-op handler. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -950,7 +913,7 @@ export async function dispatchV2Message(
       // Already handled at routing layer, ignore duplicate
       break;
     case 'reconnect':
```
dimakis, line 969: 🔵 **style**: Dead code: `case 'reconnect':` can never be reached because `reconnect` was removed from the `IncomingWsMessageV2` union. The Zod parse on line 900 rejects reconnect messages before this switch. Remove this case to avoid confusion. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +204,21 @@ export class SseConnection implements ChatConnection {
...
+      // Fire reconnect POST (fire-and-forget) if reconnecting with sessions.
+      // No need to defer _connected — handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
       if (this._isReconnect && this.seqBySession.size > 0) {
-        this.doReconnectPost(welcomeConnectionId, welcomeEs);
-      } else {
-        this._connected = true;
-        this.flushPendingSends();
-        this.listener?.({ type: '_open' });
+        this.doPost('reconnect', {
```
dimakis: 🟡 **unsafe_assumptions**: Fire-and-forget reconnect POST means sends can arrive at the server before handleReconnect runs. handleSendV2 calls `watch()` (line 541 of ws-handler-v2.ts) but does NOT call `resetCursor()`, so the cursor starts at 0 for that connection+session. If a broadcast fires during the race window (between send and reconnect POST processing), the cursor advances. When handleReconnect then calls `resetCursor(lastSeq)`, it could roll the cursor backward, potentially causing duplicate event delivery on the next reconnect. Consider adding a `resetCursor` call in the handleSendV2 takeover/reattach path, or documenting why this is acceptable. `[fixable]`

dimakis:
Rebased onto main, fixed CI (dead reconnect case in WS handler switch, missing setSessionState mocks in tests, removed reconnect from protocol union test). Cherry-picked interrupt guard from #396. Ready for review.

dimakis:
## Centaur Review

### `server/index.ts`
_Clean simplification that removes ~400 lines of reconnect complexity. The main risk is the fire-and-forget reconnect POST: if it fails, the server never replays missed events or resets cursors, yet the client considers itself connected and flushes queued sends. The old code had explicit failure recovery (scheduleReconnect); the new code relies on EventSource auto-reconnect which only fires on connection loss, not POST failure._

- 🔵 **style** (L984): Stale comment: `// Stop periodic sync + clear state` — periodic sync was removed in this PR. Should be `// Clear connection state` or similar. `[fixable]`


_Clean simplification that removes ~400 lines of reconnect complexity. The main risk is the fire-and-forget reconnect POST: if it fails, the server never replays missed events or resets cursors, yet the client considers itself connected and flushes queued sends. The old code had explicit failure recovery (scheduleReconnect); the new code relies on EventSource auto-reconnect which only fires on connection loss, not POST failure._

- 🟡 **unsafe_assumptions** (L210): Fire-and-forget reconnect POST has no visibility into failure. If the POST to `/api/chat/reconnect` fails (network error, 500), the server never runs handleReconnect — meaning no EventStore replay, no cursor reset, no suspend resume. The client marks itself connected and flushes pending sends, but the server hasn't restored session state. Unlike before (where failure triggered scheduleReconnect), there's now no recovery path. EventSource auto-reconnect would fire a new welcome, but only if the connection actually drops — a POST failure alone won't trigger it. Consider at minimum logging the failure in doPost for reconnect calls. `[fixable]`
- 🟡 **regressions** (L218): Pending sends are now flushed immediately on welcome, before the reconnect POST completes. Previously, sends were held until the server confirmed the reconnect (cursor reset, replay, reattach). Now a queued `send` message can race with the reconnect POST — the server may process the `send` before `handleReconnect` has reset cursors and replayed missed events. This is mitigated by handleSendV2 doing ownership on first message, but the ordering guarantee (reconnect-then-send) is no longer enforced client-side.


_Clean simplification that removes ~400 lines of reconnect complexity. The main risk is the fire-and-forget reconnect POST: if it fails, the server never replays missed events or resets cursors, yet the client considers itself connected and flushes queued sends. The old code had explicit failure recovery (scheduleReconnect); the new code relies on EventSource auto-reconnect which only fires on connection loss, not POST failure._

- 🔵 **missing_tests**: The PR removed 12 tests covering reconnect failure/retry/race scenarios and added only 3 simplified replacements. There's no test verifying behavior when EventSource auto-reconnects (fires multiple welcome events rapidly) — duplicate reconnect POSTs could be sent. A test for the auto-reconnect scenario would increase confidence in the fire-and-forget design. `[fixable]`


_Clean simplification that removes ~400 lines of reconnect complexity. The main risk is the fire-and-forget reconnect POST: if it fails, the server never replays missed events or resets cursors, yet the client considers itself connected and flushes queued sends. The old code had explicit failure recovery (scheduleReconnect); the new code relies on EventSource auto-reconnect which only fires on connection loss, not POST failure._

- 🔵 **regressions** (L133): Comment says `reconnect replay will cover the gap` when broadcast send fails, but if the reconnect POST itself fails (fire-and-forget), there's no replay either. The comment is accurate for the normal case but understates the failure mode. `[fixable]`


_Clean simplification that removes ~400 lines of reconnect complexity. The main risk is the fire-and-forget reconnect POST: if it fails, the server never replays missed events or resets cursors, yet the client considers itself connected and flushes queued sends. The old code had explicit failure recovery (scheduleReconnect); the new code relies on EventSource auto-reconnect which only fires on connection loss, not POST failure._

- 🔵 **style** (L112): Added `setSessionState: vi.fn()` mock but no test in this file exercises it. Likely added to satisfy a TypeScript interface requirement — harmless but worth confirming it's needed.


---
_Reviewed at 263f7bcef7261478c803f6efcc7249478b77bf15_

dimakis:
All three findings addressed: restored WS reconnect in union + handler, documented cursor race as benign (single-threaded + client seq dedup), cherry-picked interrupt guard from #396. Please re-review.

dimakis:
## Centaur Review

_Solid simplification that removes ~300 lines of complex reconnect-POST-gating logic. The fire-and-forget approach is architecturally sound given EventStore replay and client-side seq dedup, but the silent failure mode of reconnect POSTs (no retry, no logging, no event replay on failure) deserves at minimum a warning log and documentation of the recovery path._

- 🟡 **unsafe_assumptions** (L219): Client marks `_connected = true` and flushes pending sends before the reconnect POST completes. If the POST fails (network error, server 500), the server never processes `handleReconnect` — no cursor reset, no event replay, no boot context re-send. The client thinks it's connected and sends messages into a session the server hasn't re-established context for. The comment in `doPost` says 'SSE stream will reconnect and replay' but that only applies to EventSource-level reconnect, not the reconnect POST itself. The 48h detach TTL means the session won't be killed, but the user may see a gap in replayed events until they trigger another reconnect cycle.
- 🟡 **unsafe_assumptions** (L210): Reconnect POST and pending sends (`flushPendingSends` at line 220) both fire as concurrent fire-and-forget POSTs. No ordering guarantee exists — the server may process a queued `send` before the `reconnect` POST. The comment at ws-handler-v2.ts:542-546 acknowledges this and claims Node's event loop prevents true interleaving, but HTTP requests from the same client can arrive on different TCP connections and be processed out of order at the Express layer. In practice, `handleSendV2` does call `watch()` and handles ownership independently, so this may be benign, but it's worth noting that cursor won't be set correctly until the reconnect POST arrives.
- 🔵 **style** (L266): The empty catch block in `doPost` (line 266-268) silently swallows all errors including reconnect POST failures. Consider at minimum a `console.warn` for reconnect failures specifically, since a failed reconnect POST means the server won't replay events or re-send boot context. The comment 'POST failures non-fatal' understates the impact for reconnect specifically. `[fixable]`


### `server/ws-handler-v2.ts`
_Solid simplification that removes ~300 lines of complex reconnect-POST-gating logic. The fire-and-forget approach is architecturally sound given EventStore replay and client-side seq dedup, but the silent failure mode of reconnect POSTs (no retry, no logging, no event replay on failure) deserves at minimum a warning log and documentation of the recovery path._

- 🟡 **regressions** (L255): Removing reattach from `handleReconnect` means a detached session stays detached until the user sends a message or interrupt. Previously, reconnect would reattach immediately, cancelling the detach timer (indirectly via `reattachChat`). Now, if a user reconnects and passively observes an agent turn in progress (no send/interrupt), the session remains detached. The 48h detach TTL (from session-registry.ts) is long enough that this is unlikely to cause session death in practice, but the agent's transport stays stale — new events flow via `connRegistry.broadcast()` (watch-based), not via the session's attached transport. If any code path uses the session's attached transport instead of broadcasting, those events would be lost.
- 🔵 **style** (L542): The 5-line comment block (542-546) explaining why `resetCursor` isn't called from `handleSendV2` references 'Node's single-threaded event loop prevents true interleaving' which is misleading — two HTTP requests (reconnect POST and send POST) can arrive as separate events on the event loop and interleave at the await boundaries. The actual safety net is client-side seq dedup (mentioned last), which deserves to be the lead explanation. `[fixable]`


_Solid simplification that removes ~300 lines of complex reconnect-POST-gating logic. The fire-and-forget approach is architecturally sound given EventStore replay and client-side seq dedup, but the silent failure mode of reconnect POSTs (no retry, no logging, no event replay on failure) deserves at minimum a warning log and documentation of the recovery path._

- 🔵 **missing_tests**: The tests removed 11 async reconnect scenarios (deferred connect, stale POST guard, POST failure recovery, queued send survival across retries, etc.) and replaced them with 3 simpler synchronous tests. While this matches the simplified fire-and-forget design, there's no test verifying that a failed reconnect POST doesn't prevent event replay via SSE — the core assumption that makes fire-and-forget safe. A test showing SSE events still arrive after reconnect POST failure would document this guarantee. `[fixable]`


---
_Reviewed at 66ef2cf18176f14a5889bfdc9b340d49adbc26fa_

dimakis (commented):
## Centaur Review

Found **7** issue(s) (1 critical) (5 warning).


_Critical bug: `doPost` doesn't check `res.ok`, so HTTP 500 from the reconnect endpoint silently clears the retry state — a regression from the old `doReconnectPost` which explicitly handled non-ok responses. Secondary issues: `_pendingReconnectSessions` is not cleared on `disconnect()` and never reconciled with `seqBySession`, and the new retry mechanism has zero test coverage._

- 🔴 **bugs** (L274): `doPost` does not check `res.ok`. `fetch()` only rejects on network errors — an HTTP 500 resolves the promise successfully. The reconnect POST `.then()` handler (line 224) clears `_pendingReconnectSessions` on resolve, so a server 500 is treated as success: cursor is never reset, replay never happens, and the retry mechanism is silently disarmed. The old `doReconnectPost` explicitly checked `res.ok` and called `scheduleReconnect()` on non-ok responses. Fix: check `res.ok` in `doPost` and throw on non-ok, or check it at the reconnect call site. `[fixable]`
- 🟡 **bugs** (L60): `disconnect()` does not clear `_pendingReconnectSessions`. If a reconnect POST failed (setting `_pendingReconnectSessions`), then the user disconnects and later reconnects, the stale sessions array will be retried on the next welcome — even if those sessions no longer exist locally (cleared via `clearSession`). Add `this._pendingReconnectSessions = null;` in `disconnect()`. `[fixable]`
- 🟡 **unsafe_assumptions** (L212): `_pendingReconnectSessions` is used as a retry payload but is never reconciled with `seqBySession`. If a session is cleared (`clearSession`) while `_pendingReconnectSessions` holds a reference to it, the next welcome retries a reconnect POST for a session the client no longer tracks. The stale entry will cause unnecessary server work and log noise. Consider filtering `_pendingReconnectSessions` against current `seqBySession` keys before retrying. `[fixable]`
- 🔵 **style** (L86): Bare `.catch(() => {})` on four `doPost` calls silently swallows errors. Now that `doPost` logs a `console.warn` before throwing, the catch is functionally correct, but the pattern is fragile — a future caller might assume `doPost` doesn't throw. Consider having `doPost` not throw (log-only) for fire-and-forget calls, and a separate `doPostOrThrow` for the reconnect path that needs error discrimination.


_Critical bug: `doPost` doesn't check `res.ok`, so HTTP 500 from the reconnect endpoint silently clears the retry state — a regression from the old `doReconnectPost` which explicitly handled non-ok responses. Secondary issues: `_pendingReconnectSessions` is not cleared on `disconnect()` and never reconciled with `seqBySession`, and the new retry mechanism has zero test coverage._

- 🟡 **missing_tests**: No test covers the HTTP 500 (non-ok) response case for reconnect POST. The old suite had two tests for this (`stays disconnected when reconnect POST fails`, `recovers after failed reconnect when EventSource auto-reconnects`). With the new fire-and-forget model, the behavior for non-ok responses is different from network errors — but no test verifies which path is taken. Given the `doPost` bug above, a 500 silently clears `_pendingReconnectSessions`. `[fixable]`
- 🟡 **missing_tests**: No test verifies the `_pendingReconnectSessions` retry-on-next-welcome mechanism — the key new behavior this PR introduces. A test should verify: (1) reconnect POST fails → `_pendingReconnectSessions` retained, (2) EventSource auto-reconnects → next welcome retries with the saved sessions, (3) success clears `_pendingReconnectSessions`. `[fixable]`


### `server/ws-handler-v2.ts`
_Critical bug: `doPost` doesn't check `res.ok`, so HTTP 500 from the reconnect endpoint silently clears the retry state — a regression from the old `doReconnectPost` which explicitly handled non-ok responses. Secondary issues: `_pendingReconnectSessions` is not cleared on `disconnect()` and never reconciled with `seqBySession`, and the new retry mechanism has zero test coverage._

- 🟡 **regressions** (L259): `handleReconnect` now only reattaches sessions where `ownerConnection === connectionId`. When a user's device gets a new connectionId (e.g. app restart), the session's clientId still has the old connectionId prefix. In this scenario, reconnect won't reattach (owner mismatch), and the agent's transport stays stale until the user sends a message (triggering `handleSendV2` takeover). During this gap, the agent may try to write to a dead transport — permission prompts, for example, would be lost. The old code handled this by checking `ownerGone` and reattaching eagerly. `[fixable]`


---
_Reviewed at 13153d68a7a9065ca1ff04e1d01f07c08cef02fe_

packages/client/src/sse-connection.ts
```diff
@@ -262,66 +271,6 @@ export class SseConnection implements ChatConnection {
...
-      this.es.close();
-      this.es = null;
-    }
-    this.reconnectTimer = setTimeout(() => {
-      this.reconnectTimer = null;
-      this.doConnect();
-    }, this.config.reconnectDelayMs);
-  }
-
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
```
dimakis, line 280: 🔴 **bugs**: `doPost` does not check `res.ok`. `fetch()` only rejects on network errors — an HTTP 500 resolves the promise successfully. The reconnect POST `.then()` handler (line 224) clears `_pendingReconnectSessions` on resolve, so a server 500 is treated as success: cursor is never reset, replay never happens, and the retry mechanism is silently disarmed. The old `doReconnectPost` explicitly checked `res.ok` and called `scheduleReconnect()` on non-ok responses. Fix: check `res.ok` in `doPost` and throw on non-ok, or check it at the reconnect call site. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -60,10 +59,6 @@ export class SseConnection implements ChatConnection {
 
   disconnect(): void {
```
dimakis, line 60: 🟡 **bugs**: `disconnect()` does not clear `_pendingReconnectSessions`. If a reconnect POST failed (setting `_pendingReconnectSessions`), then the user disconnects and later reconnects, the stale sessions array will be retried on the next welcome — even if those sessions no longer exist locally (cleared via `clearSession`). Add `this._pendingReconnectSessions = null;` in `disconnect()`. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +206,38 @@ export class SseConnection implements ChatConnection {
...
-      // external send() from bypassing the pending queue mid-reconnect.
-      // Capture both connectionId and ES instance for the staleness guard.
-      const welcomeConnectionId = this._connectionId;
-      const welcomeEs = this.es;
-      if (this._isReconnect && this.seqBySession.size > 0) {
-        this.doReconnectPost(welcomeConnectionId, welcomeEs);
+      // Fire reconnect POST if reconnecting with sessions, or retry a
+      // previously failed reconnect. handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
+      const sessions =
```
dimakis: 🟡 **unsafe_assumptions**: `_pendingReconnectSessions` is used as a retry payload but is never reconciled with `seqBySession`. If a session is cleared (`clearSession`) while `_pendingReconnectSessions` holds a reference to it, the next welcome retries a reconnect POST for a session the client no longer tracks. The stale entry will cause unnecessary server work and log noise. Consider filtering `_pendingReconnectSessions` against current `seqBySession` keys before retrying. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -88,12 +83,12 @@ export class SseConnection implements ChatConnection {
     if (!endpoint) return false;
 
     if (this._connected && this._connectionId) {
-      this.doPost(endpoint, msg);
+      this.doPost(endpoint, msg).catch(() => {});
```
dimakis, line 87: 🔵 **style**: Bare `.catch(() => {})` on four `doPost` calls silently swallows errors. Now that `doPost` logs a `console.warn` before throwing, the catch is functionally correct, but the pattern is fragile — a future caller might assume `doPost` doesn't throw. Consider having `doPost` not throw (log-only) for fire-and-forget calls, and a separate `doPostOrThrow` for the reconnect path that needs error discrimination.

server/ws-handler-v2.ts
```diff
@@ -247,49 +247,27 @@ export function handleReconnect(
...
-              if (found.clientId !== newClientId) {
-                rekeyChat(found.clientId, newClientId);
-                log.info('rekeyed session to new connection', {
-                  connectionId,
-                  sessionId: entry.sessionId,
-                  oldClientId: found.clientId,
-                  newClientId,
-                });
-              }
+        if (found && ctx.sessionRegistry.isActive(found.clientId)) {
```
dimakis, line 272: 🟡 **regressions**: `handleReconnect` now only reattaches sessions where `ownerConnection === connectionId`. When a user's device gets a new connectionId (e.g. app restart), the session's clientId still has the old connectionId prefix. In this scenario, reconnect won't reattach (owner mismatch), and the agent's transport stays stale until the user sends a message (triggering `handleSendV2` takeover). During this gap, the agent may try to write to a dead transport — permission prompts, for example, would be lost. The old code handled this by checking `ownerGone` and reattaching eagerly. `[fixable]`

packages/client/src/sse-connection.ts
dimakis, line 60: 🟡 **bugs**: `disconnect()` does not clear `_pendingReconnectSessions`. If `disconnect()` is called while a reconnect POST is in-flight, the `.then()` callback can still fire and call `flushPendingSends()` after the connection is torn down. Additionally, the stale sessions array persists in memory. Add `this._pendingReconnectSessions = null;` to `disconnect()`. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +206,38 @@ export class SseConnection implements ChatConnection {
...
+        (this._isReconnect && this.seqBySession.size > 0
+          ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+              sessionId,
+              lastSeq,
+            }))
+          : null);
+      this._connected = true;
+      if (sessions) {
+        this._pendingReconnectSessions = sessions;
+        this.doPost('reconnect', { type: 'reconnect', sessions }).then(
```
dimakis, line 226: 🟡 **unsafe_assumptions**: The reconnect POST's `.then()` callback clears `_pendingReconnectSessions` and flushes sends without checking whether `_connectionId` has changed since the POST was initiated. If a new welcome arrives while the POST is in-flight (EventSource auto-reconnect), the new welcome captures `_pendingReconnectSessions` (line 213), starts a second POST, and then the stale `.then()` clears the flag. This is benign only if the server handles duplicate reconnect POSTs with different connection IDs idempotently — which it does (handleReconnect is stateless per call) — but the double `flushPendingSends()` could cause sends to fire twice. Consider capturing `_connectionId` before the POST and comparing in the callback. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -247,49 +247,24 @@ export function handleReconnect(
...
+          const ownerConnection = found.clientId.split(':')[0];
```
dimakis: 🔵 **style**: Uses `found.clientId.split(':')[0]` while every other ownership check in this file (lines 519, 681, 819, 868) uses the `getOwnerConnection()` utility defined at line 93. Use `getOwnerConnection(found.clientId)` for consistency — they return the same result but the utility is the convention and would track any format changes. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -247,49 +247,24 @@ export function handleReconnect(
...
-              if (found.clientId !== newClientId) {
-                rekeyChat(found.clientId, newClientId);
-                log.info('rekeyed session to new connection', {
-                  connectionId,
-                  sessionId: entry.sessionId,
-                  oldClientId: found.clientId,
-                  newClientId,
-                });
-              }
+        if (found && ctx.sessionRegistry.isActive(found.clientId)) {
```
dimakis, line 272: 🟡 **regressions**: handleReconnect now only reattaches sessions owned by the reconnecting connection (`ownerConnection === connectionId`). Previously, it also handled cross-connection reattach when the owner was gone. This is deferred to handleSendV2, but creates a window where a non-owner reconnecting client sees events via broadcast while the agent's transport remains stale. Passive observers (watch-only, never send) will never trigger the handleSendV2 ownership path. This is an intentional design choice (documented in comments), but worth noting as a behavioral change.

dimakis (commented):
## Centaur Review

_Sound simplification — fire-and-forget reconnect and periodic sync removal are architecturally justified. Main concern is `_pendingReconnectSessions` not being cleared in `disconnect()`, which can cause post-teardown side effects, plus missing test coverage for the new retry mechanism._

- 🟡 **bugs** (L60): `disconnect()` does not clear `_pendingReconnectSessions`. If `disconnect()` is called while a reconnect POST is in-flight, the `.then()` callback can still fire and call `flushPendingSends()` after the connection is torn down. Additionally, the stale sessions array persists in memory. Add `this._pendingReconnectSessions = null;` to `disconnect()`. `[fixable]`
- 🟡 **unsafe_assumptions** (L223): The reconnect POST's `.then()` callback clears `_pendingReconnectSessions` and flushes sends without checking whether `_connectionId` has changed since the POST was initiated. If a new welcome arrives while the POST is in-flight (EventSource auto-reconnect), the new welcome captures `_pendingReconnectSessions` (line 213), starts a second POST, and then the stale `.then()` clears the flag. This is benign only if the server handles duplicate reconnect POSTs with different connection IDs idempotently — which it does (handleReconnect is stateless per call) — but the double `flushPendingSends()` could cause sends to fire twice. Consider capturing `_connectionId` before the POST and comparing in the callback. `[fixable]`


### `server/ws-handler-v2.ts`
_Sound simplification — fire-and-forget reconnect and periodic sync removal are architecturally justified. Main concern is `_pendingReconnectSessions` not being cleared in `disconnect()`, which can cause post-teardown side effects, plus missing test coverage for the new retry mechanism._

- 🔵 **style** (L260): Uses `found.clientId.split(':')[0]` while every other ownership check in this file (lines 519, 681, 819, 868) uses the `getOwnerConnection()` utility defined at line 93. Use `getOwnerConnection(found.clientId)` for consistency — they return the same result but the utility is the convention and would track any format changes. `[fixable]`
- 🟡 **regressions** (L259): handleReconnect now only reattaches sessions owned by the reconnecting connection (`ownerConnection === connectionId`). Previously, it also handled cross-connection reattach when the owner was gone. This is deferred to handleSendV2, but creates a window where a non-owner reconnecting client sees events via broadcast while the agent's transport remains stale. Passive observers (watch-only, never send) will never trigger the handleSendV2 ownership path. This is an intentional design choice (documented in comments), but worth noting as a behavioral change.


_Sound simplification — fire-and-forget reconnect and periodic sync removal are architecturally justified. Main concern is `_pendingReconnectSessions` not being cleared in `disconnect()`, which can cause post-teardown side effects, plus missing test coverage for the new retry mechanism._

- 🔵 **missing_tests**: No test verifies that `disconnect()` during an in-flight reconnect POST doesn't cause side effects (the old test `bails out if disconnect() called during in-flight reconnect POST` was removed). The fire-and-forget model changes the expected behavior, but there should be a test confirming that the `.then()` callback doesn't flush sends or emit `_open` after `disconnect()`. `[fixable]`
- 🔵 **missing_tests**: No test verifies the retry path: reconnect POST fails (setting `_pendingReconnectSessions`), then a new welcome arrives and retries with the stored sessions. The `_pendingReconnectSessions` retry mechanism is the key new behavior replacing the removed `scheduleReconnect`, but it has no dedicated test. `[fixable]`


---
_Reviewed at 1737898d5d9b6791a1edb802fbe54ff3d4c86e63_

dimakis:
## Centaur Review

### `server/ws-handler-v2.ts`
_The fire-and-forget reconnect simplification introduces send-before-reconnect race conditions predicated on a client-side seq dedup mechanism that does not exist — this is the critical gap that needs addressing before merge._

- 🔴 **bugs** (L566): Comment claims 'client-side seq dedup (store.ts) drops events with seq <= lastProcessedSeq' but this mechanism does not exist. No code in the client package (store.ts, protocol-parser.ts, sse-connection.ts, connection.ts) filters incoming events by seq. `seqBySession` only tracks the last seen seq for reconnect coordination — events are always forwarded unconditionally to the listener. When send POST arrives before reconnect POST (a real possibility since `_connected=true` fires immediately on welcome), `handleSendV2` calls `watch()` without `resetCursor()`, and subsequent broadcasts may deliver duplicate events to a client that cannot deduplicate them. `[fixable]`
- 🟡 **regressions** (L271): handleReconnect now reattaches without rekeying. When a device restarts (new connectionId), the session is reattached to the new transport but clientId stays as `old-conn:sess-1`. The subsequent handleSendV2 call will detect ownership mismatch and do a full takeover (reattach + rekey), causing a redundant double-reattach. More importantly, if the user never sends (passive observer), the stale clientId persists for the session's lifetime, and log entries will attribute agent activity to the old connection. `[fixable]`


_The fire-and-forget reconnect simplification introduces send-before-reconnect race conditions predicated on a client-side seq dedup mechanism that does not exist — this is the critical gap that needs addressing before merge._

- 🔴 **bugs** (L299): doPost comment states 'Client-side seq dedup prevents duplicate delivery' but no seq dedup exists on the client. This same false claim appears in ws-handler-v2.ts:566. The architectural assumption that duplicates are harmless is incorrect — the message reducer processes every event unconditionally, so duplicate block_delta/block_start events will corrupt the UI state (double text, duplicate tool blocks). `[fixable]`
- 🟡 **unsafe_assumptions** (L226): Setting `_connected = true` before the reconnect POST resolves means `send()` (line 86-88) can fire user messages to the server before handleReconnect runs. The queued sends in `flushPendingSends()` are correctly ordered (they wait for the .then), but NEW `send()` calls from the user bypass the queue entirely. The reconnect POST body comment (line 232-234) acknowledges ordering matters but only controls pre-queued sends, not live calls. `[fixable]`
- 🟡 **regressions** (L240): When the reconnect POST fails, `flushPendingSends()` is called immediately (line 240). These queued messages will POST to the server, but the server never ran handleReconnect for this cycle — no cursor reset, no replay. The server's handleSendV2 handles ownership independently, but the sends arrive without the context the reconnect would have established (boot_context replay, cursor positioning). Previously, a failed reconnect POST kept the client disconnected, naturally preventing premature sends.


_The fire-and-forget reconnect simplification introduces send-before-reconnect race conditions predicated on a client-side seq dedup mechanism that does not exist — this is the critical gap that needs addressing before merge._

- 🟡 **missing_tests**: No test covers the race where a user calls `send()` between `_connected = true` and reconnect POST completion. The existing 'flushes pending sends after reconnect POST completes' test only covers pre-queued sends. A test should verify that a live `send()` call during the reconnect POST window does not cause ordering violations. `[fixable]`


### `server/index.ts`
_The fire-and-forget reconnect simplification introduces send-before-reconnect race conditions predicated on a client-side seq dedup mechanism that does not exist — this is the critical gap that needs addressing before merge._

- 🔵 **style** (L984): Stale comment: `connRegistry.dispose(); // Stop periodic sync + clear state` — periodic sync was removed in this PR. Four additional stale 'periodic sync' references remain in connection-registry.ts:113, connection-registry.test.ts:222/:281, and ws-handler-v2.ts:412. `[fixable]`


_The fire-and-forget reconnect simplification introduces send-before-reconnect race conditions predicated on a client-side seq dedup mechanism that does not exist — this is the critical gap that needs addressing before merge._

- 🔵 **missing_tests**: The deleted 'handleReconnect ownership guard' test block covered the case where a different connection owns the session AND is still alive. The old behavior correctly refused reattach. The new code (ws-handler-v2.ts:263-264) checks `ownerGone` but there's no test verifying that reattach is skipped when the original owner is still registered in connRegistry. `[fixable]`


---
_Reviewed at 76c40403c5fe4ec688fedb87b8c0f978425fa8a4_

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +207,43 @@ export class SseConnection implements ChatConnection {
...
-      const welcomeConnectionId = this._connectionId;
-      const welcomeEs = this.es;
-      if (this._isReconnect && this.seqBySession.size > 0) {
-        this.doReconnectPost(welcomeConnectionId, welcomeEs);
+      // Fire reconnect POST if reconnecting with sessions, or retry a
+      // previously failed reconnect. handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
+      // Filter pending retries against current seqBySession — sessions may
+      // have been cleared (clearSession) since the retry was queued.
+      const pending = this._pendingReconnectSessions?.filter((s) =>
```
dimakis: 🟡 **bugs**: When `_pendingReconnectSessions` retries on the next welcome, the stored `lastSeq` values are from the original failed attempt. Between failure and retry, `seqBySession` may have advanced via `trackSeq()` as SSE events arrived. The filter (line 215-216) only checks `seqBySession.has()` for existence but doesn't refresh `lastSeq` from the current map. Consider replacing the pending entries' `lastSeq` with current values: `pending.map(s => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }))`. Functionally safe due to seq dedup but causes unnecessary replay of events the client already has. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +207,43 @@ export class SseConnection implements ChatConnection {
...
+      );
+      const sessions =
+        (pending && pending.length > 0 ? pending : null) ??
+        (this._isReconnect && this.seqBySession.size > 0
+          ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+              sessionId,
+              lastSeq,
+            }))
+          : null);
+      this._connected = true;
```
dimakis: 🔵 **unsafe_assumptions**: Setting `_connected = true` before the reconnect POST completes means `send()` (line 86-88) will fire user messages immediately via `doPost()`, potentially arriving at the server before the reconnect POST. The comment at ws-handler-v2.ts:563-569 documents this is safe due to client-side seq dedup, but there is no server-side ordering guarantee. If a user message arrives before `handleReconnect` sets up the cursor, `broadcast()` events may be delivered at cursor=0 until the reconnect POST arrives. This is self-healing but could cause a burst of duplicate events.

server/ws-handler-v2.ts
```diff
@@ -247,48 +247,31 @@ export function handleReconnect(
...
+          const ownerConnection = found.clientId.split(':')[0];
```
dimakis: 🔵 **style**: Uses `found.clientId.split(':')[0]` instead of the existing `getOwnerConnection(found.clientId)` helper (defined at line 93 and used elsewhere in the same file, e.g. lines 527, 689, 827). Inconsistent — use the helper for uniformity and to benefit from its `indexOf` fast-path. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -247,48 +247,31 @@ export function handleReconnect(
...
-              if (found.clientId !== newClientId) {
-                rekeyChat(found.clientId, newClientId);
-                log.info('rekeyed session to new connection', {
-                  connectionId,
-                  sessionId: entry.sessionId,
-                  oldClientId: found.clientId,
-                  newClientId,
-                });
-              }
+        if (found && ctx.sessionRegistry.isActive(found.clientId)) {
```
dimakis, line 272: 🔵 **regressions**: Reconnect no longer calls `rekeyChat()` — ownership transfer is deferred to `handleSendV2`. Between reconnect and first send, the session's `clientId` still references the old connectionId while the transport points to the new connection. This is fine for event delivery (transport is reattached), but if the agent emits logs or metrics using `clientId`, they'll reference the stale connectionId until the first send triggers rekey. This is a deliberate design choice (documented in the test renames) — flagging for awareness, not as a bug.

dimakis (commented):
## Centaur Review

_Clean simplification that removes periodic sync and defers ownership operations to handleSendV2. Architecturally sound — seq dedup guarantees correctness. One minor inefficiency where retry reconnect uses stale lastSeq, and one style inconsistency with getOwnerConnection._

- 🟡 **bugs** (L215): When `_pendingReconnectSessions` retries on the next welcome, the stored `lastSeq` values are from the original failed attempt. Between failure and retry, `seqBySession` may have advanced via `trackSeq()` as SSE events arrived. The filter (line 215-216) only checks `seqBySession.has()` for existence but doesn't refresh `lastSeq` from the current map. Consider replacing the pending entries' `lastSeq` with current values: `pending.map(s => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }))`. Functionally safe due to seq dedup but causes unnecessary replay of events the client already has. `[fixable]`
- 🔵 **unsafe_assumptions** (L226): Setting `_connected = true` before the reconnect POST completes means `send()` (line 86-88) will fire user messages immediately via `doPost()`, potentially arriving at the server before the reconnect POST. The comment at ws-handler-v2.ts:563-569 documents this is safe due to client-side seq dedup, but there is no server-side ordering guarantee. If a user message arrives before `handleReconnect` sets up the cursor, `broadcast()` events may be delivered at cursor=0 until the reconnect POST arrives. This is self-healing but could cause a burst of duplicate events.


### `server/ws-handler-v2.ts`
_Clean simplification that removes periodic sync and defers ownership operations to handleSendV2. Architecturally sound — seq dedup guarantees correctness. One minor inefficiency where retry reconnect uses stale lastSeq, and one style inconsistency with getOwnerConnection._

- 🔵 **style** (L260): Uses `found.clientId.split(':')[0]` instead of the existing `getOwnerConnection(found.clientId)` helper (defined at line 93 and used elsewhere in the same file, e.g. lines 527, 689, 827). Inconsistent — use the helper for uniformity and to benefit from its `indexOf` fast-path. `[fixable]`
- 🔵 **regressions** (L259): Reconnect no longer calls `rekeyChat()` — ownership transfer is deferred to `handleSendV2`. Between reconnect and first send, the session's `clientId` still references the old connectionId while the transport points to the new connection. This is fine for event delivery (transport is reattached), but if the agent emits logs or metrics using `clientId`, they'll reference the stale connectionId until the first send triggers rekey. This is a deliberate design choice (documented in the test renames) — flagging for awareness, not as a bug.


_Clean simplification that removes periodic sync and defers ownership operations to handleSendV2. Architecturally sound — seq dedup guarantees correctness. One minor inefficiency where retry reconnect uses stale lastSeq, and one style inconsistency with getOwnerConnection._

- 🔵 **missing_tests**: No test verifies that `_pendingReconnectSessions` are filtered correctly when a session is cleared via `clearSession()` between the failed reconnect POST and the retry welcome. The filter at line 215 depends on this, and a test like 'retried reconnect excludes sessions cleared between attempts' would validate the guard. `[fixable]`

dimakis:
## Centaur Review

_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🟡 **bugs** (L228): Setting `_connected = true` before the reconnect POST completes means `send()` calls between welcome and POST completion bypass the pending queue and hit the server directly. This can cause user messages to arrive before `handleReconnect` runs (cursor reset, replay), violating the ordering invariant documented in lines 239-240. The old code deferred `_connected` until POST success specifically to prevent this. The comment on line 239 says "Flush pending sends AFTER reconnect so the server processes handleReconnect (cursor reset, replay) before user messages" — but only *queued* sends are gated behind the POST; new `send()` calls during the POST's in-flight window go straight through. `[fixable]`
- 🟡 **regressions** (L248): `flushPendingSends()` is called in *both* the `.then()` and `.catch()` callbacks of the reconnect POST. When the POST fails (server never ran `handleReconnect` — no cursor reset, no replay), flushing sends means user messages are delivered without the server having set up the session watch/cursor. Previously, failed POSTs left sends queued and scheduled a delayed reconnect. The comment says "handleSendV2 handles ownership independently" but the concern is cursor state and replay, not ownership. `[fixable]`


_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🟡 **missing_tests** (L366): The test 'flushes pending sends after reconnect POST completes' only verifies sends queued *before* the welcome are flushed after the POST. There is no test for a `send()` called *after* welcome (when `_connected = true`) but while the reconnect POST is still in-flight — this is the race window from the bug above. If the design intentionally allows this (relying on server-side seq dedup), a test should document that intent. `[fixable]`


_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🔵 **style** (L113): Stale comment: "Updates delivery cursor on success so periodic sync can retry failures" — periodic sync has been removed. Should say something like "Updates delivery cursor on success so reconnect replay covers the correct range." `[fixable]`


### `server/index.ts`
_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🔵 **style** (L984): Stale comment: `connRegistry.dispose(); // Stop periodic sync + clear state` — periodic sync is removed, comment should just say `// Clear state`. `[fixable]`


_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🔵 **style** (L222): Two test comments still reference periodic sync (lines ~222 and ~281): "can't inspect directly, but periodic sync will use it" and "verified by periodic sync behavior". These should be updated since periodic sync no longer exists. `[fixable]`


### `server/ws-handler-v2.ts`
_The fire-and-forget reconnect simplification removes significant complexity but introduces a send-ordering race (sends can bypass the reconnect POST) and a transport/clientId mismatch window — both depend on client-side seq dedup for correctness, which works but should be tested and documented._

- 🟡 **unsafe_assumptions** (L563): The comment on `handleSendV2` says "No resetCursor here — handleReconnect (fire-and-forget POST) sets cursor to lastSeq when it arrives." But the reconnect POST and send POST arrive on separate event loop ticks in any order. If `send` arrives first and the server calls `watch()` (line 562) + `broadcast()` before the reconnect POST arrives, the cursor starts at 0 (default), so broadcasts of events the client already has will be delivered. The comment says "duplicate delivery is always harmless thanks to seq dedup" — this is true for correctness, but it's worth noting this is a bandwidth/performance trade-off, not zero-cost.
- 🟡 **bugs** (L271): Reconnect reattaches detached sessions (refreshes transport) but no longer rekeys the clientId. If the reconnecting connection has a different ID than the original owner (device restart: `ownerGone = true`), the clientId stays as `old-conn:sess-1` while the transport belongs to `new-conn`. Subsequent `handleSendV2` will rekey on the first send, but between reattach and first send, `getOwnerConnection(found.clientId)` returns `old-conn` — any code checking ownership during this window sees a stale owner. The comment says ownership is deferred to `handleSendV2`, but the reattach here creates an intermediate state where transport and clientId disagree. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -262,80 +287,28 @@ export class SseConnection implements ChatConnection {
...
-      this.es = null;
-    }
-    this.reconnectTimer = setTimeout(() => {
-      this.reconnectTimer = null;
-      this.doConnect();
-    }, this.config.reconnectDelayMs);
-  }
-
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
     if (!this._connectionId) return;
```
dimakis: 🟡 **bugs**: doPost silently resolves (doesn't throw) when _connectionId is null. When called from the reconnect .then() chain (line 236), if _connectionId were somehow cleared between the welcome event and the fetch, the success handler would fire without the POST ever being sent — marking _connected=true and flushing sends into the void. In practice this path is unreachable (connectionId is set at line 208 before the POST), but the contract is misleading: callers assume doPost either succeeds or throws, yet this early return is a silent success. `[fixable]`

dimakis (commented):
## Centaur Review

_Well-structured simplification that replaces periodic sync with on-demand reconnect replay. The stale-callback guard and zombie-deferral design are sound, but the stale-callback race and zombie-reattach window deserve test coverage, and the doPost silent-resolve on null connectionId is a subtle contract mismatch._

- 🟡 **bugs** (L291): doPost silently resolves (doesn't throw) when _connectionId is null. When called from the reconnect .then() chain (line 236), if _connectionId were somehow cleared between the welcome event and the fetch, the success handler would fire without the POST ever being sent — marking _connected=true and flushing sends into the void. In practice this path is unreachable (connectionId is set at line 208 before the POST), but the contract is misleading: callers assume doPost either succeeds or throws, yet this early return is a silent success. `[fixable]`
- 🔵 **style** (L220): The session-selection logic (lines 217-227) is dense: a null-coalescing chain of filtered/mapped pending sessions falling back to a ternary for fresh sessions. Consider extracting to a private method like `getReconnectSessions()` for readability. `[fixable]`


### `server/ws-handler-v2.ts`
_Well-structured simplification that replaces periodic sync with on-demand reconnect replay. The stale-callback guard and zombie-deferral design are sound, but the stale-callback race and zombie-reattach window deserve test coverage, and the doPost silent-resolve on null connectionId is a subtle contract mismatch._

- 🟡 **unsafe_assumptions** (L571): The comment explains that between watch() (line 570) and the reconnect POST arriving, the cursor starts at 0, so broadcasts may deliver duplicates. This relies entirely on client-side seq dedup. If a client implementation doesn't dedup (e.g. a future non-Zustand consumer), it will see duplicate events. The comment documents the trade-off well, but the gap is structural — consider adding a TODO for P4 to unify cursor reset.
- 🟡 **regressions** (L267): handleReconnect now reattaches detached sessions without checking EventStore state. Previously, it cross-referenced storeState (ENDED/CLOSING) to detect zombies and called registry.remove() before attempting reattach. Now it reattaches any active+detached session unconditionally. If a session's EventStore state is ENDED but the SessionRegistry hasn't been cleaned up yet (race window), handleReconnect will reattach a zombie. The zombie is only caught later if the user sends a message (handleSendV2 line 586). For observe-only reconnects (no send), the zombie stays reattached until the next tick or server restart. `[fixable]`


_Well-structured simplification that replaces periodic sync with on-demand reconnect replay. The stale-callback guard and zombie-deferral design are sound, but the stale-callback race and zombie-reattach window deserve test coverage, and the doPost silent-resolve on null connectionId is a subtle contract mismatch._

- 🟡 **missing_tests**: The stale-callback guard for racing welcome events is no longer tested. The old test 'ignores stale reconnect POST when a newer welcome arrives' verified that when two welcome events arrive in quick succession, only the second (current) POST's callback marks connected. The new code still has this guard (line 238/245: connectionId !== postConnectionId), but no test exercises it. This was a real-world race condition worth covering. `[fixable]`
- 🔵 **missing_tests**: No test verifies that SSE events (es.onmessage) update seqBySession while the reconnect POST is in-flight. The old test 'dispatches SSE events to listener while reconnect POST is in-flight' was removed. The new code refreshes lastSeq from seqBySession on retry (line 219), so verifying that SSE events advance the seq map during the in-flight window is important for the retry-with-fresh-seq guarantee. `[fixable]`


### `packages/protocol/src/ws-schemas-v2.ts`
_Well-structured simplification that replaces periodic sync with on-demand reconnect replay. The stale-callback guard and zombie-deferral design are sound, but the stale-callback race and zombie-reattach window deserve test coverage, and the doPost silent-resolve on null connectionId is a subtle contract mismatch._

- 🔵 **style** (L127): The added comment says 'WS clients still send it over WS until P4 removes the WS chat transport' — this is a phase-plan reference that will become stale. Consider phrasing it as current behavior ('WS clients may also send reconnect over WS') without the roadmap reference. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,13 +207,48 @@ export class SseConnection implements ChatConnection {
...
+      // previously failed reconnect. handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
+      // Filter pending retries against current seqBySession — sessions may
+      // have been cleared (clearSession) since the retry was queued.
+      // Refresh lastSeq from current map — SSE events may have advanced it
+      // since the original failure, avoiding unnecessary replay.
+      const pending = this._pendingReconnectSessions
+        ?.filter((s) => this.seqBySession.has(s.sessionId))
+        .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
+      const sessions =
```
dimakis: 🔵 **style**: The session-selection logic (lines 217-227) is dense: a null-coalescing chain of filtered/mapped pending sessions falling back to a ternary for fresh sessions. Consider extracting to a private method like `getReconnectSessions()` for readability. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -576,6 +568,14 @@ export function handleSendV2(
             }
             applySkillPolicy(activeClientId);
             ctx.connRegistry.watch(connectionId, sessionId);
+            // No resetCursor here — handleReconnect (fire-and-forget POST) sets
```
dimakis, line 585: 🟡 **unsafe_assumptions**: The comment explains that between watch() (line 570) and the reconnect POST arriving, the cursor starts at 0, so broadcasts may deliver duplicates. This relies entirely on client-side seq dedup. If a client implementation doesn't dedup (e.g. a future non-Zustand consumer), it will see duplicate events. The comment documents the trade-off well, but the gap is structural — consider adding a TODO for P4 to unify cursor reset.

server/ws-handler-v2.ts
```diff
@@ -247,48 +247,39 @@ export function handleReconnect(
...
-          log.info('removing stale session from registry (state-based)', {
-            connectionId,
-            sessionId: entry.sessionId,
-            clientId: found!.clientId,
-            storeState,
-          });
-          ctx.sessionRegistry.remove(found!.clientId);
-        }
-        if (found && running && !ctx.sessionRegistry.isAttached(found.clientId)) {
+        if (found && ctx.sessionRegistry.isActive(found.clientId)) {
```
dimakis, line 272: 🟡 **regressions**: handleReconnect now reattaches detached sessions without checking EventStore state. Previously, it cross-referenced storeState (ENDED/CLOSING) to detect zombies and called registry.remove() before attempting reattach. Now it reattaches any active+detached session unconditionally. If a session's EventStore state is ENDED but the SessionRegistry hasn't been cleaned up yet (race window), handleReconnect will reattach a zombie. The zombie is only caught later if the user sends a message (handleSendV2 line 586). For observe-only reconnects (no send), the zombie stays reattached until the next tick or server restart. `[fixable]`

packages/protocol/src/ws-schemas-v2.ts
```diff
@@ -124,6 +124,8 @@ export const V2SetModeMessage = z.object({
 
 // ─── Union ──────────────────────────────────────────────────────────────────
 
+// ReconnectMessage is primarily handled via REST POST (SSE transport), but WS
```
dimakis, line 127: 🔵 **style**: The added comment says 'WS clients still send it over WS until P4 removes the WS chat transport' — this is a phase-plan reference that will become stale. Consider phrasing it as current behavior ('WS clients may also send reconnect over WS') without the roadmap reference. `[fixable]`

PR #445, opened 2026-07-25
refactor(client): remove client-side message queueing (P2)

dimakis:
## Summary

- Remove `wasRunning`, `pendingSend`, `PENDING_SEND_TIMEOUT_MS`, and `clearPendingSendTimer()` from the client store
- Remove `onSendQueued` callback and `pendingSend` from `ProtocolParserState`
- `sendMessage()` now always sends immediately — no branching on running state
- Server-side `storeAndEchoIfNew()` deduplicates via `clientMsgId`; SDK's `inputQueue` handles follow-up queueing

**Depends on:** #440 (P3)

**Net change:** -153 lines across 4 files (2 source, 2 test)

## Test plan

- [x] `npx tsc -p packages/client/tsconfig.json --noEmit` passes
- [x] `npx vitest run packages/client` — 335 tests pass
- [ ] Verify send-while-running sends immediately and server echoes once
- [ ] Verify rapid follow-up messages queue in SDK inputQueue without protocol errors
- [ ] iOS: verify no UI state mismatch after reconnect (relies on P3 session_state_changed replay)

### DIFF
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -47,9 +47,6 @@ export interface ProtocolCallbacks {
   /** @deprecated v1 only — called to mark the WS pool entry as running/not-running. */
   setWsRunning?(poolKey: string, running: boolean): void;
 
-  /** @deprecated v1 only — called to send a queued message after session_end. */
-  sendQueued?(poolKey: string, msg: unknown): void;
-
   /** v2: Called with token data from session_switched response. */
   onTokensHydrated?(tokens: Record<string, unknown>): void;
 
--- a/packages/client/src/protocol-parser.ts
+++ b/packages/client/src/protocol-parser.ts
@@ -50,9 +50,6 @@ export interface ProtocolCallbacks {
   /** @deprecated v1 only — called to send a queued message after session_end. */
   sendQueued?(poolKey: string, msg: unknown): void;
 
-  /** v2: Called when a queued message should be sent after session_end. */
-  onSendQueued?(msg: Record<string, unknown>): void;
-
   /** v2: Called with token data from session_switched response. */
   onTokensHydrated?(tokens: Record<string, unknown>): void;
 
@@ -65,9 +62,6 @@ export interface ProtocolCallbacks {
 export interface ProtocolParserState {
   /** Currently tracked session ID (used for expiry detection). */
   currentSessionId: string | undefined;
-
-  /** Queued messages to send after current session ends (FIFO). */
-  pendingSend: Record<string, unknown>[];
 }
 
 // ─── Parser result ───────────────────────────────────────────────────────────
@@ -364,18 +358,6 @@ export function parseServerMessage(
       if (msg.sessionId && !state.currentSessionId) {
         callbacks.onSessionAssigned(msg.sessionId as string);
       }
-      // Drain first queued message. Optimistic running=true avoids UI flicker
-      // between the send and the server's session_state_changed confirmation.
-      const pending = state.pendingSend.shift();
-      if (pending) {
-        result.messagesActions.push({ type: 'SESSION_STATE_CHANGED', state: 'running' });
-        if (callbacks.onSendQueued) {
-          callbacks.onSendQueued(pending);
-        } else {
-          callbacks.setWsRunning?.(poolKey, true);
-          callbacks.sendQueued?.(poolKey, pending);
-        }
-      }
       break;
     }
 
@@ -450,7 +432,6 @@ export function parseServerMessage(
       const errorMsg = msg.error as string;
 
       callbacks.setWsRunning?.(poolKey, false);
-      state.pendingSend = [];
       result.messagesActions.push({
         type: 'ERROR',
         error: errorMsg || 'Unknown error',
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -232,7 +232,14 @@ export class SseConnection implements ChatConnection {
         return;
       }
 
+      // Seq-based dedup: if the reconnect POST fails, the server may re-deliver
+      // events using a stale cursor. The frontend reducer is not block-level
+      // idempotent (BLOCK_DELTA concatenates), so skip already-seen events.
       if (typeof msg.seq === 'number' && typeof msg.sessionId === 'string') {
+        const lastSeq = this.seqBySession.get(msg.sessionId as string);
+        if (lastSeq !== undefined && (msg.seq as number) <= lastSeq) {
+          return; // Already seen — skip duplicate
+        }
         this.seqBySession.set(msg.sessionId as string, msg.seq as number);
       }
 
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -23,7 +23,6 @@ export interface SseConnectionConfig {
   fetch: (url: string, init?: RequestInit) => Promise<Response>;
   /** Factory for EventSource — allows injection for testing. */
   createEventSource?: (url: string) => EventSource;
-  reconnectDelayMs?: number;
   /** URL for the sendBeacon suspend fallback. */
   suspendUrl?: string;
 }
@@ -38,7 +37,6 @@ export class SseConnection implements ChatConnection {
   private listener: ConnectionListener | null = null;
   private seqBySession = new Map<string, number>();
   private pendingSends: Array<{ endpoint: string; body: Record<string, unknown> }> = [];
-  private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
   private boundOnVisibility: (() => void) | null = null;
   private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
   private boundOnPageHide: (() => void) | null = null;
@@ -47,7 +45,6 @@ export class SseConnection implements ChatConnection {
   constructor(config: SseConnectionConfig) {
     this.config = {
       createEventSource: (url: string) => new EventSource(url),
-      reconnectDelayMs: 500,
       suspendUrl: '',
       ...config,
     };
@@ -60,10 +57,6 @@ export class SseConnection implements ChatConnection {
 
   disconnect(): void {
     this.removeBrowserListeners();
-    if (this.reconnectTimer) {
-      clearTimeout(this.reconnectTimer);
-      this.reconnectTimer = null;
-    }
     if (this.es) {
       this.es.close();
       this.es = null;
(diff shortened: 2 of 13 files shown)


dimakis:
## Centaur Review

### `server/index.ts`
_Clean simplification that correctly moves message queueing to the server. Two stale code artifacts (comment, dead switch case) and a minor gap in test coverage for the fire-and-forget reconnect race, but no bugs or behavioral regressions._

- 🔵 **style** (L984): Stale comment: `// Stop periodic sync + clear state` — periodic sync was removed in this PR. The `dispose()` method now only clears connections and cursors. Should read `// Clear connection state`. `[fixable]`


### `server/ws-handler-v2.ts`
_Clean simplification that correctly moves message queueing to the server. Two stale code artifacts (comment, dead switch case) and a minor gap in test coverage for the fire-and-forget reconnect race, but no bugs or behavioral regressions._

- 🟡 **regressions** (L938): `case 'reconnect':` is dead code after this PR. `ReconnectMessage` was removed from `IncomingWsMessageV2` (ws-schemas-v2.ts), so `safeParse` at line 923 rejects reconnect messages before they reach the switch. The case can never match. Either remove the case entirely or add a comment explaining it's retained for a future cleanup (P4). `[fixable]`


_Clean simplification that correctly moves message queueing to the server. Two stale code artifacts (comment, dead switch case) and a minor gap in test coverage for the fire-and-forget reconnect race, but no bugs or behavioral regressions._

- 🟡 **unsafe_assumptions**: The reconnect POST is now fire-and-forget, and `flushPendingSends()` runs immediately without waiting for it. If the reconnect POST arrives at the server AFTER a flushed send, `handleSendV2` processes the user message before `handleReconnect` sets up cursors/replay. The inline comment at ws-handler-v2.ts:558-564 documents that client-side seq dedup makes this safe, but there is no integration-level test covering this race (send arriving before reconnect POST). Consider adding one to prevent regressions if the seq dedup behavior changes.


### `packages/client/__tests__/sse-connection.test.ts`
_Clean simplification that correctly moves message queueing to the server. Two stale code artifacts (comment, dead switch case) and a minor gap in test coverage for the fire-and-forget reconnect race, but no bugs or behavioral regressions._

- 🔵 **missing_tests**: The PR removes ~300 lines of reconnect-POST-gating tests (deferred connection, stale POST guards, retry-on-failure, queued-sends-survive-failure). The replacement tests cover the happy path well, but there's no test verifying that `flushPendingSends` is called even when `doPost('reconnect', ...)` rejects (the fire-and-forget error path). The current implementation calls `doPost` without awaiting so the flush always runs, but a future refactor could accidentally gate it. `[fixable]`


---
_Reviewed at c00e233477136ec78e9b823d67519e8368ee9b2c_

dimakis:
## Centaur Review

_Clean simplification that correctly shifts queueing to the server. One potential race between fire-and-forget reconnect POST and immediate pending-send flush deserves consideration, though handleSendV2's independent watch() call mitigates it in practice. A few stale comments reference the removed periodic sync._

- 🟡 **unsafe_assumptions** (L210): Race condition: reconnect POST (line 211) and flushPendingSends (line 220) are both fire-and-forget. If a queued send arrives at the server before the reconnect POST is processed, the server hasn't called connRegistry.watch() yet for that session, so broadcast() won't route new events to this connection until handleSendV2 independently calls watch(). In practice this is mitigated because handleSendV2 also calls watch() (line 541/561 in ws-handler-v2.ts), but there's a window where events generated between the send and the watch setup are lost. The old code serialized these by awaiting the reconnect POST before flushing. `[fixable]`
- 🟡 **regressions**: reconnectDelayMs was removed from SseConnectionConfig but still exists in MitzoConnectionConfig (packages/client/src/connection.ts:15). This is intentional (WS vs SSE), but callers that pass reconnectDelayMs to SseConnection configs will now have it silently ignored since it's no longer in the interface. If any call site passes it, TypeScript will flag a compile error — verify no call sites outside the test files set this property.


_Clean simplification that correctly shifts queueing to the server. One potential race between fire-and-forget reconnect POST and immediate pending-send flush deserves consideration, though handleSendV2's independent watch() call mitigates it in practice. A few stale comments reference the removed periodic sync._

- 🔵 **style** (L113): Stale JSDoc: "Updates delivery cursor on success so periodic sync can retry failures" — periodic sync was removed in this PR. Should read something like "so reconnect replay can cover gaps." `[fixable]`


_Clean simplification that correctly shifts queueing to the server. One potential race between fire-and-forget reconnect POST and immediate pending-send flush deserves consideration, though handleSendV2's independent watch() call mitigates it in practice. A few stale comments reference the removed periodic sync._

- 🔵 **missing_tests**: The old test suite thoroughly covered reconnect failure/retry scenarios (POST failure → stays disconnected → retries on next welcome). The new fire-and-forget model has only one failure test ('POST failure does not affect connection state'). Consider adding a test verifying that if the reconnect POST fails, SSE events are still dispatched correctly to the listener (this was previously tested in 'dispatches SSE events to listener while reconnect POST is in-flight' which was deleted). `[fixable]`


_Clean simplification that correctly shifts queueing to the server. One potential race between fire-and-forget reconnect POST and immediate pending-send flush deserves consideration, though handleSendV2's independent watch() call mitigates it in practice. A few stale comments reference the removed periodic sync._

- 🔵 **style** (L50): The @deprecated v1 callback `sendQueued` (line 50) is still declared in ProtocolCallbacks, but the only v2 caller (onSendQueued) was removed in this PR. The v1 path that used sendQueued is also dead — all session_end queue-drain code was removed. Consider cleaning up this deprecated declaration if v1 is no longer supported. `[fixable]`


---
_Reviewed at c99a9da1e75722e86836b3dfe65d9ba5473f1124_

packages/client/src/sse-connection.ts
```diff
@@ -217,18 +204,21 @@ export class SseConnection implements ChatConnection {
...
-        this.listener?.({ type: '_open' });
+        this.doPost('reconnect', {
+          type: 'reconnect',
+          sessions: Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+            sessionId,
+            lastSeq,
+          })),
+        });
       }
+      this._connected = true;
```
dimakis, line 219: 🟡 **unsafe_assumptions**: Fire-and-forget reconnect POST means the server never resets the cursor (via `resetCursor` in `handleReconnect`) if the POST fails. Since the client marks `_connected = true` immediately and flushes pending sends, the server's `broadcast()` will use the stale cursor from the previous connection and may re-deliver events. The frontend's message reducer is NOT idempotent at the block level — `BLOCK_START` appends to `blockOrder` without dedup, and `BLOCK_DELTA` concatenates text unconditionally, so duplicate delivery causes visible corruption (doubled text, duplicated blocks). The old code gated `_connected` on POST success specifically to prevent this. Consider adding seq-based dedup in the `onmessage` handler (skip events with `seq <= seqBySession.get(sessionId)`) as a safety net. `[fixable]`

dimakis (commented):
## Centaur Review

_Clean removal of client-side queueing with correct server-side dedup via storeAndEchoIfNew, but the fire-and-forget reconnect POST introduces a duplicate-event risk when the POST fails (frontend reducer is not block-level idempotent) and a detach-TTL race window before the first user message reattaches the session._

- 🟡 **unsafe_assumptions** (L219): Fire-and-forget reconnect POST means the server never resets the cursor (via `resetCursor` in `handleReconnect`) if the POST fails. Since the client marks `_connected = true` immediately and flushes pending sends, the server's `broadcast()` will use the stale cursor from the previous connection and may re-deliver events. The frontend's message reducer is NOT idempotent at the block level — `BLOCK_START` appends to `blockOrder` without dedup, and `BLOCK_DELTA` concatenates text unconditionally, so duplicate delivery causes visible corruption (doubled text, duplicated blocks). The old code gated `_connected` on POST success specifically to prevent this. Consider adding seq-based dedup in the `onmessage` handler (skip events with `seq <= seqBySession.get(sessionId)`) as a safety net. `[fixable]`


### `server/ws-handler-v2.ts`
_Clean removal of client-side queueing with correct server-side dedup via storeAndEchoIfNew, but the fire-and-forget reconnect POST introduces a duplicate-event risk when the POST fails (frontend reducer is not block-level idempotent) and a detach-TTL race window before the first user message reattaches the session._

- 🟡 **regressions**: The ownership dance (reattach/rekey/zombie cleanup) was removed from `handleReconnect` and deferred to `handleSendV2`. This is architecturally sound (verified: `handleSendV2` does handle all three), but creates a window where the session stays detached between reconnect and first user message. During this window, the detach TTL timer (10 min) is still ticking. If a user reconnects but doesn't send a message quickly, the session could be aborted by the TTL despite the user being connected and watching events. The old code reattached eagerly on reconnect, stopping the TTL clock immediately. `[fixable]`


_Clean removal of client-side queueing with correct server-side dedup via storeAndEchoIfNew, but the fire-and-forget reconnect POST introduces a duplicate-event risk when the POST fails (frontend reducer is not block-level idempotent) and a detach-TTL race window before the first user message reattaches the session._

- 🔵 **style** (L51): `sendQueued` callback is declared in `ProtocolCallbacks` (marked `@deprecated v1 only`) but is now completely dead — the PR removed `onSendQueued` (the v2 caller) and the queue drain code path that called `sendQueued`. The v1 code path that used it was already gone. Consider removing it entirely to avoid confusion, since the test still mocks it and asserts it's not called. `[fixable]`


_Clean removal of client-side queueing with correct server-side dedup via storeAndEchoIfNew, but the fire-and-forget reconnect POST introduces a duplicate-event risk when the POST fails (frontend reducer is not block-level idempotent) and a detach-TTL race window before the first user message reattaches the session._

- 🔵 **missing_tests**: The deleted tests for reconnect POST failure recovery (retry via EventSource auto-reconnect, stale POST guards, queued sends surviving POST failure) represented real edge cases. The replacement tests verify fire-and-forget semantics but don't cover the scenario where a queued send races with a failed reconnect POST — e.g., the send arrives at the server before `handleReconnect` establishes the watch, so the server has no cursor and can't deduplicate. A test verifying that `doPost('reconnect', ...)` is called before `flushPendingSends()` (ordering guarantee) would document this invariant. `[fixable]`

packages/client/src/protocol-parser.ts
```diff
@@ -50,9 +50,6 @@ export interface ProtocolCallbacks {
   /** @deprecated v1 only — called to send a queued message after session_end. */
   sendQueued?(poolKey: string, msg: unknown): void;
```
dimakis: 🔵 **style**: `sendQueued` callback is declared in `ProtocolCallbacks` (marked `@deprecated v1 only`) but is now completely dead — the PR removed `onSendQueued` (the v2 caller) and the queue drain code path that called `sendQueued`. The v1 code path that used it was already gone. Consider removing it entirely to avoid confusion, since the test still mocks it and asserts it's not called. `[fixable]`

PR #446, opened 2026-08-07
feat(protocol): Symposium Phase 1 — types, event-store, tests

dimakis:
## Summary
- New types: `SessionType`, `TurnMode`, `InterceptMode`, `SeatConfig`, `TurnRules`, `SymposiumConfig`
- Event-store migration: `session_type` + `symposium_config` columns
- `upsertSession` wiring for symposium config
- 14 comprehensive tests covering type shapes and config composition

## Test Plan
- [x] All protocol package tests pass (257/257)
- [x] Types importable from `@mitzo/protocol`
- [x] Migration runs cleanly on existing databases
- [x] Symposium config serializes/deserializes correctly

### DIFF
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -225,33 +225,35 @@ export class SseConnection implements ChatConnection {
               lastSeq,
             }))
           : null);
-      this._connected = true;
       if (sessions) {
         this._pendingReconnectSessions = sessions;
         // Capture connectionId to detect stale callbacks — if a new welcome
         // arrives while this POST is in-flight, the callback should no-op
         // to avoid double-flushing pending sends.
         const postConnectionId = this._connectionId;
+        // Don't mark connected until POST succeeds — prevents send() from
+        // bypassing the pending queue and arriving before cursor setup.
         this.doPost('reconnect', { type: 'reconnect', sessions }).then(
           () => {
             if (this._connectionId !== postConnectionId) return; // stale callback
             this._pendingReconnectSessions = null;
-            // Flush pending sends AFTER reconnect so the server processes
-            // handleReconnect (cursor reset, replay) before user messages.
+            this._connected = true;
             this.flushPendingSends();
+            this.listener?.({ type: '_open' });
           },
           () => {
             if (this._connectionId !== postConnectionId) return; // stale callback
             // doPost already logs the warning. Keep _pendingReconnectSessions
             // so the next EventSource reconnect retries automatically.
-            // Still flush — handleSendV2 handles ownership independently.
-            this.flushPendingSends();
+            // Don't flush — server hasn't set up cursor/replay. Don't mark
+            // connected — sends stay queued until next successful reconnect.
           },
         );
       } else {
+        this._connected = true;
         this.flushPendingSends();
+        this.listener?.({ type: '_open' });
       }
-      this.listener?.({ type: '_open' });
       this._isReconnect = true;
     });
 
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -214,17 +214,7 @@ export class SseConnection implements ChatConnection {
       // have been cleared (clearSession) since the retry was queued.
       // Refresh lastSeq from current map — SSE events may have advanced it
       // since the original failure, avoiding unnecessary replay.
-      const pending = this._pendingReconnectSessions
-        ?.filter((s) => this.seqBySession.has(s.sessionId))
-        .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
-      const sessions =
-        (pending && pending.length > 0 ? pending : null) ??
-        (this._isReconnect && this.seqBySession.size > 0
-          ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
-              sessionId,
-              lastSeq,
-            }))
-          : null);
+      const sessions = this.getReconnectSessions();
       if (sessions) {
         this._pendingReconnectSessions = sessions;
         // Capture connectionId to detect stale callbacks — if a new welcome
@@ -288,7 +278,9 @@ export class SseConnection implements ChatConnection {
   }
 
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
-    if (!this._connectionId) return;
+    if (!this._connectionId) {
+      throw new Error('doPost called without connectionId');
+    }
     try {
       const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
@@ -352,6 +344,30 @@ export class SseConnection implements ChatConnection {
       default:
         return null;
     }
+  }
+
+  /**
+   * Build the session list for a reconnect POST.
+   *
+   * Priority: (1) pending retry sessions (filtered against current seqBySession,
+   * with refreshed lastSeq), (2) all tracked sessions if this is a reconnect.
+   * Returns null when there's nothing to reconnect.
+   */
+  private getReconnectSessions(): Array<{ sessionId: string; lastSeq: number }> | null {
+    const pending = this._pendingReconnectSessions
+      ?.filter((s) => this.seqBySession.has(s.sessionId))
+      .map((s) => ({ ...s, lastSeq: this.seqBySession.get(s.sessionId)! }));
+
+    if (pending && pending.length > 0) return pending;
+
+    if (this._isReconnect && this.seqBySession.size > 0) {
+      return Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
+        sessionId,
+        lastSeq,
+      }));
+    }
+
+    return null;
   }
 
   // ─── Browser lifecycle ─────────────────────────────────────────────────────
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -64,6 +64,7 @@ export class SseConnection implements ChatConnection {
       this.es = null;
     }
     this._connected = false;
+    this._pendingReconnectSessions = null;
   }
 
   /**
@@ -209,8 +210,13 @@ export class SseConnection implements ChatConnection {
       // Fire reconnect POST if reconnecting with sessions, or retry a
       // previously failed reconnect. handleSendV2 handles ownership on first
       // message, and replayed events arrive via SSE regardless.
+      // Filter pending retries against current seqBySession — sessions may
+      // have been cleared (clearSession) since the retry was queued.
+      const pending = this._pendingReconnectSessions?.filter((s) =>
+        this.seqBySession.has(s.sessionId),
+      );
       const sessions =
-        this._pendingReconnectSessions ??
+        (pending && pending.length > 0 ? pending : null) ??
         (this._isReconnect && this.seqBySession.size > 0
           ? Array.from(this.seqBySession.entries()).map(([sessionId, lastSeq]) => ({
               sessionId,
@@ -274,7 +280,7 @@ export class SseConnection implements ChatConnection {
   private async doPost(endpoint: string, body: Record<string, unknown>): Promise<void> {
     if (!this._connectionId) return;
     try {
-      await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
+      const res = await this.config.fetch(`${this.config.baseUrl}/api/chat/${endpoint}`, {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
@@ -282,6 +288,9 @@ export class SseConnection implements ChatConnection {
         },
         body: JSON.stringify(body),
       });
+      if (!res.ok) {
+        throw new Error(`HTTP ${res.status}`);
+      }
     } catch (err) {
       // POST failures are non-fatal — the server may be temporarily
       // unreachable. SSE EventSource auto-reconnects and replays missed events
--- a/packages/client/src/sse-connection.ts
+++ b/packages/client/src/sse-connection.ts
@@ -23,7 +23,6 @@ export interface SseConnectionConfig {
   fetch: (url: string, init?: RequestInit) => Promise<Response>;
   /** Factory for EventSource — allows injection for testing. */
   createEventSource?: (url: string) => EventSource;
-  reconnectDelayMs?: number;
   /** URL for the sendBeacon suspend fallback. */
   suspendUrl?: string;
 }
@@ -38,7 +37,6 @@ export class SseConnection implements ChatConnection {
   private listener: ConnectionListener | null = null;
   private seqBySession = new Map<string, number>();
   private pendingSends: Array<{ endpoint: string; body: Record<string, unknown> }> = [];
-  private reconnectTimer: ReturnType<typeof setTimeout> | null = null;
   private boundOnVisibility: (() => void) | null = null;
   private boundOnPageShow: ((e: PageTransitionEvent) => void) | null = null;
   private boundOnPageHide: (() => void) | null = null;
@@ -47,7 +45,6 @@ export class SseConnection implements ChatConnection {
   constructor(config: SseConnectionConfig) {
     this.config = {
       createEventSource: (url: string) => new EventSource(url),
-      reconnectDelayMs: 500,
       suspendUrl: '',
       ...config,
     };
@@ -60,10 +57,6 @@ export class SseConnection implements ChatConnection {
 
   disconnect(): void {
     this.removeBrowserListeners();
-    if (this.reconnectTimer) {
-      clearTimeout(this.reconnectTimer);
-      this.reconnectTimer = null;
-    }
     if (this.es) {
       this.es.close();
       this.es = null;
@@ -92,8 +85,8 @@ export class SseConnection implements ChatConnection {
       return true;
     }
 
-    // Queue if reconnecting
-    if (this.reconnectTimer || this.es) {
+    // Queue if EventSource exists (reconnecting)
+    if (this.es) {
       if (this.pendingSends.length >= MAX_PENDING_SENDS) {
         this.pendingSends.shift();
       }
@@ -175,7 +168,6 @@ export class SseConnection implements ChatConnection {
    */
   checkAndReconnect(force = false): void {
     if (!force && this._connected) return;
-    if (this.reconnectTimer) return;
     if (this.es) {
       this.es.close();
       this.es = null;
@@ -192,11 +184,6 @@ export class SseConnection implements ChatConnection {
 
   private doConnect(): void {
     if (this.es) return;
-
-    if (this.reconnectTimer) {
-      clearTimeout(this.reconnectTimer);
-      this.reconnectTimer = null;
-    }
 
     // Always use the base URL — reconnect sessions are sent via POST in the
     // welcome handler. This avoids the bug where EventSource auto-reconnect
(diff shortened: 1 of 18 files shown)


packages/client/src/sse-connection.ts
```diff
@@ -217,13 +207,38 @@ export class SseConnection implements ChatConnection {
...
+      // Fire reconnect POST if reconnecting with sessions, or retry a
+      // previously failed reconnect. handleSendV2 handles ownership on first
+      // message, and replayed events arrive via SSE regardless.
+      // Filter pending retries against current seqBySession — sessions may
+      // have been cleared (clearSession) since the retry was queued.
+      // Refresh lastSeq from current map — SSE events may have advanced it
+      // since the original failure, avoiding unnecessary replay.
+      const sessions = this.getReconnectSessions();
+      if (sessions) {
+        this._pendingReconnectSessions = sessions;
```
dimakis, line 219: 🟡 **unsafe_assumptions**: When the reconnect POST fails, `_pendingReconnectSessions` retains the sessions for retry on the next welcome. However, `checkAndReconnect()` (line 172) does NOT clear `_pendingReconnectSessions` — it only sets `_connected = false` and creates a new EventSource. This means `getReconnectSessions()` will use the stale pending list on the next welcome, which is the intended design. But if an `onerror` fires (EventSource auto-reconnect, line 267), `_connected` is set to false but `_pendingReconnectSessions` is also preserved. This is correct for the retry intent, but if the server-side session has been cleaned up between failures, the client will keep retrying a reconnect POST for a session that no longer exists — there's no TTL or attempt limit on `_pendingReconnectSessions`. Consider adding a retry cap or TTL to prevent infinite reconnect POST attempts for dead sessions. `[fixable]`

packages/client/src/sse-connection.ts
```diff
@@ -217,13 +207,38 @@ export class SseConnection implements ChatConnection {
...
+        // Capture connectionId to detect stale callbacks — if a new welcome
+        // arrives while this POST is in-flight, the callback should no-op
+        // to avoid double-flushing pending sends.
+        const postConnectionId = this._connectionId;
+        // Don't mark connected until POST succeeds — prevents send() from
+        // bypassing the pending queue and arriving before cursor setup.
+        this.doPost('reconnect', { type: 'reconnect', sessions }).then(
```
dimakis, line 226: 🟡 **regressions**: The old `doReconnectPost` had a staleness guard that checked BOTH `this.es !== welcomeEs` (EventSource instance identity) AND `this._connectionId !== welcomeConnectionId`. The new inline handler only checks `this._connectionId !== postConnectionId`. This drops the ES-identity guard. If `checkAndReconnect()` is called while a reconnect POST is in-flight, it closes the old ES and creates a new one, but `_connectionId` doesn't change until the new welcome arrives. During that window, if the stale POST resolves, the connectionId check alone won't catch it — `_connectionId` is still the old value because no new welcome has arrived yet. The old code would have bailed via the `es !== welcomeEs` check. This could cause `_connected = true` and `flushPendingSends()` to fire on a stale EventSource. `[fixable]`

packages/protocol/__tests__/event-store.test.ts
```diff
@@ -851,6 +851,56 @@ describe('EventStore', () => {
...
+      const parsed = JSON.parse(session!.symposiumConfig!);
+      expect(parsed.seats).toHaveLength(2);
+    });
+
+    it('defaults symposium_config to null', () => {
+      store.upsertSession({ sessionId: 'chat-2', summary: 'Chat' });
+      expect(store.getSession('chat-2')!.symposiumConfig).toBeNull();
+    });
+
+    it('events table supports seat_id in payload', () => {
```
dimakis, line 901: 🔵 **style**: The test 'events table supports seat_id in payload' is misleading — `seatId` is just a JSON field inside the `payload` TEXT column, not a database column. The events table stores payload as a JSON string, so any key can be stored. This test doesn't validate any schema change; it only confirms that JSON round-trips correctly, which is already covered by existing event append/read tests. Consider removing or renaming to clarify it tests payload serialization of symposium-specific fields. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -576,6 +582,17 @@ export function handleSendV2(
             }
             applySkillPolicy(activeClientId);
             ctx.connRegistry.watch(connectionId, sessionId);
+            // No resetCursor here — handleReconnect (fire-and-forget POST) sets
```
dimakis, line 585: 🔵 **style**: The 11-line TODO comment block (lines 585–595) explaining the cursor/seq-dedup trade-off is excessive inline documentation. The key insight ('client-side seq dedup handles duplicates between watch and reconnect POST') could be a 2-line comment. The detailed analysis of event loop ordering and HTTP request interleaving belongs in a design doc or commit message, not in the source code where it will rot. `[fixable]`

dimakis (commented):
## Centaur Review

_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🟡 **unsafe_assumptions** (L219): When the reconnect POST fails, `_pendingReconnectSessions` retains the sessions for retry on the next welcome. However, `checkAndReconnect()` (line 172) does NOT clear `_pendingReconnectSessions` — it only sets `_connected = false` and creates a new EventSource. This means `getReconnectSessions()` will use the stale pending list on the next welcome, which is the intended design. But if an `onerror` fires (EventSource auto-reconnect, line 267), `_connected` is set to false but `_pendingReconnectSessions` is also preserved. This is correct for the retry intent, but if the server-side session has been cleaned up between failures, the client will keep retrying a reconnect POST for a session that no longer exists — there's no TTL or attempt limit on `_pendingReconnectSessions`. Consider adding a retry cap or TTL to prevent infinite reconnect POST attempts for dead sessions. `[fixable]`
- 🟡 **regressions** (L226): The old `doReconnectPost` had a staleness guard that checked BOTH `this.es !== welcomeEs` (EventSource instance identity) AND `this._connectionId !== welcomeConnectionId`. The new inline handler only checks `this._connectionId !== postConnectionId`. This drops the ES-identity guard. If `checkAndReconnect()` is called while a reconnect POST is in-flight, it closes the old ES and creates a new one, but `_connectionId` doesn't change until the new welcome arrives. During that window, if the stale POST resolves, the connectionId check alone won't catch it — `_connectionId` is still the old value because no new welcome has arrived yet. The old code would have bailed via the `es !== welcomeEs` check. This could cause `_connected = true` and `flushPendingSends()` to fire on a stale EventSource. `[fixable]`


_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🟡 **missing_tests**: The PR removes tests for several important edge cases that were previously covered: (1) `bails out if disconnect() called during in-flight reconnect POST` — the new code clears `_pendingReconnectSessions` on disconnect, but there's no test that the in-flight POST's `.then()` callback is a no-op after disconnect. (2) `ignores stale reconnect POST when a newer welcome arrives` — the new code uses connectionId-based staleness, but the only test for this (`marks connected after reconnect POST succeeds`) doesn't test the race with a second welcome. (3) `dispatches SSE events to listener while reconnect POST is in-flight` — removed entirely with no replacement. SSE event delivery during the reconnect window is a critical path. `[fixable]`


_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🟡 **regressions**: Removing periodic sync entirely means there is no server-side mechanism to retry delivery of events that failed during `broadcast()`. The old code would retry via periodic sync every 5s. The new design relies entirely on client-initiated reconnect (EventSource auto-reconnect → welcome → replay from EventStore). This is a deliberate architectural change documented in comments, but it means that a transient `broadcast()` failure (e.g., WebSocket buffer full) on a still-connected client will result in a permanently missed event until the client disconnects and reconnects. The comment at line 133 says 'reconnect replay will cover the gap' but that only happens on reconnect, not on transient send failures where the connection stays alive.


_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🔵 **style** (L897): The test 'events table supports seat_id in payload' is misleading — `seatId` is just a JSON field inside the `payload` TEXT column, not a database column. The events table stores payload as a JSON string, so any key can be stored. This test doesn't validate any schema change; it only confirms that JSON round-trips correctly, which is already covered by existing event append/read tests. Consider removing or renaming to clarify it tests payload serialization of symposium-specific fields. `[fixable]`


### `server/ws-handler-v2.ts`
_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🔵 **style** (L585): The 11-line TODO comment block (lines 585–595) explaining the cursor/seq-dedup trade-off is excessive inline documentation. The key insight ('client-side seq dedup handles duplicates between watch and reconnect POST') could be a 2-line comment. The detailed analysis of event loop ordering and HTTP request interleaving belongs in a design doc or commit message, not in the source code where it will rot. `[fixable]`


### `packages/protocol/__tests__/types.test.ts`
_Solid Phase 1 PR combining symposium types with a significant transport simplification (periodic sync removal, reconnect POST refactoring). The symposium additions are clean and complete. The main risk is the SSE reconnect refactoring: the staleness guard lost the ES-identity check (regression), and several edge-case tests were removed without replacement._

- 🔵 **bugs** (L215): The symposium type tests (lines 215–319) only verify TypeScript type assignability at compile time — they don't validate any runtime behavior, Zod schemas, or serialization. For example, `SessionType[] = ['chat', 'symposium']` proves the type accepts these literals, but doesn't verify exhaustiveness (a third value added to the union wouldn't fail this test). These tests provide marginal value over the TypeScript compiler itself. Consider adding Zod schema validation tests if runtime validation is planned.

packages/protocol/__tests__/types.test.ts
```diff
@@ -206,4 +212,108 @@ describe('protocol types', () => {
     expect(meta.promptCount).toBe(3);
     expect(meta.totalCostUsd).toBe(0.01);
   });
+
```
dimakis, line 215: 🔵 **bugs**: The symposium type tests (lines 215–319) only verify TypeScript type assignability at compile time — they don't validate any runtime behavior, Zod schemas, or serialization. For example, `SessionType[] = ['chat', 'symposium']` proves the type accepts these literals, but doesn't verify exhaustiveness (a third value added to the union wouldn't fail this test). These tests provide marginal value over the TypeScript compiler itself. Consider adding Zod schema validation tests if runtime validation is planned.

packages/protocol/src/event-store.ts
```diff
@@ -781,6 +809,8 @@ function rowToSession(row: SessionRow): SessionMeta {
     lastStateChange: row.last_state_change ?? null,
     agentName: row.agent_name ?? null,
     bootContext: row.boot_context ?? null,
+    sessionType: (row.session_type as SessionMeta['sessionType']) ?? 'chat',
```
dimakis, line 812: 🟡 **unsafe_assumptions**: Unsafe cast of `row.session_type` to `SessionMeta['sessionType']` without validation. If the database contains a value other than 'chat' or 'symposium' (e.g. from a manual edit or future migration rollback), the cast silently passes through an invalid string. Consider validating the value: `row.session_type === 'symposium' ? 'symposium' : 'chat'`. `[fixable]`

server/chat.ts
```diff
@@ -1270,7 +1270,17 @@ export async function interruptChat(
       );
       await Promise.allSettled(stops);
     }
-    await session.queryInstance.interrupt();
+    try {
```
dimakis, line 1273: 🔵 **missing_tests**: The new try/catch around `queryInstance.interrupt()` has no dedicated test. Adding a test where `interrupt()` rejects (e.g. ProcessTransport 'not ready for writing') would verify the function continues to queue the message and returns true despite the error. `[fixable]`

server/ws-handler-v2.ts
```diff
@@ -576,6 +582,17 @@ export function handleSendV2(
             }
             applySkillPolicy(activeClientId);
             ctx.connRegistry.watch(connectionId, sessionId);
```
dimakis, line 584: 🟡 **unsafe_assumptions**: After watch() on handleSendV2's running-session path, cursor starts at 0 (default) until the reconnect POST arrives and calls resetCursor. The inline comment acknowledges this relies on client-side seq dedup for correctness. While the TODO(P4) notes this, the window is real: if the server broadcasts a high-seq event between watch() and resetCursor(), the client receives it (fine), but a concurrent reconnect replay could also deliver the same events. The comment says 'bandwidth trade-off, not correctness' but duplicate events could confuse non-idempotent client logic (e.g. incrementing counters). Verify that all client message handlers are truly idempotent under seq dedup.

server/ws-handler-v2.ts
```diff
@@ -247,50 +247,55 @@ export function handleReconnect(
...
+        // watch/broadcast but the agent's transport stays stale.
+        //
+        // Note: reattach refreshes the transport but does NOT rekey the
+        // clientId. If ownerGone=true (device restart), the clientId still
+        // references the old connectionId until the first send triggers
+        // rekeyChat via handleSendV2. During this window, getOwnerConnection()
+        // returns a stale value. This is safe — event delivery uses the
+        // transport (refreshed), and ownership checks on send/interrupt
+        // handle the rekey atomically.
         const found = ctx.sessionRegistry.findBySessionId(entry.sessionId);
```
dimakis, line 266: 🟡 **regressions**: Reconnect now reattaches detached sessions (refreshing transport) but no longer rekeys the clientId. The inline comment (lines 259-265) documents this is safe because 'event delivery uses the transport (refreshed)'. However, between reattach and the first send, `getOwnerConnection()` returns the stale connectionId. If another connection attempts a send/interrupt for the same session during this window, the ownership check sees the stale owner as 'gone' (not in connRegistry) and performs a takeover — which is correct behavior but may trigger spurious takeover logs. Not a bug, but worth being aware of in production log analysis.

packages/client/src/sse-connection.ts
dimakis, line 226: 🔵 **style**: The reconnect POST `.then(successCb, failureCb)` form is harder to read than async/await with try/catch, especially since the success handler is 5 lines and failure handler is 4 lines. However, since the welcome event handler is synchronous and the POST is intentionally fire-and-forget, the `.then()` form avoids making the handler async. This is acceptable — just noting the trade-off.

packages/protocol/__tests__/event-store.test.ts
```diff
@@ -851,6 +851,60 @@ describe('EventStore', () => {
...
```
dimakis, line 901: 🔵 **missing_tests**: The test 'events table supports seat_id in payload' only verifies that seatId round-trips through the JSON payload column. It doesn't test that the session associated with 'symp-4' was auto-created (upsertSession wasn't called). This works because `append()` auto-creates session rows, but makes the test implicitly dependent on that behavior. Consider calling `store.upsertSession({ sessionId: 'symp-4', sessionType: 'symposium' })` first for clarity. `[fixable]`

dimakis (commented):
## Centaur Review

_Well-structured PR that cleanly adds Symposium Phase 1 types/migration and simplifies reconnect by removing periodic sync and deferring ownership to send/interrupt handlers. The main concern is an unsafe cast of session_type from the database; the reconnect simplification is sound with proper stale-callback guards._

- 🟡 **unsafe_assumptions** (L812): Unsafe cast of `row.session_type` to `SessionMeta['sessionType']` without validation. If the database contains a value other than 'chat' or 'symposium' (e.g. from a manual edit or future migration rollback), the cast silently passes through an invalid string. Consider validating the value: `row.session_type === 'symposium' ? 'symposium' : 'chat'`. `[fixable]`


### `server/chat.ts`
_Well-structured PR that cleanly adds Symposium Phase 1 types/migration and simplifies reconnect by removing periodic sync and deferring ownership to send/interrupt handlers. The main concern is an unsafe cast of session_type from the database; the reconnect simplification is sound with proper stale-callback guards._

- 🔵 **missing_tests** (L1273): The new try/catch around `queryInstance.interrupt()` has no dedicated test. Adding a test where `interrupt()` rejects (e.g. ProcessTransport 'not ready for writing') would verify the function continues to queue the message and returns true despite the error. `[fixable]`


### `server/ws-handler-v2.ts`
_Well-structured PR that cleanly adds Symposium Phase 1 types/migration and simplifies reconnect by removing periodic sync and deferring ownership to send/interrupt handlers. The main concern is an unsafe cast of session_type from the database; the reconnect simplification is sound with proper stale-callback guards._

- 🟡 **unsafe_assumptions** (L584): After watch() on handleSendV2's running-session path, cursor starts at 0 (default) until the reconnect POST arrives and calls resetCursor. The inline comment acknowledges this relies on client-side seq dedup for correctness. While the TODO(P4) notes this, the window is real: if the server broadcasts a high-seq event between watch() and resetCursor(), the client receives it (fine), but a concurrent reconnect replay could also deliver the same events. The comment says 'bandwidth trade-off, not correctness' but duplicate events could confuse non-idempotent client logic (e.g. incrementing counters). Verify that all client message handlers are truly idempotent under seq dedup.
- 🟡 **regressions** (L266): Reconnect now reattaches detached sessions (refreshing transport) but no longer rekeys the clientId. The inline comment (lines 259-265) documents this is safe because 'event delivery uses the transport (refreshed)'. However, between reattach and the first send, `getOwnerConnection()` returns the stale connectionId. If another connection attempts a send/interrupt for the same session during this window, the ownership check sees the stale owner as 'gone' (not in connRegistry) and performs a takeover — which is correct behavior but may trigger spurious takeover logs. Not a bug, but worth being aware of in production log analysis.


_Well-structured PR that cleanly adds Symposium Phase 1 types/migration and simplifies reconnect by removing periodic sync and deferring ownership to send/interrupt handlers. The main concern is an unsafe cast of session_type from the database; the reconnect simplification is sound with proper stale-callback guards._

- 🔵 **style** (L226): The reconnect POST `.then(successCb, failureCb)` form is harder to read than async/await with try/catch, especially since the success handler is 5 lines and failure handler is 4 lines. However, since the welcome event handler is synchronous and the POST is intentionally fire-and-forget, the `.then()` form avoids making the handler async. This is acceptable — just noting the trade-off.


_Well-structured PR that cleanly adds Symposium Phase 1 types/migration and simplifies reconnect by removing periodic sync and deferring ownership to send/interrupt handlers. The main concern is an unsafe cast of session_type from the database; the reconnect simplification is sound with proper stale-callback guards._

- 🔵 **missing_tests** (L901): The test 'events table supports seat_id in payload' only verifies that seatId round-trips through the JSON payload column. It doesn't test that the session associated with 'symp-4' was auto-created (upsertSession wasn't called). This works because `append()` auto-creates session rows, but makes the test implicitly dependent on that behavior. Consider calling `store.upsertSession({ sessionId: 'symp-4', sessionType: 'symposium' })` first for clarity. `[fixable]`

